PR #10, opened 2018-12-11
support for OpenVINO R4 release

dtrawins:


requirements.txt
```diff
@@ -1,7 +1,7 @@
 setuptools==34.3.0
 grpcio==1.14.1
 grpcio-tools==1.14.1
-numpy==1.14.5
+numpy
```
mareklew-Intel: I think it`s will be better and safer if we still have specified version of numpy in requirements.txt

PR #15, opened 2019-01-19
auto batch size based on input data

dtrawins:
model server detects the batch size in the request data and reload the model accordingly
first request of the new batch size will be slower due to extra model reloading
sequential requests with the same batch size are executed without delay

### DIFF
--- a/ie_serving/models/ir_engine.py
+++ b/ie_serving/models/ir_engine.py
@@ -108,7 +108,6 @@ class IrEngine():
 
     def infer(self, data: dict, batch_size=None):
         if batch_size is not self.net.batch_size:
-            #self.exec_net.requests[0].set_batch(batch_size)
             self.net.batch_size = batch_size
             self.exec_net = self.plugin.load(network=self.net)
         results = self.exec_net.infer(inputs=data)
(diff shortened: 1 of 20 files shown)


ie_serving/models/ir_engine.py
```diff
@@ -40,19 +42,36 @@ def __init__(self, model_xml, model_bin, mapping_config, exec_net,
...
     @classmethod
-    def build(cls, model_xml, model_bin, mapping_config):
+    def build(cls, model_xml, model_bin, mapping_config, batch_size):
         plugin = IEPlugin(device=DEVICE, plugin_dirs=PLUGIN_DIR)
         if CPU_EXTENSION and 'CPU' in DEVICE:
             plugin.add_cpu_extension(CPU_EXTENSION)
         net = IENetwork.from_ir(model=model_xml, weights=model_bin)
-        input_blob = next(iter(net.inputs))
-        batch_size = net.inputs[input_blob].shape[0]
+        if batch_size is not None:
```
mareklevv: This whole if-else statement should be moved to separate function. In our unit tests all paths should be tested. Especially with non int input.

PR #19, opened 2019-02-08
add support for s3 with anonymous access

dtrawins:


ie_serving/models/s3_model.py
```diff
@@ -42,13 +45,19 @@ def s3_download_file(cls, path):
...
         bucket_name = parsed_path.netloc
         file_path = parsed_path.path[1:]
         tmp_path = os.path.join('/tmp', file_path.split(os.sep)[-1])
-        s3_transfer = boto3.s3.transfer.S3Transfer(s3_client)
-        s3_transfer.download_file(bucket_name, file_path, tmp_path)
+        try:
+            s3_client = boto3.client('s3', endpoint_url=s3_endpoint)
+            s3_transfer = boto3.s3.transfer.S3Transfer(s3_client)
+            s3_transfer.download_file(bucket_name, file_path, tmp_path)
+        except exceptions.ClientError:
```
mzegla: I believe ClientError is also thrown when for example bucket does not exist. Maybe we could specify that we want to retry if credentials are the issue.

dtrawins: because there is the same error for missing file and incorrect permissions it might be had to distinguish? Currently we retry with anonymous client if the standard client fails to connect (for example missing or invalid aws variables).

mzegla: I was thinking about some ClientError subclass maybe, but it seems there are no explicit ClientError subclasses. In this case I don't have better solution for this at the moment. Maybe some error code check, but it wouldn't improve much.

PR #22, opened 2019-03-15
Switching to asynchronous processing

mzegla:
Changing from synchronous to asynchronous inference. Enabling processing with multiple workers for throughput oriented use cases. Employing [OpenVINO CPU Streams](https://software.intel.com/en-us/articles/OpenVINO-Inference-Engine-Optimization-Guide#cpu-streams).
New setting setting: 
number of gRPC server workers == number of [InferRequests](https://software.intel.com/en-us/articles/OpenVINO-InferEngine#inferrequest-class) for network == number of CPU Streams


mzegla:
Closing due to lack of advantage over multi-instance with single workers approach. 
Time variations present in server specific use cases are bad for Asynchronous API and multiple streams usage. In order to maximize performance with multiple streams there should be as much inferences running in parallel as possible. 
Different requests arrival and processing time in gRPC server, their dispatch to particular workers, images preprocessing and inference time create those variations, which significantly impact overall performance in asynchronous mode with multiple workers and streams.

PR #23, opened 2019-03-18
Add custom minio configuration

mareklevv:


ie_serving/config.py
```diff
@@ -30,3 +30,11 @@
 
 
 MAPPING_CONFIG_FILENAME = 'mapping_config.json'
+
+
+class S3_Config:
+    S3_ENDPOINT = os.environ.get('S3_ENDPOINT', 's3.us-west-2.amazonaws.com')
```
dtrawins: default endpoint should be None

PR #36, opened 2019-04-16
2019 r1

dtrawins:


Dockerfile_binary_openvino
```diff
@@ -19,14 +19,14 @@ RUN cd $TEMP_DIR && pwd && ls && \
     cd l_openvino_toolkit* && \
     sed -i 's/decline/accept/g' silent.cfg && \
     pwd | grep -q openvino_toolkit_p ; \
-    if [ $? = 0 ];then sed -i 's/COMPONENTS=DEFAULTS/COMPONENTS=;intel-openvino_base__noarch;intel-dldt_base__noarch;intel-setupvars__noarch;intel-inference_engine_sdk__noarch;intel-inference_engine_rt__noarch;intel-inference_engine_cpu__noarch;intel-inference_engine_vpu__noarch;intel-inference_engine_gna__noarch;intel-inference_engine_dlia__noarch;intel-openvino_base-pset/g' silent.cfg; fi && \
+    if [ $? = 0 ];then sed -i 's/COMPONENTS=DEFAULTS/COMPONENTS=intel-openvino-base__noarch;intel-openvino-dldt-base__noarch;intel-openvino-setupvars__x86_64;intel-openvino-ie-sdk-ubuntu-xenial__x86_64;intel-openvino-ie-rt__x86_64;intel-openvino-ie-rt-core-ubuntu-xenial__x86_64;intel-openvino-ie-rt-cpu-ubuntu-xenial__x86_64;intel-openvino-ie-rt-vpu-ubuntu-xenial__x86_64;intel-openvino-ie-rt-hddl-ubuntu-xenial__x86_64;intel-openvino-gfx-driver__x86_64;intel-openvino-base-pset/g' silent.cfg; fi && \
```
joakimr-axis: For a oneliner like this, it would be simpler to use

[ $? != 0 ] || sed

joakimr-axis: ...but the best would actually be to do this with the shells case functionality. I will fix a separate patch for that.

joakimr-axis: (Added such functionality in #40.)

Dockerfile_binary_openvino
```diff
@@ -20,15 +20,23 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+    sed -i "s/COMPONENTS=DEFAULTS/COMPONENTS=$COMPONENTS/g" silent.cfg && \    
     ./install.sh -s silent.cfg --ignore-signature && \
     rm -Rf $TEMP_DIR $INSTALL_DIR/install_dependencies $INSTALL_DIR/uninstall* /tmp/* $DL_INSTALL_DIR/documentation $DL_INSTALL_DIR/inference_engine/samples
 
-ENV PYTHONPATH="$INSTALL_DIR/python/python3.5/ubuntu16:$INSTALL_DIR/python/python3.5"
-ENV LD_LIBRARY_PATH="$DL_INSTALL_DIR/inference_engine/external/gna/lib:$DL_INSTALL_DIR/inference_engine/external/mkltiny_lnx/lib:$DL_INSTALL_DIR/inference_engine/lib/ubuntu_16.04/intel64:$DL_INSTALL_DIR/inference_engine/external/omp/lib"
+ENV PYTHONPATH="$INSTALL_DIR/python/python3.5"
+ENV LD_LIBRARY_PATH="$DL_INSTALL_DIR/inference_engine/external/tbb/lib:$DL_INSTALL_DIR/inference_engine/external/mkltiny_lnx/lib:$DL_INSTALL_DIR/inference_engine/lib/intel64"
 
 ENV https_proxy $https_proxy
```
mareklevv: I think set up proxy envs on build stage is wrong approach, because image built with http-proxy args will require to unset this proxies on container start.

Dockerfile
```diff
@@ -32,16 +32,22 @@ RUN apt-get update && apt-get install -y \
             vim \
             wget
 
+RUN wget https://cmake.org/files/v3.14/cmake-3.14.3.tar.gz && \
```
mareklevv: In first `RUN` in this Dockerfile you also install cmake.

PR #40, opened 2019-04-17
Use shell built-in case to determine toolkit version

### LINKED ISSUE #39: Simplify the pwd | grep -> check result part of Dockerfile_binary_openvino
The

pwd | grep -q openvino_toolkit_p ; \
    if [ $? = 0 ];then sed -i 's [...]

parts of Dockerfile_binary_openvino are not, IMO, easily read. But what they do is to determine from pwd what toolkit (p/fpga) is at hand and selects components in silent.cfg accordingly. I reckon such a check is a typical job for the shell's built-in case statement, which would not only increase readability but also be easier to maintain. (If, say, the components would differ between toolkit versions, handling of multiple versions is easily added as case statements.)

joakimr-axis:
In Dockerfile_binary_openvino, the use of pwd piped to grep and checking
grep's result code is here replaced with a simple shell native case
instead. This increases readability and maintainability; should there be
difference between toolkit version it is quite simple to extend the
case. (A bonus is increased performance by using shell natives, even if
that is hardly noticable on a modern computer.)

Fixes #39

Change-Id: I40ad3ac1e98c45e8b197498130412ebe8f03705a

### DIFF
--- a/Dockerfile_binary_openvino
+++ b/Dockerfile_binary_openvino
@@ -20,10 +20,24 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 ADD l_openvino_toolkit*.tgz $TEMP_DIR/
 RUN cd $TEMP_DIR/l_openvino_toolkit* && \
     sed -i 's/decline/accept/g' silent.cfg && \
-    pwd | grep -q openvino_toolkit_p ; \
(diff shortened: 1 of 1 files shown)


dtrawins:
those changes will be included in PR https://github.com/IntelAI/OpenVINO-model-server/pull/36 
but components are configured only for the releases 2019*
The reason is that in 2019 there were introduced changes which makes the dockerfiles incompatible with older binary releases.

PR #43, opened 2019-05-07
start_server.sh: POSIX compatible, no need for bash

### LINKED ISSUE #42: start_server.sh is bash-specific but could be POSIX compliant
There is nothing in start_server.sh that needs bash-specific functionality.
Hence it would be better off as POSIX compliant.

joakimr-axis:
Using '.' instead of 'source' in start_server.sh makes it POSIX
compatible. There is no reason not to have it POSIX compatible.

Fixes #42

Change-Id: I655d391979c5a624b3a79c3409542a1b301c1762

### DIFF
--- a/start_server.sh
+++ b/start_server.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 #
 # Copyright (c) 2018 Intel Corporation
 #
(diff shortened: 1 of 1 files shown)


joakimr-axis:
...well, there is a reason to use bash since the OpenVINO SDK setupvars.sh has bash-specific content. But none of that bash-specific content in setupvars.sh needs to be there either since it adds no additional functionality over POSIX. Changing the [[ to [, use '.' instead of 'source' and set /bin/sh in the shebang and that one would be good to go.

Does anybody know if there is a place to place pull requests for the OpenVINO SDK?

PR #46, opened 2019-05-15
Let make choose amount of parallelization threads

### LINKED ISSUE #28: Dockerfile uses hard-coded -j4 option
The hard coded value 4 works well for CPUs with 4 threads. On a CPU with more than that, which is found in many modern computers, the CPU is under used with the -j4 setting. Using e.g. the outcome of `nproc` instead would be more efficient and shorten build times on CPUs capable of more than 4 threads.

joakimr-axis:
Using 'make -j' without a certain number will leave the decision how
many threads to use to make. Typically, make will extract this from the
system's CPU information and choose the amount of CPU threads. (This
should then give the same result as using 'make -j$(nproc)'.)

Fixes #28

Change-Id: I030b39e647b758781a099c09aa05566bbbc233c2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN wget https://cmake.org/files/v3.14/cmake-3.14.3.tar.gz && \
     tar -xvzf cmake-3.14.3.tar.gz && \
     cd cmake-3.14.3/  && \
     ./configure && \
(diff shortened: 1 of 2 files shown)


dtrawins:
I tend to reject this PR because it is breaking our CI system due to too much RAM consumption.

PR #64, opened 2019-07-05
performance tuning

dtrawins:


ie_serving/server/service.py
```diff
@@ -73,13 +73,13 @@ def Predict(self, request, context):
             context.set_details(inference_input)
             logger.debug("PREDICT, problem with input data. Exit code {}"
                          .format(code))
-            self.models[model_name].engines[version].in_use.release()
             return predict_pb2.PredictResponse()
-
+        self.models[model_name].engines[version].in_use.acquire()
```
mzegla: Moving this lock here might be dangerous. There is a risk of performing inference on different model than it was meant for.
If user added new version and updating thread performed the update exactly when worker thread executes code between `prepare_input_data()` and `infer()`, then inference is made on the wrong servable.

PR #78, opened 2019-07-15
More unit tests for GetModelStatus

mzegla:
Unit tests for interfaces check


tests/unit/server/test_get_model_status_grpc.py
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright (c) 2018 Intel Corporation
```
dtrawins: change to 2019

PR #85, opened 2019-07-25
Upgrade to 2019.R2

dtrawins:


Dockerfile
```diff
@@ -1,80 +1,29 @@
-FROM ubuntu:16.04 as DEV
-RUN apt-get update && apt-get install -y \
-            autoconf \
-            automake \
-            build-essential \
+FROM ubuntu:18.04
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
mzegla: Since Myriad plugin for NCS is public we can include `usbutils` package that is needed for Movidius Sticks to work. 
We are already having it in binary build dockerfile to enable NCS.

Dockerfile
```diff
@@ -1,80 +1,29 @@
...
 
-RUN . .venv/bin/activate && pip3 install .
-
-COPY --from=DEV /2019_R1.1/inference-engine/bin/intel64/Release/lib/*.so /usr/local/lib/
-COPY --from=DEV /2019_R1.1/inference-engine/ie_bridges/python/bin/intel64/Release/python_api/python3.5/openvino/ /usr/local/lib/openvino/
-COPY --from=DEV /2019_R1.1/mklml_lnx_2019.0.3.20190220/lib/lib*.so /usr/local/lib/
-ENV LD_LIBRARY_PATH=/usr/local/lib
-ENV PYTHONPATH=/usr/local/lib
-
+RUN . .venv/bin/activate && pip3 install .
```
mzegla: New line 😄

PR #87, opened 2019-08-19
HDDL Support

mzegla:


docs/docker_container.md
```diff
@@ -310,6 +310,34 @@ ie-serving-py:latest /ie-serving-py/start_server.sh ie_serving model --model_pat
...
+
+In order to run container that is using HDDL accelerator, _hddldaemon_ must
+ run on host machine. It's  required to set up environment 
+ (the OpenVINO package must be pre-installed) and start _hddldaemon_ on the
+  host before starting a container. Refer to the steps from [OpenVINO documentation](https://docs.openvinotoolkit.org/latest/_docs_install_guides_installing_openvino_docker_linux.html#build_docker_image_for_intel_vision_accelerator_design_with_intel_movidius_vpus).
+
+To start server with HDDL you can use command similar to:
+
+```
+docker run --rm -it --device=/dev/ion:/dev/ion -v /var/tmp:/var/tmp -v /opt/model:/opt/model -v /dev:/dev -e DEVICE=HDDL -p 9001:9001 \
```
dtrawins: Mounting /dev seems to be unnecessary. There is already /dev/ion character device mounted.

Dockerfile_binary_openvino
```diff
@@ -15,7 +15,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     python3-pip \
     python3-venv \
     usbutils \
-    virtualenv
+    virtualenv \
+    libboost-filesystem1.58.0 \
+    libboost-thread1.58.0
 
 ADD l_openvino_toolkit*.tgz $TEMP_DIR/
 RUN cd $TEMP_DIR/l_openvino_toolkit* && \
```
asutic: OV R2 publishes binaries as apt packages. It will simplify installation is we just go with that.

Dockerfile_binary_openvino
```diff
@@ -35,8 +37,9 @@ RUN cd $TEMP_DIR/l_openvino_toolkit* && \
     ./install.sh -s silent.cfg --ignore-signature && \
     rm -Rf $TEMP_DIR $INSTALL_DIR/install_dependencies $INSTALL_DIR/uninstall* /tmp/* $DL_INSTALL_DIR/documentation $DL_INSTALL_DIR/inference_engine/samples
 
+ENV HDDL_INSTALL_DIR="$DL_INSTALL_DIR/inference_engine/external/hddl"
 ENV PYTHONPATH="$INSTALL_DIR/python/python3.5"
-ENV LD_LIBRARY_PATH="$DL_INSTALL_DIR/inference_engine/external/tbb/lib:$DL_INSTALL_DIR/inference_engine/external/mkltiny_lnx/lib:$DL_INSTALL_DIR/inference_engine/lib/intel64"
+ENV LD_LIBRARY_PATH="$DL_INSTALL_DIR/inference_engine/external/tbb/lib:$DL_INSTALL_DIR/inference_engine/external/mkltiny_lnx/lib:$DL_INSTALL_DIR/inference_engine/external/hddl/lib:$DL_INSTALL_DIR/inference_engine/lib/intel64"
```
asutic: Simpler is to execute the setupvars.sh if possible in the container.

Dockerfile_binary_openvino
mzegla: Sounds good, but I think I could apply it later on, in separate pull request.

asutic: right, and it looks like apt is based on the open-source version, it doesn't have hddl support.

Dockerfile_binary_openvino
mzegla: It actually spoils some envs and imports. Perhaps we will investigate it in the future.

PR #88, opened 2019-09-03
Reshaping feature support

mzegla:


ie_serving/models/ir_engine.py
```diff
@@ -140,9 +146,98 @@ def set_keys(self, mapping_config):
...
+                reshape_required = True
+        return reshape_required, inputs_shapes
+
+    def reshape(self, reshape_param):
+        if type(reshape_param) is dict:
+            return self._reshape(reshape_param)
+        elif type(reshape_param) is int:
+            return self._change_batch_size(reshape_param)
+        else:
+            message = "Unknown error occurred in input reshape preparation"
```
dtrawins: why don't we catch the error from IE?

ie_serving/models/ir_engine.py
```diff
@@ -140,9 +146,98 @@ def set_keys(self, mapping_config):
...
+                             "required for input: {}. Current "
+                             "shape: {}. Required shape: {}"
+                             .format(self.model_name, self.model_version,
+                                     input_name, net_input_shape,
+                                     input_data.shape))
+                reshape_required = True
+        return reshape_required, inputs_shapes
+
+    def reshape(self, reshape_param):
+        if type(reshape_param) is dict:
```
dtrawins: could you also add in the debug mode duration time to the logs for batch size change and shape operation?

mzegla: It's currently done on service level:
https://github.com/IntelAI/OpenVINO-model-server/pull/88/files#diff-9ae4a243234a022f94664844406f2dabR100

ie_serving/models/ir_engine.py
mzegla: It's just a fail-safe exit here. At this point this `else `should never get executed as reshape_param shall be either dict with shapes or integer with new batch size.

ie_serving/server/predict_utils.py
```diff
@@ -24,31 +24,32 @@
...
+                    REST: falcon.HTTP_BAD_REQUEST},
 }
 
 
 def prepare_input_data(models, model_name, version, data, rest):
     model_inputs_in_input_request = list(dict(data).keys())
     input_keys = models[model_name].engines[version].input_key_names
     inference_input = {}
-    request_type = 'grpc' if not rest else 'rest'
+    request_type = GRPC if not rest else REST
```
mareklevv: `rest` variable should be renamed to `service_type`

ie_serving/server/predict_utils.py
```diff
@@ -61,32 +62,46 @@ def prepare_input_data(models, model_name, version, data, rest):
...
+            if models[model_name].engines[version].batching_info.mode != \
+                BatchingMode.AUTO and shape_required_in_model[0] != \
+                    tensor_input.shape[0]:
+                code = statusCodes['invalid_arg'][request_type]
+                message = INVALID_BATCHSIZE.format(
+                    tensor_input.shape[0],
+                    models[model_name].engines[
+                        version].batching_info.batch_size)
+                logger.debug("PREDICT error,Invalid batchsize:{}".format(message))
+                return True, message, code
```
mareklevv: We can delete first return value `True/False` and instead use `code` variable, if is it `None` keep going if not, return message and code

ie_serving/server/predict_utils.py
```diff
@@ -61,32 +62,46 @@ def prepare_input_data(models, model_name, version, data, rest):
                 message = str(e)
                 logger.debug("PREDICT prepare_input_data make_ndarray error: "
                              "{}".format(message))
-                return True, message, None, code
+                return True, message, code
         else:
             tensor_input = np.asarray(data[requested_input_blob])
+        # Validate shape if shape not in auto mode
+        if models[model_name].engines[version].shape_info.mode != ShapeMode.AUTO:
```
mareklevv: Ehhh I think it's time to pass to `prepare_input_data` only three variables :inference engine object, data, service_type.

After that change readable of our code will increase :)

ie_serving/models/shape_management/reshaper.py
```diff
@@ -0,0 +1,79 @@
...
+            self._prepare_error_response(error_message, response_context)
+            return True
+        duration = \
+            (reshape_end_time - reshape_start_time).total_seconds() * 1000
+        logger.debug(
+            "PREDICT; network reshape completed; {}; {}; {}ms".format(
+                engine.model_name, engine.model_version, duration))
+        return False
+
+    def _prepare_error_response(self, error_message, response_context):
```
mareklevv: I think it's wrong approach, `Reshaper` should focus only on reshaping, without changing request state.
It should only return codes and messages

ie_serving/models/shape_management/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    @classmethod
+    def build(cls, shape_param):
+        if shape_param is not None:
+            if shape_param == 'auto':
+                shape_mode = ShapeMode.AUTO
+            else:
+                shape_mode = ShapeMode.DEFAULT
+        else:
+            shape_mode = ShapeMode.DISABLED
+        return cls(shape_mode)
```
mareklevv: New line :)

ie_serving/models/shape_management/reshaper.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def detect_shapes_incompatibility(self, engine, inference_input):
+        # Compares workload shapes with engine inputs shapes. If different,
+        # returns True, reshape_param.
+        # reshape_param is inputs shapes dictionary (input_name:shape pairs)
+        # for reshapable models and batch size for non-reshapable
+
+        reshape_required, inputs_shapes = engine.scan_input_shapes(
+            inference_input)
+        reshape_param = inputs_shapes
```
mareklevv: Unnecessary assignment

ie_serving/models/ir_engine.py
```diff
@@ -140,9 +123,98 @@ def set_keys(self, mapping_config):
...
+        except Exception as e:
+            message = "Error occurred while loading network: {}".format(
+                str(e))
+            logger.debug("[Model: {}, version: {}] --- {}".format(
+                self.model_name, self.model_version, message))
+            return True, message
+        logger.debug("[Model: {}, version: {}] --- Network loaded "
+                     "successfully".format(self.model_name,
+                                           self.model_version))
+        return False, None
```
mareklevv: Like in `predict_utils.py` file, return only `message` or `None` should be enough

ie_serving/models/ir_engine.py
```diff
@@ -140,9 +123,98 @@ def set_keys(self, mapping_config):
...
+            inputs_shapes[input_name] = input_data.shape
+            if net_input_shape != input_data.shape:
+                logger.debug("[Model: {}, version: {}] --- Shape change "
+                             "required for input: {}. Current "
+                             "shape: {}. Required shape: {}"
+                             .format(self.model_name, self.model_version,
+                                     input_name, net_input_shape,
+                                     input_data.shape))
+                reshape_required = True
+        return reshape_required, inputs_shapes
```
mareklevv: The same situation, if inputs_shapes dict is empty then reshape is not required.

ie_serving/models/shape_management/reshaper.py
mzegla: Here I assume that model is reshapable and by default reshape parameter will be dict of new input shapes. 
Right below, if model is not reshapable, I assign batch_size to reshape_param, so reshape_param can be either dict of shapes or integer depending on shape mode.

ie_serving/main.py
```diff
@@ -62,20 +63,41 @@ def check_config_structure(configs):
         sys.exit()
 
 
+def get_model_spec(config):
+
+    model_name = config.get('name', config.get('model_name', None))
```
mareklevv: This lines are specific only for `one model` path which should be in `parse_one_model` function.

PR #96, opened 2019-09-18
Fixed shape

mzegla:


dtrawins:
new tests are skipped, fixed shape feature should be merged with those tests

PR #102, opened 2019-10-01
Multiple workers and asynchronous inference

mzegla:
Enabling multiple server workers for receiving and processing requests in parallel alongside with asynchronous processing for inference parallelism.


ie_serving/server/service.py
```diff
@@ -80,51 +85,41 @@ def Predict(self, request, context):
...
-            (serialization_end_time -
-             inference_end_time).total_seconds() * 1000
-        logger.debug("PREDICT; inference results serialization completed;"
-                     " {}; {}; {}ms".format(model_name, version, duration))
-
         return response
 
+    def wait_for_results(self, target_engine, inference_id, interval=0.001):
+        while True:
+            time.sleep(interval)
```
jpowie01: As Marek suggested, maybe using Event will be better and faster:
https://docs.python.org/3/library/threading.html#event-objects

Sleeping and checking for results in the loop may be a bottleneck or potential performance issue one day. Do a quick POC and check if it will works as well as current solution :)

ie_serving/server/start.py
```diff
@@ -72,10 +72,10 @@ def serve(models, max_workers: int=1, port: int=9000):
         sys.exit(0)
 
 
-def start_web_rest_server(models, rest_port):
+def start_web_rest_server(models, rest_port, num_threads):
     d = PathInfoDispatcher({'/': create_rest_api(models)})
     server = WSGIServer(('0.0.0.0', rest_port), d,
-                        numthreads=1, request_queue_size=50)
+                        numthreads=num_threads, request_queue_size=50)
```
jpowie01: Use constant `REQUESTS_QUEUE_SIZE` from `config.py`.

mzegla:
@jpowie01 
Sleep replaced with Events will be done in the future commits alongside with changes for rest service.

ie_serving/server/start.py
mzegla: No sure if `REQUESTS_QUEUE_SIZE `will be a good fit there. 
It defines max size of requests queue inside IR Engine. Both gRPC and Rest workers contribute to engine's requests queue. Also there might be multiple models and engines.
I suggest having another constant for queue sizes for servers with `REQUESTS_QUEUE_SIZE` only for IR Engines.

ie_serving/server/service.py
```diff
@@ -80,49 +85,25 @@ def Predict(self, request, context):
...
-                context.set_code(code)
-                context.set_details(error_message)
-                target_engine.in_use.release()
-                return predict_pb2.PredictResponse()
-        ################################################
-        inference_start_time = datetime.datetime.now()
-        inference_output, error_message = target_engine.infer(
-            inference_input)
-        if error_message is not None:
+        inference_id = str(uuid.uuid1())
```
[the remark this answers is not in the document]
mzegla: How about having a pool of IDs? 
If we specify max number of workers for the server, then we know that the number of requests handled in parallel would not exceed number of workers. We could generate that pools on OVMS start-up and then every worker that starts processing request would take free ID from the pool and give it back after it finishes instead of generating new ID per request. 
What do you think? Would it be safe or am I missing something?

ie_serving/server/service.py
```diff
@@ -80,49 +82,25 @@ def Predict(self, request, context):
...
+        target_engine.requests_queue.put(inference_request)
+        inference_output, used_ireq_index = inference_request.wait_for_result()
+        if inference_output is str:
             code = statusCodes['invalid_arg'][GRPC]
             context.set_code(code)
-            context.set_details(error_message)
-            target_engine.in_use.release()
+            context.set_details(inference_output)
+            logger.debug("PREDICT, problem during inference execution. Exit "
+                         "code {}".format(code))
```
jpowie01: You need to release your `ireq_index` back to the queue in order to not lose it in case of error:
```python
target_engine.free_ireq_index_queue.put(used_ireq_index)
```

ie_serving/config.py
```diff
@@ -16,22 +16,25 @@
 
 import os
 
-DEVICE = os.environ.get('DEVICE', "CPU")
```
mareklevv: Please update licence header ;)

ie_serving/config.py
```diff
@@ -16,22 +16,25 @@
...
+    'file_system_poll_wait_seconds': float(
+        os.getenv('FILE_SYSTEM_POLL_WAIT_SECONDS', 1)),
+    'mapping_config_filename': 'mapping_config.json',
+    'rest_requests_queue_size': os.getenv('REST_REQUESTS_QUEUE_SIZE', 100),
+    'engine_requests_queue_size': os.getenv('ENGINE_REQUESTS_QUEUE_SIZE',
+                                            None)
+}
 
 
 class S3_Config:
```
mareklevv: Maybe is a good idea to keep all configs in the same way, as dicts?

ie_serving/models/constants.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+class InferenceStatus:
```
mareklevv: This status container can be placed in ```ie_serving/models/__init__.py`` file

ie_serving/schemas.py
```diff
@@ -33,3 +33,34 @@
...
+            'type': 'object',
+            'required': ['config'],
+            'properties': {
+                'config': {
+                    'type': 'object',
+                    'required': ['name', 'base_path'],
+                    'properties': {
+                        'name': {'type': 'string'},
+                        'base_path': {'type': 'string'},
+                        'batch_size': {'type': ['integer', 'string']},
```
mareklevv: Batch size should be only `int` type

mareklevv: Please also update type of `batch_size` parameter for one model path

ie_serving/server/start.py
```diff
@@ -72,10 +72,12 @@ def serve(models, max_workers: int=1, port: int=9000):
         sys.exit(0)
 
 
-def start_web_rest_server(models, rest_port):
+def start_web_rest_server(models, rest_port, num_threads):
     d = PathInfoDispatcher({'/': create_rest_api(models)})
     server = WSGIServer(('0.0.0.0', rest_port), d,
-                        numthreads=1, request_queue_size=50)
+                        numthreads=num_threads,
```
mareklevv: Please check which parameter in that WSGIServer exactly mean how many request can be handle in one time.

ie_serving/schemas.py
mzegla: `batch_size` value can also be  "auto" for automatic batch size changes

ie_serving/server/start.py
mzegla: Looking at:
https://docs.cherrypy.org/projects/cheroot/en/latest/pkg/cheroot.wsgi.html#cheroot.wsgi.Server

numthreads - "Set minimum number of threads."
For maximum number of threads, by default, there's no limit.
Can't find any other parameter related to parallelism or multiple server threads.

PR #103, opened 2019-10-01
Multiple inputs sample

subashp:
This basic example shows how to input the frames from cameras
& disk files to inference engine running in OVMS.

Signed-off-by: Patel, Subash <subash.patel@intel.com>

### DIFF
--- a/example_client/multi_inputs.py
+++ b/example_client/multi_inputs.py
@@ -1,0 +1,290 @@
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Multi-threaded sample to run a RMNet & SSDMobilenet v2 that will
+# detect only person, bike and vehicle (change the output parsing
+# for more classes)
+#
+# Example usage:
+# RMNet: python3.6 multi_inputs.py -n "RMNet" -l "data" -o "detection_out"
+# -d 1024 -i 127.0.0.1 -p 9001 -c 1
+# -f /var/repos/github/sample-videos/person-bicycle-car-detection.mp4
+# SSDMobileNet: python3.6 multi_inputs.py -n "SSDMobileNet" -l "image_tensor"
+# -o "DetectionOutput" -d 300 -i 127.0.0.1 -p 9001 -c 1
+# -f /var/repos/github/sample-videos/person-bicycle-car-detection.mp4
+
+from __future__ import print_function
+from argparse import ArgumentParser, SUPPRESS
+from tensorflow_serving.apis import predict_pb2
+from tensorflow_serving.apis import prediction_service_pb2_grpc
+from time import time, sleep
+
+import sys
+import os
+import cv2
+import grpc
+import threading
+import numpy as np
+import logging as log
+import tensorflow.contrib.util as tf_contrib_util
+
+# global data (shared between threads & main)
+CLASSES = ["None", "Pedestrian", "Vehicle", "Bike", "Other"]
+COLORS = [(255, 255, 255), (255, 0, 0), (0, 255, 0), (0, 0, 255), (128, 128, 128)]
+SRC_TYPE = ["Camera", "Video"]
+
+exit_ok = False             # manage thread loop
+CAM_WIDTH = 640             # camera width
+CAM_HEIGHT = 480            # camera height
+CAM_FPS = 30                # camera speed
+CONFIDENCE_THRESHOLD = 0.75 # detection confidence
+
+#####################################################################################
+
+def build_argparser():
+  parser = ArgumentParser(add_help=False)
+  args = parser.add_argument_group('Options')
+  args.add_argument('-h', '--help', action='help', default=SUPPRESS,
+        help='Show this help message and exit.')
+  args.add_argument('-n', '--network_name', required=True,
+        type=str, help='Network name')
+  args.add_argument('-l', '--input_layer', required=True,
+        type=str, help='Input layer name')
+  args.add_argument('-o', '--output_layer', required=True,
+        type=str, help='Output layer name')
+  args.add_argument('-d', '--frame_size', required=True,
+        type=int, help='Input frame width and height that matches used model')
+  args.add_argument('-c', '--num_cameras', help='Number of cameras to be used',
+        required=False, type=int, default=1)
+  args.add_argument('-f', '--file', help='Path to the video file',
+        required=False, type=str)
+  args.add_argument('-i', '--ip', help='ip address of the ovms', required=True)
+  args.add_argument('-p', '--port', help='port of the ovms', required=True)
+
+  return parser
+
+# Decoding idea based on the link below. Not very accurate. So pls implement yours
+# https://github.com/opencv/open_model_zoo/blob/master/intel_models/\
+# person-vehicle-bike-detection-crossroad-0078/\
+# description/person-vehicle-bike-detection-crossroad-0078.md
+def parse_output(thr_id, res, frame):
+  for batch, data in enumerate(res):
+    pred = data[0]
+    for values in enumerate(pred):
+      # tuple
+      index = values[0]
+      l_pred = values[1]
+
+      # actual predictions
+      img_id = l_pred[0]
+      label = l_pred[1]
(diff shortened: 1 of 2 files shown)


example_client/multi_inputs.py
```diff
@@ -0,0 +1,279 @@
...
+      img_id = l_pred[0]
+      label = l_pred[1]
+      conf = l_pred[2]
+      x_min = l_pred[3]
+      y_min = l_pred[4]
+      x_max = l_pred[5]
+      y_max = l_pred[6]
+
+      # preventing any wrong array indexing (for RMNet)
+      if label >= 4:
```
mzegla: If `l_pred[1]` is 4, then you have` label = 4 `which I believe is okay, but then it gets into this conditional block, logs error and again assigns `label = 4`. It might be confusing. 
I suggest changing this condition to `if label > 4`

example_client/multi_inputs.py
```diff
@@ -0,0 +1,279 @@
...
+      # restart the video file when it reaches the end
+      if not ret:
+        cam.set(cv2.CAP_PROP_POS_FRAMES, 0)
+        continue
+      # normalize the video frame dimension to that of the camera
+      else:
+        # to maintain the frame inferencing parity with the cameras, lets sleep
+        # here to maintain cam_fps speed
+        sleep((1000 / cam_fps) / 1000)
+        frame = cv2.resize(frame, (cam_width, cam_height))
```
mzegla: Why do you resize here to `(cam_width, cam_height)` and few lines below to `(input_dimension, input_dimension) `?
Is it necessary to do both?

example_client/multi_inputs.py
```diff
@@ -0,0 +1,279 @@
...
+  args = parser.add_argument_group('Options')
+  args.add_argument('-h', '--help', action='help', default=SUPPRESS,
+        help='Show this help message and exit.')
+  args.add_argument('-n', '--network_name', required=True,
+        type=str, help='Network name')
+  args.add_argument('-l', '--input_layer', required=True,
+        type=str, help='Input layer name')
+  args.add_argument('-o', '--output_layer', required=True,
+        type=str, help='Output layer name')
+  args.add_argument('-d', '--input_dimension', required=True,
```
mzegla: Not sure about this naming. I was thinking about something like `--frame-size` with description in help like 
`"Input frame width and height that matches used model"`. 
I think "dimension" might be confused with numpy arrays shapes. What do you think?

example_client/multi_inputs.py
subashp: This is to resize the video frames read from the files.

example_client/multi_inputs.py
```diff
@@ -0,0 +1,279 @@
...
+import logging as log
+import tensorflow.contrib.util as tf_contrib_util
+
+# global data (shared between threads & main)
+CLASSES = ["None", "Pedestrian", "Vehicle", "Bike", "Other"]
+COLORS = [(255, 255, 255), (255, 0, 0), (0, 255, 0), (0, 0, 255), (128, 128, 128)]
+SRC_TYPE = ["Camera", "Video"]
+
+exit_ok = False         # manage thread loop
+cam_width = 640         # camera width
```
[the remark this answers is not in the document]
subashp: Done

example_client/multi_inputs.py
subashp: Done

example_client/multi_inputs.py
subashp: Done

example_client/multi_inputs.py
```diff
@@ -0,0 +1,280 @@
...
+      # restart the video file when it reaches the end
+      if not ret:
+        cam.set(cv2.CAP_PROP_POS_FRAMES, 0)
+        continue
+      # normalize the video frame dimension to that of the camera
+      else:
+        # to maintain the frame inferencing parity with the cameras, lets sleep
+        # here to maintain cam_fps speed
+        sleep((1000 / CAM_FPS) / 1000)
+        frame = cv2.resize(frame, (CAM_WIDTH, CAM_HEIGHT))
```
mzegla: I know I mentioned it before, but I'm still not sure if this statement will be effective as you have line 165:
`image = cv2.resize(frame, (input_dimension, input_dimension))`
So, here you resize your frame to (CAM_WIDTH, CAM_HEIGHT) and then again to (input_dimension, input_dimension).
Given that you don't use `frame` between those lines, it seems to me that this first resize is not really effective.

example_client/multi_inputs.py
```diff
@@ -0,0 +1,280 @@
+#!/usr/bin/env python
+#
+# Copyright 2019, Intel Corporation
```
mareklevv: Please update licence header. 
It isn't a full Apache licence header

example_client/multi_inputs.py
```diff
@@ -0,0 +1,280 @@
...
+# global data (shared between threads & main)
+CLASSES = ["None", "Pedestrian", "Vehicle", "Bike", "Other"]
+COLORS = [(255, 255, 255), (255, 0, 0), (0, 255, 0), (0, 0, 255), (128, 128, 128)]
+SRC_TYPE = ["Camera", "Video"]
+
+exit_ok = False             # manage thread loop
+CAM_WIDTH = 640             # camera width
+CAM_HEIGHT = 480            # camera height
+CAM_FPS = 30                # camera speed
+CONFIDENCE_THRESHOLD = 0.75 #detection confidence
```
mareklevv: Please add space after `#`

example_client/multi_inputs.py
```diff
@@ -0,0 +1,280 @@
...
+            video_files.append(r + f_)
+    elif os.path.isfile(vid_src):
+      if '.mp4' in vid_src:
+        video_files.append(vid_src)
+
+  # ovms connection
+  channel = grpc.insecure_channel("{}:{}".format(args.ip, args.port))
+  stub = prediction_service_pb2_grpc.PredictionServiceStub(channel)
+
+  # lock to manage the request_queue
```
mareklevv: GRPC library is thread safe, so this lock is unnecessary 
https://github.com/googleapis/google-cloud-python/issues/3272

mzegla: Yeah and I think it will not take advantage of multiple worker OVMS since only one thread will communicate with model server at a time.

example_client/multi_inputs.py
```diff
@@ -0,0 +1,280 @@
...
+          # only mp4 files supported as of now
+          if '.mp4' in f_:
+            video_files.append(r + f_)
+    elif os.path.isfile(vid_src):
+      if '.mp4' in vid_src:
+        video_files.append(vid_src)
+
+  # ovms connection
+  channel = grpc.insecure_channel("{}:{}".format(args.ip, args.port))
+  stub = prediction_service_pb2_grpc.PredictionServiceStub(channel)
```
mzegla: @mareklevv correct me if I'm wrong, but wouldn't it be better to have `stub` (or event `channel` as well) created per thread?
From: https://www.grpc.io/docs/guides/concepts/
"On the client side, the client has a local object known as stub (for some languages, the preferred term is client)"

I'm thinking of taking the most out of upcoming multiple worker release.

example_client/multi_inputs.py
subashp: There are two things that are happening here:
a) Network needs input in certain resolution - line 165 which you refer above
b) The results are displayed on windows using the inference results - In this case, camera is already asked to provide the input "frame" in CAM_WIDTH & CAM_HEIGHT dimension. However our example also supports the video files. To maintain the output window parity across the camera and video files, we do an extra resize when we read one frame from the video file. This is "frame" which will be used to show the result.

This is the reason why two resizes exist. It does have effect. If you comment out line 153 above, the output frame from the video file would be displayed with whatever dimension it has been read. I will remove it for your satisfaction although the output window looks inconsistent(esp on small monitors)

example_client/multi_inputs.py
```diff
@@ -0,0 +1,279 @@
...
+      x_max = l_pred[5]
+      y_max = l_pred[6]
+
+      # preventing any wrong array indexing (for RMNet)
+      if label >= 4:
+        log.error('Unsupported class detected in camera {}'.format(thr_id))
+        label = 4
+
+      # Do you want confidence level to be passed from command line?
+      if img_id != -1 and conf >= 0.75:
```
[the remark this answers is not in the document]
subashp: Done

example_client/multi_inputs.py
subashp: Done

example_client/multi_inputs.py
subashp: Done

example_client/multi_inputs.py
subashp: Good point. Removed

example_client/multi_inputs.py
subashp: makes sense. modified code to create local stub per thread

subashp:
I was tied up with other work and hence couldnt rework earlier. I have taken care of all comments. Pls let me know if you need any other changes to be made.

PR #109, opened 2019-10-25
R3 - readme update

mzegla:


mzegla:
Moved to:
https://github.com/IntelAI/OpenVINO-model-server/pull/111

PR #111, opened 2019-10-29
Readme updates

dtrawins:


docs/performance_tuning.md
```diff
@@ -0,0 +1,105 @@
...
+(**grpc_wokers >= nireq** and **rest_workers >= nireq**).
+
+**Note**: When using CPU plugin, the vertical scalability (adding mode cores to the OVMS instnace) is a bit less efficient 
+then horizontal (adding more smaller OVMS instances). Add more instances if you need to maximize the throughput and you
+don't have restrictions of the number of instances and their RAM usage.
+
+### HDDL accelerators
+
+A single worker on OpenVINO Model Server could use only a single VPU (Vision Processing Unit) from HDDL accelerators at a time.
+Multiple workers can increase the throughput. With parallel handling of the requests, it is possible to utilize all available VPUs.
```
mzegla: I don't know how detailed we want our explanation here, but it's more like all VPUs are utilized either way, but with single worker they're utilized very poorly (in my experiments it was less than 15% utilization on each VPU).
Multiple worker setup also utilizes all VPUs, but it's far more effecient (more than 95% utilization per VPU). 
So, it might be a little confusing to say _"A single worker on OpenVINO Model Server could use only a single VPU"_. 
I would understand it like for card with 8x VPUs, with single worker OVMS I get 1x VPU fully utilized and the other 7x VPUs not used at all.

mzegla:
I would be nice to add new configuration options in example here:
https://github.com/IntelAI/OpenVINO-model-server/blob/readme-updates/docs/docker_container.md#starting-docker-container-with-a-configuration-file

It would explicitly show that user can define number of infer request, plugin configuration and target device for each model independently and for example run multiple models on multiple devices with one OVMS instance.

docs/docker_container.md
```diff
@@ -256,7 +255,9 @@ It uses `json` format as shown in the example below:
          "config":{
              "name":"model_name5",
              "base_path":"s3://bucket/models/model5",
-             "shape": "auto"
+             "shape": "auto",
+             "target_device": "HDDL",
+             "plugin_config": {"CPU_THROUGHPUT_STREAMS": "CPU_THROUGHPUT_AUTO"}
```
mzegla: I know it's just and example, but to avoid confusion please move one of these parameters to another model's config.
`CPU_*` params won't work with HDDL as a target device.

mzegla: Adding `nireq` would also be cool 😄

PR #136, opened 2020-02-10
Fix getting model version based on model name parsing.

PatrykMatyjasek:


mzegla:
Please add similar scenarios to S3 and GS methods unit tests

PR #137, opened 2020-02-10
Added support for OV binary package 2020.1

joannanosek:
* Added backward compatibility to cpu extensions

### DIFF
--- a/ie_serving/config.py
+++ b/ie_serving/config.py
@@ -17,10 +17,10 @@
 import os
 
 DEFAULT_CPU_EXTENSION_PATHS = ["/usr/local/lib/libcpu_extension.so",
-                              "/usr/lib64/libcpu_extension.so",
-                              "/opt/intel/openvino/deployment_tools/"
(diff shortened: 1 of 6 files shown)


ie_serving/config.py
```diff
@@ -16,12 +16,16 @@
 
 import os
 
-if os.path.isfile("/usr/local/lib/libcpu_extension.so"):
-    default_cpu_extension = "/usr/local/lib/libcpu_extension.so"
+LOCAL_LIB_CPU_EXTENSION = "/usr/local/lib/libcpu_extension.so"
```
dtrawins: I would suggest to use a list of paths in a variable DEFAULT_CPU_EXTENSION_PATHS. It would include values depending on the package and distribution:
/usr/local/lib/libcpu_extension.so
inference_engine/lib/intel64/libcpu_extension_avx2.so
/usr/lib64/libcpu_extension.so
With that, dockerfile for clearlinux could be simplified - ENV dropped

PR #140, opened 2020-02-13
reduce number of docker image flavors

dtrawins:


docs/docker_container.md
```diff
@@ -22,10 +21,6 @@ or
 ```bash
 make docker_build_clearlinux http_proxy=$http_proxy https_proxy=$https_proxy
 ```
-or
-```bash
-make docker_build_src_intelpython http_proxy=$http_proxy https_proxy=$https_proxy
-```
 
 **Note:** You can use also publicly available docker image from [dockerhub](https://hub.docker.com/r/intelaipg/openvino-model-server/)
 based on intelpython base image.
```
mzegla: based on **intelpython** base image.

Guess this one is outdated

joannanosek:
What about examples? 
This notebook uses dropped Docker image https://github.com/IntelAI/OpenVINO-model-server/blob/master/example_client/REST_age_gender.ipynb
Also here there's intelpython reference https://github.com/IntelAI/OpenVINO-model-server/blob/multiprocessing-poc/example_kubeflow_pipelines/openvino/container/Dockerfile

PR #141, opened 2020-02-14
Initial version of internal ZMQ messaging protocol

mzegla:
Setting up initial version of inter-process communication protocol.


ie_serving/messaging/zmq_messages.proto
```diff
@@ -0,0 +1,58 @@
...
+message EndpointResponse{
+    oneof type{
+        PredictResponse predict_response = 1;
+        ModelMetadataResponse model_metadata_response = 2;
+        ModelStatusResponse model_status_response = 3;
+        ErrorResponse error_response = 4;
+    }
+}
+
+message PredictRequest{
```
dtrawins: We need to include the socket name for the callback function

ie_serving/messaging/zmq_messages.proto
```diff
@@ -0,0 +1,58 @@
...
+    message Data{
+        string input_name = 1;
+        string shm_name = 2;
+    }
+    repeated Data inputs = 1;
+}
+
+message PredictResponse{
+    message Data{
+        string output_name = 1;
```
dtrawins: Is there a need to pass info about input type to read it correctly? list numpy or string or binary?

ie_serving/messaging/zmq_messages.proto
```diff
@@ -0,0 +1,58 @@
...
+    }
+    repeated Data outputs = 1;
+}
+
+message ModelMetadataRequest{
+    bool metadata_field = 1;
+}
+
+message ModelMetadataResponse{
+    //TODO
```
dtrawins: list of inputs and outputs, both including tensorinfo (name,shape(list of ints),type)

ie_serving/messaging/zmq_messages.proto
```diff
@@ -0,0 +1,58 @@
...
+}
+
+message ModelStatusRequest{
+    bool status_field = 1;
+}
+
+message ModelStatusResponse{
+    //TODO
```
dtrawins: just string should be fine

PR #143, opened 2020-02-19
Initial abstraction layer for engine pluginability

PatrykMatyjasek:
With basic stub implementation for OpenVINO


dtrawins:
Few comments:
We need a method to load the model including downloading from external storage.
There should be callback function included, which return prediction
Instead of run infetence there should be submit_inference - it will be async mode

mzegla:
About inference function name, in https://github.com/IntelAI/OpenVINO-model-server/pull/148/files#diff-a7b6e12a068292328e01647cdafc8dc2R122
I assumed the name `predict` as we call this endpoint `Predict`. Perhaps this name would be okay?

I don't think inference callback function is in the scope of the abstraction layer. Callbacks will be used by OpenVINO backend and are backend specific. I believe that in Engine class we should just call `predict/run_inference/submit_inference` and let the underlying backend do the job itself, whether it's asynchronous OpenVINO or synchronous Tensorflow inference.

PR #146, opened 2020-02-20
[2020R2] Start new processes for grpc server and models

joannanosek:


ie_serving/models/gs_model.py
```diff
@@ -123,13 +124,14 @@ def get_engine_for_version(cls, model_name, version_attributes):
         logger.info('Downloaded files from GCS')
 
         engine_spec = cls._get_engine_spec(model_name, version_attributes)
-        engine = IrEngine.build(**engine_spec)
+        engine_process = multiprocessing.Process(target=IrEngine.build,
+                                                 args=engine_spec)
```
mzegla: The engine process must start somewhere. No matter where you decide to call `start()` on this handler you can't call `cls.delete_local_mirror(...)` like you did below. It would cause files deletion before engine process can properly load them. 
Consider moving files download and deletion to engine process.

ie_serving/models/s3_model.py
```diff
@@ -132,13 +133,13 @@ def get_engine_for_version(cls, model_name, version_attributes):
         logger.info('Downloaded files from S3')
 
         engine_spec = cls._get_engine_spec(model_name, version_attributes)
-        engine = IrEngine.build(**engine_spec)
-
+        engine_process = multiprocessing.Process(target=IrEngine.build,
+                                                 args=engine_spec)
```
mzegla: same as for gs

ie_serving/main.py
```diff
@@ -145,16 +144,20 @@ def parse_one_model(args):
...
         process_thread.setDaemon(True)
         process_thread.start()
-    start_server(models=models, max_workers=args.grpc_workers,
-                 port=args.port)
+
+    num_servers = int(os.getenv("NUM_SERVERS", 1))
+    server_processes_handlers = []
+    for i in range(num_servers):
+        server_process_handler = multiprocessing.Process(
+            target=start_server, args=(models, args.grpc_workers, args.port))
```
mzegla: Server processes will not be aware of the models. So variable models among start_server arguments is unnecessary.

ie_serving/models/local_model.py
```diff
@@ -52,7 +54,9 @@ def _get_mapping_config(cls, version):
         return None
 
     @classmethod
-    def get_engine_for_version(cls, model_name, version_attributes):
+    def get_engine_process_for_version(cls, model_name, version_attributes):
         engine_spec = cls._get_engine_spec(model_name, version_attributes)
-        engine = IrEngine.build(**engine_spec)
-        return engine
+        engine_process = multiprocessing.Process(target=IrEngine.build,
```
mzegla: Can we use `_start_engine_process_for_version` here as well to make it more consistent? (this one suggests that only model building will happen in this process)
I can see you made `_start_engine_process_for_version` for remote storage only (as it contains downloaded files cleanup). Can we have single entry point i.e. pass `_start_engine_process_for_version` for every model subclass?

requirements.txt
```diff
@@ -6,3 +6,4 @@ boto3==1.10.2
 jsonschema==3.1.1
 falcon==2.0.0
 cheroot==8.2.1
+pyzmq==18.1.1
```
mzegla: We probably gonna need to change grpcio package installation. 
To make it possible for gRPC Server to use SO_REUSE_PORT option you might need to install it with `--no-binary` option like:
`grpcio==1.25.0 --no-binary grpcio`
Unless it's available by default in latest grpcio release. Please check that out.

joannanosek: It seems this issue is still open https://github.com/grpc/grpc/issues/18210 and is addresed in also still open PR https://github.com/grpc/grpc/pull/21854
I will update grpcio to 1.27.1 in `requirements.txt` and add workaround `grpcio==1.27.1 --no-binary grpcio --force-reinstall` in dockerfiles until there's a release with fixed `SO_REUSEPORT` flag in grpcio

PR #151, opened 2020-03-02
Use Jenkins, rewrite tests

kromanow94:


Jenkinsfile
```diff
@@ -0,0 +1,24 @@
+pipeline {
```
dtrawins: I would prefer the stages to be triggered by make command like is used to be. 
preparing virtualenv shouldn't be a separate stage. I would see it more like a dependency for other stages.
I suggest to configure separate jobs and statuses for independent tasks like style, unit tests, functional tests. This way in github statues there would be better visibility which task failed.

tests/scripts/unit-tests.sh
```diff
@@ -0,0 +1,35 @@
...
+    -t ${OVMS_TESTS_IMAGE} \
+    -f `dirname $0`/Dockerfile.openvino \
+    --build-arg OPEN_VINO_DOCKER_IMAGE=${OPEN_VINO_DOCKER_IMAGE} \
+    `mktemp -d`
+
+# Check python version in OV Docker Image
+PYTHON_VERSION=`docker run --rm $OVMS_TESTS_IMAGE \
+    /usr/bin/env python3 --version | awk -F"[ .]" '{ print $2"."$3 }'`
+
+PYTHONPATH=/opt/intel/openvino/python/python${PYTHON_VERSION}
```
dtrawins: source /opt/intel/openvino/bin/setupvars.sh is the recommended way to set envs now

joannanosek:
Is this for closing?

PR #152, opened 2020-03-03
Update paths for gs bucket

joannanosek:


dtrawins:
@joannanosek 
also tests need to be adjusted to use different model https://github.com/IntelAI/OpenVINO-model-server/blob/master/tests/functional/test_single_model_gc.py#L31 - replace input name to data and output name to prob. The same in test test_get_model_metadata. In test test_get_model_status change the expected version to 1. Also change the fixture https://github.com/IntelAI/OpenVINO-model-server/blob/master/tests/functional/fixtures/server_remote_models_fixtures.py#L24 to use new model path. Input and output names should be changed in https://github.com/IntelAI/OpenVINO-model-server/blob/master/tests/functional/test_multi_models.py#L76 
Finally merge the latest changes from jenkins-staging to test whole CI process.

PR #154, opened 2020-03-05
Switching to Jenkins CI and test models from OpenVINO model zoo

kromanow94:


tests/functional/utils/model_management.py
```diff
@@ -34,6 +36,9 @@ def wait_endpoint_setup(container):
         except Exception as e:
             time.sleep(1)
     print("Logs from container: ", logs)
+    # NOTE(kromanow): this sleep was added to check if occasionall fail on
```
dtrawins: this comment sounds like work in process. are the checks concluded.
there is not need to add the note author ;-)

tests/scripts/functional-tests-apt-ubuntu.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -ex
+
+TEST_DIRS=tests
+
+make docker_build_apt_ubuntu
+
+. .venv-jenkins/bin/activate
+
+py.test ${TEST_DIRS}/functional/ -v --test_dir=/var/jenkins_home/test_models_mzeglars
```
dtrawins: avoid folder name with login name . how about /var/jenkins_home/test-ovms-models ?

tests/scripts/functional-tests-bin.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -ex
+
+TEST_DIRS=tests
+
+make docker_build_bin dldt_package_url=${OPENVINO_DOWNLOAD_LINK_2020_1}
+
+. .venv-jenkins/bin/activate
+
+py.test ${TEST_DIRS}/functional/ -v --test_dir=/var/jenkins_home/test_models_mzeglars
```
dtrawins: avoid folder name with login name . how about /var/jenkins_home/test-ovms-models ?

tests/scripts/functional-tests-clearlinux.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -ex
+
+TEST_DIRS=tests
+
+make docker_build_clearlinux
+
+. .venv-jenkins/bin/activate
+
+py.test ${TEST_DIRS}/functional/ -v --test_dir=/var/jenkins_home/test_models_mzeglars
```
dtrawins: avoid folder name with login name . how about /var/jenkins_home/test-ovms-models ?

Makefile
```diff
@@ -64,7 +65,7 @@ coverage: venv
 
 test: venv
 	@echo "Executing functional tests..."
-	@. $(ACTIVATE); py.test $(TEST_DIRS)/functional/
+	@. $(ACTIVATE); py.test $(TEST_DIRS)/functional/ --test_dir $(TEST_MODELS_DIR)
 
 test_local_only: venv
 	@echo "Executing functional tests with only local models..."
```
dtrawins: now also those test can be added as local only:
tests/functional/test_model_version_policy.py
tests/functional/test_model_versions_handling.py
tests/functional/test_reshaping.py

Makefile
```diff
@@ -105,7 +106,7 @@ docker_build_clearlinux:
 	@echo "Building docker image"
 	@echo OpenVINO Model Server version: $(OVMS_VERSION) > version
 	@echo Git commit: `git rev-parse HEAD` >> version
-	@echo OpenVINO version: 2019_R2 clearlinux >> version
+	@echo OpenVINO version: 2020_R1 clearlinux >> version
```
dtrawins: that will be the case  shortly but for now it is 2019_R3

tests/functional/utils/model_management.py
kromanow94: I consider this as a good practice to add the note author. I've seen this in many repositories.
About this sleep: this resolved previously occurring problem.
This comment don't necessary state about WIP but about known problem that we're willing to fix in future. Having sleep for a few seconds is workaround, it would be better to implement mechanism ensuring server readiness.
I vote to leave it. Or if not at least create a JIRA task.

tests/scripts/functional-tests-apt-ubuntu.sh
kromanow94: :+1: 
Beside that in future we're going to move this to Makefile anyway so this will be in a variable.
Also lets keep to the syntax and use underscores.

Makefile
kromanow94: We should parametrize it anyway.
Also there are some steps repeating between docker builds in Makefile so we could make them in separate target and add as dependency.

tests/functional/fixtures/model_conversion_fixtures.py
```diff
@@ -0,0 +1,61 @@
...
+    with open(target_file, 'wb') as output:
+        output.write(response.content)
+
+    return target_file
+
+
+@pytest.fixture(autouse=True, scope="session")
+def resnet_multiple_batch_sizes(get_test_dir, get_docker_context):
+    tensorflow_model = \
+        download_file('https://download.01.org/opencv/public_models/012020/resnet-50-tf/resnet_v1-50.pb', # noqa
```
kromanow94: Lets parametrize this.

tests/functional/test_mapping.py
```diff
@@ -102,45 +96,43 @@ def test_run_inference_rest(self, resnet_2_out_model_downloader,
...
-            print("output shape", output['mask'].shape)
-            print("output shape", output['output'].shape)
-            assert output['mask'].shape == (1, 2048, 7, 7), ERROR_SHAPE
-            assert output['output'].shape == (1, 2048, 7, 7), ERROR_SHAPE
-
-    def test_get_model_metadata_rest(self, resnet_2_out_model_downloader,
+        print("Downloaded model files:", age_gender_model_downloader)
+
+        imgs_v1_224 = np.ones((1, 3, 62, 62))
+        rest_url = 'http://localhost:5556/v1/models/age_gender:predict'
```
kromanow94: Is rest url never going to be changed? Maybe some parametrization?

PR #157, opened 2020-03-10
Fix single model functional tests to use public models

dkalinowski:
Fix following test suites:
- test_batching.py
- test_mapping.py
- test_single_model.py

### DIFF
--- a/tests/functional/fixtures/server_with_batching_fixtures.py
+++ b/tests/functional/fixtures/server_with_batching_fixtures.py
@@ -44,6 +44,32 @@ def start_server_batch_model(request, get_image, get_test_dir,
 
 
 @pytest.fixture(scope="class")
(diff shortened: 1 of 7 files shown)


tests/functional/fixtures/server_with_batching_fixtures.py
```diff
@@ -91,3 +143,29 @@ def start_server_batch_model_bs4(request, get_image, get_test_dir,
     assert running is True, "docker container was not started successfully"
 
     return container
+
+
+@pytest.fixture(scope="class")
+def start_server_batch_model_auto_bs4_2out(request, get_image, get_test_dir,
```
mzegla: We set batch size to 4 in this fixture. "auto" in name might be a little confusing.

PR #159, opened 2020-03-11
Parallel CI pipeline

joannanosek:


tests/functional/utils/parametrization.py
```diff
@@ -0,0 +1,29 @@
...
+        try:
+            sock.bind(('', min_port))
+            sock.close()
+            return min_port
+        except OSError:
+            min_port += 1
+    raise IOError('no free ports')
+
+
+def get_ports_for_fixture():
```
dtrawins: I think it might be sensitive to hazards when triggered in parallel. There is a risk that different test will get the same port assigned. Maybe more reliable would be assigning different last 2 digits for each test and 2 digits for each image target?
This verification could be handy to avoid conflicts with permanently occupied ports on the host running the tests.

joannanosek: This is why I made some randomization with seed but I agree that this not fully prevents us from potential doubled assignment. Your proposition sounds nice.

PR #164, opened 2020-03-16
OVMS image with OpenVINO base image

dtrawins:


Dockerfile_openvino_base
```diff
@@ -0,0 +1,22 @@
+FROM openvino/ubuntu18_runtime
+USER root
+RUN apt-get update && apt-get install -y python3-dev
+RUN pip3 install virtualenv
+WORKDIR /ie-serving-py
+RUN chown openvino .
+USER openvino
+COPY requirements.txt /ie-serving-py/
+RUN virtualenv -p python3 .venv && \
```
kromanow94: Maybe we should start using `python3 -m venv`? It has better support, its built into python3 in most distros (not Ubuntu, still need to install `python3-vevn`) and it's suggested to use with Python3.3+.

Dockerfile_openvino_base
```diff
@@ -0,0 +1,22 @@
...
+
+RUN . .venv/bin/activate && pip3 install .
+ENV PYTHONPATH=/opt/intel/openvino/python/python3.6:/opt/intel/openvino/python/python3
+ARG OV_PATH=/opt/intel/openvino/deployment_tools
+ENV LD_LIBRARY_PATH=\
+$OV_PATH/ngraph/lib:\
+$OV_PATH/inference_engine/external/gna/lib:\
+$OV_PATH/inference_engine/external/mkltiny_lnx/lib:\
+$OV_PATH/inference_engine/external/tbb/lib:\
+$OV_PATH/inference_engine/lib/intel64
```
kromanow94: This should be formatted to human-readable indents or use += 
Also end with newline.

PR #167, opened 2020-03-18
Minio S3 setup for tests.

rasapala:
Minio S3 setup for tests.

### DIFF
--- a/tests/functional/utils/model_management.py
+++ b/tests/functional/utils/model_management.py
@@ -21,42 +21,34 @@ import time
 from pathlib import Path
 
 
-def wait_endpoint_setup(container):
+def minio_condition(container):
+    if "created" in container.status:
+        return True
+    return False
+
+
+def serving_condition(container):
+    logs = str(container.logs())
+    if "server listens on port" in logs:
+        return True
+    return False
+
+
+def wait_endpoint_setup(container, condition=serving_condition, timeout=900):
     start_time = time.time()
     tick = start_time
     running = False
-    logs = ""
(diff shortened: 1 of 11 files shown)


tests/functional/utils/model_management.py
```diff
@@ -40,6 +40,23 @@ def wait_endpoint_setup(container):
     time.sleep(2)
     return running
 
+def wait_minio_endpoint_setup(container):
```
joannanosek: This function and `wait_endpoint_setup` from this file could be merged to one with a parameter as a message we want to have in logs.

rasapala: wait_endpoint_setup checks for specific line in logs and has a big 900 timeout while wait_minio_endpoint_setup checks container status and has smaller timeout thus I would prefer to leave it as separate functions.

joannanosek: So the timeout can be parametrized too.
As long as `if` is checking for boolean value, still the conditions as well as timeout can be passed as parameters, so the signature will looks like`def wait_endpoint_setup(condition, timeout)`.

rasapala: Done.

tests/functional/utils/model_management.py
```diff
@@ -21,21 +21,32 @@
 from pathlib import Path
 
 
-def wait_endpoint_setup(container):
+def minio_condition(container):
+    if "created" in container.status:
```
joannanosek: `return "created" in container.status` will be enough, the same for `serving_condition`

PR #182, opened 2020-04-09
Improve error handling and cleanup containers for model conversion

joannanosek:


tests/functional/utils/cleanup.py
```diff
@@ -49,19 +49,21 @@ def kill_container(container):
...
-        else:
-            raise
+        raise_or_pass(e)
+
+
+def raise_or_pass(docker_error):
+    # Such error can occur if container
+    # is being killed or removed by other entity
+    # Container is not running or No such container
+    if docker_error.status_code == 409 or docker_error.status_code == 404:
```
kromanow94, line 66: It'd be better to have a map of allowed error codes. Something like code=>description. If code in map, print description. Else - raise.

tests/functional/utils/cleanup.py
```diff
@@ -49,19 +49,21 @@ def kill_container(container):
     try:
         container.kill()
     except APIError as e:
-        # Conflict: Container is not running
-        if e.status_code == 409:
-            pass
-        else:
-            raise
+        raise_or_pass(e)
```
kromanow94, line 52: I'd prefer the name "handle_cleanup_exception".

PR #190, opened 2020-04-22
Preprocess binary input images

mateusz-ciesielski:
Adding a simple function for preprocessing binary images.
Before running tests, run `get_test_images.sh` script.


example_ams/ams/preprocess_binary_images.py
```diff
@@ -0,0 +1,82 @@
...
+#
+
+
+from typing import Tuple
+
+import tensorflow as tf
+import numpy as np
+
+
+def preprocess_binary_image(image: bytes, channels: int = None,
```
adam-marek, line 24: Let's not propagate the underlying exceptions out of this function. It's supposed to provide an abstraction for the pre-processing functionality however currently it leaks the details of the underlying implementation by propagating the exceptions from the specific library used under the hood - if that library changes, the surrounding code that uses that function will also need to change. Instead let's identify the possible failure modes for the preprocessing functionality, define them as possible status codes in the contract of that function and then make sure that the errors reported by the underlying library are correctly interpreted and reported.

mateusz-ciesielski: Only one exception may be raised by this function, as it is written in the docstring below: ValueError. Therefore, no library specific exceptions have to be expected by the caller, and potential underlying library changes will have no effect on the code that is using this function. There is a clear contract between this function and its callers - caller has to only expect ValueError - which will contain all the information about the actual cause of error, so no information will be lost. Moreover, ValueError is widely used for such purposes and it is a recommended way to indicate an error which was probably caused by invalid value passed to the function - in this case, we expect that errors will be caused by invalid value of passed image.

If you are not familiar with the concept of exception chaining, see: https://www.python.org/dev/peps/pep-3134

adam-marek: Thanks, I'm familiar with the concept of exception chaining. And the concern here **is not possible loss of information**. The point here is that ValueError semantics (specifically the circumstances under which the exception can be thrown) will rely on the specific underlying implementation of the libraries used and anyone using that function who wants to communicate meaningful errors to their users will need to refer to that underlying implementation to understand which situations may result in exceptions and how to meaningfully tell one category of problems from another (data error, format error, parameter error, etc.). This is shifting the responsibility of understanding the possible failure modes in that function from its author to the users of that function which is not good encapsulation.

mateusz-ciesielski: I don't agree that providing caller with full information about the cause of failure is good encapsulation - it is widely considered as good practice to provide caller with possibility to look at full stack trace in case of error. Although, it is a good point that more granular information about the cause of error would be beneficial. So, my proposition is to handle it in following manner:
```python
try:
        decoded_image = tf.io.decode_image(image, channels=channels, dtype=dtype)
except Exception as e:
        raise ValueError('Failed to decode provided binary image.') from e

try:
        if target_size:
            height, width = target_size
            decoded_image = tf.image.resize_with_crop_or_pad(decoded_image, height, width)
except Exception as e:
      raise ValueError('Failed to resize provided binary image') from e
 (...)
```
That way caller would get more detailed information about the cause of error, without the need to create a new abstraction around returning statuses instead of raising proper exceptions. What do you think about this?

mateusz-ciesielski: By the way, please look at the documentation of https://www.tensorflow.org/api_docs/python/tf/io/decode_image - it defines exactly the same contract as in my solution - `ValueError` is the only exception that should be expected from this function.

extras/ams_wrapper/src/preprocessing/preprocess_image.py
```diff
@@ -0,0 +1,107 @@
...
+            height, width = target_size
+            decoded_image = tf.image.resize_with_crop_or_pad(decoded_image, height, width)
+        except Exception as e:
+            raise ImageResizeError('Failed to resize provided binary image from: {} '
+                                   'to: {}.'.format(tf.shape(decoded_image), target_size)) from e
+
+    try:
+        image_array = decoded_image.numpy()
+        if standardization:
+            decoded_image = tf.image.per_image_standardization(decoded_image)
```
mzegla, line 78: Does this call have an effect?
 `decoded_image` is not used anymore after this call and at the end of the function, the return value will be `image_array`.
Will standardization be applied to the returned array?

extras/ams_wrapper/src/preprocessing/preprocess_image.py
```diff
@@ -0,0 +1,107 @@
...
+        if reverse_input_channels:
+            image_array = image_array[..., ::-1]
+        if channels_first:
+            image_array = np.transpose(image_array, [2, 0, 1])
+        if scale:
+            image_array = image_array * scale
+    except Exception as e:
+        raise ImagePreprocessError('Failed to preprocess binary image, '
```
mzegla, line 86: I'm not sure about this error name. Above we have ImageResizeError. If we consider resizing as part of preprocessing then maybe this error name could be more specific. 

Also the part _"Failed to preprocess **binary** image..."_ in error message indicates working with binary format of the image while we start this `try...except...` block with `image_array = decoded_image.numpy()` and `decoded_image` is `Tensor` object.

extras/ams_wrapper/src/preprocessing/preprocess_image.py
```diff
@@ -0,0 +1,107 @@
...
+                                   'check if provided parameters are correct.') from e
+
+    return image_array
+
+
+
+if __name__ == "__main__":
+    img_path = '<path to the image>'
```
mzegla, line 94: Could we read image path from command line? It would be easier to use this function as standalone script.

extras/ams_wrapper/src/preprocessing/preprocess_image.py
mateusz-ciesielski: Nope, it's my mistake, I'll fix that

extras/ams_wrapper/src/preprocessing/preprocess_image.py
mateusz-ciesielski: Sure, it would be better way to get this from command line, I'll change that

extras/ams_wrapper/src/preprocessing/preprocess_image.py
mateusz-ciesielski: I'll think about more specific name, it is true that it may be a little bit too generic

extras/ams_wrapper/src/preprocessing/preprocess_image.py
```diff
@@ -0,0 +1,136 @@
...
+            height, width = target_size
+            decoded_image = tf.image.resize_with_crop_or_pad(decoded_image, height, width)
+        except Exception as e:
+            raise ImageResizeError('Failed to resize provided binary image from: {} '
+                                   'to: {}.'.format(tf.shape(decoded_image), target_size)) from e
+
+    try:
+        if standardization:
+            decoded_image = tf.image.per_image_standardization(decoded_image)
+        image_array = decoded_image.numpy()
```
rasapala, line 101: Dostaję tutaj:
    image_array = decoded_image.numpy()                 
AttributeError: 'Tensor' object has no attribute 'numpy'

extras/ams_wrapper/src/preprocessing/preprocess_image.py
```diff
@@ -0,0 +1,136 @@
...
+        if reverse_input_channels:
+            image_array = image_array[..., ::-1]
+        if channels_first:
+            image_array = np.transpose(image_array, [2, 0, 1])
+        if scale:
+            image_array = image_array * scale
+    except Exception as e:
+        raise ImageTransformError('Failed to preprocess image, '
```
rasapala, line 109: Need to print oryginal e exception mesage sothat user knows what is going on.

extras/ams_wrapper/src/preprocessing/preprocess_image.py
rasapala: Works with TF 2.0++

PR #191, opened 2020-04-23
CVS-30106: kubernetization, deployment docs

pskindel:
Whoever is going to merge that - please **squash** commits before merge.


deploy/ovms/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        env:
+        - name: "GOOGLE_APPLICATION_CREDENTIALS"
+          value: /secret/gcp-creds.json
+        command: ["/ie-serving-py/start_server.sh"]
+        args: ["ie_serving", "model",
+               "--model_path", "gs://models-repository",
+               "--model_name", "icnet-camvid-ava-0001",
+               "--port", "8080",
```
jcchr, line 36: I propose to add the model name as a parameter in the "values.yaml" file - as right now this deployment can serve this one model only. Please add also an information how to pass this name while starting the deployment - in a similar manner to a description of passing a number of replicas.

deploy/ovms/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+               "--rest_port", "8081",
+               "--batch_size", "auto"]
```
mzegla, line 38: I would remove batch_size parameter and keep it default. I think it would be better for generic example as not all models behave as expected with batch_size set to auto. For example as far as I know SSD object detection models batch size shall be done via reshape method.

deploy/README.md
```diff
@@ -0,0 +1,158 @@
...
+```shell script
+gsutil mb gs://model-repository
+```
+
+You can download the model from the link provided above nad upload it to GCS:
+```shell script
+gsutil cp -r 1 gs://model-repository/1
+```
+
+## Bucket permissions
```
dtrawins, line 31: This section could be simplified if both the cluster and bucket is in the same GCP account. In that case credentials are not required. Just storage API on the cluster level.

deploy/ovms/templates/deployment.yaml
ryanloney: Yes, please make sure we provide instructions for multiple models.

deploy/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+Please refer to: https://helm.sh/docs/intro/install for Helm installation.
+
+## Model repository
+
+If you already have a model repository you may use that with this helm chart. If you don't, you can use any model
+from https://download.01.org/opencv/2020/openvinotoolkit/2020.2/open_model_zoo/models_bin .
+
+OpenVINO Model Server needs models that it will make available for inferencing. For example you can 
+use Google Cloud Storage bucket:
```
ryanloney, line 21: Based on feedback last week, please add AWS S3 and Azure Blob examples. https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction. (We have this already for S3, but should explain how to use: https://github.com/openvinotoolkit/model_server/blob/master/example_k8s/openvino_model_server_s3_rc.yaml)

pskindel: I can't add Azure Blob examples, because OVMS doesn't support this storage so far. I created a task for adding support for Azure storage: https://jira.devtools.intel.com/browse/CVS-31091 .

Going to add S3 instructions.

deploy/README.md
```diff
@@ -0,0 +1,158 @@
...
+## Using OpenVINO Model Server
+
+Now that the server is running you can send HTTP or gRPC requests to it to perform inferencing. 
+By default, the service is exposed with a LoadBalancer service type. Use the following to find the 
+external IP for the server:
+```shell script
+pskindel@ovms_dev:~$ kubectl get svc
+NAME                    TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)                         AGE
+openvino-model-server   LoadBalancer   10.121.14.253   34.65.153.232   8080:30043/TCP,8081:32606/TCP   59m
+
```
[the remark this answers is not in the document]
pskindel: done

deploy/ovms/templates/deployment.yaml
pskindel: done

deploy/README.md
ryanloney: Sounds good.

deploy/README.md
```diff
@@ -0,0 +1,182 @@
...
+```shell script
+$ helm install ovms ovms --set model_name=resnet50-binary-0001,model_path=gs://models-repository,replicas=3
+```
+
+
+## Deploy the Model Server with configuration file
+
+To serve multiple models you can run the Model Server with configuration file as described in:
+https://github.com/openvinotoolkit/model_server/tree/master/example_client#submitting-grpc-requests-based-on-a-dataset-from-a-list-of-jpeg-files
```
dtrawins, line 90: wrong link...

deploy/ovms/values.yaml
```diff
@@ -0,0 +1,12 @@
+replicas: 1
+
+config_configmap_name: ""
+
+model_name: ""
+model_path: ""
+
+gcp_creds_secret_name: ""
+
+aws_access_key_id: ""
```
dtrawins, line 10: do we want to support also minio? If so, it would be good to added endpoint_url param.

deploy/ovms/templates/deployment.yaml
pskindel: Done. I parametrized many things in _values.yaml_ and provided instruction for multiple models in _Deploy the Model Server with configuration file_ section.

deploy/README.md
pskindel: Done. Added _If the bucket is public or you are going to run Model Server **on the same GCP or AWS account as storage** then no additional changes are needed and you can proceed to _Deploy the Model Server_ section._

deploy/README.md
pskindel: done

deploy/ovms/values.yaml
pskindel: added.

pskindel:
Whoever is going to merge that - please **squash** commits before merge.

deploy/README.md
```diff
@@ -0,0 +1,188 @@
...
+If you already have a model repository you may use that with this helm chart. If you don't, you can use any model
+from https://download.01.org/opencv/2020/openvinotoolkit/2020.2/open_model_zoo/models_bin .
+
+OpenVINO Model Server needs models that it will make available for inferencing. For example you can 
+use Google Cloud Storage bucket:
+```shell script
+$ gsutil mb gs://model-repository
+```
+
+You can download the model from the link provided above nad upload it to GCS:
```
ryanloney, line 26: @pskindel should say "and" instead of "nad"

PR #194, opened 2020-04-27
Basic AMS dockerfiles and image building scripts 

mzegla:
Providing additional Dockerfile for running OVMS and AMS service in the same container. 
Added `docker_build_ams` rule in `Makefile` which uses `Dockerfile_ams` to build docker image.

`Dockerfile_ams` is similar to `Dockerfile_binary_openvino`. It builds OV from binary package, but only with CPU support and includes AMS service in the image. 
Just like `Dockerfile_binary_openvino`, `Dockefile_ams` requires `dldt_package_url` parameter.

`Dockerfile_ams` requires models and config.json file under `extras/models` directory. Those models will be embedded in the image. 

To run container, you can use command:
`docker run ams:latest /ams_wrapper/start_ams.sh <ams_port> <ovms_port>`

When running `start_ams.sh` script without arguments, the AMS will listen on port 5000 and OVMS, serving models from `extras/models` directory, will listen on port 9000.


extras/ams_wrapper/start_ams.sh
```diff
@@ -0,0 +1,15 @@
...
+#   * Consider parametrizing paths
+#   * Might require providing mapping for neural network output
+
+export FILE_SYSTEM_POLL_WAIT_SECONDS=0
+
+AMS_PORT=${1:-5000}
+OVMS_PORT=${2:-9000}
+
+. /ie-serving-py/.venv/bin/activate
+/ie-serving-py/start_server.sh ie_serving config --config_path /opt/models/config.json --port $OVMS_PORT &
```
mateusz-ciesielski, line 14: Just a note - we will not get non-zero exit code from `start_ams.sh` if this background command fails.

extras/ams_wrapper/start_ams.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
mateusz-ciesielski, line 1: Adding `-e` flag here would be nice, so script will exit immediately in case of error

PR #196, opened 2020-04-28
Ams grpc

PatrykMatyjasek:
Predic GRPC call implementation

### DIFF
--- a/extras/ams_wrapper/src/api/ovms_connector.py
+++ b/extras/ams_wrapper/src/api/ovms_connector.py
@@ -17,6 +17,12 @@
 import grpc
 from tensorflow_serving.apis import predict_pb2
 from tensorflow_serving.apis import prediction_service_pb2_grpc
+from tensorflow import make_tensor_proto, make_ndarray
+import numpy as np
+
+
+class ModelNotFoundError(ValueError):
+    pass
 
 class OvmsConnector():
     def __init__(self, ovms_port, ovms_model_info):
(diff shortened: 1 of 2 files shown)


extras/ams_wrapper/src/api/ovms_connector.py
```diff
@@ -25,12 +31,35 @@ def __init__(self, ovms_port, ovms_model_info):
...
+            tensor_proto = make_tensor_proto(inference_input, shape=(inference_input.shape))
+        except TypeError as e:
+            raise TypeError("Unsupported data type") from e
+        except ValueError as e:
+            raise ValueError("Invalid arguments") from e
+        request.inputs[self.input_name].CopyFrom(tensor_proto)
+        try:
+            result = self.stub.Predict(request, 10.0)
+            output = make_ndarray(result.outputs[self.output_name])
+            numpy_array = np.array(output)
```
mateusz-ciesielski, line 53: Just a question: isn't `make_ndarray` already returning a numpy.ndarray?

PatrykMatyjasek: good catch ;)

extras/ams_wrapper/src/api/ovms_connector.py
```diff
@@ -25,12 +31,35 @@ def __init__(self, ovms_port, ovms_model_info):
...
+        except grpc.RpcError as e:
+            if e.code() == grpc.StatusCode.INVALID_ARGUMENT:
+                raise ValueError("Invalid argument: {}".format(e.details())) from e
+            elif e.code() == grpc.StatusCode.NOT_FOUND:
+                raise ModelNotFoundError("Requested model not found") from e
+            else:
```
mzegla, line 59: I think we should explicitly handle errors like SOCKET_CLOSED for situations when OVMS is not accessible by the AMS.

PR #202, opened 2020-04-29
[AMS] Detection type results handling

mzegla:
Changes made:

- added types.py file containing types used in post-processing of inference output in AMS service
- added load_labels method for loading json file with mapping of class ids to class names (might be changed in the future to load more info about the model)
- provided examplary code (commented in `postprocess_inference_output` of `ExampleModel`) doing inference output conversion to json response based on loaded labels and `Entity` types (not tested, just an example of the flow)

### DIFF
--- a/extras/ams_wrapper/src/api/types.py
+++ b/extras/ams_wrapper/src/api/types.py
@@ -1,0 +1,112 @@
+from typing import List
+from abc import ABC
+
+class Tag:
+    def __init__(self, value: str, confidence: float):
+        self.value = value
+        self.confidence = confidence
+
+    def as_dict(self):
+        result_dict = {
+            "value": self.value,
+            "confidence": self.confidence
+        }
+        return result_dict
+
+class Attribute:
+    def __init__(self, name: str, value: str, confidence: float):
+        self.name = name
+        self.value = value
+        self.confidence = confidence
+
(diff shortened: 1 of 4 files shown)


extras/ams_wrapper/src/api/types.py
```diff
@@ -0,0 +1,129 @@
...
+    def __init__(self, value: str, confidence: float):
+        self.value = value
+        self.confidence = confidence
+
+    def as_dict(self):
+        result_dict = {
+            "value": self.value,
+            "confidence": self.confidence
+        }
+        return result_dict
```
mateusz-ciesielski, line 31: How about `return vars(self)` ? We could potentially avoid defining attributes both in `__init__` and `as_dict` that way.

mzegla: I was thinking of a single point (one method name) for all types that would convert them to dict type. This explicit dictionary definition is consistent among all types classes. 
Your idea is cool for simple types, but for example for Entity I don't put entities list as is, but I do:  `"entities": [entity.as_dict() for entity in self.entities]` and I couldn't do just 
`return vars(self)` here right?

Also, if in the future types objects will require some internal fields that shall not be included in output dictionary or naming change will be required then we would still have to define output dict explicitly.

PR #206, opened 2020-04-29
Ams postprocesing test

rasapala:


extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.json
```diff
@@ -0,0 +1,3 @@
+{
+  "1.0":"car"
```
mzegla, line 2: Can we hold keys in the OV format? I guess label id from model output is either float or integer. 
Maybe it could be like?
`{ 1 : "car" }`
or
`{ 1.0 : "car" }`

Also, I belive https://github.com/openvinotoolkit/model_server/pull/208/files will provide labels for this model.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
```diff
@@ -0,0 +1,77 @@
...
+import json
+from logger import get_logger
+from api.models.model import Model
+
+logger = get_logger(__name__)
+
+
+class VehicleDetectionAdas(Model):
+
+    def load_default_labels(self):
```
mzegla, line 27: This shall be changed when https://github.com/openvinotoolkit/model_server/pull/208/files is merged. We could define labels file path here:
https://github.com/openvinotoolkit/model_server/blob/ams/extras/ams_wrapper/src/config.py#L23

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
```diff
@@ -0,0 +1,77 @@
...
+                data = json.load(labels_file)                                         
+                self.labels = data
+        except Exception as e:                                                        
+            logger.exception("Error occurred while opening labels file: {}".format(e))
+            sys.exit(1)                                                               
+        return                                                                   
+
+
+    def postprocess_inference_output(self, inference_output: dict) -> str:
+        from api.types import Tag, Rectangle, SingleEntity, Entity
```
mzegla, line 40: Could we move those imports up and place them among all the other import?

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
```diff
@@ -0,0 +1,77 @@
...
+
+        result_array = inference_output["detection_out"]
```
mzegla, line 42: We could consider adding output name to the model configuration, like we do with input:
https://github.com/openvinotoolkit/model_server/blob/ams/extras/ams_wrapper/src/config.py#L23

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # model with output shape (1,1,200,7) 
+        # with last dimension containg detection details
+        detection_threshold = 0.5
+        detections = []
+        for detection in result_array[0][0]:
+            label = str(detection[1].item())
```
mzegla, line 49: As I mentioned above. Do we need this conversion to string? Perhaps just `detection[1].item()` would be enough?

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
```diff
@@ -0,0 +1,77 @@
...
+            # End of detections
+            if label == "0.0":
+                break
+
+            if not label in self.labels:
+                label_value = "unknown"
+            else:
+                label_value = self.labels[label]
```
mzegla, line 57: This situation is clear mismatch between model output and labels file and I think we should define exception for it and raise it here if returned label is not among labels defined in labels file.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
```diff
@@ -0,0 +1,77 @@
...
+            y_max = detection[6].item()
+
+            if conf >= detection_threshold:
+                tag = Tag(label_value, conf)
+
+                box = Rectangle(x_min, y_min, abs(x_max-x_min), abs(y_max-y_min))
+
+                detection = SingleEntity(tag, box)
+                detections.append(detection)
+
```
mzegla, line 73: Here you could check if there are any detections at all. If the `detections` list is empty you could raise exception so the processing loop will catch it and return appropriate code. 
This exception should be defined at least on the Model class level, or somewhere else (maybe separate file for errors definitions?).

tests/functional/test_single_model_vehicle.py
```diff
@@ -14,7 +14,9 @@
 # limitations under the License.
 #
 import sys
-
+import cv2
```
mzegla, line 17: If we need this package, OpenCV should be included in requirements-dev.txt.
Although you can consider using tensorflow functions for image processing as it's already our dependency and we wouldn't need to include OpenCV.
Default preprocessing function uses tensorflow, you could check it out and see if you could apply it here.

https://github.com/openvinotoolkit/model_server/blob/ams/extras/ams_wrapper/src/preprocessing/preprocess_image.py

tests/functional/test_single_model_vehicle.py
```diff
@@ -51,23 +61,150 @@ def test_run_inference(self, vehicle_adas_model_downloader,
...
+
+        """
+
+        _, ports = start_server_single_vehicle_model
+        imgs_path =  os.path.join(vehicle_adas_data_downloader, "data", "annotation_val_images")
+
+        img_files = os.listdir(imgs_path)
+        imgs = np.zeros((0,3,384,672), np.dtype('<f'))
+        input_img = self.load_image(os.path.join(imgs_path,"image_000015.jpg"), 672, 384)
+        imgs = np.append(imgs, input_img, axis=0)
```
mzegla, line 108: You could go with:
```
input_img = self.load_image(os.path.join(img_path,"image_000015.jpg"), 672, 384)
img = np.expand_dims(input_img, axis=0)
```
If you just want to have additional dimension.

tests/functional/test_single_model_vehicle.py
```diff
@@ -51,23 +61,150 @@ def test_run_inference(self, vehicle_adas_model_downloader,
...
+        print("result:" + str(result))
+        for x in range(0, imgs.shape[0] - batch_size + 1, batch_size):
+            img = imgs[x:(x + batch_size)]       
+            print("img: " + str(img))
+            for y in range(0,img.shape[0]):
+                for i in range(0, 200*batch_size-1):
+                    detection = result[:,:,i,:]
+                    print("detection: " + str(detection))
+                    if detection[0,0,2] > 0.5 and int(detection[0,0,0]) == y:
+                        detections_sum+=1
```
mzegla, line 135: We do inference on only one image and the batch size is hard-coded to 1. 
Could you simplify this block? I think we only need one loop here. Could you check something like 
this:
```
for i in range(0, 200):
    detection = result[0,0,i]
    if detection[2] > 0.5:
        detections_sum+=1
```

I believe it's enough for this case. Also you could use this 2 in a constant (i.e. `CONFIDENCE=2` and then `detection[CONFIDENCE]`)

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.json
rasapala: Json load accepts only double quotes. It has to be "1.0"
We will merge with json model definitions afterwords.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
rasapala: Yes we will merge it in next iteration.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
rasapala: Done.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
rasapala: Done.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
rasapala: Done.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
rasapala: Should we throw exceptions in such cases ? Is it not normal for an image to not detect an object ?

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
rasapala: Done.

tests/functional/test_single_model_vehicle.py
rasapala: Done.

tests/functional/test_single_model_vehicle.py
rasapala: Done.

tests/functional/test_single_model_vehicle.py
rasapala: We will use tf and preprocess when we move tests to ams_wrapper if that ok.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
```diff
@@ -0,0 +1,77 @@
...
+    def load_default_labels(self):
+        labels_path = os.path.abspath(__file__).replace(".py",".json")
+        try:                                                                          
+            with open(labels_path, 'r') as labels_file:                               
+                data = json.load(labels_file)                                         
+                self.labels = data
+        except Exception as e:                                                        
+            logger.exception("Error occurred while opening labels file: {}".format(e))
+            sys.exit(1)
```
PatrykMatyjasek, line 36: Is it ok to exit here after catching exception?

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
```diff
@@ -0,0 +1,77 @@
...
+
+class VehicleDetectionAdas(Model):
+
+    def load_default_labels(self):
+        labels_path = os.path.abspath(__file__).replace(".py",".json")
+        try:                                                                          
+            with open(labels_path, 'r') as labels_file:                               
+                data = json.load(labels_file)                                         
+                self.labels = data
+        except Exception as e:
```
PatrykMatyjasek, line 34: I think there are two possible ways to fail here. One is when program is unable to open file, second one is when JSON is not valid and unable to load.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
PatrykMatyjasek: Is this dictionary somewhere validated? Is it possible that the key does not exist?

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
mzegla: Perhaps you're right, but either way we need to handle that situation. If not with exceptions, then maybe we could return `None`  to let processing loop know that not a single object has been detected and that it should return 204 for this request?

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
rasapala: Result is validated in functional test:
boxes_count = str(json_response).count("box")

print("detected boxes:" + str(boxes_count))
assert boxes_count == 2

try:
    format_check = json.loads(json_response)
except Exception as e:
    print("json loads exception:" + str(e))
    assert False

print("format_check:" + str(format_check))
assert format_check["subtype"] == "vehicle-detection"

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
rasapala: This will be changed once we switch to model templates from your PR.
We will add error codes for this situation.

extras/ams_wrapper/src/api/models/vehicle_detection_adas_model.py
rasapala: This will be changed once we switch to model templates from your PR.
We will add error codes for this situation.

PR #209, opened 2020-04-30
Ams wrapper functional tests

mateusz-ciesielski:
* Added a basic facilities for performing functional tests of ams wrapper
* Added a temporary mock server as a target for these tests, we should replace it when we will have a fully working flow in ams wrapper available
* changes in CI, ams unit tests/image build/functional tests will be executed automatically


mateusz-ciesielski:
@mzegla Initially I was storing tests in extras/ams_wrapper directory, but I moved it to tests/ diretory to allow reuse of already existing fixtures in the future, e.g. for starting docker containers, downloading test models etc

PR #210, opened 2020-05-05
[AMS] Initial integration with vehicle detection model enablement

mzegla:
Building command expects: 

- OVMS models and `config.json` with their configuration for OpenVINO Model Server under `extras/models`
- AMS models configuration files under `extras/ams_models` for AMS Service

Example directory structure for vehicle detection:
```
extras
├── ams_models
│   └── vehicle_detection_adas_model.json
└── models
    ├── config.json
    └── vehicle_detection_adas
        └── 1
            ├── vehicle-detection-adas-0002.bin
            └── vehicle-detection-adas-0002.xml
```
With `config.json` defining  `vehicle_detection_adas` model configuration for OVMS.

Having that done you can run:
1. `make docker_build_ams dldt_package_url=<OV_PACKAGE_URL>`
2. `docker run --rm -it -p 5000:5000 ams:latest /ams_wrapper/start_ams.sh`

That will start AMS Service and backing OVMS. By default AMS Service listens on port 5000.


mzegla:
`extras/models` directory is no longer required
`vehicle-detection-adas-binary` model is now downloaded from OV models zoo
OVMS starts with that single model with name `vehicle_detection_adas` located under `/opt/models/vehicle_detection_adas` in container filesystem

extras/ams_wrapper/src/api/models/model.py
```diff
@@ -18,39 +18,38 @@
...
-                data = json.load(labels_file)
-        except Exception as e:
+        self.model_name = model_name
+        self.labels = self.load_labels(labels_path)
+
+    def load_labels(self, labels_path):
+        try:                                                                          
+            with open(labels_path, 'r') as labels_file:                               
+                data = json.load(labels_file)                                         
+                labels = data['outputs'][0]['classes']
```
PatrykMatyjasek, line 40: I think we should put here a comment that only one output is currently supported

PR #211, opened 2020-05-05
Parse model's input configuration

mateusz-ciesielski:
* Assuming that there will be one configuration file for each model
* Configuration will accept all currently supported preprocessing options
* `input_name` is the only required field in input configuration, rest is optional

### DIFF
--- a/extras/ams_wrapper/src/api/models/model.py
+++ b/extras/ams_wrapper/src/api/models/model.py
@@ -28,7 +28,7 @@ import numpy as np
 from abc import ABC, abstractmethod
 from logger import get_logger
 from preprocessing.preprocess_image import preprocess_binary_image as default_preprocessing
-from api.ovms_connector import OvmsUnavailableError, ModelNotFoundError
+from api.ovms_connector import OvmsUnavailableError, ModelNotFoundError, OvmsConnector
 from .input_config import ModelInputConfiguration, \
     ModelInputConfigurationSchema, ValidationError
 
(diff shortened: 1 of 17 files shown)


extras/ams_wrapper/src/api/models/model.py
```diff
@@ -14,24 +14,35 @@
...
+from src.api.models.input_config import ModelInputConfiguration, \
+    ModelInputConfigurationSchema, ValidationError
 
 logger = get_logger(__name__)
 
+
 class Model(ABC):
 
-    def __init__(self, ovms_connector):
+    def __init__(self, ovms_connector, config_file_path: str = None):
```
mzegla, line 36: Maybe we could already pass inputs configuration here?
Right now, we have model config defined in config.py and it's imported here:
https://github.com/openvinotoolkit/model_server/blob/ams/extras/ams_wrapper/src/api/dispatcher.py#L20
in the process of creating model's objects and dispatcher. 
Perhaps we could extend wrapper parameters to include directory with models configs and then load all of them at the AMS startup. Then just pass configs to appropriate models constructors.

mateusz-ciesielski:
Exemplary input config that should work with vehicle detection model (it is very simple due to model being compatible with default preprocessing values, assuming that resizing will be done on OV side): 
```json
(...)
"inputs": [
  {
    "input_name": "detection_out"
  }
]
```
or more explicitly:
```json
(...)
"inputs": [
  {
    "input_name": "detection_out",
    "input_format": "NCHW",
    "color_format": "BGR"
  }
]
```

PR #213, opened 2020-05-05
[AMS] CVS-31283 performance test basic

janekmichalik:


extras/ams_wrapper/start_ams.sh
```diff
@@ -51,5 +56,5 @@ esac
 done
 
 . /ie-serving-py/.venv/bin/activate
-/ie-serving-py/start_server.sh ie_serving model --model_path /opt/models/vehicle_detection_adas --model_name vehicle_detection_adas --port $OVMS_PORT & 
-python /ams_wrapper/src/wrapper.py --port $AMS_PORT
+/ie-serving-py/start_server.sh ie_serving model --model_path /opt/models/vehicle_detection_adas --grpc_workers 10 --nireq 10 --model_name vehicle_detection_adas --port $OVMS_PORT &
```
mzegla, line 59: We would probably want to parameterize  OVMS `grpc_workers` and `nireq` too.
It will change though, when we will enable multiple models in AMS.

tests/performance/ams_client.py
```diff
@@ -0,0 +1,34 @@
...
+ROOT_PATH = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+DATASET = os.path.join(ROOT_PATH, "tests", "functional", "fixtures", "test_images")
+IMAGE = "single_car_small_reshaped.png"
+
+ITERATIONS = 1000
+
+for num in range(ITERATIONS):
+    with open(os.path.join(DATASET, IMAGE), mode='rb') as image_file:
+        image_bytes = image_file.read()
+        response = requests.post("http://localhost:5000/vehicleDetection",
```
mzegla, line 32: Checkout: https://requests.readthedocs.io/en/master/user/advanced/#session-objects
You could try to have something like:
`s = requests.Session()`
outside of this `for` loop and then do `s.post(...)` here.
It should significantly increase performance.

tests/performance/test_performance.py
```diff
@@ -0,0 +1,178 @@
...
+
+from tabulate import tabulate
+
+ROOT_PATH = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+DATASET = os.path.join(ROOT_PATH, "tests", "functional", "fixtures", "test_images", "performance")
+OVMS_CLIENT = os.path.join(ROOT_PATH, "example_client", "face_detection_performance.py")
+OVMS_DATASET = os.path.join(DATASET, "test_dir")
+
+CONFIG_4 = {"config": 4}
+CONFIG_32 = {"config": 32}
```
mzegla, line 34: It's not clear what this config means. Maybe we could change those two constants to a list:
`CORES_NUMBER = [4, 32]` ?

tests/performance/test_performance.py
```diff
@@ -0,0 +1,178 @@
...
+class TestPerformance:
+
+    @staticmethod
+    def inference(image, iterations: int):
+        responses = []
+        for num in range(iterations):
+            with open(os.path.join(DATASET, image), mode='rb') as image_file:
+                image_bytes = image_file.read()
+                start_time = datetime.datetime.now()
+                response = requests.post("http://localhost:5000/vehicleDetection",
```
mzegla, line 102: Same thing that I wrote above. Please try with requests.Session().

PR #217, opened 2020-05-07
Error handling in main model loop.

PatrykMatyjasek:

### DIFF
--- a/extras/ams_wrapper/src/api/models/model.py
+++ b/extras/ams_wrapper/src/api/models/model.py
@@ -23,7 +23,7 @@ import numpy as np
 from abc import ABC, abstractmethod
 from logger import get_logger
 from preprocessing.preprocess_image import preprocess_binary_image as default_preprocessing
-from api.ovms_connector import OvmsUnavailableError
+from api.ovms_connector import OvmsUnavailableError, ModelNotFoundError
 
 logger = get_logger(__name__)
 
@@ -98,6 +98,12 @@ class Model(ABC):
             resp.status = falcon.HTTP_400
             resp.body = json.dumps(body)
             return
+        except ModelNotFoundError as ex:
+            logger.exception("Model not found")
+            body = {"message": str(ex)}
+            resp.status = falcon.HTTP_500
+            resp.body = json.dumps(body)
+            return
         except OvmsUnavailableError as ex:
             logger.exception("OVMS unavailable")
             body = {"message": str(ex)}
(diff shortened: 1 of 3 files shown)


extras/ams_wrapper/src/api/models/model.py
```diff
@@ -63,38 +64,64 @@ def postprocess_inference_output(self, inference_output: dict) -> str:
...
+            resp.body = json.dumps(body)
+            return
+        except Exception as ex:
+            body = {"message": str(ex)}
+            resp.status = falcon.HTTP_500
+            resp.body = json.dumps(body)
+            return
+
+        # If model did not found results
+        if inference_output is None:
```
mzegla, line 106: Is it true for any situation? 
If inference were done right, then it's output shouldn't be None. 
I think that determining if the inference produced any meaningful results is done in postprocessing.

extras/ams_wrapper/src/api/models/model.py
```diff
@@ -63,38 +64,64 @@ def postprocess_inference_output(self, inference_output: dict) -> str:
...
-                    connection_start_time).total_seconds() * 1000
-        logger.debug(f"OVMS request handling time: {duration} ms")
+        try:
+            connection_start_time = datetime.datetime.now()
+            inference_ouput = self.ovms_connector.send(input_image)
+            duration = (datetime.datetime.now() -
+                        connection_start_time).total_seconds() * 1000
+            logger.debug(f"OVMS request handling time: {duration} ms")
+        except OvmsUnavailableError as ex:
+            body = {"message": str(ex)}
```
mateusz-ciesielski, line 95: I'd add `logger.exception()` call here, so full stack trace will be visible in the logs (perhaps it would be helpful to log exceptions in other try except blocks here).

extras/ams_wrapper/src/api/models/model.py
PatrykMatyjasek: Good catch!

extras/ams_wrapper/src/api/models/model.py
```diff
@@ -63,38 +64,74 @@ def postprocess_inference_output(self, inference_output: dict) -> str:
...
 
         # Send inference request to corresponding model in OVMS
-        connection_start_time = datetime.datetime.now()
-        inference_ouput = self.ovms_connector.send(input_image)
-        duration = (datetime.datetime.now() -
-                    connection_start_time).total_seconds() * 1000
-        logger.debug(f"OVMS request handling time: {duration} ms")
+        try:
+            connection_start_time = datetime.datetime.now()
+            inference_ouput = self.ovms_connector.send(input_image)
```
mzegla, line 91: typo 😃 
inference_ou**t**put

PR #225, opened 2020-05-08
Add output configuration parsing, integrate config management

mateusz-ciesielski:
* Add classes for loading and parsing model's output configuration
* Load input/output config in the model
* `Model.preprocess_binary_image` will now use preprocessing settings defined in model's configuration


extras/ams_wrapper/src/api/models/model.py
```diff
@@ -28,43 +28,34 @@
...
-                    labels[output["output_name"]] = output['classes']
-        except Exception as e:                                                        
-            logger.exception("Error occurred while opening labels file: {}".format(e))
-            sys.exit(1)
-        return labels
+        #self.labels = self.load_labels(config_file_path)
+        self.input_configs = self.load_input_configs(config_file_path)
+        self.output_configs = self.load_output_configs(config_file_path) 
+        self.labels = {output_name: {index: label for label, index in self.output_configs[output_name].classes.items()}
+                       for output_name in self.output_configs.keys()}
```
mateusz-ciesielski, line 47: I'm not sure if we still need separate `labels` property - for now I've left it as it is used quite often in `Model` subclasses

extras/ams_wrapper/src/api/models/model_config.py
```diff
@@ -57,4 +66,33 @@ def validate_type(self, data, **kwargs):
...
+
+class ModelOutputConfiguration:
+    def __init__(self, output_name: str, value_index_mapping: dict = None,
+                 classes: dict = None, confidence_threshold: float = None,
+                 top_k_results: int = None):
+                 self.output_name = output_name
+                 self.value_index_mapping = value_index_mapping
+                 self.classes = classes
+                 self.confidence_threshold = confidence_threshold
+                 self.top_k_results = top_k_results
```
mzegla, line 80: I'm wondering if maybe we could deal with those two parameters as HTTP request parameters so it could be different for each request?

extras/ams_wrapper/src/api/models/model.py
rasapala: Lets leave it for clarity.

extras/ams_wrapper/src/api/models/model.py
```diff
@@ -28,43 +28,34 @@
...
-            with open(labels_path, 'r') as labels_file:
-                data = json.load(labels_file)
-                labels = dict()
-                for output in data['outputs']: 
-                    labels[output["output_name"]] = output['classes']
-        except Exception as e:                                                        
-            logger.exception("Error occurred while opening labels file: {}".format(e))
-            sys.exit(1)
-        return labels
+        #self.labels = self.load_labels(config_file_path)
```
rasapala, line 43: Remove commented code.

extras/ams_wrapper/src/api/models/model.py
```diff
@@ -28,43 +28,34 @@
...
-            preprocessed_image = default_preprocessing(binary_image)
+        try: 
+            # Assuming single input for now
+            preprocessing_config = next(iter(self.input_configs.values()))
+            preprocessing_config = preprocessing_config.as_preprocessing_options()
+            preprocessed_image = default_preprocessing(binary_image, **preprocessing_config)
             preprocessed_image = np.expand_dims(preprocessed_image, axis=0)
         except Exception as e:
             # TODO: Error handling
+            logger.exception('Failed to preprocess binary image')
```
rasapala, line 58: Lets add e message.

mzegla: I think it's the part conflicting with target branch. 
This exception here shall be re-raised.

extras/ams_wrapper/src/api/models/model_config.py
mateusz-ciesielski: I think it would be a good idea to handle these parameters in HTTP request handling, but we are not yet using these config variables, so they can be safely deleted in case when we will implement handling them as HTTP paramters

extras/ams_wrapper/src/api/models/model.py
mateusz-ciesielski: Exception is now re-raised, regarding `e` message - it is automatically included when using `logger.exception()`.

extras/ams_wrapper/src/api/models/model.py
mateusz-ciesielski: fixed

PR #230, opened 2020-05-11
Classification config

rasapala:


extras/ams_wrapper/src/config.py
```diff
@@ -42,4 +42,13 @@
         },
     "config_path": "/opt/ams_models/vehicle_attributes_model.json"
     },
+    {"name": "emotionsRecognition", 
+    "class": VehicleAttributes,
```
mzegla, line 46: I think this class name is a little misleading 😄

extras/ams_models/emotions_recognition_model.json
```diff
@@ -0,0 +1,25 @@
...
+            {
+                "output_name": "prob_emotion",
+                "classes": {
+                    "neutral": 0.0,
+                    "happy": 1.0,
+                    "sad": 2.0,
+                    "surprise": 3.0,
+                    "anger": 4.0
+                }
+            }
```
mzegla, line 23: Don't we also need value_index mapping here?

extras/ams_wrapper/src/config.py
rasapala: Done.

extras/ams_models/emotions_recognition_model.json
rasapala: We do not have any complex mapping here. Just output type and indexed class names. So I do not see the need to index mapping. Maybe in more complex classification classes ? Info in config json file should be enough.

extras/ams_wrapper/src/api/models/classification_attributes_model.py
```diff
@@ -26,29 +26,28 @@
...
+        for output_name in self.labels.keys():
             attributes = []
             highest_prob = 0.0
-            for position in outputs[output].keys():
-                class_name = outputs[output][position]
-                probability = inference_output[type_name][0,int(float(position)),0,0].item()
+
+            if output_name not in inference_output:
+                message = 'Model configuration label- {}'
+                ' not found in model output - {}'.format(output_name, inference_output)
```
mzegla, line 40: It's a little unclear here in my opinion. Could we change this message to something like:
"Output name from model config does not match model outputs" ?

PR #232, opened 2020-05-12
Adding functional tests for 3 models.

rasapala:


tests/functional/test_ams_inference.py
```diff
@@ -157,6 +171,122 @@ def test_ams_with_wrong_model_name(self, start_ams_service_with_wrong_model_name
...
+        assert tag_value == "vehicle"
+        assert detections_count == 11
+        assert highest_box["w"] + epsilon >= 0.034460186958313
+        assert highest_box["l"] + epsilon >= 0.783527314662933
+        assert highest_box["h"] + epsilon >= 0.0431380569934845
+        assert highest_box["t"] + epsilon >= 0.173053205013275
+        assert highest_box["w"] - epsilon <= 0.034460186958313
+        assert highest_box["l"] - epsilon <= 0.783527314662933
+        assert highest_box["h"] - epsilon <= 0.0431380569934845
+        assert highest_box["t"] - epsilon <= 0.173053205013275
```
mateusz-ciesielski, line 287: I think it would be more clear to assert expected range in one statement, e.g.:
```python
assert 0.034460186958313 - epsilon <= highest_box["w"]  <= 0.034460186958313 + epsilon
```
or:
```python
assert abs(highest_box["w"] - 0.034460186958313) <= epsilon 
```

I think we should not test to have such precision in bounding box coordinates - it may differ due to rounding errors on different hardware, so perhaps we should have more tolerance for the results in tests.

tests/functional/test_ams_inference.py
```diff
@@ -157,6 +171,122 @@ def test_ams_with_wrong_model_name(self, start_ams_service_with_wrong_model_name
...
+
+        for classification in response_json["classifications"][0]["attributes"]:
+            if classification["confidence"] > highest_probability:
+                highest_probability = classification["confidence"]
+                highest_value = classification["value"]
+                type_name = classification["name"]
+
+        assert highest_probability > 0.547
+        assert highest_value == "red"
+        assert type_name == "color"
```
mzegla, line 230: What do you think about changing name to `attribute_name` here? 
IHMO it fits nice and would also match with class name (ClassificationAttributes)
If you decide to change it, please change all appearances.

PR #238, opened 2020-05-13
Adding ams ut coverage report.

rasapala:
Local usage:
pip install pytest-cov
cd extras/ams_wrapper
pytest --cov-config=.coveragerc --cov=src tests/ --cov-report=html
open ams_coverage_report/index.html file in browser

Web usage:
visit url:
http://htmlpreview.github.com/?https://github.com/openvinotoolkit/model_server/blob/ut_coverage_ams/extras/ams_wrapper/ams_coverage_report/index.html


mateusz-ciesielski:
I think we should not add html/js files of coverage report - it would be best to generate such report on demand, e.g. by running some make target

PR #239, opened 2020-05-13
[AMS] CVS-30726 - performance for all models

janekmichalik:


tests/performance/data/performance_constants.py
```diff
@@ -0,0 +1,78 @@
...
+AMS_CLIENT_PATH = os.path.join(ROOT_PATH, "tests", "performance", "ams_client.py")
+OVMS_CLIENT_PATH = os.path.join(ROOT_PATH, "tests", "performance", "ovms_client.sh")
+
+# OTHERS PARAMS
+ITERATIONS = 1000
+AMS_ADDRESS = "localhost"
+OVMS_PORT_FOR_AMS = "9000"
+AMS_PORT = "5000"
+OVMS_PORT = "9007"
+DLDT_PACKAGE = "http://registrationcenter-download.intel.com/akdlm/irc_nas/16612/l_openvino_toolkit_p_2020.2.120_online.tgz"
```
dtrawins, line 40: make it configurable from env variable instead of reading from src

PR #243, opened 2020-05-15
Ams clearlinux image

mateusz-ciesielski:


mateusz-ciesielski:
@rasapala Sounds good, I'll modify test script to run on both images

PR #245, opened 2020-05-15
[AMS] Refactor

mzegla:
* introduced ModelBuilder class
* moved configuration loading logic to ModelBuilder
* added top level config schema validation
* moved configuration details from config.py to models json files
* changed ovms_connector so it could include multiple inputs in the request
* adjusted unit tests


extras/ams_wrapper/src/config.py
```diff
@@ -20,62 +20,7 @@
...
-    "class": DetectionModel, 
-    "ovms_mapping": {
-        "model_name": "face_detection_adas",
-        "model_version": DEFAULT_VERSION,
-        "input_name": "data",
-        },
-    "config_path": "/opt/ams_models/face_detection_adas.json"
-    },
-]
+TYPE_CLASS_MAPPING = {
```
rasapala, line 23: Maybe we can move this definition elsewhere to remove the config.py file ? wrapper.py ?

extras/ams_models/person_vehicle_bike_detection_model.json
```diff
@@ -1,5 +1,6 @@
 {
-    "model_name": "person-vehicle-bike-detection-crossroad",
+    "model_name": "personVehicleBikeDetection",
```
rasapala, line 2: Should we change this property to endpoint_name ?

extras/ams_wrapper/src/config.py
mzegla: How about just renaming it? I think it would be good to have all AMS related constants in one place.

rasapala: Ok

extras/ams_wrapper/src/api/models/model_config.py
```diff
@@ -15,6 +15,13 @@
 #
 
 from marshmallow import Schema, fields, validates_schema, ValidationError, post_load, validate
+    
+class ModelConfigurationSchema(Schema):
+    model_name = fields.String(required=True)
+    model_type = fields.String(required=True)
+    inputs = fields.List(fields.Dict(keys=fields.String(), required=True))
+    outputs = fields.List(fields.Dict(keys=fields.String(), required=True))
```
mateusz-ciesielski, line 23: Shouldn't it be: 
```python
inputs = fields.List(fields.Nested(ModelInputConfigurationSchema, required=True))
outputs = fields.List(fields.Nested(ModelOutputConfigurationSchema, required=True))
```
?

PR #250, opened 2020-05-20
Centos based AMS image

dtrawins:


tests/scripts/functional-tests-ams.sh
```diff
@@ -21,3 +20,14 @@ NO_PROXY=localhost py.test ${TEST_DIRS}/functional/test_single_model_vehicle_att
                            -v --test_dir=/var/jenkins_home/test_ovms_models-${TESTS_SUFFIX} \
                            --image ${DOCKER_OVMS_TAG}-clearlinux
```
mateusz-ciesielski, line 21: It looks like we are using ${DOCKER_OVMS_TAG}-clearlinux image before actually building it

PR #256, opened 2020-05-27
Fix clearlinux dependencies

PatrykMatyjasek:


Dockerfile_ams_clearlinux
```diff
@@ -1,14 +1,58 @@
...
+    machine-learning-tensorflow --debug && \
+    swupd clean
 
 COPY --from=dlrs /usr/local/lib/inference-engine /usr/local/lib/inference-engine
 
 WORKDIR /ie-serving-py
 
 COPY start_server.sh setup.py version requirements_clearlinux.txt /ie-serving-py/
-RUN virtualenv --system-site-packages -p python3 .venv && . .venv/bin/activate && pip3 install --no-cache-dir -r requirements_clearlinux.txt
+COPY --from=base_build /tmp/pip .
```
dtrawins, line 50: Only one wheel package is needed out of 2. GPU could be dropped from the copy. Let's name a specific package.

Dockerfile_ams_clearlinux
```diff
@@ -1,14 +1,58 @@
+FROM clearlinux as base_build
+
```
dtrawins, line 2: Could you add a comment that we build here tensorflow-serving-api  pip package for python3.8 and TFS version 2.0
The container could be named pip build for example.

PR #263, opened 2020-05-29
added support for Live Video Analytics in Azure

dtrawins:
This PR introduces support for Azure Live Video Analytics (LVA) as an AI Extension on 
IoT Edge devices. It enables easy delegation of inference requests to OpenVINO in media analytics pipelines.


ryanloney:
Not to make things more difficult, but I would really like to keep the AMS Dockerfiles out of the main directory of the repo for OVMS.

PR #275, opened 2020-06-23
Small fixes to enable CI for OVMS early checks

kromanow94:
Newest version of virtualenv was causing errors, fixing to version 20.0.20 fixed it.
Also because of some spaces in projects directory it was required to change string interpolation.
Also changed the ov_source_branch to OV_SOURCE_BRANCH. It now takes the variable also from shell environment.


kromanow94:
See https://jenkins.ovms.igk.intel.com/job/OVMS-Python%20and%20OpenVINO%20Early%20Check/job/Build%20OVMS%20with%20OpenVINO-kromanow/

PR #281, opened 2020-07-02
Updated EULA.txt for readability

ryanloney:


mzegla:
Does it completely replace Apache licence? Because we still have it mentioned in:
https://github.com/openvinotoolkit/model_server/tree/EULA-cleanup#contribution-rules

and it's present in: 
https://github.com/openvinotoolkit/model_server/blob/EULA-cleanup/LICENSE

ryanloney:
No, Apache 2.0 is still the license for our open source code. The EULA is only referenced when an end user downloads a container image from a public registry. Similar to this project (which has BSD license): https://github.com/intel/oneapi-containers

PR #285, opened 2020-07-14
lva classification models schema change

dtrawins:


extras/ams_wrapper/src/api/models/classification_attributes_model.py
```diff
@@ -71,11 +71,15 @@ def postprocess_inference_output(self, inference_output: dict) -> str:
             if current_conf.confidence_threshold:
                 attributes = [attr for attr in attributes
                               if attr.confidence >= current_conf.confidence_threshold]
-            classification = SingleClassification(attributes)
+            if (output_name=="color"):
```
dtrawins, line 74: To make this change applicable for wider range of models it would be better to avoid hardcoded subtype names. I suggest using the model output name as the subtype. That way the same model category could be used for other models like flower classification etc.

extras/ams_wrapper/src/api/types.py
```diff
@@ -46,6 +46,17 @@ def as_dict(self):
         }
         return result_dict
 
+class Attribute_fix:
```
dtrawins, line 49: It would be better to avoid duplicated classes. The original class should be just updated or if something is shared with other model categories - more meaningful name should be used.

extras/ams_wrapper/src/api/types.py
```diff
@@ -90,6 +101,20 @@ def as_dict(self):
                                          for attribute in self.attributes]
         return result_dict
 
+class SingleClassification_fix:
```
dtrawins, line 104: check my previous comment

extras/ams_wrapper/src/api/types.py
```diff
@@ -80,29 +77,30 @@ def as_dict(self):
...
 
     def as_dict(self):
         result_dict = {
-            "type": self.type_name,
-            "subtype": self.subtype_name,
-            "classifications": [classification.as_dict() for classification in self.classifications]
+            "inferences": [classification.as_dict() for classification in self.classifications]
         }
+        result_dict["inferences"][0], result_dict["inferences"][1] = \
+            result_dict["inferences"][1], result_dict["inferences"][0]
```
mzegla, line 103: What's the point of that switch?

extras/ams_wrapper/tests/unit/test_classification_attributes_model.py
```diff
@@ -72,28 +72,27 @@ def fake_output_config() -> Dict[str, ModelOutputConfiguration]:
...
-                              ]}]}
-                         )])
+    (MOCK_INFERENCE_OUTPUT,
+     {
+         "inferences": [
+             {
+                 "type": "classification",
+                 "subtype": "type",
+                 "classification": {
+                     "tag": "van"
```
mzegla, line 82: I don't see confidence value anywhere in those expected responses. Is it not required?

extras/ams_wrapper/src/api/types.py
kkadirvx: > What's the point of that switch?

This switching is just to match with the order of desired schema output(Like the order of "type" ,"color"). But going forward , it can be removed since the order might change for mode to model.

PR #288, opened 2020-07-21
Schema changes

jerinmax:
Updated schema for classification and object detection models, also updated test(Unit & Functional) scripts for the same.

### DIFF
--- a/extras/ams_wrapper/src/api/models/classification_attributes_model.py
+++ b/extras/ams_wrapper/src/api/models/classification_attributes_model.py
@@ -19,7 +19,7 @@ import json
 from src.logger import get_logger
 from src.api.models.model import Model
 from src.api.models.model_config import ModelOutputConfiguration
-from src.api.types import Attribute, SingleClassification, Classification
+from src.api.types import Tag, SingleClassification, Classification
 
 logger = get_logger(__name__)
 
@@ -31,7 +31,7 @@ class ClassificationAttributes(Model):
         classifications = []
 
         for output_name in self.labels.keys():
-            attributes = []
+            tags = []
             highest_prob = 0.0
 
             if output_name not in inference_output:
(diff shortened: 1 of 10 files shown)


extras/ams_wrapper/src/api/types.py
```diff
@@ -86,7 +87,9 @@ def as_dict(self):
         result_dict = {
             "type": self.type_name,
             "subtype": self.subtype_name,
-            "classification": self.attributes[0].as_dict()
+            "classification": {
+            	"tag": attribute.as_dict() for attribute in self.attributes
```
mzegla, line 91: Since there can be only one tag for classification why use list of attributes? 
You could have separate class `Tag `with required fields and go with something like `"tag": self.tag.as_dict()`
It would increase readability and separate tag from attribute (in case you needed to add attributes to the classification output in the future).
Despite that, there will be only one tag element in classification, so using list of attributes can be misleading as only the last element of attributes list will be present in this dict.

jerinmax: Committed the new changes. Kindly review

jerinmax:
> Please fix style errors and add entity changes.

Committed the changes. Kindly review

extras/ams_wrapper/src/api/models/classification_attributes_model.py
```diff
@@ -58,24 +58,25 @@ def postprocess_inference_output(self, inference_output: dict) -> str:
...
-                attributes = attributes[:current_conf.top_k_results]
+                tags = tags[:current_conf.top_k_results]
             if current_conf.confidence_threshold:
-                attributes = [attr for attr in attributes
-                              if attr.confidence >= current_conf.confidence_threshold]
-            classification = SingleClassification(subtype_name=output_name, attributes=attributes)
-            classifications.append(classification)
+                tags = [attr for attr in tags
+                        if attr.confidence >= current_conf.confidence_threshold]
+            if len(tags) != 0:
```
mzegla, line 74: To be consistent with detection model you could go with:
```
if not tags:
    return None
```
Then the main loop shall return 204 in case there's no classification above confidence threshold.

jerinmax: Committed the changes. kindly review the updated file

PR #291, opened 2020-08-05
fix manual tests execution

dtrawins:


tests/scripts/unit-tests.sh
```diff
@@ -29,10 +29,11 @@ docker run -t --rm \
     -w /mnt/ovms \
     -e LD_LIBRARY_PATH="${LD_LIBRARY_PATH}" \
     -e PYTHONPATH="${PYTHONPATH}" \
+    -u `id -u` \
```
kromanow94, line 32: SC2046: Quote this to prevent word splitting.
SC2006: Use $(...) notation instead of legacy backticked `...`.

kromanow94:
@mzegla please approve - changes dismissed it

PR #309, opened 2020-10-13
added OV install via yum without DLDT package url

dtrawins:


Makefile
```diff
@@ -44,7 +44,7 @@ INSTALL_RPMS_FROM_URL ?=
 #         - uncomment source build section, comment binary section
 #         - adjust binary version path - version variable is not passed to WORKSPACE file!
 OV_SOURCE_BRANCH ?= 2020.4
```
mzegla, line 46: Can you also change it to 2021.1?

dtrawins: OV building from sources was not tested recently. It would require a few changes to enable it back. For now it is low priority.

mzegla (approved):

Dockerfile.centos
```diff
@@ -109,15 +109,21 @@ RUN if [ "$ov_use_binary" == "0" ] ; then true ; else exit 0 ; fi ; make --jobs=
...
     cd l_openvino_toolkit* && \
     sed -i 's/decline/accept/g' silent.cfg && \
     ./install.sh -s silent.cfg --ignore-signature && \
     rm -Rf $TEMP_DIR $INSTALL_DIR/uninstall* /tmp/* $DL_INSTALL_DIR/documentation $DL_INSTALL_DIR/inference_engine/samples
 
+# yum package
+RUN if [ "$ov_use_binary" == "1" ] && [ -n $DLDT_PACKAGE_URL ] ; then true ; else exit 0 ; fi ; \
+    yum-config-manager --add-repo https://yum.repos.intel.com/openvino/2021/setup/intel-openvino-2021.repo && \
+    rpm --import https://yum.repos.intel.com/openvino/2021/setup/RPM-GPG-KEY-INTEL-OPENVINO-2021 && \
+    yum install -y intel-openvino-runtime-centos7
```
intel-rrozestw, line 125: version should be probably more explicit here; seems like 2021 repo will be upgraded.

intel-rrozestw (changes requested):

intel-rrozestw (changes requested):
Please add a parameter to a makefile and pass it here, in order to have a single source of truth.

intel-rrozestw (approved):

PR #311, opened 2020-10-16
CVS-33877 Adding ovms version switch.

rasapala:
Adding ovms version switch.


dtrawins:
Can we add also OV backend version? 
Maybe just one parameter would be sufficient for the version name?
I think it would be the best to have the default version name including git sha - if someone build the image from latest sources, it will report the same name like the official release.
Also current version should not be named as patch1.

rasapala:
Command: ./ovms --version
Example output:
OpenVINO Model Server 2021.1.b702587
OpenVINO backend 2021.1.105

Dockerfile.centos
```diff
@@ -32,7 +37,7 @@ ARG YUM_OV_PACKAGE=intel-openvino-runtime-centos7-2021.1.110.x86_64
 ARG build_type=dbg
 ARG debug_bazel_flags=--strip=never\ --copt="-g"\ -c\ dbg
 ENV HDDL_INSTALL_DIR=/opt/intel/openvino/deployment_tools/inference_engine/external/hddl
-
+
```
michalkulakowski, line 40: Unnecessary spaces

Dockerfile.centos
```diff
@@ -188,9 +192,15 @@ WORKDIR /ovms
 
 COPY src/ /ovms/src/
 COPY release_files/ /ovms/release_files/
+COPY set_version.sh /tmp
 
 ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/intel/openvino/deployment_tools/inference_engine/lib/intel64/:/opt/intel/openvino/deployment_tools/ngraph/lib/:/opt/intel/openvino/inference_engine/external/tbb/lib/
 RUN ln -s /opt/intel/openvino_2021 /opt/intel/openvino
+RUN bash /tmp/set_version.sh REPLACE_PROJECT_NAME ${PROJECT_NAME}
+RUN if [ "$ov_use_binary" == "1" ] && [ -n $DLDT_PACKAGE_URL ] ; then true ; else exit 0 ; fi ;bash /tmp/set_version.sh REPLACE_OPENVINO_NAME `echo "$YUM_OV_PACKAGE" | sed -e 's/intel-openvino-runtime-centos7-//g' | sed 's/.x86_64//g'`
```
dtrawins, line 200: you could use just a single sed like:
sed -r  's/.*-([0-9]+\.[0-9]+\.[0-9]+).x86_64/\1/'

set_version.sh
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#*****************************************************************************
+first="$1"
+shift
+rest="${@}"
+sed -i -e 's|'$first'|'"$rest"'|g' /ovms/src/version.hpp
```
dtrawins, line 20: do we need this script at all while the dockerfile is using sed command anyway. Maybe the dockerfile should edit the version.hpp directly. It should be simpler.

dtrawins (approved):

michalkulakowski (approved):

PR #317, opened 2020-10-20
CVS-4107 Create logging policy

michalkulakowski:


docs/contributing.md
```diff
@@ -121,3 +121,75 @@ Prerequisites: dpkg -i http://archive.ubuntu.com/ubuntu/pool/universe/c/checksec
...
 Prerequisites: pip3 install bandit
+
+## Logging Policy
+
+### Levels
+
+</br>
+
+#### `ERROR`
+Log at this level errors of every ovms operation that failed
```
mzegla, line 132: What do you mean by "failed"? 
Handling of the client's requests can fail (wrong shape, inputs etc.) and OVMS can potentially handle thousands of requests. If someone would flood OVMS with badly formed requests we would see nothing but errors in logs. 

Perhaps we could define that on ERROR log level we should log only errors that are related to the internal model server state (like when a model could not be loaded for some reason). 
And errors that are results of handling users requests are logged on DEBUG level?

docs/contributing.md
```diff
@@ -121,3 +121,75 @@ Prerequisites: dpkg -i http://archive.ubuntu.com/ubuntu/pool/universe/c/checksec
...
+
+Example:
+```
+    "Failed to create a temporary path PATH"
+```
+
+</br>
+
+#### `WARN`
+Log at this level everything that may lead to an error/need to be fix by administrator.
```
mzegla, line 143: We could also log warning when we detect using deprecated or partially valid configuration (like when someone passes shape and batch_size). Something that user should be notified about. Not only stuff that may lead to error, but also stuff that can lead to behavior not really wanted or expected by the user.

docs/contributing.md
```diff
@@ -121,3 +121,75 @@ Prerequisites: dpkg -i http://archive.ubuntu.com/ubuntu/pool/universe/c/checksec
...
+
+Example:
+```
+    "Expected version directory to be in number format. Got:DIR_NAME"
+```
+
+</br>
+
+#### `INFO` 
+Log at this level all actions that are user-driven, or any others that changes state of application:
```
mzegla, line 154: What do you mean by user-driven? 
I agree about the "changes state of application"

docs/contributing.md
```diff
@@ -121,3 +121,75 @@ Prerequisites: dpkg -i http://archive.ubuntu.com/ubuntu/pool/universe/c/checksec
...
+</br>
+
+- processing requests
```
mzegla, line 157: Do we want to log **every** request processing at INFO level?
Perhaps it's something that we should do on the DEBUG level?

mzegla (commented):
I propose setting convention/requirement for log structure. So we can align to the same standard.

For example:
[Model: resnet; Version: 2] Invalid input shape for input data. Expected ... Actual: ...

[Pipeline: my_pipeline; Node: dl_classify] Wrong tensor content. Expected: ... Actual ...

I think it might be valuable to enforce providing certain information on developer like:
- which model/pipeline/node it is related to
- what are expected and actual values for errors
- what are previous and next/current state for internal state changes (like changing default version served)
etc.

docs/contributing.md
michalkulakowski: This is what i meant by user-driven :p but youre right - i will move it to the DEBUG level

docs/contributing.md
```diff
@@ -129,62 +129,50 @@ Prerequisites: pip3 install bandit
...
 
 ####  `TRACE`
 Log at this level very specific information about what happens in the program, including code-related names. (Only for development purpose)
 </br>
 
 Example:
-```
-    "GCS: directory contents fetched, items: ITEMS_COUNT"
-```
+
```
mzegla, line 173: + intermediate states of the objects

docs/contributing.md
```diff
@@ -129,62 +129,50 @@ Prerequisites: pip3 install bandit
...
-Log at this level errors of every ovms operation that failed 
+Log at this level errors related to the internal model server state.
 </br>
-
 Example:
-```
-    "Failed to create a temporary path PATH"
-```
 
+- exceptions thrown during accesing local files
```
mzegla, line 136: + errors during models updates (failed to load/reload model)

docs/contributing.md
```diff
@@ -129,62 +129,50 @@ Prerequisites: pip3 install bandit
...
-</br>
-
 #### `DEBUG`
 Log at this level information about what happens in the program that may help during debbuging.
 </br>
 
 Example:
-```
-    "Getting default version for model:MODEL_NAME, DEFAULT_VERSION"
-```
```
mzegla, line 164: + duration of specific operations (deserialization, serialization, inference etc.)

rasapala (approved):
LGTM

mzegla (approved):

PR #318, opened 2020-10-22
CVS-41121 onnx-support

mgumowsk:


rasapala (approved):
LGTM

src/modelinstance.cpp
```diff
@@ -226,7 +226,7 @@ std::unique_ptr<InferenceEngine::CNNNetwork> ModelInstance::loadOVCNNNetworkPtr(
 }
 
 Status ModelInstance::loadOVCNNNetwork() {
-    auto& modelFile = modelFiles[".xml"];
+    auto& modelFile = modelFiles[0];
```
mzegla, line 229: For IR format. Are we sure **.xml** file will be the first element of this vector every time and not the **.bin** one?

mgumowsk: yes, see OV_MODEL_FILES_EXTENSIONS in modelinstance.hpp

mzegla: oh, I see, it's all good in this case

mzegla (approved):

PR #339, opened 2020-11-09
Cvs 35968 better handling of anonymous shape param

michalkulakowski:


src/modelconfig.cpp
```diff
@@ -358,7 +358,11 @@ Status ModelConfig::parseNode(const rapidjson::Value& v) {
                             shapeInfo.shape.push_back(sh.GetUint64());
                         }
                     }
-                    this->addShape(s.name.GetString(), shapeInfo);
+                    if (s.name.GetString() != ANONYMOUS_INPUT_NAME) {
+                        this->addShape(s.name.GetString(), shapeInfo);
+                    } else {
+                        SPDLOG_WARN("Provided one of provided shapes name: {} is forbidden and will be omitted", ANONYMOUS_INPUT_NAME);
```
rasapala, line 364: Provided shape name: {} is forbidden

rasapala (changes requested):

rasapala (approved):
LGTM

src/test/modelconfig_test.cpp
```diff
@@ -439,3 +439,32 @@ TEST(ModelConfig, modelVersionPolicyIncorrect) {
...
 }
+
+TEST(ModelConfig, ConfigParseNodeWithForbiddenShapeName) {
+    const char* config = R"#({
+        "model_config_list": [
+            {
+                "config": {
+                    "name": "alpha",
+                    "base_path": "/tmp/models/dummy1",
+                    "shape": {"ANONYMOUS_INPUT_NAME": "(1, 3, 600, 600)"}
```
dkalinowski, line 450: Can we somehow avoid hardcoding ANONYMOUS_INPUT_NAME? We already have it defined somewhere else under alias `ANONYMOUS_INPUT_NAME`. Maybe we could somehow use it here?

dkalinowski (changes requested):
Please also create task for long term solution - without forbidding any input name.

dkalinowski (approved):

Issue #303, opened 2020-09-11
gRPC client example jpeg_classification.py passes RBG image while RESNET expects RGB


smcope:
Line 48 of jpeg_classification.py transposes BGR-->RBG. RESNET expects RGB format. 

Proposed change:
Line 48: img = img.transpose(2,1,0).reshape(1,3,size,size)

dtrawins:
@smcope indeed for models expecting RGB format there should be added a conversion.
It should be something like this:
img = img[:, :, [2, 1, 0]]
The command you listed would rather convert on different dimensions from HWC to CHW.
We will be adding improvements in the exemplary clients shortly.

rasapala:
@smcope We are using the resnet50 model in the example from: gs://ovms-public-eu/resnet50
That is https://github.com/openvinotoolkit/open_model_zoo/blob/master/models/intel/resnet50-binary-0001/description/resnet50-binary-0001.md model that expects the following format:

A blob with a BGR image in the format: [B, C=3, H=224, W=224], where:

B – batch size
C – number of channels
H – image height
W – image width

img = cv2.imdecode(img, cv2.IMREAD_COLOR) reads an image in BGR channel format with HWC dimensions.
Thus we need to transpose it to CHW using img = img.transpose(2,0,1) method.
No color channels conversion is required here. But as you noticed if model input requires different channels order a channel reorder would be required.
We have added an option to convert to RGB in https://github.com/openvinotoolkit/model_server/pull/345/files PR using the --rgb_image 1 command line option in the jpeg_classification example.

PR #345, opened 2020-11-10
Example readme fix

rasapala:
Issue https://github.com/openvinotoolkit/model_server/issues/303 resolution.


example_client/README.md
```diff
@@ -18,6 +18,11 @@ In the examples listed below, OVMS can be started using a command:
 docker run -d --rm -p 8000:8000 -p 9000:9000 openvino/model_server:latest --model_name resnet --model_path gs://ovms-public-eu/resnet50 --port 9000 --rest_port 8000
 ```
 
+Since access to www resources is needed for the samples to operate add proxy settings variables to the above command if required:
```
dtrawins, line 21: I suggest avoiding duplicate command with proxy settings. It should be sufficient to add a note that access to Google Cloud Storage might require proper configuration of https_proxy in the docker engine or in the docker container. Or combine the commands in one with https_proxy env. `-e https_proxy=${https_proxy}` might be a bit more elegant.

dtrawins (changes requested):

dtrawins (approved):

mzegla (approved):

PR #360, opened 2020-11-18
Building OV from source.

rasapala:


mzegla (commented):
Users might want to build OVMS with directly with `docker build` command. 
Shall we write a note in documentation that if they're not using `make`, they should rename desired workspace file to WORKSPACE? 
Or do we push users to use our Makefile?

rasapala:
@mzegla As checked offline we do not use docker build in our instructions. Using this command directly is not recommended nor supported.

mzegla (approved):

docs/age_gender_guide.md
```diff
@@ -13,6 +13,13 @@ curl --create-dirs https://download.01.org/opencv/2021/openvinotoolkit/2021.1/op
 
 ```
 
+Grant proper access rights to downloaded files
```
mzegla, line 16: Are you sure it's going to be required always?
If not, then perhaps you could add: _"You may need to..."_ to the beginning of that sentence?

docs/age_gender_guide.md
```diff
@@ -13,6 +13,13 @@ curl --create-dirs https://download.01.org/opencv/2021/openvinotoolkit/2021.1/op
...
+
+```Bash
+chmod -R 777 model
```
mzegla, line 19: Perhaps `chmod 744` will be enough?

rasapala: 744 not  enough.

docs/ovms_quickstart.md
```diff
@@ -55,6 +55,13 @@ Download the model components to the `model/1` directory. Example command using
 curl --create-dirs https://download.01.org/opencv/2021/openvinotoolkit/2021.1/open_model_zoo/models_bin/1/face-detection-retail-0004/FP32/face-detection-retail-0004.xml https://download.01.org/opencv/2021/openvinotoolkit/2021.1/open_model_zoo/models_bin/1/face-detection-retail-0004/FP32/face-detection-retail-0004.bin -o model/1/face-detection-retail-0004.xml -o model/1/face-detection-retail-0004.bin
 ```
 
+You may need to grant proper access rights to downloaded files
+
+```Bash
+chmod -R 755 model
```
dtrawins, line 61: this change is not needed because ovms is started with -u $(id -u):$(id -g)

docs/age_gender_guide.md
```diff
@@ -13,6 +13,13 @@ curl --create-dirs https://download.01.org/opencv/2021/openvinotoolkit/2021.1/op
 
 ```
 
+You may need to grant proper access rights to downloaded files
+
+```Bash
+chmod -R 755 model
```
dtrawins, line 19: this change is not needed because ovms is started with -u $(id -u):$(id -g)

BINARY_OV_WORKSPACE
```diff
@@ -0,0 +1,23 @@
+# OPENVINO DEFINITION FOR BUILDING FROM BINARY RELEASE: ##########################
+new_local_repository(
+    name = "openvino",
+    build_file_content = """
+cc_library(
+    name = "openvino",
+    srcs = glob([
```
dtrawins, line 7: shouldn't we add also plugins?

dtrawins (changes requested):

.gitignore
```diff
@@ -16,3 +16,4 @@ __pycache__
 *user_config*
 *.swp
 dist/
+WORKSPACE
```
mgumowsk, line 19: WORKSPACE is required for bazel, shouldn't be in git ignore

BINARY_OV_WORKSPACE
mgumowsk: no, plugins are loaded dynamically

SOURCE_OV_WORKSPACE
```diff
@@ -0,0 +1,8 @@
+# OPENVINO DEFINITION FOR BUILDING FROM SOURCE:
+
+new_local_repository(
+    name = "openvino",
+    build_file = "@//third_party/openvino:BUILD",
```
mgumowsk, line 5: Wasn't it easier to keep a single WORKSPACE file?
The definition of build_file/build_file_content is the same.
All you need to do is to keep the same path. For example, keep the definition for binary version (inference_engine/lib/intel64/libinference_engine_legacy.so) and in case of compiling from sources create a symbolic link from /openvino/bin/intel64/Release/lib/ to /opt/intel/openvino/deployment_tools/inference_engine/lib/intel64/
The second symbolic link is for includes.
That way, we'll have a single WORKSPACE file with no special hacks in Makefile

mgumowsk (changes requested):

dtrawins (approved):

lib_search.py
```diff
@@ -48,7 +48,8 @@ def check_dir(start_dir):
                      'Doxyfile', 'clang-format','net_http.patch', 'tftext.patch', 'tf.patch', 'client_requirements.txt',
                      'openvino.LICENSE.txt', 'c-ares.LICENSE.txt', 'zlib.LICENSE.txt', 'boost.LICENSE.txt',
                      'libuuid.LICENSE.txt', 'input_images.txt', 'REST_age_gender.ipynb', 'dummy.xml', 'listen.patch', 'add.xml',
-                     'requirements.txt', 'missing_headers.txt', 'libevent/BUILD', 'azure_sdk.patch', 'rest_sdk_v2.10.16.patch', ]
+                     'requirements.txt', 'missing_headers.txt', 'libevent/BUILD', 'azure_sdk.patch', 'rest_sdk_v2.10.16.patch',
+                     'BINARY_OV_WORKSPACE', 'SOURCE_OV_WORKSPACE',]
```
mgumowsk, line 52: you forgot to remove it

WORKSPACE
```diff
@@ -235,3 +202,28 @@ http_archive(
...
+    name = "openvino",
+    srcs = glob([
+        "inference_engine/lib/intel64/libinference_engine_legacy.so",
+        "inference_engine/lib/intel64/libinference_engine.so",
+        "inference_engine/lib/intel64/libinference_engine_c_api.so",
+        "ngraph/lib/libngraph.so"
+    ]),
+    hdrs = glob([
+        "inference_engine/include/**/*.*"
+    ]),
```
mgumowsk, line 221: Can we move the build_file_content to third_party/openvino/ and use build_file instead (similarly to awssdk above) ?

mgumowsk (approved):

PR #376, opened 2020-11-26
CVS-35586 log error when model name is doubled

Ficol:


rasapala (approved):
LGTM

dkalinowski:
@Ficol I think it would be nice to add unit test for this scenario. Let me know if you need help with it. Other than that the change looks OK.

dkalinowski (changes requested):
@Ficol I think it would be nice to add unit test for this scenario. Let me know if you need help with it. Other than that the change looks OK.

atobiszei (approved):

src/test/modelmanager_test.cpp
```diff
@@ -350,6 +350,39 @@ TEST(ModelManager, ConfigReloadingWithWrongInputName) {
...
+      }
+    }]})";
+    std::filesystem::create_directories(model_1_path);
+    std::filesystem::create_directories(model_2_path);
+    std::string fileToReload = "/tmp/ovms_config_file2.json";
+    createConfigFileWithContent(configWithTwoSameNames, fileToReload);
+    modelMock = std::make_shared<MockModel>();
+    MockModelManager manager;
+
+    EXPECT_CALL(*modelMock, addVersion(_))
```
dkalinowski, line 375: Here you expect StatusCode::OK one time, is it possible to expect error status the second time when name is duplicated?

src/test/modelmanager_test.cpp
```diff
@@ -350,6 +350,39 @@ TEST(ModelManager, ConfigReloadingWithWrongInputName) {
...
+    {
+      "config": {
+        "name": "same_name",
+        "base_path": "/tmp/models/dummy1"
+      }
+    },
+    {
+      "config": {
+        "name": "same_name",
+        "base_path": "/tmp/models/dummy2"
```
dkalinowski, line 365: This base path is invalid, but it does not matter because duplicate name check is done before actual model loading, right?

src/test/modelmanager_test.cpp
Ficol: I think no, because when name is duplicated, it just skips model without calling reloadModelWithVersions() and doesn't return anything.

dkalinowski (approved):

PR #378, opened 2020-11-27
CVS-33484-cpu_extension_loader_draft

intel-rrozestw:


src/server.cpp
```diff
@@ -248,6 +249,10 @@ int server_main(int argc, char** argv) {
         auto& config = ovms::Config::instance().parse(argc, argv);
         configure_logger(config.logLevel(), config.logPath());
 
+        if(config.cpuExtensionLibraryPath() != "") {
```
intel-rrozestw, line 252: should be moved to some other place, where we create OV::Core during normal flow. Just once.

mgumowsk (approved):

src/modelinstance.cpp
```diff
@@ -236,6 +236,13 @@ uint ModelInstance::getNumOfParallelInferRequests(const ModelConfig& modelConfig
 
 void ModelInstance::loadOVEngine() {
     engine = std::make_unique<InferenceEngine::Core>();
+    if (ovms::Config::instance().cpuExtensionLibraryPath() != "") {
+        SPDLOG_INFO("Loading custom CPU extension from {}", ovms::Config::instance().cpuExtensionLibraryPath());
+        auto extension_ptr = InferenceEngine::make_so_pointer<InferenceEngine::IExtension>(ovms::Config::instance().cpuExtensionLibraryPath().c_str());
```
mgumowsk, line 241: Does it throw an exception? (make_so_pointer and engine->AddExtenstion)? If yes, maybe we should add try catch

dtrawins (approved):

PR #380, opened 2020-11-27
Cvs 40375 shape segfault fix

rasapala:


src/modelconfig.cpp
```diff
@@ -266,8 +266,14 @@ Status ModelConfig::parseShape(ShapeInfo& shapeInfo, const std::string& str) {
 
     auto tokens = tokenize(s, shapeDelimeter);
     shapeInfo.shape.clear();
-    std::transform(tokens.begin(), tokens.end(), std::back_inserter(shapeInfo.shape),
-        [](const std::string& str) { return std::stoi(str); });
+    try {
+        std::transform(tokens.begin(), tokens.end(), std::back_inserter(shapeInfo.shape),
+            [](const std::string& str) { return std::stoi(str); });
+    } catch (const std::out_of_range& e) {
+        SPDLOG_DEBUG("ERROR: Parsing model shape string out of range: {}, error: {}", str, e.what());
```
dkalinowski, line 273: Why appending ERROR into the string? If you decide it is error we should use `SPDLOG_ERROR` instead.

intel-rrozestw (changes requested):
change logging level and handle error in addition to logging it.

dkalinowski (approved):

intel-rrozestw (approved):

PR #387, opened 2020-12-03
CVS-39279 fix default version selection

mgumowsk:


rasapala (approved):
LGTM

src/model.hpp
```diff
@@ -38,9 +38,11 @@ class Model {
     mutable std::shared_mutex modelVersionsMtx;
 
     /**
-         * @brief Update default version
-         */
-    void updateDefaultVersion();
+      * @brief Update default version
+      *
+      * @param ignoredVersion Version that should not be taken into account
```
dtrawins, line 43: I would be good to improve this description. Why a version should be ignored?

src/prediction_service_utils.cpp
```diff
@@ -74,24 +72,14 @@ Status getModelInstance(ovms::ModelManager& manager,
...
             return StatusCode::MODEL_VERSION_MISSING;
         }
-        status = modelInstance->waitForLoaded(WAIT_FOR_MODEL_LOADED_TIMEOUT_MS, modelInstanceUnloadGuardPtr);
-        if (status == StatusCode::MODEL_VERSION_NOT_LOADED_ANYMORE) {
-            SPDLOG_INFO("The default version from model {} is retired. Retrying.", modelName);
-        }
     }
 
-    return status;
+    return modelInstance->waitForLoaded(WAIT_FOR_MODEL_LOADED_TIMEOUT_MS, modelInstanceUnloadGuardPtr);
```
dtrawins, line 82: why we need to wait for loading the model to get the modelInstance reference? it would have consequences in many other functions.

mgumowsk: Because we don't want to work on instance that is unstable (eg. during loading), otherwise it would cause inference request to fail.

src/model.hpp
mgumowsk: How would you like to improve the description?

dtrawins: maybe like this: to exclude them from being selected as the default version (for example when the version is to be retired)

src/prediction_service_utils.cpp
dtrawins: I was expecting this method to be used also to report model status. My worry was that it would prevent reporting Loading state. I checked that this method is used only in prediction path so that should be ok.

src/model.hpp
mgumowsk: changed

dtrawins (approved):

PR #400, opened 2020-12-09
CVS-44880 Reset model subscriptions on initial pipeline loading fail

dkalinowski:


atobiszei (approved):
Depending on schedule either:
add test with this MR or
add MR with that test afterwards.

Scenario:
first have config with wrong eg. model_path (non existing) with pipeline in config as well.
Then fix path in config.

dkalinowski:
Unit test added

dtrawins (approved):

atobiszei (approved):

PR #405, opened 2020-12-10
Cvs 44987 jobs param

rasapala:


intel-rrozestw (approved):
LGTM, great idea.

mzegla (commented):
You changed existing `--jobs` parameters, but we should also add `--jobs=$(JOBS)` to `bazel build ` and `bazel test` commands.

mzegla (approved):

PR #409, opened 2020-12-11
Stateful example upgrade

rasapala:


example_client/grpc_stateful_client.py
```diff
@@ -60,10 +71,12 @@ def CalculateUtteranceError(referenceArray, resultArray):
 parser.add_argument('--model_name', default='rm_lstm4f', help='Define model name, must be same as is in service. default: rm_lstm4f',
                     dest='model_name')
 
-parser.add_argument('--utterances', required=False, default=10, help='How many utterances to process from ark file. default 10')
-parser.add_argument('--samples', required=False, default=500, help='How many samples to process from each utterance file. default 500')
+parser.add_argument('--utterances', required=False, default=0, help='How many utterances to process from ark file. default 0 meaning no limit')
+parser.add_argument('--samples', required=False, default=0, help='How many samples to process from each utterance file. default 0 meaning no limit')
 parser.add_argument('--debug', required=False, default=0, help='Enabling debug prints. default 0')
 
+print('### Starting grpc_stateful_client.py client processing ###')
```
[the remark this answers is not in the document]
rasapala: I think it is useful to debug any error messages - thanks to this we know previous errors are not from the client data processing. What do you think ?

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+    print('\tContext window left width cw_l: {}'.format(cw_l))
+    print('\tContext window right width cw_r: {}'.format(cw_r))
+    print('Start processing:')
+    print('\tModel name: {}'.format(args.get('model_name')))
+
+    initial_ark_reader, ark_scores, input_names, output_names, numberOfKeys, inputArrays, referenceArrays = PrepareProcessingData(args)
+
+    SEQUENCE_START = 1
+    SEQUENCE_END = 2
+    sequence_id = 2
```
mzegla, line 185: Can we parametrize it? Since it's sequential and synchronous client can use same ID for all sequences in ARK file. 
We could let user define sequence_id in command line.

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+    meanErrGlobal = 0.0
+
+    for key, obj in initial_ark_reader:
+        batch_size = obj.shape[0]
```
mzegla, line 189: Perhaps `sequence_size`?

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+        print('\n\tInput name: {}'.format(key))
+        printDebug('\tInput in shape: {}'.format(obj.shape))
+        printDebug('\tInput batch size: {}'.format(batch_size))
```
mzegla, line 192: Those prints will show only partial data. Not all inputs will be described here.

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+            request.model_spec.name = args.get('model_name')
+
+            # Set input data index
+            input_index = x
+            # Input for context window
+            if x < cw_l:
+                # Left context window
+                input_index = 0
+            elif x >= cw_l and x < batch_size + cw_l:
+                # Middle window
```
mzegla, line 210: "Middle window" might be misleading. Can we remove this comment?

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+        for x in range(0, batch_size + cw_l + cw_r):
+            printDebug('\tExecution: {}\n'.format(x))
+            request = predict_pb2.PredictRequest()
+            request.model_spec.name = args.get('model_name')
+
+            # Set input data index
+            input_index = x
+            # Input for context window
+            if x < cw_l:
+                # Left context window
```
mzegla, line 207: Perhaps more describing comment would be:
"Filling left side of the context window with the first request"

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+                input_index = 0
+            elif x >= cw_l and x < batch_size + cw_l:
+                # Middle window
+                input_index = x-cw_l
+            else:
+                # Right context window
```
mzegla, line 213: Perhaps more describing comment would be:
"Filling left side of the context window with the first request"

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+                input_index = batch_size - 1
+
+            # Setting request input
+            for input_name in input_names:
+                inputSubArray = inputArrays[input_name]
+                inputData = inputSubArray[key][input_index]
```
mzegla, line 219: Refering to `key` here assumes that all input ark's has utterances with exactly the same names.

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+            if x == 0:
+                request.inputs['sequence_control_input'].CopyFrom(make_tensor_proto(SEQUENCE_START, dtype="uint32"))
+        
+            # Set sequence id
+            request.inputs['sequence_id'].CopyFrom(make_tensor_proto(sequence_id, dtype="uint64"))
 
-    samples_limit = int(args.get('samples'))
-    for x in range(0, batch_size):
+            # Add sequence end
+            if x == batch_size + cw_l + cw_r - 1:
```
mzegla, line 230: When there's only one frame in the sequence it can't be handled with this condition.

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+            if score_index >= 0:
+                # Loop over reference output results
+                errPerNameSum = 0.0
 
-        if x == 0:
-            request.inputs['sequence_control_input'].CopyFrom(make_tensor_proto(SEQUENCE_START, dtype="uint32"))
-        
-        request.inputs['sequence_id'].CopyFrom(make_tensor_proto(sequence_id, dtype="uint64"))
+                for output_name in output_names:
+                    scoreData = referenceArrays[output_name][key][score_index]
```
mzegla, line 248: Refering to `key` here assumes that all input and score arks' have utterances with exactly the same names.

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
-        end_time = datetime.datetime.now()
-        if args['output_name'] not in result.outputs:
-            print("ERROR: Invalid output name", args['output_name'])
-            print("Available outputs:")
-            for Y in result.outputs:
-                print(Y)
-            exit(1)
+                    # Calculate error
+                    meanErr = CalculateUtteranceError(scoreData, resultsArray[0])
+                    errPerNameSum += meanErr
```
mzegla, line 255: To make it more clear please change `meanErr` to `avg_rms_error` as what we are calculating here is Average Root Mean Square Error.

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
 
-        if 'sequence_id' not in result.outputs:
-            print("ERROR: Missing sequence_id in model output")
-            print("Available outputs:")
-            for Y in result.outputs:
-                print(Y)
-            exit(1)
+                    # Statistics
+                    printDebug('Output name: {} Mean error: {:.10f}\n'.format(output_name,meanErr))
```
mzegla, line 258: Please refer to average rms error

mzegla (changes requested):

example_client/grpc_stateful_client.py
```diff
@@ -32,154 +32,250 @@ def printDebug(msg):
...
+    print('\tModel name: {}'.format(args.get('model_name')))
+
+    initial_ark_reader, ark_scores, input_names, output_names, numberOfKeys, inputArrays, referenceArrays = PrepareProcessingData(args)
+
+    SEQUENCE_START = 1
+    SEQUENCE_END = 2
+    sequence_id = 2
+    meanErrGlobal = 0.0
+
+    for key, obj in initial_ark_reader:
```
mzegla, line 188: Can we extract sequence name and sequence size for every sequence since each input and expected output will have the same number of sequences with the same names and the same number of requests/samples in it?

example_client/grpc_stateful_client.py
mzegla: If we don't want to support single request sequences, then you could throw error when user provides such sequence.

example_client/grpc_stateful_client.py
```diff
@@ -26,160 +26,356 @@
...
+        output_name = output_names[name_index]
+        score_ojects = {k: m for k, m in ark_score}
+        reference_scores[output_name] = score_ojects
+        name_index += 1
+
+    # First ark file is the one we will iterate through for all input ark
+    # files per inference request input
+    data_iterator = dict()
+    for key, obj in input_files[0]:
+        data_iterator[key] = obj
```
mzegla, line 195:
```suggestion
    data_iterator = dict()
    for key, obj in input_files[0]:
        data_iterator[key] = obj
```
```suggestion
    sequence_size_map= dict()
    for sequence_name, data in input_files[0]:
        sequence_size_map[sequence_name] = data.shape[0]
```
With that change below code might be more readable.

example_client/grpc_stateful_client.py
```diff
@@ -26,160 +26,356 @@
...
+        # Validate reference output scores data for existing keys and shapes
+        for name in reference_scores:
```
mzegla, line 197: We should have similar validation for inputs.

example_client/grpc_stateful_client.py
```diff
@@ -26,160 +26,356 @@
...
+    SEQUENCE_START = 1
+    SEQUENCE_END = 2
+
+    global_avg_rms_error_sum = 0.0
+
+    # Main inference loop
+    for key in data_iterator:
+        obj = data_iterator[key]
+        sequence_size = obj.shape[0]
+        print('\n\tInput name: {}'.format(key))
```
mzegla, line 259:
```suggestion
        print('\n\Sequence name: {}'.format(key))
```

example_client/grpc_stateful_client.py
```diff
@@ -26,160 +26,356 @@
...
+
+    # Input control tokens
+    SEQUENCE_START = 1
+    SEQUENCE_END = 2
+
+    global_avg_rms_error_sum = 0.0
+
+    # Main inference loop
+    for key in data_iterator:
+        obj = data_iterator[key]
```
mzegla, line 257: You can reduce it to one line:
`for key, obj in data_iterator`

example_client/grpc_stateful_client.py
```diff
@@ -26,160 +26,356 @@
...
+        sequence_size = obj.shape[0]
+        print('\n\tInput name: {}'.format(key))
+        print_debug('\tInput in shape: {}'.format(obj.shape))
```
mzegla, line 260: Can we do this dim expansion here? 
So we get for example "Input in shape: (1, 490)" instead of "Input in shape: (39,490)" ? 
First dimension is sequence size and we already print it explicitly below.

example_client/grpc_stateful_client.py
```diff
@@ -26,160 +26,356 @@
...
+        print_debug('\tInput sequence size: {}'.format(sequence_size))
+        print_debug('\tSequence id: {}'.format(sequence_id))
+
+        if sequence_size == 1:
+            print('\nERROR: Sequence size equal {} in input data must be bigger than 1.'.format(
```
mzegla, line 265: I suggest message like:
"ERROR: Detected sequence with only one frame. Every sequence must contain at least 2 frames"

example_client/grpc_stateful_client.py
```diff
@@ -26,160 +26,356 @@
...
+        for x in range(0, sequence_size + cw_l + cw_r):
+            print_debug('\tExecution: {}\n'.format(x))
+            request = predict_pb2.PredictRequest()
+            request.model_spec.name = args.get('model_name')
+
+            # Set input data index
+            input_index = x
+            # Input for context window
+            if x < cw_l:
+                # Repeating first input data infer request cw_l times for the
```
mzegla, line 281: Maybe instead of "first input data infer request" we could go with "first frame"?

example_client/grpc_stateful_client.py
```diff
@@ -26,160 +26,356 @@
...
+                # context window model start
+                input_index = 0
+            elif x >= cw_l and x < sequence_size + cw_l:
+                # Standard processing
+                input_index = x - cw_l
+            else:
+                # Repeating last input data infer request cw_r times for the
```
mzegla, line 288: Maybe instead of "last input data infer request" we could go with "last frame"?

mzegla (approved):

PR #419, opened 2020-12-16
CVS-45311Add class to hold additional information necessary during request processing

mzegla:
Introducing new classes to wrap data that needs to be passed back and forth between methods during request handling. 
Added ProcessingSpec that for now will hold only SequenceProcessingSpec, but might be used to keep other types of data that we may need to maintain throughout request processing. 
It contains pointer to SequenceProcessingSpec type that might point to valid object if request is processed by stateful model or nullptr if request is processed by stateless model.

### DIFF
--- a/src/BUILD
+++ b/src/BUILD
@@ -80,6 +80,7 @@ cc_library(
         "prediction_service.hpp",
         "prediction_service_utils.hpp",
         "prediction_service_utils.cpp",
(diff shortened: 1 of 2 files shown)


rasapala (approved):
LGTM

michalkulakowski (approved):

PR #431, opened 2020-12-20
CVS-44761 fix handling errors in adding version

dtrawins:


src/model.cpp
```diff
@@ -130,9 +130,12 @@ Status Model::addVersion(const ModelConfig& config) {
     return StatusCode::OK;
 }
 
-Status Model::addVersions(std::shared_ptr<model_versions_t> versionsToStart, ovms::ModelConfig& config, std::shared_ptr<FileSystem>& fs) {
+Status Model::addVersions(std::shared_ptr<model_versions_t> versionsToStart, ovms::ModelConfig& config, std::shared_ptr<FileSystem>& fs, std::shared_ptr<model_versions_t>& versionsFailed) {
```
dkalinowski, line 133: Inconsistent parameter passing, versionToStart is` shared_ptr` but versionsFailed is `shared_ptr&` (reference)

src/model.cpp
```diff
@@ -130,9 +130,12 @@ Status Model::addVersion(const ModelConfig& config) {
...
     Status result = StatusCode::OK;
     downloadModels(fs, config, versionsToStart);
+    SPDLOG_INFO("clearning vector versionsFailed");
```
dkalinowski, line 136: Do we need these logs?

src/model.cpp
```diff
@@ -192,7 +196,7 @@ void Model::retireAllVersions() {
     subscriptionManager.notifySubscribers();
 }
 
-Status Model::reloadVersions(std::shared_ptr<model_versions_t> versionsToReload, ovms::ModelConfig& config, std::shared_ptr<FileSystem>& fs) {
+Status Model::reloadVersions(std::shared_ptr<model_versions_t> versionsToReload, ovms::ModelConfig& config, std::shared_ptr<FileSystem>& fs, std::shared_ptr<model_versions_t>& versionsFailed) {
```
dkalinowski, line 199: Inconsistent parameter passing aswell

src/model.cpp
```diff
@@ -224,6 +228,8 @@ Status Model::reloadVersions(std::shared_ptr<model_versions_t> versionsToReload,
                 version,
                 status.string());
             result = status;
+            // modelVersion->unloadModel();  // invalidate version when reloading fails due to corrupted or missing model files
```
dkalinowski, line 231: To remove?

src/modelmanager.cpp
```diff
@@ -661,15 +661,16 @@ Status ModelManager::reloadModelWithVersions(ModelConfig& config) {
...
+    Status blocking_status = readAvailableVersions(fs, config.getBasePath(), availableVersions);
     if (!blocking_status.ok()) {
         return blocking_status;
     }
-    requestedVersions = config.getModelVersionPolicy()->filter(requestedVersions);
+    auto requestedVersions = config.getModelVersionPolicy()->filter(availableVersions);
     std::shared_ptr<model_versions_t> versionsToStart;
     std::shared_ptr<model_versions_t> versionsToReload;
     std::shared_ptr<model_versions_t> versionsToRetire;
+    std::shared_ptr<model_versions_t> versionsFailed = std::make_shared<model_versions_t>(0);
```
dkalinowski, line 673: If intention was to create empty vector, then you do not need to pass `0`, just call:  
`std::make_shared<model_versions_t>()`

src/modelmanager.cpp
```diff
@@ -698,18 +699,39 @@ Status ModelManager::reloadModelWithVersions(ModelConfig& config) {
             return StatusCode::OK;
         }
     }
+
     getVersionsToChange(config, model->getModelVersions(), requestedVersions, versionsToStart, versionsToReload, versionsToRetire);
 
-    if (versionsToStart->size() > 0) {
-        auto blocking_status = addModelVersions(model, fs, config, versionsToStart);
+    while (versionsToStart->size() > 0) {
```
dkalinowski, line 705: Will OVMS hang if user starts application with one corrupted model version?

src/test/modelmanager_test.cpp
```diff
@@ -350,6 +351,90 @@ TEST(ModelManager, ConfigReloadingWithWrongInputName) {
     ASSERT_EQ(status, ovms::StatusCode::CONFIG_SHAPE_IS_NOT_IN_NETWORK);
 }
 
+class dummyModel {
```
dkalinowski, line 354: Use `DummyModel` name convention

src/test/modelmanager_test.cpp
```diff
@@ -350,6 +351,90 @@ TEST(ModelManager, ConfigReloadingWithWrongInputName) {
...
+private:
+    std::string model_source_path;
```
dkalinowski, line 356: Use camel case: `modelSourcePath`

src/test/modelmanager_test.cpp
dkalinowski: This is not really a model - more like `DummyModelDirectoryCreator`?

src/test/modelmanager_test.cpp
```diff
@@ -350,6 +351,90 @@ TEST(ModelManager, ConfigReloadingWithWrongInputName) {
...
+    ASSERT_EQ(status, ovms::StatusCode::OK);
+    ASSERT_EQ(modelInstance2->getStatus().getState(), ovms::ModelVersionState::AVAILABLE);
+    modelInstanceUnloadGuard.reset();
+    status = ovms::getModelInstance(manager, model.name, 3, modelInstance3, modelInstanceUnloadGuard);
+    modelInstanceUnloadGuard.reset();
+    ASSERT_EQ(status, ovms::StatusCode::MODEL_VERSION_MISSING);
+
+    model.removeVersion(3);
+    model.removeVersion(2);
+    std::cout << "Removed versions 3 and 2" << std::endl;
```
dkalinowski, line 411: Remove std::couts

src/modelmanager.cpp
mzegla, line 705: Why are we using while here? Is second iteration possible here?
If it's just to check condition maybe we don't need a loop

dtrawins: multiple loops are possible. In situation when several latest versions will be invalid, there will be attempt to add the latests only. When it fails, ovms need to reassess the version policy excluding the failed latest version. It means  the second loop which can still include corrupted version. There could be as many loops as many corrupted versions get detected. We should continue the loop till there are no more versions to add (all corrupted) or the load is successful.

src/model.cpp
```diff
@@ -130,9 +130,10 @@ Status Model::addVersion(const ModelConfig& config) {
     return StatusCode::OK;
 }
 
-Status Model::addVersions(std::shared_ptr<model_versions_t> versionsToStart, ovms::ModelConfig& config, std::shared_ptr<FileSystem>& fs) {
+Status Model::addVersions(std::shared_ptr<model_versions_t> versionsToStart, ovms::ModelConfig& config, std::shared_ptr<FileSystem>& fs, std::shared_ptr<model_versions_t> versionsFailed) {
     Status result = StatusCode::OK;
     downloadModels(fs, config, versionsToStart);
+    versionsFailed->clear();
```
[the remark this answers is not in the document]
dtrawins: In situation when we have many corrupted versions to add there might be many iterations in the loop trying to add new versions. The clear operation is to avoid removing previously failed versions from availableVersions. It would be duplicated because they were removed from available versions earlier. In means versionFailed should include just the versions failed in the current addVersions call.

mzegla (approved):

dkalinowski (approved):

PR #437, opened 2020-12-23
[WIP]Add a mention about config numerical values handling inconsistency to…

michalkulakowski:
… docs


mzegla (approved):

docs/docker_container.md
```diff
@@ -206,6 +206,7 @@ Here the numerical values depict the version number of the model.
 | `"target_device"` | `"CPU"/"HDDL"/"GPU"/"NCS"/"MULTI"/"HETERO"` |  Device name to be used to execute inference operations. Refer to AI accelerators support below. ||
 
 #### To know more about batch size and shape parameters refer [Batch Size and Shape document](shape_and_batch_size.md)
+#### NOTICE: For CLI numerical values in quotes are allowed but for JSON config aren't.
```
dtrawins, line 209: I don't understand this notice. Maybe it would be clearer with:
`In OVMS CLI all numerical parameters are allowed to be with or without quotes. The configuration file must be compliant with the json schema so numerical values should be without quotes.`

PR #440, opened 2020-12-23
Cvs 45355 stateful config

rasapala:


src/test/stateful_config_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+        {
+            "config": {
+                "name": "dummy",
+                "base_path": "/ovms/src/test/dummy",
+                "target_device": "CPU",
+                "model_version_policy": {"latest": {"num_versions":1}},
+                "nireq": 100,
+                "stateful": true,
+                "low_latency_transformation": true,
+                "sequence_timeout": 120,
```
dtrawins, line 83: add the unit to the name, it should be clear if the timeout is in sec, h or ms.

src/test/stateful_config_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+                "max_sequence_number": 1000,
```
dtrawins, line 84: why do we need to limit the sequence length?

rasapala: The stateful pipeline parameter setting was added in the design so that user has the possibility to control sequences.

mzegla: It's actually not sequence length limit. Sequence can have unlimited amount of requests. If user doesn't send SEQUENCE_END signal the sequence can potentially be infinite. 

**max_sequence_number** parameter purpose is to **limit the number of sequences handled by the model instance concurrently**. As handling each sequence (holding its state or whole infer request) requires resources, setting this parameter can limit number of sequences that can be processed concurrently by the model, so it doesn't consume too much memory.

src/test/stateful_config_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+        {
+            "config": {
+                "name": "dummy",
+                "base_path": "/ovms/src/test/dummy",
+                "target_device": "CPU",
+                "model_version_policy": {"latest": {"num_versions":1}},
+                "nireq": 100,
+                "stateful": true,
+                "low_latency_transformation": true,
+                " sequence_timeout_seconds": 120,
```
mzegla, line 83:
```suggestion
                "sequence_timeout_seconds": 120,
```

mzegla (approved):

dtrawins (approved):

PR #443, opened 2020-12-28
Cvs 45498 blob copy

rasapala:


src/ov_utils.cpp
```diff
@@ -64,4 +64,46 @@ Status blobClone(InferenceEngine::Blob::Ptr& destinationBlob, const InferenceEng
     return StatusCode::OK;
 }
 
+Status constBlobClone(InferenceEngine::Blob::Ptr& destinationBlob, const InferenceEngine::Blob::CPtr sourceBlob) {
```
dkalinowski, line 67: We could make the second parameter templated, since both implementations are the same?

dkalinowski (approved):

kromanow94 (approved):
I'm approving this PR because there is no other dev team member available.
From my understanding everything seems all right.
I trust @dkalinowski and @rasapala that this code was reviewed accoarding to present coding standards.

Live long and prosper 🖖

PR #450, opened 2021-01-11
CVS-46541 Implement Sequence class

mzegla:


src/sequence.cpp
```diff
@@ -0,0 +1,58 @@
+//*****************************************************************************
+// Copyright 2020 Intel Corporation
```
rasapala, line 2: 2021

src/sequence.hpp
```diff
@@ -0,0 +1,52 @@
+//*****************************************************************************
+// Copyright 2020 Intel Corporation
```
rasapala, line 2: 2021

michalkulakowski (approved):

src/test/sequence_test.cpp
```diff
@@ -0,0 +1,33 @@
...
+//*****************************************************************************
+#include <memory>
+#include <mutex>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "../sequence.hpp"
+
+TEST(Sequence, MovedMutexNullified) {
```
rasapala, line 24: I think adding ULT for  updateMemoryState is crucial here. Also we should check last modification time was updated.

src/test/sequence_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+    std::vector<float> state1(elementsCount1);
+    std::iota(state1.begin(), state1.end(), 0);
+    addState(newState, "state1", shape1, state1);
+
+    ovms::Sequence sequence;
+    auto time1 = sequence.getLastActivityTime();
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    sequence.updateMemoryState(newState);
+    auto time2 = sequence.getLastActivityTime();
+    EXPECT_NE(time1, time2);
```
rasapala, line 98: Check grather than

rasapala (approved):
LGTM

PR #463, opened 2021-01-18
Cvs 46546 stateful input tests

rasapala:


src/statefulmodelinstance.cpp
```diff
@@ -19,16 +19,20 @@ using namespace InferenceEngine;
 
 namespace ovms {
 
-uint64_t extractSequenceId(const tensorflow::TensorProto& proto) {
-    if (proto.uint64_val_size() == 1)
-        return proto.uint64_val(0);
-    return 0;
+const Status extractSequenceId(const tensorflow::TensorProto& proto, uint64_t* sequence_id) {
```
mzegla, line 22: Can we use reference instead of a pointer in `sequence_id`? 
Also please use camel case - `sequenceId`

src/statefulmodelinstance.cpp
```diff
@@ -19,16 +19,20 @@ using namespace InferenceEngine;
...
+        return StatusCode::OK;
+    }
+    return StatusCode::SEQUENCE_ID_BAD_TYPE;
 }
 
-uint32_t extractSequenceControlInput(const tensorflow::TensorProto& proto) {
-    if (proto.uint32_val_size() == 1)
-        return proto.uint32_val(0);
-    return 0;
+const Status extractSequenceControlInput(const tensorflow::TensorProto& proto, uint32_t* sequenceControlInput) {
```
mzegla, line 30: Can we use reference instead of a pointer in sequenceControlInput?

src/status.cpp
```diff
@@ -53,6 +53,8 @@ const std::map<const StatusCode, const std::string> Status::statusMessageMap = {
     {StatusCode::SEQUENCE_ALREADY_EXISTS, "Sequence with provided ID already exists"},
     {StatusCode::SEQUENCE_ID_NOT_PROVIDED, "Sequence ID has not been provided in request inputs"},
     {StatusCode::INVALID_SEQUENCE_CONTROL_INPUT, "Unexpected value of sequence control input"},
+    {StatusCode::SEQUENCE_ID_BAD_TYPE, "Sequence ID has been provided with bad proto type"},
+    {StatusCode::SEQUENCE_CONTROL_INPUT_BAD_TYPE, "Sequence control input was provided with bad proto type"},
```
mzegla, line 57: It may be misleading since we get the same code when those values are not passed at all. So maybe a better message would be something like:
"Could not find sequence id / sequence control input in expected tensor proto field - uint64_val / uint32_val" 
This would cover both situations - when value is not provided at all and when it's been written to unexpected field like tensor_content.

mzegla (approved):

michalkulakowski (approved):

PR #465, opened 2021-01-19
Cvs 46549 stateful postprocessing

rasapala:


src/test/stateful_modelinstance_test.cpp
```diff
@@ -264,37 +285,92 @@ TEST_F(StatefulModelInstance, PreprocessingFirstRequest) {
...
+
+    std::vector<float> currentBlobIrData;
+    currentBlobIrData.assign((float*)stateCloneBlob->buffer(), ((float*)stateCloneBlob->buffer()) + elementsCount);
+    EXPECT_EQ(currentBlobIrData, currentState);
+
+    tensorflow::serving::PredictResponse response;
+
+    modelInstance->postInferenceProcessing(&response, inferRequest, sequenceProcessingSpec);
+
+    auto& output = (*response.mutable_outputs())["sequence_id"];
```
mzegla, line 359: This might be confusing because if response does not contain `sequence_id` entry, it will create one.
I think we should check if (*response.mutable_outputs()) contains `sequence_id` beforehand.

src/test/stateful_modelinstance_test.cpp
```diff
@@ -315,15 +391,20 @@ TEST_F(StatefulModelInstance, PreprocessingIntermediateRequest) {
...
-        modelInstance.preInferenceProcessing(inferRequest, sequenceProcessingSpec);
+        tensorflow::serving::PredictResponse response;
+
+        modelInstance->postInferenceProcessing(&response, inferRequest, sequenceProcessingSpec);
+
+        auto& output = (*response.mutable_outputs())["sequence_id"];
+        EXPECT_EQ(output.uint64_val_size(), 1);
+        EXPECT_EQ(output.uint64_val(0), sequenceId);
 
         // Check if InferRequest memory state has been updated to sequence memory state
```
mzegla, line 404: In case of postprocessing InferRequest memory state does not get updated, sequence memory state does. Here we should check if sequence contains the same memory state as InferRequest.

mzegla (approved):

michalkulakowski (approved):

PR #468, opened 2021-01-20
Stateful tests rewrite

rasapala:


src/test/stateful_modelinstance_test.cpp
```diff
@@ -81,12 +82,37 @@ class StatefulModelInstanceTempDir : public TestWithTempDir {
...
         modelInput.clear();
     }
 };
 
+class MockedValidateStatefulModelInstance : public ovms::StatefulModelInstance {
+public:
+    MockedValidateStatefulModelInstance(const std::string& name, ovms::model_version_t version) :
+        StatefulModelInstance(name, version) {}
+
+    const ovms::Status mockValidate(const tensorflow::serving::PredictRequest* request, ovms::ProcessingSpec* processingSpecPtr) {
```
mzegla, line 111: Can we use SequenceProcessingSpec reference? 
If necessary please change validate method to take SequenceProcessingSpec& instead of ProcessingSpec*

mzegla (commented):
Can we also move validate() method in ModelInstance to protected and test it like we test it StatefulModelInstance?

mzegla (approved):

michalkulakowski (approved):

PR #470, opened 2021-01-20
CVS-46551 StatefulModelInstance - implement infer method

mzegla:


src/processing_spec.hpp
```diff
@@ -19,12 +19,19 @@
 
 namespace ovms {
 
-struct SequenceProcessingSpec {
-    const uint32_t sequenceControlInput;
-    const uint64_t sequenceId;
+class SequenceProcessingSpec {
+    uint32_t sequenceControlInput;
```
rasapala, line 23: We do not need const no more ?

src/sequence_manager.cpp
```diff
@@ -79,6 +79,55 @@ Status SequenceManager::removeTimedOutSequences(std::chrono::steady_clock::time_
     return StatusCode::OK;
 }
 
+Status SequenceManager::hasSequence(SequenceProcessingSpec sequenceProcessingSpec, std::unique_ptr<std::mutex>* sequenceMutexPtr) {
+    const uint64_t& sequenceId = sequenceProcessingSpec.getSequenceId();
+    if (!sequenceExists(sequenceId))
+        return StatusCode::SEQUENCE_MISSING;
+    sequenceMutexPtr = sequences.at(sequenceId).getMutexPtr();
+    if (*sequenceMutexPtr == nullptr)
+        return StatusCode::SEQUENCE_ENDED;
```
rasapala, line 88: FINISHED ?

src/processing_spec.hpp
mzegla: Because now it's a private member and we provide public getter and setter for it.

src/sequence_manager.cpp
```diff
@@ -79,6 +79,63 @@ Status SequenceManager::removeTimedOutSequences(std::chrono::steady_clock::time_
...
 
+Status SequenceManager::hasSequence(const uint64_t& sequenceId, MutexPtr& sequenceMutexPtr) {
+    if (!sequenceExists(sequenceId))
+        return StatusCode::SEQUENCE_MISSING;
+
+    if (sequences.at(sequenceId).isTerminated())
+        return StatusCode::SEQUENCE_TERMINATED;
+
+    sequenceMutexPtr = sequences.at(sequenceId).getMutexPtr();
+
```
michalkulakowski, line 90: unnecessary empty line

michalkulakowski (approved):

PR #472, opened 2021-01-21
Cvs 45632 refactor config reloading

michalkulakowski:


rasapala (approved):
LGTM

src/modelmanager.cpp
```diff
@@ -482,10 +482,13 @@ void ModelManager::watcher(std::future<void> exit) {
     while (exit.wait_for(std::chrono::milliseconds(1)) == std::future_status::timeout) {
         std::this_thread::sleep_for(std::chrono::seconds(watcherIntervalSec));
         SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Watcher thread check cycle begin");
+
+        std::lock_guard<std::recursive_mutex> loadingLock(configMtx);
```
mzegla, line 486: Maybe we can use `std::unique_lock` if we want to manually unlock this mutex it later?

mzegla (approved):

PR #476, opened 2021-01-21
Cvs 47114 laod model impl

rasapala:


src/statefulmodelinstance.hpp
```diff
@@ -30,18 +30,21 @@ const uint32_t SEQUENCE_END = 2;
...
     /**
          * @brief A default constructor
          */
     StatefulModelInstance(const std::string& name, model_version_t version) :
-        ModelInstance(name, version) {}
+        ModelInstance(name, version) {
+        sequenceManager = std::make_unique<SequenceManager>(config.getSequenceTimeout(), config.getMaxSequenceNumber());
+    }
 
     SequenceManager& getSequenceManager() {
```
mzegla, line 45: Can we return reference to the pointer that we store as StatefulModelInstance member?

mzegla (approved):

michalkulakowski (approved):

PR #487, opened 2021-02-02
CVS-46632 Parse stateful specific inputs in http handler

mzegla:


src/rest_parser.cpp
```diff
@@ -49,6 +49,41 @@ void RestParser::removeUnusedInputs() {
...
+            proto.add_uint32_val(value.GetUint());
+        else
+            return false;
+    }
+    return true;
+}
+
+bool RestParser::parseSpecialInput(rapidjson::Value& doc, tensorflow::TensorProto& proto, const std::string& tensorName) {
+    // Special tensors are given in 1 dimentional array
+    if (doc.GetArray()[0].IsArray())
```
dkalinowski, line 76: Should we ensure that GetArray returns an array with at least one element?

src/statefulmodelinstance.cpp
```diff
@@ -27,15 +27,41 @@ using namespace InferenceEngine;
...
+
     if (proto.uint64_val_size() == 1) {
         sequenceId = proto.uint64_val(0);
         return StatusCode::OK;
     }
     return StatusCode::SEQUENCE_ID_BAD_TYPE;
 }
 
-const Status extractSequenceControlInput(const tensorflow::TensorProto& proto, uint32_t& sequenceControlInput) {
+const Status StatefulModelInstance::extractSequenceControlInput(const tensorflow::TensorProto& proto, uint32_t& sequenceControlInput) {
```
dkalinowski, line 51: These methods: `extractSequenceControlInput` and `extractSequenceId` are very similar and can be used for different purposes than handling control input/sequence id too. If the only difference is type (uint32 vs uint64) then maybe name them accordingly?

dkalinowski: Like `extractOneUint32` and `extractOneUint64`?

src/rest_utils.cpp
```diff
@@ -48,59 +57,107 @@ Status makeJsonFromPredictResponse(
...
+                    return status;
+            } else {
+                for (size_t i = 0; i < tensor.tensor_content().size(); i += sizeof(uint8_t))
+                    tensor.add_int_val(*reinterpret_cast<uint8_t*>(tensor.mutable_tensor_content()->data() + i));
             }
             break;
         case DataType::DT_INT64:
-            for (size_t i = 0; i < tensor.tensor_content().size(); i += sizeof(int64_t)) {
-                tensor.add_int64_val(*reinterpret_cast<int64_t*>(tensor.mutable_tensor_content()->data() + i));
+            if (seekDataInValField) {
```
dkalinowski, line 134: Please add additional unit tests for `makeJsonFromPredictResponse`

dkalinowski: `rest_utils_test.cpp`

rasapala (approved):
LGTM

src/statefulmodelinstance.cpp
mzegla: I don't see any other usage of these methods than in sequence_id and sequence_control_input extraction in the near future. 
Can we stay with those very specific names to make the purpose of these methods obvious for now and when any other use case for these methods appears we would rename them?

dkalinowski: Seems fair

dkalinowski (approved):

PR #488, opened 2021-02-03
add control api function

michalkulakowski:


src/http_rest_api_handler.cpp
```diff
@@ -325,4 +326,36 @@ Status HttpRestApiHandler::processModelStatusRequest(
...
+    Status status;
+    auto& config = ovms::Config::instance();
+    auto& manager = ModelManager::getInstance();
+    bool isConfigFileReloadNeeded = manager.configFileReloadNeeded();
+    if (isConfigFileReloadNeeded) {
+        status = manager.loadConfig(config.configPath());
+        if (!status.ok()) {
+            return status;
+        }
+    }
```
mzegla, line 340: What if config didn't change? Don't we want to scan filesystem for changes in models versions?
In case user didn't set periodic filesystem scans.

src/status.hpp
```diff
@@ -202,6 +202,9 @@ enum class StatusCode {
     NODE_LIBRARY_LOAD_FAILED_SYM,
     NODE_LIBRARY_MISSING,
 
+    // Model control API
+    OK_RELOAD_NOT_NEEDED, /*!< Operation succeeded but no config reload was needed */
```
mzegla, line 206: Can we change name to make it clear that config reload was not required?
Not to confuse with model reload (due to changes in versions or dynamic shape/batch size)

src/test/http_rest_api_handler_test.cpp
```diff
@@ -0,0 +1,82 @@
...
+    ovms::Config::instance().parse(arg_count, n_argv);
+
+    auto handler = ovms::HttpRestApiHandler(10);
+    std::string response;
+    auto status = handler.processModelControlApiRequest(response);
+
+    EXPECT_EQ(status, ovms::StatusCode::FILE_INVALID);
+}
+
+TEST(ModelControlApi, positive) {
```
mzegla, line 54: Do we have test to check if reload actually happened? (check the response content?)

mzegla (approved):

rasapala (approved):
LGTM

PR #492, opened 2021-02-05
Stateful model instance - infer synchronization issue fix

mzegla:


rasapala (approved):
LGTM

src/sequence.cpp
```diff
@@ -25,6 +25,10 @@ void Sequence::updateLastActivityTime() {
     lastActivityTime = std::chrono::steady_clock::now();
 }
 
+const uint64_t& Sequence::getId() const {
```
dkalinowski, line 28: Uint64 is not a big overhead to copy. Maybe just return `uint64_t`?

dkalinowski: This also applies to all other functions that take and return const reference to uint64_t which I don't think is neccessary

src/sequence_manager.cpp
```diff
@@ -79,67 +56,67 @@ Status SequenceManager::removeTimedOutSequences(std::chrono::steady_clock::time_
...
-    sequenceMutexPtr = sequences.at(sequenceId).getMutexPtr();
-    if (sequenceMutexPtr == nullptr)
-        return StatusCode::INTERNAL_ERROR;
-
     return StatusCode::OK;
 }
 
-Status SequenceManager::createSequence(const uint64_t& sequenceId, MutexPtr& sequenceMutexPtr) {
+Status SequenceManager::createSequence(const uint64_t& sequenceId) {
     /* TO DO: Generate unique ID if not provided by the client
```
dkalinowski, line 70: Remove space from todo word so SDL checks can detect it before release

src/sequence_manager.cpp
```diff
@@ -79,67 +56,67 @@ Status SequenceManager::removeTimedOutSequences(std::chrono::steady_clock::time_
...
-    auto status = addSequence(sequenceId);
-    if (!status.ok())
-        return status;
-
-    sequenceMutexPtr = sequences.at(sequenceId).getMutexPtr();
-    if (sequenceMutexPtr == nullptr)
-        return StatusCode::INTERNAL_ERROR;
-
+    if (sequenceExists(sequenceId)) {
+        spdlog::debug("Sequence with provided ID already exists");
```
dkalinowski, line 75: Use SPDLOG_LOGGER_DEBUG

src/sequence_manager.cpp
```diff
@@ -79,67 +56,67 @@ Status SequenceManager::removeTimedOutSequences(std::chrono::steady_clock::time_
...
+    } else {
+        spdlog::debug("Sequence with provided ID does not exists");
+        return StatusCode::SEQUENCE_MISSING;
+    }
     return StatusCode::OK;
 }
 
-Status SequenceManager::getSequenceMutexPtr(SequenceProcessingSpec& sequenceProcessingSpec, MutexPtr& sequenceMutexPtr) {
+Status SequenceManager::processRequestedSpec(SequenceProcessingSpec& sequenceProcessingSpec) {
     const uint32_t& sequenceControlInput = sequenceProcessingSpec.getSequenceControlInput();
```
dkalinowski, line 109: I don't see a point why we use reference instead of copy here

src/statefulmodelinstance.cpp
```diff
@@ -107,13 +107,16 @@ Status StatefulModelInstance::infer(const tensorflow::serving::PredictRequest* r
...
         return status;
 
-    MutexPtr sequenceMutexPtr = nullptr;
     std::unique_lock<std::mutex> sequenceManagerLock(sequenceManager->getMutex());
-    status = sequenceManager->getSequenceMutexPtr(sequenceProcessingSpec, sequenceMutexPtr);
+    status = sequenceManager->processRequestedSpec(sequenceProcessingSpec);
     if (!status.ok())
         return status;
+    const uint64_t& sequenceId = sequenceProcessingSpec.getSequenceId();
+    if (!sequenceManager->sequenceExists(sequenceId))
```
dkalinowski, line 115: We have 2 methods that do the same - `sequenceExists` and `hasSequence`. In HasSequence you return INTERNAL_ERROR if sequence is missing, why not to use it here? And get rid of one of the methods.

mzegla: sequenceExists is the low level method to just check if element exists in the sequences container (returning true or false).

hasSequence has more context and returns appropriate status depending on sequence presence in sequences container as well as its termination status (if the last request of the sequence is being processed, the sequence is not available for any other requests).

dkalinowski (approved):

PR #498, opened 2021-02-09
Cvs 46636 stateful timeout

rasapala:


src/sequence_manager.cpp
```diff
@@ -44,7 +46,9 @@ bool SequenceManager::sequenceExists(const uint64_t sequenceId) const {
     return sequences.count(sequenceId);
 }
 
-Status SequenceManager::removeTimedOutSequences(std::chrono::steady_clock::time_point currentTime) {
+Status SequenceManager::removeTimedOutSequences() {
+    std::unique_lock<std::mutex> sequenceManagerLock(mutex);
+    std::chrono::steady_clock::time_point currentTime = std::chrono::steady_clock::now();
     for (auto it = sequences.cbegin(); it != sequences.cend();) {
         auto& sequence = it->second;
         auto timeDiff = currentTime - sequence.getLastActivityTime();
```
mzegla, line 54: Here we access sequence properties. Moreover, in the next line we remove the sequence from sequences container. 
What if there is ongoing inference on that sequence? Locking sequence manager will not prevent removal of sequence that is being used.

src/sequence_manager.cpp
```diff
@@ -44,25 +46,50 @@ bool SequenceManager::sequenceExists(const uint64_t sequenceId) const {
...
 }
 
-Status SequenceManager::removeTimedOutSequences(std::chrono::steady_clock::time_point currentTime) {
+Status SequenceManager::removeTimedOutSequences() {
+    std::unique_lock<std::mutex> sequenceManagerLock(mutex);
     for (auto it = sequences.cbegin(); it != sequences.cend();) {
-        auto& sequence = it->second;
-        auto timeDiff = currentTime - sequence.getLastActivityTime();
-        if (std::chrono::duration_cast<std::chrono::seconds>(timeDiff).count() > timeout)
+        Sequence& seqRef = getSequence(it->second.getId());
```
mzegla, line 54:
```suggestion
        Sequence& sequence = getSequence(it->second.getId());
```
I think `sequence` is enough. I think we use it everywhere else in the code instead of `seqRef `or `sequenceRef`.

src/sequence_manager.cpp
```diff
@@ -44,25 +46,50 @@ bool SequenceManager::sequenceExists(const uint64_t sequenceId) const {
...
+        std::unique_lock<std::mutex> sequenceLock(seqRef.getMutex());
+        if (seqRef.isTimedout()) {
+            SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Removing timeouted sequence - Id: {}", seqRef.getId());
             it = sequences.erase(it);
-        else
+            sequenceLock.unlock();
```
mzegla, line 59: This can result in undefined behavior. You delete mutex while still locking it.

src/sequence_manager.cpp
```diff
@@ -44,25 +46,50 @@ bool SequenceManager::sequenceExists(const uint64_t sequenceId) const {
...
             ++it;
     }
     return StatusCode::OK;
 }
 
+Status SequenceManager::checkForTimedOutSequences() {
+    std::unique_lock<std::mutex> sequenceManagerLock(mutex);
+    std::chrono::steady_clock::time_point currentTime = std::chrono::steady_clock::now();
+    for (auto it = sequences.cbegin(); it != sequences.cend();) {
+        Sequence& seqRef = getSequence(it->second.getId());
```
mzegla, line 71:
```suggestion
        Sequence& sequence = getSequence(it->second.getId());
```

src/sequence.hpp
```diff
@@ -39,13 +39,15 @@ class Sequence {
     std::chrono::steady_clock::time_point lastActivityTime;
     std::mutex mutex;
     bool terminated;
+    bool timedout;
```
mzegla, line 42:
```suggestion
    bool timedOut;
```

src/sequence.hpp
```diff
@@ -54,6 +56,8 @@ class Sequence {
     std::mutex& getMutex();
     bool isTerminated() const;
     void setTerminated();
+    bool isTimedout() const;
```
mzegla, line 59:
```suggestion
    bool isTimedOut() const;
```

src/sequence.hpp
```diff
@@ -54,6 +56,8 @@ class Sequence {
...
+    void setTimedout();
```
mzegla, line 60:
```suggestion
    void setTimedOut();
```

michalkulakowski (approved):

src/sequence_manager.cpp
```diff
@@ -44,14 +46,35 @@ bool SequenceManager::sequenceExists(const uint64_t sequenceId) const {
...
+    return StatusCode::OK;
+}
+
+Status SequenceManager::checkForTimedOutSequences() {
+    std::unique_lock<std::mutex> sequenceManagerLock(mutex);
+    std::chrono::steady_clock::time_point currentTime = std::chrono::steady_clock::now();
+    for (auto it = sequences.cbegin(); it != sequences.cend();) {
+        Sequence& sequence = getSequence(it->second.getId());
+        std::unique_lock<std::mutex> sequenceLock(sequence.getMutex());
+        if (!sequence.isTerminated()) {
```
mzegla, line 69: Sequence gets set to terminated state in infer method during special inputs processing phase with sequence manager mutex locked. Since it doesn't get set to terminated state during request processing (within sequence critical section) I think we don't need to lock sequence before checking if it's terminated (we don't in infer() method). 
Can we move this condition up so it gets check before sequence locking attempt?

mzegla (changes requested):

src/sequence_manager.cpp
```diff
@@ -44,14 +46,35 @@ bool SequenceManager::sequenceExists(const uint64_t sequenceId) const {
...
 }
 
-Status SequenceManager::removeTimedOutSequences(std::chrono::steady_clock::time_point currentTime) {
+Status SequenceManager::removeTimedOutSequences() {
+    std::unique_lock<std::mutex> sequenceManagerLock(mutex);
     for (auto it = sequences.cbegin(); it != sequences.cend();) {
-        auto& sequence = it->second;
-        auto timeDiff = currentTime - sequence.getLastActivityTime();
-        if (std::chrono::duration_cast<std::chrono::seconds>(timeDiff).count() > timeout)
+        Sequence& sequence = getSequence(it->second.getId());
```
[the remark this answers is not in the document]
rasapala: it is const, cant modify it afterwards.

src/sequence_manager.cpp
```diff
@@ -44,14 +46,35 @@ bool SequenceManager::sequenceExists(const uint64_t sequenceId) const {
...
+        }
+    }
+    return StatusCode::OK;
+}
+
+Status SequenceManager::checkForTimedOutSequences() {
+    std::unique_lock<std::mutex> sequenceManagerLock(mutex);
+    std::chrono::steady_clock::time_point currentTime = std::chrono::steady_clock::now();
+    for (auto it = sequences.cbegin(); it != sequences.cend();) {
+        Sequence& sequence = getSequence(it->second.getId());
```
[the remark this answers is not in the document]
rasapala: it is const, cant modify it afterwards.

src/sequence_manager.hpp
```diff
@@ -71,8 +94,34 @@ class SequenceManager {
 
     Status removeSequence(const uint64_t sequenceId);
 
-    Status removeTimedOutSequences(std::chrono::steady_clock::time_point currentTime);
+    Status checkForTimedOutSequences();
```
mzegla, line 97: Why changing the method name? It seems now that we are actually removing timed out sequences in that method so previous name looks like a better fit to me.

src/sequence_manager.hpp
```diff
@@ -47,9 +55,24 @@ class SequenceManager {
 
 public:
     SequenceManager() = default;
-    SequenceManager(uint32_t timeout, uint32_t maxSequenceNumber) :
+    SequenceManager(uint32_t timeout, uint32_t maxSequenceNumber, std::string modelName) :
```
mzegla, line 58: If you're adding modelName to sequence manager fields for logging purposes you could also add model version here.
Or even pass it as std::pair to have less parameters and track the together.

src/sequence.hpp
```diff
@@ -54,6 +52,7 @@ class Sequence {
     std::mutex& getMutex();
     bool isTerminated() const;
     void setTerminated();
+    void updateLastActivityTime();
```
mzegla, line 55: I think we should keep this method private or protected and get back to calling it from the inside of some public method.

src/status.hpp
```diff
@@ -76,6 +76,7 @@ enum class StatusCode {
     INVALID_SEQUENCE_CONTROL_INPUT,  /*!< Unexpected value of sequence control input */
     SEQUENCE_CONTROL_INPUT_BAD_TYPE, /*!< Sequence control input in bad type */
     SEQUENCE_TERMINATED,             /*!< Sequence last request is being processed and it's not available anymore */
+    SEQUENCE_TIMEDOUT,               /*!< Sequence has timed out and it's not available anymore */
```
mzegla, line 79: Do we still need this status?

src/statefulmodelinstance.cpp
```diff
@@ -27,6 +27,10 @@ using namespace InferenceEngine;
 
 namespace ovms {
 
+StatefulModelInstance::~StatefulModelInstance() {
+    sequenceManager->join();
+}
```
mzegla, line 32: Do we need this? Sequence manager destructor will  be called along with this one and will also call `join()` method

src/test/stateful_modelinstance_test.cpp
```diff
@@ -221,6 +363,283 @@ void RunStatefulPredicts(const std::shared_ptr<ovms::ModelInstance> modelInstanc
...
+        std::this_thread::sleep_for(std::chrono::seconds(sequenceTimeoutSleepSeconds));
+
+    waitAfterSequenceStarted.set_value();
+    if (sequenceTimeoutScenario == SequenceTimeoutScenarios::WAIT_BEFORE_SEQUENCE_UNLOCKED)
+        std::this_thread::sleep_for(std::chrono::seconds(sequenceTimeoutSleepSeconds));
+
+    waitBeforeSequenceFinished.set_value();
+
+    if (sequenceTimeoutScenario == SequenceTimeoutScenarios::WAIT_AFTER_SEQUENCE_UNLOCKED)
+        std::this_thread::sleep_for(std::chrono::seconds(sequenceTimeoutSleepSeconds));
```
mzegla, line 400: Inconsistent new lines

mzegla (approved):

PR #513, opened 2021-02-18
Cvs 46635 stateful watcher thread

rasapala:


src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+ovms::Status GlobalSequencesViewer::addVersions(std::shared_ptr<ovms::Model>& model,
+    std::shared_ptr<model_versions_t> versionsToAdd,
+    std::shared_ptr<model_versions_t> versionsFailed) {
+    for (const auto version : *versionsToAdd) {
+        if (std::count(versionsFailed->begin(), versionsFailed->end(), version))
+            continue;
+        auto modelInstance = model->getModelInstanceByVersion(version);
+        auto stetefulModelInstance = std::static_pointer_cast<StatefulModelInstance>(modelInstance);
+        std::string managerId = model->getName() + std::to_string(version);
```
mzegla, line 40: I think we need separator here. What if someone has model with name "model" with version 12 and "model1" with version 2? They will be translated to the same key and will result in conflict at some point.

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+            return status;
+    }
+
+    updateThreadInterval();
+    return ovms::StatusCode::OK;
+}
+
+ovms::Status GlobalSequencesViewer::retireVersions(std::shared_ptr<ovms::Model>& model, std::shared_ptr<model_versions_t> versionsToRetire) {
+    for (const auto version : *versionsToRetire) {
+        std::string managerId = model->getName() + std::to_string(version);
```
mzegla, line 53: I think we need separator here. What if someone has model with name "model" with version 12 and "model1" with version 2? They will be translated to the same key and will result in conflict at some point.

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+    return ovms::StatusCode::OK;
+}
+
+ovms::Status GlobalSequencesViewer::reloadVersions(std::shared_ptr<ovms::Model>& model,
+    std::shared_ptr<model_versions_t> versionsToReload,
+    std::shared_ptr<model_versions_t> versionsFailed) {
+    for (const auto version : *versionsToReload) {
+        if (std::count(versionsFailed->begin(), versionsFailed->end(), version))
+            continue;
+        std::string managerId = model->getName() + std::to_string(version);
```
mzegla, line 69: I think we need separator here. What if someone has model with name "model" with version 12 and "model1" with version 2? They will be translated to the same key and will result in conflict at some point.

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+static bool sequenceWatcherStarted = false;
+
+ovms::Status GlobalSequencesViewer::addVersions(std::shared_ptr<ovms::Model>& model,
+    std::shared_ptr<model_versions_t> versionsToAdd,
+    std::shared_ptr<model_versions_t> versionsFailed) {
+    for (const auto version : *versionsToAdd) {
+        if (std::count(versionsFailed->begin(), versionsFailed->end(), version))
+            continue;
+        auto modelInstance = model->getModelInstanceByVersion(version);
```
mzegla, line 38: Potentially returns nullptr. Need to check that.

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+    std::shared_ptr<model_versions_t> versionsFailed) {
+    for (const auto version : *versionsToReload) {
+        if (std::count(versionsFailed->begin(), versionsFailed->end(), version))
+            continue;
+        std::string managerId = model->getName() + std::to_string(version);
+        auto status = unregisterManager(managerId);
+        if (status.getCode() != ovms::StatusCode::OK)
+            return status;
+
+        auto modelInstance = model->getModelInstanceByVersion(version);
```
mzegla, line 74: Potentially returns nullptr. Need to check that.

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+        status = registerManager(managerId, stetefulModelInstance->getSequenceManager().get());
+        if (status.getCode() != ovms::StatusCode::OK)
+            return status;
+    }
+    updateThreadInterval();
+    return ovms::StatusCode::OK;
+}
+
+void GlobalSequencesViewer::updateThreadInterval() {
+    uint32_t lowestHalfTimeoutInterval = DEFAULT_SEQUENCE_TIMEOUT_SECONDS / 2;
```
mzegla, line 86: I think we could set it to uint32 max value at the start up

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+    for (auto const& [key, val] : registeredSequenceManagers) {
+        auto sequenceManager = val;
+        uint32_t newInterval = sequenceManager->getTimeout() / 2;
```
mzegla, line 89: Values of timeout of 0 and 1 will evaluate to 0 here and we won't have any interval

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    sequenceWatcherIntervalSec = lowestHalfTimeoutInterval;
+    SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Sequence watcher thread interval set to {}", sequenceWatcherIntervalSec);
+}
+
+ovms::Status GlobalSequencesViewer::registerManager(std::string managerId, SequenceManager* sequenceManager) {
+    std::unique_lock<std::mutex> viewerLock(mutex);
+    if (registeredSequenceManagers.count(managerId)) {
+        SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence manager {} already exists", managerId);
+        return StatusCode::SEQUENCE_ALREADY_EXISTS;
```
mzegla, line 102: This should not happen. I think we could return INTERNAL_ERROR here and log it at error level, not debug.

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+    } else {
+        SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence manager {} added", managerId);
```
mzegla, line 104:
```suggestion
        SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence manager {} registered in the sequence timeout watcher", managerId);
```

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+        registeredSequenceManagers.emplace(managerId, sequenceManager);
+    }
+
+    return StatusCode::OK;
+}
+
+ovms::Status GlobalSequencesViewer::unregisterManager(std::string managerId) {
+    std::unique_lock<std::mutex> viewerLock(mutex);
+    if (registeredSequenceManagers.count(managerId)) {
+        SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence manager {} removed", managerId);
```
mzegla, line 114:
```suggestion
        SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence manager {} unregistered from the sequence timeout watcher", managerId);
```

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+        registeredSequenceManagers.erase(managerId);
+    } else {
+        SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence manager {} already exists", managerId);
+        return StatusCode::SEQUENCE_MISSING;
```
mzegla, line 118: This shouldn't happen. We could return with INTERNAL_ERROR.

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    return ovms::StatusCode::OK;
+}
+
+void GlobalSequencesViewer::sequenceWatcher(std::future<void> exit) {
+    SPDLOG_LOGGER_INFO(modelmanager_logger, "Started sequence timeout watcher thread");
+
+    while (exit.wait_for(std::chrono::milliseconds(1)) == std::future_status::timeout) {
+        std::this_thread::sleep_for(std::chrono::seconds(sequenceWatcherIntervalSec));
+        SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Sequence watcher thread check cycle begin");
```
mzegla, line 141: Can we change logging level to trace?

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+        removeTimedOutSequences();
+
+        SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Sequence watcher thread check cycle end");
```
mzegla, line 145: Can we change logging level to trace?

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+    SPDLOG_LOGGER_ERROR(modelmanager_logger, "Exited sequence timeout watcher thread");
```
mzegla, line 147: Why error level?

src/global_sequences_viewer.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+void GlobalSequencesViewer::sequenceWatcher(std::future<void> exit) {
+    SPDLOG_LOGGER_INFO(modelmanager_logger, "Started sequence timeout watcher thread");
+
+    while (exit.wait_for(std::chrono::milliseconds(1)) == std::future_status::timeout) {
+        std::this_thread::sleep_for(std::chrono::seconds(sequenceWatcherIntervalSec));
+        SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Sequence watcher thread check cycle begin");
+
+        removeTimedOutSequences();
```
mzegla, line 143: This call is locking the viewer, but do adding, removing and reloading models?

src/global_sequences_viewer.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+         * @brief An exit signal to notify watcher thread to exit
+         */
+    std::promise<void> exit;
+
+    /**
+         * Time interval between each sequence timeout check
+         */
+    uint sequenceWatcherIntervalSec = 1;
```
mzegla, line 56: Maybe we could start with uint32 max value?
Also please use uint32_t

src/modelmanager.cpp
```diff
@@ -82,6 +83,12 @@ void ModelManager::startWatcher() {
         watcherStarted = true;
         monitor = std::move(t);
     }
+
+    sequenceViewer->startWatcher();
+}
+
+void ModelManager::startSequenceWatcher() {
```
mzegla, line 90: Is this method used?

src/modelmanager.cpp
```diff
@@ -727,6 +736,16 @@ Status ModelManager::addModelVersions(std::shared_ptr<ovms::Model>& model, std::
             SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error occurred while loading model: {} versions; error: {}",
                 config.getName(),
                 status.string());
+            return status;
+        }
+        if (config.isStateful()) {
+            status = sequenceViewer->addVersions(model, versionsToStart, versionsFailed);
+            if (!status.ok()) {
+                SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error occurred while adding version in sequence viewer: {}; versions; error: {}",
```
mzegla, line 744:
```suggestion
                SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error occurred while registering model versions to sequence timeout watcher: {}; versions; error: {}",
```

src/modelmanager.cpp
```diff
@@ -743,6 +762,17 @@ Status ModelManager::reloadModelVersions(std::shared_ptr<ovms::Model>& model, st
             SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error occurred while reloading model: {}; versions; error: {}",
                 config.getName(),
                 status.string());
+
+            return status;
+        }
+        if (config.isStateful()) {
+            status = sequenceViewer->reloadVersions(model, versionsToReload, versionsFailed);
+            if (!status.ok()) {
+                SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error occurred while reloading version in sequence viewer: {}; versions; error: {}",
```
mzegla, line 771:
```suggestion
                SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error occurred while reloading version in sequence timeout watcher: {}; versions; error: {}",
```

src/modelmanager.cpp
```diff
@@ -854,6 +884,16 @@ Status ModelManager::reloadModelWithVersions(ModelConfig& config) {
             SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error occurred while unloading model: {}; versions; error: {}",
                 config.getName(),
                 status.string());
+            return status;
+        }
+        if (config.isStateful()) {
+            status = sequenceViewer->retireVersions(model, versionsToRetire);
+            if (!status.ok()) {
+                SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error occurred while retiring version in sequence viewer: {}; versions; error: {}",
```
mzegla, line 892:
```suggestion
                SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error occurred while unregistering model versions from sequence timeout watcher: {}; versions; error: {}",
```

src/sequence_manager.cpp
```diff
@@ -61,25 +61,28 @@ bool SequenceManager::sequenceExists(const uint64_t sequenceId) const {
...
-
+    SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence watcher thread for model {} version {} Manager lock acquired", modelName, modelVersion);
     for (auto it = sequences.begin(); it != sequences.end();) {
         Sequence& sequence = it->second;
         // Non blocking try to get mutex
         std::unique_lock<std::mutex> sequenceLock(sequence.getMutex(), std::try_to_lock);
         if (!sequence.isTerminated() && sequenceLock.owns_lock()) {
+            SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence watcher thread for model {} version {} Sequence lock acquired Id: {}", modelName, modelVersion, sequence.getId());
             sequenceLock.unlock();
+            SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence watcher thread for model {} version {} Sequence unlocked Id: {}", modelName, modelVersion, sequence.getId());
```
mzegla, line 72: I think all this logging about locking and unlocking should be removed before merge

src/sequence_manager.cpp
```diff
@@ -61,25 +61,28 @@ bool SequenceManager::sequenceExists(const uint64_t sequenceId) const {
...
-                SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence watcher thread for model {} version {} Sequence timeouted and removed - Id: {}", modelName, modelVersion, sequence.getId());
+                SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence watcher thread for model {} version {} Sequence timeouted and removed Id: {}", modelName, modelVersion, sequence.getId());
                 it = sequences.erase(it);
                 continue;
             }
         }
         ++it;
     }
 
+    SPDLOG_LOGGER_DEBUG(sequence_manager_logger, "Sequence watcher thread for model {} version {} Manager lock released", modelName, modelVersion);
```
mzegla, line 85: I think we should remove this log before merge

src/global_sequences_viewer.cpp
rasapala: It is safer to set it to sane value.

src/global_sequences_viewer.cpp
rasapala: register and unregister does lock it.

src/global_sequences_viewer.hpp
rasapala: I propose: DEFAULT_SEQUENCE_TIMEOUT_SECONDS / 2

src/modelmanager.cpp
rasapala: In tests.

src/global_sequences_viewer.hpp
rasapala: Nope must be limits.

src/global_sequences_viewer.cpp
rasapala: Limits is ok.

src/modelmanager.cpp
mzegla: Can we then either use it instead of calling `sequenceViewer->startWatcher();` 
or add it in mocked class so we don't have code unused in main logic?

mzegla (approved):
New methods should be also tested in isolation. We agreed to do that in a separate PR.

michalkulakowski (approved):

PR #516, opened 2021-02-19
CVS-48153 config validation custom library loading corectness

Ficol:


src/custom_node_library_manager.hpp
```diff
@@ -25,6 +25,7 @@ namespace ovms {
 
 class CustomNodeLibraryManager {
     std::unordered_map<std::string, NodeLibrary> libraries;
+    std::unordered_map<std::string, std::string> libraryBasePaths;
```
dkalinowski, line 28: How about keeping base path in `NodeLibrary` class field? This way we could have just one unordered_map.

atobiszei (approved):

dkalinowski (approved):

PR #520, opened 2021-02-24
Cvs 49703 stateful client example docs

rasapala:


dtrawins (changes requested):
I don't think it is a good idea to move all existing client examples to basic folder. It could break external links to example clients and the name basic is not describing the content correctly.
It would be better to add stateful examples to the same content or create a subfolder for stateful models if there is expected big number of them.

PR #530, opened 2021-02-26
Adding error information.

rasapala:


michalkulakowski (approved):

src/gcsfilesystem.cpp
```diff
@@ -144,9 +144,11 @@ StatusCode GCSFileSystem::isDirectory(const std::string& path,
         client_.ListObjects(bucket, gcs::Prefix(appendSlash(object)))) {
         if (meta) {
             *is_directory = true;
-            break;
+            return StatusCode::OK;
         }
     }
+
+    SPDLOG_LOGGER_ERROR(gcs_logger, "Invalid or missing GCS credentials, or directory does not exist - {}", path);
```
dtrawins, line 151: why the return status is OK while the directory does not exist?

src/s3filesystem.cpp
```diff
@@ -208,7 +208,7 @@ StatusCode S3FileSystem::isDirectory(const std::string& path, bool* is_dir) {
 
     auto head_bucket_outcome = client_.HeadBucket(head_request);
     if (!head_bucket_outcome.IsSuccess()) {
-        SPDLOG_LOGGER_ERROR(s3_logger, "Couldn't get MetaData for bucket with name {}", bucket);
+        SPDLOG_LOGGER_ERROR(s3_logger, "Invalid or missing S3 credentials, or bucket does not exist - {}", bucket);
```
dtrawins, line 211: why two error messages for the same scenario?

src/gcsfilesystem.cpp
```diff
@@ -144,9 +144,11 @@ StatusCode GCSFileSystem::isDirectory(const std::string& path,
         client_.ListObjects(bucket, gcs::Prefix(appendSlash(object)))) {
         if (meta) {
             *is_directory = true;
-            break;
+            return StatusCode::OK;
```
dtrawins, line 147: in the log SPDLOG_LOGGER_ERROR(gcs_logger, "Path is empty -> always a directory");
the message is not clear. It would be better with "Directory {} is empty. Model versions are missing"

src/gcsfilesystem.cpp
rasapala: because the api was designed to use bool* is_directory as flag determing the existance of the directory.

src/s3filesystem.cpp
rasapala: Because the is no API to check which of the errors occurred. It is either 3 of those in this case.

dtrawins: I meant the message below. The error message head_bucket_outcome.GetError().GetMessage() could be in the same line like above.

dtrawins (approved):

PR #531, opened 2021-02-27
Added opencv and building process

dtrawins:


dkalinowski (approved):

src/BUILD
```diff
@@ -237,9 +237,9 @@ cc_binary(
 )
 
 cc_binary(
-    name = "libnode_resize_opencv.so",
+    name = "libcustom-node-east-ocr.so",
```
dkalinowski, line 240: Use `_` instead of `-` to follow convention

PR #533, opened 2021-03-02
CVS-49720 Example custom node - east model postprocessing implementation

dkalinowski:


src/custom_nodes/east_ocr/east_ocr.cpp
```diff
@@ -14,89 +14,207 @@
...
-    //         }
-    //     }
-    // }
-    uint64_t H = input->dims[1];
-    uint64_t W = input->dims[2];
-    cv::Mat image(H, W, CV_32FC3);
-    // std::memcpy((void*)image.data, (void*)data.data(), input->dataLength);
-    std::memcpy((void*)image.data, (void*)input->data, input->dataLength);
-    return image;
+template<typename T>
```
dtrawins, line 45: It would be good to move common functions to utils.hpp. They could be more reusable that way.

src/custom_nodes/east_ocr/nms.hpp
```diff
@@ -0,0 +1,221 @@
+#pragma once
```
dtrawins, line 1: Add the original license

dkalinowski: Added license from original github repository to nms.hpp header

src/custom_nodes/east_ocr/east_ocr.cpp
dkalinowski: Done

michalkulakowski (approved):

dtrawins (approved):

PR #536, opened 2021-03-03
CVS-47716 customloader-internal-server-error-message-when-model-file-not-found

Ficol:


mzegla (approved):
LGTM. In the future, you can consider mapping CustomLoaderStatus to StatusCode if that's possible.

atobiszei (approved):

PR #566, opened 2021-03-12
Mkulakow/model control api single model

michalkulakowski:


mzegla (approved):

src/test/http_rest_api_handler_test.cpp
```diff
@@ -56,6 +56,12 @@ class ConfigApi : public TestWithTempDir {
     void RemoveConfig() {
         std::filesystem::remove(configFilePath);
     }
+
+    void SetUpSingleModel(std::string modelPath, std::string modelName) {
+        char* n_argv[] = {"ovms", "--model_path", &modelPath[0], "--model_name", &modelName[0], "--file_system_poll_wait_seconds", "0"};
```
dkalinowski, line 61: modelPath.data() or modelPath.c_str() to get pointer of underlying char array

src/test/http_rest_api_handler_test.cpp
```diff
@@ -56,6 +56,12 @@ class ConfigApi : public TestWithTempDir {
     void RemoveConfig() {
         std::filesystem::remove(configFilePath);
     }
+
+    void SetUpSingleModel(std::string modelPath, std::string modelName) {
```
dkalinowski, line 60: Pass by const reference

src/test/http_rest_api_handler_test.cpp
dkalinowski: Are you sure that buffers do not need to exist after parsing? Lifetime of modelPath/modelName ends at the end oof SetUpSingleModel function

src/test/http_rest_api_handler_test.cpp
```diff
@@ -128,6 +134,36 @@ TEST_F(ConfigReload, startWith1DummyThenReload) {
     EXPECT_EQ(status, ovms::StatusCode::OK_RELOADED);
 }
 
+TEST_F(ConfigReload, singleModel) {
+    ModelManagerTest manager;
+    SetUpSingleModel("/ovms/src/test/dummy", "dummy");
+    manager.startFromConfig();
```
dkalinowski, line 140: This is single model test but you call `startFromConfig` which is called by OVMS when started from config.jon

dkalinowski (approved):

PR #608, opened 2021-03-26
CVS-51585 test added for ovms start with model files missing

Ficol:


src/test/modelmanager_test.cpp
```diff
@@ -461,6 +461,17 @@ TEST(ModelManager, StartFromFile) {
     modelMock.reset();
 }
 
+TEST(ModelManager, StartFromFileWhenModelFilesMissing) {
+    std::filesystem::create_directories(model_1_path + "/1");
```
michalkulakowski, line 465: model_1_path already contains version dir

src/test/modelmanager_test.cpp
```diff
@@ -461,6 +461,17 @@ TEST(ModelManager, StartFromFile) {
...
+    std::string fileToReload = "/tmp/ovms_config_file1.json";
```
michalkulakowski, line 466: Maybe adding assertion of model version directory emptiness(filesystem::is_empty) would be a good idea since you are using same dir as in other tests.

michalkulakowski (approved):

dkalinowski (approved):

PR #613, opened 2021-04-09
CVS-51577 ovms does not validate correcty dag with model used as demultiplexer node

Ficol:


src/pipelinedefinition.cpp
```diff
@@ -617,8 +618,7 @@ class NodeValidator {
                 return result;
             }
 
-            if (
-                (dependantNodeInfo.kind == NodeKind::DL || dependantNodeInfo.kind == NodeKind::CUSTOM) &&
+            if (dependantNodeInfo.kind != NodeKind::ENTRY &&
```
dkalinowski, line 621: Why this change? We allow checking connection metadata correctness for EXIT node, which has no defined input metadata.

src/pipelinedefinition.cpp
```diff
@@ -469,16 +469,17 @@ class NodeValidator {
...
-        if (dependencyNodeInfo.demultiplyCount) {
-            auto result = influenceShapeWithDemultiplexer(tensorOutputShape, dependencyNodeInfo);
-            if (!result.ok()) {
-                return result;
-            }
+        auto result = influenceShapeWithDemultiplexer(tensorOutputShape, dependencyNodeInfo);
+        if (!result.ok()) {
+            return result;
+        }
+        if (dependantNodeInfo.kind == NodeKind::EXIT) {
```
dkalinowski, line 478: This is kind of hack to force demultiplexer demultiplyCount check in `influenceShapeWithDemultiplexer`.
Maybe extract this check out of `influenceShapeWithDemultiplexer` into separate method and call outside of `checkConnectionMetadataCorrectness`?

atobiszei (commented):
Could we test this with mocked modelinstance which shape could support demultiplexing?

dkalinowski (approved):

atobiszei (approved):

PR #614, opened 2021-04-12
CVS-52491 Set inputs info after reshape

mzegla:


src/modelinstance.cpp
```diff
@@ -140,6 +117,35 @@ Status ModelInstance::loadInputTensors(const ModelConfig& config, const DynamicM
         SPDLOG_DEBUG("model: {}, version: {}; reshaping inputs is not required", getName(), getVersion());
     }
 
+    for (const auto& pair : networkInputs) {
```
dkalinowski, line 120: In case of loading error we will be missing logs with precision/layout of inputs from OV which is usually helpful.

src/modelinstance.cpp
```diff
@@ -80,25 +80,11 @@ Status ModelInstance::loadInputTensors(const ModelConfig& config, const DynamicM
...
-        // Data from network
-        auto precision = input->getPrecision();
-        auto layout = input->getLayout();
         auto shape = input->getTensorDesc().getDims();
-
-        // Data from config
-        if (config.getLayout().size()) {
-            // Single layout for all inputs
-            layout = TensorInfo::getLayoutFromString(config.getLayout());
-        } else if (config.getLayouts().count(name)) {
```
dkalinowski, line 96: You are removing dead code from our source regarding layout switch. There might be more places with layout dead code. Have you checked that?

dkalinowski: Ok I see, you do not remove but move it.

src/modelinstance.cpp
```diff
@@ -140,6 +117,35 @@ Status ModelInstance::loadInputTensors(const ModelConfig& config, const DynamicM
...
+        const auto& name = pair.first;
+        auto input = pair.second;
+
+        // Data from network
+        auto precision = input->getPrecision();
+        auto layout = input->getLayout();
+        auto shape = input->getTensorDesc().getDims();
+
+        // Data from config
+        if (config.getLayout().size()) {
```
dkalinowski, line 130: Is it possible to get rid of layout dead code in this MR?

mzegla: Can we do it in separate PR? Here I'd like to have just clean fix for the issue with wrong input details reading order. 
Refactor with layout param removal could be done separately don't you think?

dkalinowski: Right

dkalinowski (approved):

rasapala (approved):
LGTM

PR #632, opened 2021-04-22
CVS-53368 binary input grpc client

Ficol:


rasapala (approved):
LGTM

dtrawins:
Could you make adjustments in the script to take similar parameters like in jpeg_classification.py?
Specifically we should be able to pass a list of images, change the batch size and also track response time and accuracy.
You could generate the data via:
```    image_data = []
    for image in FLAGS.image.split(','):
      with open(image, 'rb') as f:
        image_data.append(f.read())```
The request shape must be equal to the batch size.
You could test it with TFS and model http://download.tensorflow.org/models/official/20181001_resnet/savedmodels/resnet_v2_fp32_savedmodel_NHWC_jpg.tar.gz

dtrawins (approved):

PR #633, opened 2021-04-22
CVS-53367 binary input rest client

Ficol:


rasapala (approved):
LGTM

dtrawins:
Could you make adjustments in the script to take similar parameters like in rest_serving_client.py?
Specifically we should be able to pass a list of images, change the batch size, request format and also track response time and accuracy.
For noname row you could generate the data via:
```        predict_request = '{"instances" : [{"b64": "%s"}, {"b64": "%s"}]}' % (jpeg1_bytes, jpeg2_bytes)
```
You could test it with TFS and model http://download.tensorflow.org/models/official/20181001_resnet/savedmodels/resnet_v2_fp32_savedmodel_NHWC_jpg.tar.gz

dtrawins:
we should also include option request_format like in rest_serving_client.py to test those options on the server:
parser.add_argument('--request_format', default='row_noname', help='Request format according to TF Serving API: row_noname,row_name,column_noname,column_name',
                    choices=["row_noname", "row_name", "column_noname", "column_name"], dest='request_format')

dtrawins (approved):

PR #634, opened 2021-04-22
Fix model_zoo_intel_object_detection example custom node

dkalinowski:
For detections outside of original image it was possible to get exception thrown and request failed.
After the fix, we will crop the boxes to fit original image.


rasapala (approved):
LGTM

dtrawins (approved):

PR #635, opened 2021-04-23
Removing openssl-devel.

rasapala:


dkalinowski (approved):

dtrawins:
which that change we should be able to drop this line DockerfileMakePackage#L54 . Without copying crypto lib ovms should work for example on ubuntu20 using binary distribution without docker.

dtrawins:
redhat image should be also updated to use version OpenSSL_1_1_1k

dtrawins (approved):

PR #636, opened 2021-04-23
added sample notebook

nyurguhun:
* added sample notebook
* added people_images_list
* added images for login


nyurguhun:
Checks probably didn't work because the jupyter notebook works only in the RHODS env.

dtrawins:
notebooks are in a dedicated github repo

PR #660, opened 2021-05-13
Python client library template

mzegla:


client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -0,0 +1,93 @@
...
+            Python scalars and lists are internally converted to their numpy counterparts before further processing.
+            String values are converted to bytes and placed in TensorProto string_val field. 
+
+        dtype (optional): tensor_pb2 DataType value. 
+            If not provided, the function will try to infer the data type from **values** argument.
+
+        shape (optional): The list of integers defining the shape of the tensor. 
+            If not provided, the function will try to infer the shape from **values** argument.
+
+        use_tensor_content (optional): Determines if the **values** should be placed in tensor_content field of TensorProto.
```
dtrawins, line 34: currently ovms doesn't support _val types it it will not be used. It will be also much slower so I would question if we need this option.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -0,0 +1,93 @@
...
+        TensorProto object filled with **values**.
+
+    Raises:
+        TypeError:  if unsupported types are provided.
+        ValueError: if arguments have inappropriate values.
+
+    Examples:
+        With python string scalar:
+
+        >>> data = "hello"
```
dtrawins, line 48: that example will not work with any known model

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -0,0 +1,93 @@
...
+        >>> data = [[1, 2, 3], [4, 5, 6]]
+        >>> tensor_proto = make_tensor_proto(data)
+        >>> print(tensor_proto)
+
+        With numpy array:
+
+        >>> data = np.array([[1, 2, 3], [4, 5, 6]], np.int32)
+        >>> tensor_proto = make_tensor_proto(data)
+        >>> print(tensor_proto)
+    '''
```
dtrawins, line 63: it would be useful to add example with binary input

client/python/lib/ovmsclient/tfs_compat/http/requests.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Examples:
+        Request to the default version of the model called "model" that has 2 inputs:  
+
+        >>> predict_request = make_predict_request(
+        ...     inputs={
+        ...         "string_input": "hello there",
+        ...         "numeric_input: np.array([[1, 2, 3], [4, 5, 6]], np.int32)
+        ...     }, 
+        ...     model_name="model")
```
dtrawins, line 89: format will be set automatically?
how binary data is to be passed? base64 should be handled manually or the same format of arguments like with grpc would be used? For simplicity it might be easier to select the binary data based on string content but for TFS it could be in some model with string input incorrect.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Returns:
+        TensorProto object filled with **values**.
+
+    Raises:
+        TypeError:  if unsupported types are provided.
+        ValueError: if arguments have inappropriate values.
+
+    Examples:
+        With python string scalar:
```
dkalinowski, line 46: What about binary image data? Can we handle it here too?

client/python/lib/ovmsclient/tfs_compat/serving_client.py
```diff
@@ -0,0 +1,230 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class ServingClient:
+    
+    def __init__(self, address, grpc_port, http_port=None):
+        raise NotImplementedError
+
+    def send_predict_request(self, request):
```
dtrawins, line 22: why don't we call it simply "Predict"

client/python/lib/ovmsclient/tfs_compat/serving_client.py
```diff
@@ -0,0 +1,230 @@
...
+        '''
+        Send PredictRequest to the server.
+
+        Args:
+            request: PredictRequest object. If the request is of type GrpcPredictRequest
```
dtrawins, line 27: I'm not sure if this is good idea to make universal predict method because REST and gRPC will use different connection type. Maybe we should have two classes for connections: for REST and for gRPC. It might be useful to control which grpc stub will be used for which predictions while for REST it is not the case.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
mzegla: Stateful models use val fields for sequence ID and control input. 
It gives user more control over created proto. In case we decide to support val fields for some reasons the solution on the client side will already be in place.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
mzegla: This is meant to work the same way it currently works in our clients. 
The binary data encoded to string will be placed in string_val field.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
mzegla: Yes, the data is invalid, but the purpose of this example is to create proto and show user where the data is placed and how the result proto object looks like.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
mzegla: Well, I can change the example above with "hello" string to contain some encoded binary data, but the examples here are meant to show the tensor proto creation with certain types of data. The content is not relevant for this example.

client/python/lib/ovmsclient/tfs_compat/http/requests.py
mzegla: Well, in REST we don't really have the possibility to set data type. The data will be set as it's provided in the function argument. Base64 encoded data shall be provided as a string.

client/python/lib/ovmsclient/tfs_compat/serving_client.py
mzegla: I chose that name to be as descriptive as possible about what a method does. 
I didn't use "predict" as it might be used by an API on the higher level.

client/python/lib/ovmsclient/tfs_compat/serving_client.py
mzegla: Could you elaborate on "It might be useful to control which grpc stub will be used for which predictions" part?
Connection objects for both HTTP and GRPC will be held inside ServingClient. User will create predict request of certain type and based on that type ServingClient will dispatch it via appropriate interface.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
dkalinowski: Yes, but maybe this api could have function with specific name to indicate this is jpeg and it would place into string_val automatically?

mzegla: For argument that holds binary data its content will be put in string_val automatically

dkalinowski (approved):

dtrawins (approved):

PR #661, opened 2021-05-13
CVS-53374 add base64 string decoding to restparser

Ficol:


michalkulakowski (approved):

WORKSPACE
```diff
@@ -216,6 +216,14 @@ http_archive(
     build_file = "@//third_party/libevent:BUILD",
 )
 
+# abseil
+http_archive(
+  name = "com_google_absl",
```
dkalinowski, line 221: Is it recomended to use both: abseil and cpp17 in one binary?

dkalinowski: Maybe we could use abseil from tensorflow? By adding another abseil we will have 2 abseils to maintain in scans. @dtrawins

PR #663, opened 2021-05-19
Mkulakow/single model binary inputs

michalkulakowski:


src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+Status resizeMat(const cv::Mat& src, cv::Mat& dst, const std::shared_ptr<TensorInfo>& tensorInfo) {
+    if(tensorInfo->getLayout() == InferenceEngine::Layout::NCHW)
+    {
+        int cols = tensorInfo->getShape()[3];
+        int rows = tensorInfo->getShape()[2];
+        cv::resize(src, dst, cv::Size(cols,rows));
+        return StatusCode::OK;
+    }
+    else if (tensorInfo->getLayout() == InferenceEngine::Layout::NHWC)
```
dkalinowski, line 100: What if user sends 1 color channel only?

dkalinowski: Have you tried sending 1 color channel image? Or 4 color channel with transparency?

src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+        return StatusCode::OK;
+    }
+
+    return StatusCode::UNSUPPORTED_LAYOUT;
+}
+
+Status validateNumberOfChannels(const std::shared_ptr<TensorInfo>& tensorInfo,
+    const cv::Mat input) {
+    // Network and input must have the same number of shape dimensions. 
+    if ((unsigned int)(input.channels()) != tensorInfo->getShape()[1]) {
```
dkalinowski, line 114: Here number of color channels does not need to be at index 1. Maybe introduce method in TensorInfo `getColorChannels()`? Depending on layout it would return correct number of channels. Or maybe OpenVINO has utility functions for this?

dkalinowski: You also do not check if tensor has at least 2 elements.

src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+        SPDLOG_DEBUG("Binary sent to input: {} has invalid number of channels. Expected: {} Actual: {}", tensorInfo->getMappedName(),tensorInfo->getShape()[1], input.dims);
+        return Status(StatusCode::INVALID_NO_OF_SHAPE_DIMENSIONS);
+    }
+    return StatusCode::OK;
+}
+
+const bool checkBatchSizeMismatch(const std::shared_ptr<TensorInfo>& tensorInfo,
+    const int batchSize) {
+    if (static_cast<size_t>(batchSize) != tensorInfo->getShape()[0])
```
dkalinowski, line 123: This conversion will yield errors in static analysis, you cast int to unsigned int which may overflow.

src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+const Status validateTensor(const std::shared_ptr<TensorInfo>& tensorInfo,
+    const tensorflow::TensorProto& src) {
+    if(tensorInfo->getLayout() == InferenceEngine::Layout::NCHW ||
+       tensorInfo->getLayout() == InferenceEngine::Layout::NHWC)
+    {
+        return StatusCode::OK;
+    }
+
+    if (checkBatchSizeMismatch(tensorInfo, src.string_val_size())){
```
dkalinowski, line 145: Batch size will be checked only if layout is not NCHW/NHWC, is it desired?

src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    return StatusCode::OK;
+}
+
+Status convertTensorToMatsMatchingTensorInfo(const tensorflow::TensorProto& src, std::vector<cv::Mat>& images, const std::shared_ptr<TensorInfo>& tensorInfo) {
+
+    for (int i = 0; i < src.string_val_size(); i++) {
+        cv::Mat image = convertStringValToMat(src.string_val(i));
+        if (image.data == NULL)
```
dkalinowski, line 157: use `nullptr` in c++

src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+            }
+            images.push_back(imageResized);
+        }
+    }
+
+    return StatusCode::OK;
+}
+
+template <typename T>
+InferenceEngine::Blob::Ptr createBlobFromMats(std::vector<cv::Mat>& images, const std::shared_ptr<TensorInfo>& tensorInfo){
```
dkalinowski, line 194: `images` param can be const since you do not modify it

src/deserialization.hpp
```diff
@@ -114,10 +115,20 @@ Status deserializePredictRequest(
...
+            InferenceEngine::Blob::Ptr blob;
 
-            InferenceEngine::Blob::Ptr blob =
-                deserializeTensorProto<TensorProtoDeserializator>(
+            if (requestInput.dtype() == tensorflow::DataType::DT_STRING) {
+                SPDLOG_DEBUG("Request contains binary inputs.");
+                Status status = convertStringValToBlob(requestInput, &blob, tensorInfo);
+                if(status != StatusCode::OK)
+                {
+                    SPDLOG_DEBUG("Binary inputs conversion failed.");
```
dkalinowski, line 125: Add specific logger instance

src/modelinstance.cpp
```diff
@@ -848,6 +848,11 @@ const Status ModelInstance::validate(const tensorflow::serving::PredictRequest*
         if (!status.ok())
             return status;
 
+        if (requestInput.dtype() == tensorflow::DataType::DT_STRING) {
+            // binary inputs will be validated during conversion to blob
```
dkalinowski, line 852: Add log

src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+                nchwVector[c * (rows * cols) + y * cols + x] = reinterpret_cast<const T*>(nhwcVector)[y * channels * cols + x * channels + c];
+            }
+        }
+    }
+    return std::move(nchwVector);
+}
+
+cv::Mat convertStringValToMat(const std::string& stringVal) {
+    std::vector<unsigned char> vectordata(stringVal.begin(), stringVal.end());
+    cv::Mat data_mat(vectordata, true);
```
mzegla, line 77: Please unify the naming of variables

src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+    return cv::imdecode(data_mat, cv::IMREAD_UNCHANGED);
+}
+
+Status convertPrecision(const cv::Mat& src, cv::Mat& dst, const InferenceEngine::Precision requestedPrecision) {
+    int type = getMatTypeFromTensorPrecision(requestedPrecision);
```
mzegla, line 83: Does the return type have any alias? Perhaps it's worth creating a type for that value if there's not?

src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+        int rows = tensorInfo->getShape()[2];
+        cv::resize(src, dst, cv::Size(cols,rows));
+        return StatusCode::OK;
+    }
+    else if (tensorInfo->getLayout() == InferenceEngine::Layout::NHWC)
+    {
+        int cols = tensorInfo->getShape()[2];
+        int rows = tensorInfo->getShape()[1];
+        cv::resize(src, dst, cv::Size(cols,rows));
+        return StatusCode::OK;
```
mzegla, line 105: Is this performed post validation? You go with `tensorInfo->getShape()[x]` in multiple places, but are you sure those vectors have the right amount of elements? 
Guess what I'm asking is if:
`tensorInfo->getLayout() == InferenceEngine::Layout::NHWC`
returns true does it guarantee that `tensorInfo->getShape()` returns valid 4-element vector?

src/binaryutils.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+            images.push_back(imageResized);
+        } else {
+            cv::Mat imageResized;
+            resizeMat(image, imageResized, tensorInfo);
+            if (status != StatusCode::OK) {
+                return status;
+            }
+            images.push_back(imageResized);
+        }
```
mzegla, line 201: I think this block can be a bit shorter. The `resizeMat` call is performed almost identically in both paths.
Could you try to go with something like:
```
cv::Mat imagePostValidation = image;
if (!isPrecisionEqual(image.depth(), tensorInfo->getPrecision())) {
      status = convertPrecision(image, imagePostValidation, tensorInfo->getPrecision());

      if (status != StatusCode::OK) 
           return status;
}
cv::Mat imageResized;
resizeMat(imagePostValidation, imageResized, tensorInfo);
if (status != StatusCode::OK)
         return status;
images.push_back(imageResized);
```

Or even not use imagePostValidation variable if its possible to perform operations in-place like:
`status = convertPrecision(image, image, tensorInfo->getPrecision());`

mzegla (approved):

src/binaryutils.cpp
```diff
@@ -0,0 +1,284 @@
...
+Status validateNumberOfChannels(const std::shared_ptr<TensorInfo>& tensorInfo,
+    const cv::Mat input) {
+    // Network and input must have the same number of shape dimensions.
+    if (tensorInfo->getLayout() == InferenceEngine::Layout::NCHW) {
+        if ((unsigned int)(input.channels()) != tensorInfo->getShape()[1]) {
+            SPDLOG_DEBUG("Binary sent to input: {} has invalid number of channels. Expected: {} Actual: {}", tensorInfo->getMappedName(), tensorInfo->getShape()[1], input.channels());
+            return StatusCode::INVALID_NO_OF_CHANNELS;
+        }
+    } else if (tensorInfo->getLayout() == InferenceEngine::Layout::NHWC) {
+        if ((unsigned int)(input.channels()) != tensorInfo->getShape()[3]) {
```
dkalinowski, line 140: Unit test missing for this behavior. The same goes for most of negative paths.

src/binaryutils.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+Status validateNumberOfChannels(const std::shared_ptr<TensorInfo>& tensorInfo,
+    const cv::Mat input) {
+    // Network and input must have the same number of shape dimensions.
+    if (tensorInfo->getLayout() == InferenceEngine::Layout::NCHW) {
+        if ((unsigned int)(input.channels()) != tensorInfo->getShape()[1]) {
+            SPDLOG_DEBUG("Binary sent to input: {} has invalid number of channels. Expected: {} Actual: {}", tensorInfo->getMappedName(), tensorInfo->getShape()[1], input.channels());
+            return StatusCode::INVALID_NO_OF_CHANNELS;
+        }
+    } else if (tensorInfo->getLayout() == InferenceEngine::Layout::NHWC) {
```
dkalinowski, line 139: My latest findings are that when model has NHWC layout it still reports old shape so it is at position 1.

dkalinowski (approved):

PR #673, opened 2021-05-26
Support binary inputs in DAGs

mzegla:


src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -76,8 +76,9 @@ class EnsembleFlowCustomNodePipelineExecutionTest : public TestWithTempDir {
     template <typename T>
     std::unique_ptr<Pipeline> prepareSingleNodePipelineWithLibraryMock() {
         const std::vector<float> inputValues{3.5, 2.1, -0.2};
+        const tensor_map_t inputsInfo{{pipelineInputName, nullptr}};
```
dkalinowski, line 79: Why does this work with uninitialized tensorinfo? (nullptr)

dkalinowski (changes requested):
Should we add negative tests for invalid request to DAG?
Edit: added

src/test/ensemble_flow_custom_node_tests.cpp
mzegla: Because it's not used in that scenario. `inputsInfo` is necessary to fill new EntryNode constructor and it's used only for binary inputs right now.

dtrawins (approved):

michalkulakowski (approved):

dkalinowski (approved):

PR #674, opened 2021-05-26
Fix RH image building

dtrawins:


dkalinowski (approved):

extras/nginx-mtls-auth/Dockerfile.redhat
```diff
@@ -20,6 +20,7 @@ USER root
 RUN     set -e ; \
         set -x ; \
         mkdir /certs ; \
+        unset LD_LIBRARY_PATH ; \
```
mzegla, line 23: Why is it needed?

dtrawins: after changes in openssl/crypto lib in ovms package and upgrading it, it clashes with  microdnf package installer so it fails to install systemd package.

mzegla (approved):

PR #676, opened 2021-05-27
CVS-53585 grpc serialization with binary output

ficol:


src/binaryutils.cpp
```diff
@@ -84,6 +84,12 @@ cv::Mat convertStringValToMat(const std::string& stringVal) {
     return cv::imdecode(dataMat, cv::IMREAD_UNCHANGED);
 }
 
+std::string convertMatToStringVal(const cv::Mat& dataMat) {
+    std::vector<unsigned char> buf;
+    cv::imencode(".jpg", dataMat, buf);
+    return std::string(buf.begin(), buf.end());
```
dkalinowski, line 90: Here you create multiple copies - once from `dataMat` to `buf`, then from `buf` to return value

dkalinowski:
Maybe we could add integration tests to test single model path and DAG execution if binary inputs are serialized properly?

michalkulakowski (approved):

docs/binary_input_ouput.md
```diff
@@ -39,17 +39,17 @@ Blob data precision from binary input decoding is set automatically based on the
 Some models or DAG pipelines return images in the response. OpenVINO model outputs are always in the form of arrays. It is possible,
 however, to configure OVMS to send the image outputs in the binary image format instead.
 
-Binary outputs can be enabled by setting the output name with `_binary` suffix. In case the model is already exported with
+Binary outputs can be enabled by setting the output name with `_bytes` suffix. In case the model is already exported with
```
dkalinowski, line 42: Have you tested if this works with DAG?

src/binaryutils.cpp
```diff
@@ -265,6 +279,34 @@ InferenceEngine::Blob::Ptr convertMatsToBlob(std::vector<cv::Mat>& images, const
     }
 }
 
+Status convertBlobToMats(std::vector<cv::Mat>& images, const InferenceEngine::Blob::Ptr& blob, const std::shared_ptr<TensorInfo>& tensorInfo) {
+    int matPrecision = getMatTypeFromTensorPrecision(tensorInfo->getPrecision());
+    int rows, cols, channels;
+    if (tensorInfo->getLayout() == InferenceEngine::Layout::NCHW) {
```
dkalinowski, line 285: TensorInfo may contain 5 dimenions and layout ANY when demultiplexer is present in DAG. I'm afraid we are missing integration with DAGs here.

src/binaryutils.cpp
```diff
@@ -265,6 +279,34 @@ InferenceEngine::Blob::Ptr convertMatsToBlob(std::vector<cv::Mat>& images, const
...
+        channels = tensorInfo->getShape()[1];
+        rows = tensorInfo->getShape()[2];
+        cols = tensorInfo->getShape()[3];
+    } else if (tensorInfo->getLayout() == InferenceEngine::Layout::NHWC) {
```
dkalinowski, line 289: Do we want different behavior in binary inputs and binary outputs?
For binary inputs we enforce user to switch model to NHWC layout. We do not proceed further if model is in wrong layout. For binary outputs we allow NCHW?

NCHW will not even work here since you do not transpose data before converting to cv::Mat.

src/binaryutils.cpp
```diff
@@ -265,6 +279,34 @@ InferenceEngine::Blob::Ptr convertMatsToBlob(std::vector<cv::Mat>& images, const
...
+        rows = tensorInfo->getShape()[1];
```
dkalinowski, line 290: This transpose behavior should be covered in unit tests, which is also absent

src/serialization.cpp
```diff
@@ -85,7 +86,13 @@ Status serializePredictResponse(
             return status;
         }
         auto& tensorProto = (*response->mutable_outputs())[networkOutput->getMappedName()];
-        auto status = serializeBlobToTensorProto(tensorProto, networkOutput, blob);
+        Status status;
+        std::string binary_suffix = "_bytes";
```
dkalinowski, line 90: This should not be variable. Make it const static and move to some utility header. Check implementation of ModelInstance::waitForLoaded to check how constants are handled there.

src/test/binaryutils_test.cpp
```diff
@@ -215,4 +215,94 @@ TEST_F(BinaryUtilsTest, positive_resizing) {
...
+    ASSERT_EQ(blob->size(), 6);
+    ASSERT_EQ(status, ovms::StatusCode::OK);
+    uint8_t* ptr = blob->buffer();
+    ASSERT_EQ(std::equal(ptr, ptr + blob->size(), rgb_batchsize_2_blob), true);
+    tensorflow::TensorProto stringValDeserialized;
+    convertBlobToStringVal(stringValDeserialized, tensorInfo, blob);
+
+    EXPECT_EQ(stringVal.string_val(0).substr(stringVal.string_val(0).find("\xFF\xC0")), stringValDeserialized.string_val(0).substr(stringValDeserialized.string_val(0).find("\xFF\xC0")));
+    EXPECT_EQ(stringVal.string_val(1).substr(stringVal.string_val(1).find("\xFF\xC0")), stringValDeserialized.string_val(1).substr(stringValDeserialized.string_val(1).find("\xFF\xC0")));
+}
```
dkalinowski, line 307: End to end tests are missing, we have plenty of fixtures to use to test behavior of single model predict path and DAGs aswell. Check `ensemble_tests.cpp` for DAGs and `prediction_service_test.cpp` for single path.

src/test/binaryutils_test.cpp
```diff
@@ -215,4 +215,94 @@ TEST_F(BinaryUtilsTest, positive_resizing) {
...
+    stringVal.set_dtype(tensorflow::DataType::DT_STRING);
+    stringVal.add_string_val(image_bytes.get(), filesize);
+    stringVal.add_string_val(image_bytes.get(), filesize);
+
+    std::shared_ptr<TensorInfo> tensorInfo = std::make_shared<TensorInfo>("", InferenceEngine::Precision::U8, shape_t{2, 3, 1, 1}, InferenceEngine::Layout::NCHW);
+
+    auto status = convertStringValToBlob(stringVal, &blob, tensorInfo);
+
+    ASSERT_EQ(blob->size(), 6);
+    ASSERT_EQ(status, ovms::StatusCode::OK);
```
dkalinowski, line 299: This is wrong to test blob size before checking status. In case status is not OK we end up with `status` variable with nullptr value. Dereferencing nullptr causes segfault. We do not want segfaults in our unit tests in case of fails but meaningful messages wherever possible.

src/test/binaryutils_test.cpp
```diff
@@ -215,4 +215,94 @@ TEST_F(BinaryUtilsTest, positive_resizing) {
...
+    DataFile.seekg(0);
+    std::unique_ptr<char[]> image_bytes(new char[filesize]);
+    DataFile.read(image_bytes.get(), filesize);
+
+    tensorflow::TensorProto stringVal;
+    InferenceEngine::Blob::Ptr blob;
+    stringVal.set_dtype(tensorflow::DataType::DT_STRING);
+    stringVal.add_string_val(image_bytes.get(), filesize);
+
+    std::shared_ptr<TensorInfo> tensorInfo = std::make_shared<TensorInfo>("", InferenceEngine::Precision::U8, shape_t{1, 1, 1, 1}, InferenceEngine::Layout::NCHW);
```
dkalinowski, line 235: There are no low level NHWC unit tests.

PR #682, opened 2021-06-04
Cvs 52148 unsafe func check

rasapala:


src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -1834,6 +1834,13 @@ TEST_F(EnsembleFlowCustomNodeAndDemultiplexerLoadConfigThenExecuteTest, Demultip
 
 struct LibraryParamControlledMetadata {
     static bool startsWith(const char* str, const char* prefix) {
+        // Ensure null terminated
+        if (str[sizeof(str) - 1] == 0) {
```
dkalinowski, line 1838: `sizeof(str)` will always be 8 because this is size of pointer

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -1834,6 +1834,18 @@ TEST_F(EnsembleFlowCustomNodeAndDemultiplexerLoadConfigThenExecuteTest, Demultip
 
 struct LibraryParamControlledMetadata {
     static bool startsWith(const char* str, const char* prefix) {
+        // Ensure null terminated
+        const int MAX = 300;
+        const char* end = str;
+        for (; *end != '\0'; ++end) {
+            if (end - str > MAX)
```
dkalinowski, line 1841: `EXPECT_LE(end-str, MAX)`

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -1834,6 +1834,18 @@ TEST_F(EnsembleFlowCustomNodeAndDemultiplexerLoadConfigThenExecuteTest, Demultip
...
+                EXPECT_EQ(1, 2);
```
dkalinowski, line 1842: EXPECT_TRUE(false);

dkalinowski (approved):

michalkulakowski (approved):

PR #694, opened 2021-06-11
Suport batch auto for binary inputs

michalkulakowski:


src/test/predict_validation_test.cpp
```diff
@@ -165,6 +165,32 @@ TEST_F(PredictValidation, RequestWrongBatchSizeAuto) {
     EXPECT_EQ(status, ovms::StatusCode::BATCHSIZE_CHANGE_REQUIRED);
 }
 
+TEST_F(PredictValidation, RequestWrongBatchSizeAutoBinaryInputs) {
+    modelConfig.setBatchingParams("auto");
+    tensorflow::serving::PredictRequest binaryInputRequest = preparePredictRequest(
+        {
+            {"Binary_Input",
+                std::tuple<ovms::shape_t, tensorflow::DataType>{{1, 224, 224, 3}, tensorflow::DataType::DT_STRING}},
```
dkalinowski, line 173: This is most likely wrong, shape of binary input has 1 dimension but you specify 4

src/test/predict_validation_test.cpp
```diff
@@ -165,6 +165,32 @@ TEST_F(PredictValidation, RequestWrongBatchSizeAuto) {
...
+        });
+    auto& input = (*binaryInputRequest.mutable_inputs())["Binary_Input"];
+    std::string val1 = "val1";
```
dkalinowski, line 176: Why can't we pass `"val1"` to `add_string_val`?

src/test/predict_validation_test.cpp
```diff
@@ -165,6 +165,32 @@ TEST_F(PredictValidation, RequestWrongBatchSizeAuto) {
     EXPECT_EQ(status, ovms::StatusCode::BATCHSIZE_CHANGE_REQUIRED);
 }
 
+TEST_F(PredictValidation, RequestWrongBatchSizeAutoBinaryInputs) {
+    modelConfig.setBatchingParams("auto");
+    tensorflow::serving::PredictRequest binaryInputRequest = preparePredictRequest(
```
dkalinowski, line 170: Why do we need to prepare request this way if we declare it manually in line 175-179?

src/test/predict_validation_test.cpp
```diff
@@ -165,6 +165,32 @@ TEST_F(PredictValidation, RequestWrongBatchSizeAuto) {
...
+            {"Binary_Input",
+                std::tuple<ovms::shape_t, tensorflow::DataType>{{1, 224, 224, 3}, tensorflow::DataType::DT_STRING}},
+        });
+    auto& input = (*binaryInputRequest.mutable_inputs())["Binary_Input"];
+    std::string val1 = "val1";
+    std::string val2 = "val2";
+    input.add_string_val(val1);
+    input.add_string_val(val2);
+
+    networkInputs.clear();
```
dkalinowski, line 181: networkInputs is empty at beginning of the test

src/test/predict_validation_test.cpp
```diff
@@ -165,6 +165,32 @@ TEST_F(PredictValidation, RequestWrongBatchSizeAuto) {
...
+    std::string inputName = "Binary_Input";
```
dkalinowski, line 182: If you declared it at the beginning of a test, you could use it in line 175 aswell

src/test/predict_validation_test.cpp
```diff
@@ -165,6 +165,32 @@ TEST_F(PredictValidation, RequestWrongBatchSizeAuto) {
...
+    modelConfig.setBatchingParams("auto");
+    tensorflow::serving::PredictRequest binaryInputRequest = preparePredictRequest(
+        {
+            {"Binary_Input",
+                std::tuple<ovms::shape_t, tensorflow::DataType>{{1, 224, 224, 3}, tensorflow::DataType::DT_STRING}},
+        });
+    auto& input = (*binaryInputRequest.mutable_inputs())["Binary_Input"];
+    std::string val1 = "val1";
+    std::string val2 = "val2";
+    input.add_string_val(val1);
```
dkalinowski, line 178: I clearly indicate what test is doing I would declare const int variable requestBatchSize = 2 and fill input using for loop. This would make it clear that model instance has getBatchSize() mocked to value 1 and we have batch=2 in request.

src/modelinstance.cpp
```diff
@@ -879,6 +890,17 @@ const Status ModelInstance::validate(const tensorflow::serving::PredictRequest*
         if (requestInput.dtype() == tensorflow::DataType::DT_STRING) {
             // binary inputs will be validated during conversion to blob
             SPDLOG_DEBUG("Received request contains binary inputs");
+            if (checkBinaryInputBatchSizeMismatch(*networkInput, requestInput)) {
+                if (batchingMode == AUTO) {
+                    return StatusCode::BATCHSIZE_CHANGE_REQUIRED;
+                } else {
```
rasapala, line 896: Please add another unit test for else condition:
requestInput.dtype() == tensorflow::DataType::DT_STRING
batchingMode != AUTO

dkalinowski (approved):

rasapala (approved):
LGTM

PR #711, opened 2021-06-16
T1178 - Do not install unnecessary packages in the container (Docker)

mkuczyns11:


Dockerfile.centos
```diff
@@ -80,13 +80,6 @@ RUN yum update -d6 -y && yum install -d6 -y epel-release centos-release-scl && y
             unzip \
             wget \
             which \
-            boost-thread \
-            boost-system \
-            boost-filesystem \
-            boost-chrono \
-            boost-program-options \
-            boost-date-time \
-            boost-atomic \
```
mzegla, line 89: Is it unnecessary? I recall such packages used to be required for HDDL support. Can you confirm that after removing those packages the image works with all supported devices?

mkuczyns11: These packages were removed due to them being duplicates in yum install section. In the very beginning of the yum install/update command you can find that all of these packages are indeed still in there.

mzegla: I see now. All good then.

mzegla (approved):

rasapala (approved):
LGTM

PR #713, opened 2021-06-17
Mkulakow/cvs 53131

michalkulakowski:


src/pipelinedefinition.cpp
```diff
@@ -477,6 +477,21 @@ class NodeValidator {
         return StatusCode::OK;
     }
 
+    bool isConnectionShapeCorrect(shape_t tensorInputShape, shape_t tensorOutputShape) {
```
dkalinowski, line 480: Here you copy shape_t which is not efficient. It is also possible to have it const since we do not modify it. Please pass it as const reference.

src/pipelinedefinition.cpp
```diff
@@ -477,6 +477,21 @@ class NodeValidator {
...
+        if (tensorInputShape.size() != tensorOutputShape.size()) {
+            return false;
+        }
+
+        for (size_t i = 0; i < tensorInputShape.size(); i++) {
+            if (tensorInputShape[i] != tensorOutputShape[i]) {
+                if (tensorInputShape[i] != 0 && tensorOutputShape[i] != 0) {
```
dkalinowski, line 487: I would merge these two if statements into one to not to create unnecessary nesting

src/pipelinedefinition.cpp
```diff
@@ -477,6 +477,21 @@ class NodeValidator {
         return StatusCode::OK;
     }
 
+    bool isConnectionShapeCorrect(const shape_t& tensorInputShape, const shape_t& tensorOutputShape) {
```
mzegla, line 480: I would change the name. I can see that this method provides comparison logic for tensor shapes. 
For easier reuse in other places maybe we could name it: areShapesEqual() or something like that to make it more generic?

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -1981,6 +1981,39 @@ TEST_F(EnsembleConfigurationValidationWithCustomNode, SuccessfulConfiguration) {
     ASSERT_EQ(pipelineDefinition->validate(manager), StatusCode::OK);
 }
 
+TEST_F(EnsembleConfigurationValidationWithCustomNode, SuccessfulConfigurationWithDynamicShape) {
```
mzegla, line 1984: Can we have a simpler test that would test new method - isConnectionShapeCorrect directly?
I assume it might be used elsewhere so it would be nice to have test just for this method and not rely only on `pipelineDefinition->validate()` call.

mzegla (approved):

dkalinowski (approved):

PR #750, opened 2021-07-05
Task grpc connection wait for ready

jasiu86:


dkalinowski:
Why do we need to add it? Is there jira task that describes it? Should all example clients be updated as well?

rasapala (approved):
LGTM

PR #759, opened 2021-07-07
CVS-57651 - Task add target device for ovms c tests

jasiu86:
Fetch target_device from ENV variable and pass it to docker.


rasapala (approved):
LGTM

atobiszei (approved):

PR #763, opened 2021-07-09
Client library - ModelStatusRequest

mkuczyns11:


client/python/lib/test/test_make_status_request_grpc.py
```diff
@@ -0,0 +1,33 @@
...
+def test_make_status_request_valid(name, version):
+    model_status_request = make_status_request(name, version)
+    assert model_status_request.model_name == name
+    assert model_status_request.model_version == version
+    assert model_status_request.raw_request.model_spec.name == name
+    assert model_status_request.raw_request.model_spec.version.value == version
+
+invalid_input = [
+    (12, 1, pytest.raises(TypeError), "12 has type int, but expected one of: bytes, unicode"),
+    (None, -1, pytest.raises(ValueError), "model_version should be positive integer, but is negative"),
```
mzegla, line 22: Can we also have a case when name is None and version is a valid value?

client/python/lib/ovmsclient/tfs_compat/grpc/requests.py
```diff
@@ -144,5 +148,24 @@ def make_status_request(model_name, model_version=0):
...
+
+    if not isinstance(model_name, str):
+        raise TypeError(f'model_name should be type string, but is type {type(model_name).__name__}')
     
-    raise NotImplementedError
+    if not isinstance(model_version, int):
+        raise TypeError(f'model_version should be type int, but is type {type(model_version).__name__}')
+
+    if not model_version.bit_length() <= 63:
+        raise ValueError(f'model_version should have max 63 bits, but has {model_version.bit_length()}')
```
mzegla, line 168: I would change this message to be easier for the user to interpret like:
`model_version should be in range <0, 9223372036854775807>`
In the code, upper bound should be evaluated (like: 2^63-1) rather than hardcoded.

client/python/lib/ovmsclient/tfs_compat/grpc/requests.py
```diff
@@ -144,5 +148,24 @@ def make_status_request(model_name, model_version=0):
...
+
+    if model_version < 0:
+        raise ValueError(f'model_version should be positive integer, but is negative')
```
mzegla, line 171: Looks like we can merge this condition to the one above with precision check. Then the message with appropriate values range would be well fitting to both scenarios and we would have value range check in one logical block.

client/python/lib/test/config.py
```diff
@@ -0,0 +1,19 @@
+
+import pytest
+
+MODEL_REQUEST_VALID = [
```
mzegla, line 4:
```suggestion
MODEL_SPEC_VALID = [
```

client/python/lib/test/config.py
```diff
@@ -0,0 +1,19 @@
...
+    ("model_name", 1),
+    ("3", 17),
+    ("*.-", 0),
+    ("model_name", 9223372036854775806),
+]
+
+MODEL_REQUEST_INVALID = [
```
mzegla, line 11:
```suggestion
MODEL_SPEC_INVALID = [
```

client/python/lib/test/config.py
```diff
@@ -0,0 +1,19 @@
...
+    (12, 1, pytest.raises(TypeError), "model_name should be type string, but is type int"),
+    (None, -1, pytest.raises(TypeError), "model_name should be type string, but is type NoneType"),
+    (None, 1, pytest.raises(TypeError), "model_name should be type string, but is type NoneType"),
+    ("model_name", "3", pytest.raises(TypeError), "model_version should be type int, but is type str"),
+    ("model_name" , None, pytest.raises(TypeError), "model_version should be type int, but is type NoneType"),
+    ("model_name", -1, pytest.raises(ValueError), "model_version should be positive integer, but is negative"),
+    ("model_name", 9223372036854775809, pytest.raises(ValueError), "model_version should have max 63 bits, but has 64"),
```
mzegla, line 18: The 3rd argument is in form `pytest.raises(<error_type>)`
Can we change it to just <error_type> and then in the test go with:
`with pytest.raises(exception) as e_info:`
It would give us more understanding of what is happening on the test level and keep just raw values in the test cases

client/python/lib/test/test_check_model_spec.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from ovmsclient.tfs_compat.grpc.requests import _check_model_spec
+from config import MODEL_REQUEST_INVALID, MODEL_REQUEST_VALID
+
+@pytest.mark.parametrize("name, version", MODEL_REQUEST_VALID)
+def test_check_model_spec_valid(name, version):
+    model_status_request = _check_model_spec(name, version)
+
+@pytest.mark.parametrize("name, version, exception, message", MODEL_REQUEST_INVALID)
```
mzegla, line 10:
```suggestion
@pytest.mark.parametrize("name, version, expected_exception, expected_message", MODEL_REQUEST_INVALID)
```

client/python/lib/test/test_make_status_request.py
```diff
@@ -0,0 +1,19 @@
...
+
+@pytest.mark.parametrize("name, version", MODEL_REQUEST_VALID)
+def test_make_status_request_valid(name, version):
+    request = model_status_request = make_status_request(name, version)
+    assert isinstance(request, GrpcModelStatusRequest)
+    assert request.model_version == version
+    assert request.model_name == name
+    assert isinstance(request.raw_request, GetModelStatusRequest)
+
+@pytest.mark.parametrize("name, version, exception, message", MODEL_REQUEST_INVALID)
```
mzegla, line 15:
```suggestion
@pytest.mark.parametrize("name, version, expected_exception, expected_message", MODEL_REQUEST_INVALID)
```

mzegla (approved):

michalkulakowski:
LGTM

michalkulakowski (approved):

PR #772, opened 2021-07-14
Client library - ModelStatusResponse

mkuczyns11:


client/python/lib/ovmsclient/tfs_compat/grpc/responses.py
```diff
@@ -26,5 +26,22 @@ def to_dict(self):
...
+        30: "AVAILABLE",
+        40: "UNLOADING",
+        50: "END"
+        }
+
     def to_dict(self):
-        raise NotImplementedError
+        result = {}
+        model_version_status = self.raw_response.model_version_status
+        for i in model_version_status:
```
mzegla, line 41: Can you rename it to be more meaningful? Especially the `i` variable.

client/python/lib/test/config.py
```diff
@@ -14,6 +14,29 @@
...
+        model_version_status = ModelVersionStatus()
+        model_version_status.version = i
+        model_version_status.state = 10
+        model_version_status.status.CopyFrom(status)
+
+        model_versions.append(model_version_status)
+    raw_response.model_version_status.extend(model_versions)
+    return raw_response
+
+MODEL_RESPONSE_CORRECT = [(generate_response(i), i) for i in range(4)]
```
mzegla, line 38: I think it's not intuitive to evaluate what the test cases will be. Also generated state and error message is hardcoded.
Maybe we could have two methods:
- create_model_status_response - that would take version, code and error message as arguments
- merge_model_status_responses - that would take a list of objects generated by the method above and merge them into one

This way the specification of the single version response can be easily read from the function parameters and we could create more complex responses (with multiple version in different states with different messages) in a more readable form.

client/python/lib/test/test_GrpcModelStatusResponse_to_dict.py
```diff
@@ -0,0 +1,33 @@
...
+    response = GrpcModelStatusResponse(raw_response)
+    response_dictionary = response.to_dict()
+    assert isinstance(response_dictionary, dict)
+    assert len(response_dictionary) == expected
+    for key in response_dictionary:
+        assert isinstance(response_dictionary[key], dict)
+        assert len(response_dictionary[key]) == 3
+        assert response_dictionary[key]['state'] == "START"
+        assert response_dictionary[key]['error_code'] == 1
+        assert response_dictionary[key]['error_message'] == "msg"
```
mzegla, line 33: This should be parametrized with different values

client/python/lib/test/test_GrpcModelStatusResponse_to_dict.py
```diff
@@ -0,0 +1,39 @@
...
+
+from ovmsclient.tfs_compat.grpc.responses import GrpcModelStatusResponse
+from config import RAW_MODEL_RESPONSE_VALID, create_model_status_response, merge_model_status_responses
+
+@pytest.mark.parametrize("model_raw_response_dict" , RAW_MODEL_RESPONSE_VALID)
+def test_GrpcModelStatusResponse_to_dict_valid(model_raw_response_dict):
+    model_raw_responses = []
+    for version in model_raw_response_dict:
+        model_raw_responses.append(create_model_status_response(version, model_raw_response_dict[version][1], model_raw_response_dict[version][2],
+         model_raw_response_dict[version][0]))
```
mzegla, line 27: 1. You can go with: `for version, status in model_raw_response_dict.items():` for easier access to the values.
2. It would be easier to read if instead of status[0], status[1], we could have status["state"], status["error_code"] etc.

client/python/lib/test/test_GrpcModelStatusResponse_to_dict.py
```diff
@@ -0,0 +1,39 @@
...
+    response = GrpcModelStatusResponse(raw_response)
+    response_dictionary = response.to_dict()
+    assert isinstance(response_dictionary, dict)
+    assert len(response_dictionary) == len(model_raw_response_dict)
+    for version in model_raw_response_dict:
+        assert version in response_dictionary
+        assert isinstance(response_dictionary[version], dict)
+        assert response_dictionary[version]['error_code'] == model_raw_response_dict[version][1]
+        assert response_dictionary[version]['error_message'] == model_raw_response_dict[version][2]
+        assert response_dictionary[version]['state'] == GrpcModelStatusResponse._STATE_TO_STRING_MAPPING[model_raw_response_dict[version][0]]
```
mzegla, line 39: The comment above would apply to this part as well

mzegla (approved):

rasapala (approved):

PR #774, opened 2021-07-14
image cropping and source type in parameters

dtrawins:


example_client/camera_client.py
```diff
@@ -129,7 +130,12 @@ def run(self):
 
 def grab_frame(cap):
     ret, frame = cap.read()
-    frame = cv2.resize(frame, (704, 704), interpolation=cv2.INTER_AREA)
+    # crop square and resize if original image is too small for the model
+    if frame.shape[0] > 704 and frame.shape[1] > 704:
+        frame = frame[0:704, 0:704]
+    else:
+        frame = frame[0:frame.shape[0], 0:frame.shape[0]]
```
dkalinowski, line 137: `frame[0:frame.shape[0], 0:frame.shape[1]]`

dtrawins: the goal here is to crop a square from a rectangle

dkalinowski: but index may get out of bounds since frame.shape[0] may be larger than width

dkalinowski (approved):

PR #777, opened 2021-07-16
Mkulakow/make ndarray

michalkulakowski:


client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -61,6 +63,39 @@ def make_tensor_proto(values, dtype=None, shape=None):
 
     raise NotImplementedError
 
+def as_numpy_dtype(tensor_dtype):
```
mzegla, line 66:
```suggestion
def _as_numpy_dtype(tensor_dtype):
```

For all methods not exposed to the user, please start the names with `_` to indicate they're private and should not participate in generating documentation.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -61,6 +63,39 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
+    elif tensor_dtype == 9:
+        return np.int64
+    elif tensor_dtype == 10:
+        return np.bool
+    elif tensor_dtype == 17:
+        return np.uint16
+    elif tensor_dtype == 22:
+        return np.uint32
+    elif tensor_dtype == 23:
+        return np.uint64
```
mzegla, line 94: Can you create enum with the names of tf types so we can have named consts here instead of numbers?

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -86,5 +121,55 @@ def make_ndarray(tensor_proto):
         >>> output = make_ndarray(tensor_proto)
         >>> print(output)
     '''
+    tensor_shape = [d.size for d in tensor_proto.tensor_shape.dim]
+    size = np.prod(tensor_shape, dtype=np.int64)
```
mzegla, line 125:
```suggestion
    num_elements = np.prod(tensor_shape, dtype=np.int64)
```
`size` can be ambiguous in this context (does it mean elements or bytes?)

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -86,5 +121,55 @@ def make_ndarray(tensor_proto):
...
+    dtype = as_numpy_dtype(tensor_proto.dtype)
```
mzegla, line 126: To avoid confusion, please either add tensor_ here and above for:
`tensor_shape, tensor_size/tensor_num_elements, tensor_dtype`
or remove it:
`shape, size/num_elements, dtype`

mzegla: Also I would suggest going withv name `np_dtype` to indicate that this variable holds type in numpy convention (not tf dtype nor basic python type)

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -86,5 +121,55 @@ def make_ndarray(tensor_proto):
         >>> output = make_ndarray(tensor_proto)
         >>> print(output)
     '''
+    tensor_shape = [d.size for d in tensor_proto.tensor_shape.dim]
```
mzegla, line 124: What if tensor_proto is invalid (does not contain shape, dtype or even data)? 
Is it possible to prepare such tensor_proto? If yes, we could have a method to validate if tensor_proto contains everything we need to make ndarray from it.

client/python/lib/test/test_make_ndarray.py
```diff
@@ -0,0 +1,55 @@
...
+# limitations under the License.
+#
+
+import pytest
+import numpy as np
+from ovmsclient.tfs_compat.grpc.tensors import make_ndarray
+from tensorflow.core.framework.tensor_shape_pb2 import *
+
+shape = TensorShapeProto(dim = [TensorShapeProto.Dim(size=3)])
+
```
mzegla, line 23: Why are you converting and comparing strings in the tests?
Is it possible to create expected ndarray with expected specification (data, shape, dtype) and compare to it?

mkuczyns11 (approved):

client/python/lib/test/test_tensors.py
```diff
@@ -0,0 +1,90 @@
...
+import pytest
+from tensorflow.core.framework.tensor_shape_pb2 import *
+from tensorflow.core.framework.tensor_pb2 import TensorProto
+from ovmsclient.tfs_compat.grpc.tensors import make_ndarray
+import numpy as np
+
+shape = TensorShapeProto(dim = [TensorShapeProto.Dim(size=3)])
+
+def test_make_ndarray_valid_int():
+    tensor_proto = TensorProto(tensor_shape=shape, dtype = 6, tensor_content = bytes([1,2,3]))
```
mzegla, line 26: Can you make const for dtype. What does 6 mean? 
Please apply to all below cases.

client/python/lib/test/test_tensors.py
```diff
@@ -0,0 +1,90 @@
...
+    array = make_ndarray(tensor_proto)
+    assert array.tolist() == [1, 2, 3]
+    assert array.dtype == np.int8
+
+def test_make_ndarray_valid_float():
+    hex_string = "0000803f0000004000004040"
```
mzegla, line 32: Is it possible we start with array and then eventually go through hex? 
It's hard to quickly tell that this value means [1, 2, 3] in float32.

Same applies to below cases.

client/python/lib/test/test_tensors.py
```diff
@@ -0,0 +1,90 @@
...
+    assert array.dtype == np.float32
+
+def test_make_ndarray_valid_string():
+    hex_string = "11111111"
+    tensor_proto = TensorProto(tensor_shape=shape, dtype = 7, string_val = [bytes.fromhex(hex_string)])
+    array = make_ndarray(tensor_proto)
+    assert array.tolist() == ['\x11\x11\x11\x11', '\x11\x11\x11\x11', '\x11\x11\x11\x11']
+    assert array.dtype == '<U4'
+
+def test_make_ndarray_valid_no_content():
```
mzegla, line 53: What if dtype == string?
String dtype is evaluated and returned before:
```
    if values.size == 0:
        return np.zeros(shape, np_dtype)
```

mzegla (approved):
Agreed offline that requested changes will be commited before merge

PR #778, opened 2021-07-16
Client library - ServingClient

mkuczyns11:


client/python/lib/ovmsclient/tfs_compat/grpc/serving_client.py
```diff
@@ -105,7 +116,111 @@ def get_model_status(self, request):
...
     @classmethod
     def _build(cls, config):
-        raise NotImplementedError
+        
+        _check_config(config)
+
+        grpc_address = f"{config['address']}:{config['port']}"
+
+        if 'tls_config' in config:
+            client_key, client_cert, server_cert = _prepare_certs(
```
mzegla, line 125: The call with parameters "to be evaluated" is hard to read. 
Could you take values extraction out and just call `_prepare_certs(server_cert_path, client_key_path, client_cert_path) `?
See: https://www.tutorialspoint.com/python/dictionary_get.htm to provide default `None` if key has not been found in the dict.

Also, the required parameters should go first and possible `Nones` should be placed at the end.

client/python/lib/ovmsclient/tfs_compat/grpc/serving_client.py
```diff
@@ -105,7 +116,111 @@ def get_model_status(self, request):
...
+        prediction_service_stub = prediction_service_pb2_grpc.PredictionServiceStub(channel)
+        model_service_stub = model_service_pb2_grpc.ModelServiceStub(channel)
+        stubs = {
+            "prediction_service_stub" : prediction_service_stub,
+            "model_service_stub" : model_service_stub
+        }
+
+        return cls(channel, stubs)
+
+def _prepare_certs(client_key, client_cert, server_cert):
```
mzegla, line 147: Please make `server_cert` first parameter and start with it in the function logic. 
Also change parameters names to end with `_path`

client/python/lib/ovmsclient/tfs_compat/grpc/serving_client.py
```diff
@@ -105,7 +116,111 @@ def get_model_status(self, request):
...
+        
+    with open(server_cert, 'rb') as f:
+        server_cert = f.read()
+    try:
+        s_cert = OpenSSL.crypto.load_certificate(
+            OpenSSL.crypto.FILETYPE_PEM,
+            server_cert
+        )
+    except OpenSSL.crypto.Error as e_info:
+        raise ValueError('server certificate file is not valid')
```
mzegla, line 176: The code is almost the same for both certificates. Please create a method like: `_check_certificate_valid(certificate_path)`
and just use it here to avoid duplication.

client/python/lib/ovmsclient/tfs_compat/grpc/serving_client.py
```diff
@@ -105,7 +116,111 @@ def get_model_status(self, request):
...
+
+def _prepare_certs(client_key, client_cert, server_cert):
+    
+    if client_key is not None:
+        with open(client_key, 'rb') as key:
+            client_key = key.read()
+        try:
+            p_key = OpenSSL.crypto.load_privatekey(OpenSSL.crypto.FILETYPE_PEM, client_key)
+        except OpenSSL.crypto.Error as e_info:
+            raise ValueError('client key file is not valid')
```
mzegla, line 155: This can be extracted to sub function like:  `_check_private_key_valid(certificate_path)`

client/python/lib/ovmsclient/tfs_compat/grpc/serving_client.py
```diff
@@ -105,7 +116,111 @@ def get_model_status(self, request):
...
+        )
+    except OpenSSL.crypto.Error as e_info:
+        raise ValueError('server certificate file is not valid')
+
+    return client_key, client_cert, server_cert
+
+def _check_config(config):
+    
+    if 'address' not in config or 'port' not in config:
+        raise Exception('The minimal config must contain address and port')
```
mzegla, line 183:
```suggestion
        raise ValueError('The minimal config must contain address and port')
```

client/python/lib/ovmsclient/tfs_compat/grpc/serving_client.py
```diff
@@ -105,7 +116,111 @@ def get_model_status(self, request):
...
+        raise TypeError(f'port type should be int, but is type {type(port).__name__}')
+
+    if port.bit_length() > 16 or port < 0:
+        raise ValueError(f'port should be in range <0, {2**16-1}>')
+
+def _check_tls_config(tls_config):
+
+    if 'server_cert_path' not in tls_config:
+        raise ValueError(f'server_cert_path is not defined in tls_config')
+    if 'client_key_path' in tls_config and 'client_cert_path' not in tls_config or 'client_cert_path' in tls_config and 'client_key_path' not in tls_config:
```
mzegla, line 213: It's a quite long condition. I wonder if something like that would be more readable:
```
if 'client_key_path' in tls_config != 'client_cert_path' in tls_config:
    raise(...)
```
Up to you 😄

client/python/lib/test/config.py
```diff
@@ -30,3 +30,232 @@
...
+]
+
+PORT_INVALID = [
+    ("9000", TypeError, "port type should be int, but is type str"),
+    ([2**16-1], TypeError, "port type should be int, but is type list"),
+    (2**16, ValueError, f"port should be in range <0, {2**16-1}>"),
+    (-1, ValueError, f"port should be in range <0, {2**16-1}>")
+]
+
+PATH_VALID = "valid_path"
```
mzegla, line 59: Please add a comment that this is just a special value for tests purposes

client/python/lib/test/config.py
```diff
@@ -30,3 +30,232 @@
...
+
+TLS_CONFIG_VALID = [
+    ({"server_cert_path" : PATH_VALID}, 1),
+    ({"client_key_path" : PATH_VALID, "client_cert_path" : PATH_VALID, "server_cert_path" : PATH_VALID}, 3),
```
mzegla, line 63: What are the numbers at the end? Can we have them in consts if they mean something?

client/python/lib/test/config.py
```diff
@@ -30,3 +30,232 @@
...
+    ({"client_key_path" : PATH_VALID, "client_cert_path" : "/very/invalid", "server_cert_path" : PATH_VALID}, ValueError, f'/very/invalid is not valid path to file', 2),
+    ({"client_key_path" : PATH_VALID, "client_cert_path" : PATH_VALID, "server_cert_path" : "third_invalid_path"}, ValueError, f'third_invalid_path is not valid path to file', 3),
+
+]
+
+CONFIG_VALID = [
+    ({
+        "address" : "localhost",
+        "port" : 9000
+    }, {"check_address" : 1, "check_port" : 1, "check_tls_config" : 0}),
```
mzegla, line 96: I assume those 1 and 0 means valid/invalid here. Can we have it named in the const and applied here and in all cases?

client/python/lib/test/test_check_tls_config.py
```diff
@@ -0,0 +1,40 @@
...
+    with pytest.raises(expected_exception) as e_info:
+        _check_tls_config(tls_config)
+        assert str(e_info.value) == expected_message
+    
+    assert mock_method.call_count ==  isfile_called_count
+
+def is_valid_path(path):
+    if path == PATH_VALID:
+        return True
+    return False
```
mzegla, line 40: I just though that we could drop this approach and remove `PATH_VALID` value. 
Negative cases could have bool value like: `is_path_valid`  and in mocker side effects you could go with:
`mocker.patch("path/to/file", side_effect=lambda path : is_path_valid)`

michalkulakowski (approved):

michalkulakowski:
LGTM

client/python/lib/ovmsclient/tfs_compat/grpc/serving_client.py
```diff
@@ -14,11 +14,22 @@
 # limitations under the License.
 #
 
+import grpc
+from validators import ipv4, domain
+import os
+import OpenSSL.crypto
```
dkalinowski, line 20: Do we want such large library dependency just for validation? Let's discuss this in next sync.

client/python/lib/test/config.py
```diff
@@ -82,6 +50,23 @@ class ErrorCode(Enum):
     # DATA_LOSS = 15
     # DO_NOT_USE_RESERVED_FOR_FUTURE_EXPANSION_USE_DEFAULT_IN_SWITCH_INSTEAD = 20
 
+MODEL_SPEC_VALID = [
+    ("model_name", 1),
```
dkalinowski, line 54: Here this is not clear what the integer values mean, maybe add single line comment with field names?

dkalinowski: The same goes for all test case definitions

client/python/lib/test/config.py
```diff
@@ -96,3 +81,443 @@ class ErrorCode(Enum):
...
+SERVER_CERT_PATH_VALID, SERVER_CERT_PATH_INVALID = "server_cert_path_valid", "server_cert_path_invalid"
+CLIENT_CERT_PATH_VALID, CLIENT_CERT_PATH_INVALID = "client_cert_path_valid", "client_cert_path_invalid"
+CLIENT_KEY_PATH_VALID, CLIENT_KEY_PATH_INVALID = "client_key_path_valid", "client_key_path_invalid"
+
+TLS_CONFIG_VALID = [
+    ({"server_cert_path" : PATH_VALID}, CallCount.ONE),
+    ({"client_key_path" : PATH_VALID, "client_cert_path" : PATH_VALID, "server_cert_path" : PATH_VALID}, CallCount.THREE),
+]
+
+TLS_CONFIG_INVALID = [
```
dkalinowski, line 123: This could be formatted to decrease number of letters in line

dkalinowski (changes requested):

dkalinowski (approved):

PR #786, opened 2021-07-20
Client library - ModelMetadataResponse

mkuczyns11:


michalkulakowski (approved):

client/python/lib/ovmsclient/tfs_compat/grpc/responses.py
```diff
@@ -15,26 +15,66 @@
...
+        tensor_shape = {key: serving_inputs[key].tensor_shape for key in serving_inputs.keys()}
+
+        input_blobs_dict = {}
+        for input_blob in input_blobs:
+            inputs_shape = [d.size for d in tensor_shape[input_blob].dim]
+            tensor_dtype = serving_inputs[input_blob].dtype
+            input_blobs_dict[input_blob] = dict([
+                ("shape", inputs_shape),
+                ("dtype", DataType.Name(tensor_dtype))
+            ])
```
mzegla, line 47: I think we can simplify this part as well as the similar one below for the output.
What if we wouldn't create `input_blobs` and `tensor_shape` dicts?
We could just have:
```
       inputs_metadata = {}
        for input_name, input_info in model_signature.inputs:
            input_shape = [d.size for d in input_info.tensor_shape.dim]
            inputs_metadata[input_name] = dict([
                ("shape", input_shape),
                ("dtype", DataType.Name(input_info.dtype))
            ])
```

This code snippet assumes, you switch from `serving_default` to `model_signature` name in assignment in line 34:
`serving_default = signature_map.ListFields()[0][1]['serving_default']`

That would look better imho.

client/python/lib/test/config.py
```diff
@@ -652,5 +653,74 @@ class CallCount(IntEnum):
...
+]
+# response_dict = {
+# 'version' : model_version,
+# 'name' : model_name,
+# 'inputs' : inputs_dict,
+# 'outputs' : outputs_dict
+# }
+MODEL_METADATA_RESPONSE_VALID = [
+    {
+        'version' : 0,
```
mzegla: Returned version number must be larger than 0

client/python/lib/test/test_grpc_responses.py
```diff
@@ -0,0 +1,155 @@
...
+    model_spec.name = model_raw_metadata_response_dict['name']
+    model_spec.version.value = model_raw_metadata_response_dict['version']
+
+    raw_response.model_spec.CopyFrom(model_spec)
+
+    metadata = SignatureDef()
+    
+    inputs = model_raw_metadata_response_dict['inputs']
+    for input_key in inputs.keys():
+        input_info = inputs[input_key]
```
mzegla, line 98:
```suggestion
    for input_name, input_info in inputs.items():
```

client/python/lib/test/test_grpc_responses.py
```diff
@@ -0,0 +1,155 @@
...
+def create_model_metadata_response(model_raw_metadata_response_dict):
+    raw_response = GetModelMetadataResponse()
+
+    model_spec = ModelSpec()
+    model_spec.name = model_raw_metadata_response_dict['name']
+    model_spec.version.value = model_raw_metadata_response_dict['version']
+
+    raw_response.model_spec.CopyFrom(model_spec)
+
+    metadata = SignatureDef()
```
mzegla, line 94: You use `metadata` name below and reassign different type to it. 
Perhaps we could go with just `signature_def` here?

client/python/lib/ovmsclient/tfs_compat/grpc/responses.py
mkuczyns11: This change causes 3 tests to fail with:
```ValueError: not enough values to unpack (expected 2, got 1)```

mkuczyns11: with ```for input_name, input_info in model_signature.inputs.items():``` now works as intended

mzegla (approved):

PR #788, opened 2021-07-20
CVS-60502_CVS-59189 Switch to using system curl & openssl

rasapala:
Till now we used several versions of openssl due to our dependencies. It was mainly due to boringssl usage in Google products which is used by both Google Cloud Storage as well as TFS/TF. Since boringssl is implementing the same interface that openssl does this caused symbol conflicts during linking/loading/running stages.

Solution for now is to mask system openssl as boringssl dependency. GCS additionally build curl with different openssl flags than the system libs which resulted in having deprecated symbol present thus the replacement of curl.

Unfortunately the plain simple "select" clause did not work for the system libs path selection so workaround for now for this is to have symbolic link in case of fedora family OS's.

### DIFF
--- a/Dockerfile.centos
+++ b/Dockerfile.centos
@@ -138,8 +138,8 @@ COPY external /ovms/external/
 COPY third_party /ovms/third_party/
 
 ENV BAZEL_LINKLIBS=-l%:libstdc++.a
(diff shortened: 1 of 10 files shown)


dkalinowski (approved):

michalkulakowski (approved):

rasapala:
Currently working on azurestorage fix - this change introdduces regression in azure storage. Cannot get rid off the boringssl symbols.

mzegla (approved):

dtrawins (approved):

PR #791, opened 2021-07-20
CVS-59402 - Incorrect error while loading pipeline using demultiplexer

mkuczyns11:


mkuczyns11:
In this solution I am utilizing previously computed tensorInputShape/tensorOutputShape, because in previous steps those two were influanced by demultiplexer, which was not the case for shapes saved in TensorInfo.

atobiszei (approved):

dkalinowski (approved):

PR #807, opened 2021-07-27
Introduce make_tensor_proto

michalkulakowski:


client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -14,11 +14,52 @@
 # limitations under the License.
 #
 
+from tensorflow.core.framework.tensor_shape_pb2 import *
```
dkalinowski, line 17: Is it good idea to import everything? What do we use from tensor_shape_pb2 anyway? @mkuczyns11

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -14,11 +14,52 @@
...
+        return np.uint16
+    elif tensor_dtype == DataType.DT_UINT32:
+        return np.uint32
+    elif tensor_dtype == DataType.DT_UINT64:
+        return np.uint64
+    else:
+        return np.uint32
+
+def _is_array_like(value):
+    if isinstance(value, list) or isinstance(value, np.ndarray):
```
dkalinowski, line 58: `return isinstance(value, list) or isinstance(value, np.ndarray)`

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
mkuczyns11: I investigated that and it seems, that we are using only TensorShapeProto.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -63,43 +104,66 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
-        return np.uint32
-
+        infered_shape = shape
+    
+    value = values
+    while dtype is None or (shape is None and _is_array_like(value)):
+        if _is_array_like(value):
+            if shape is None:
+                infered_shape.append(len(value))
+            if(len(value)>0):
```
dkalinowski, line 117: redundant `(` `)`

dkalinowski, line 117: If value contains no elements we will end up in endless loop. There is also no test for that.

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,88 @@ def test_make_ndarray_invalid_no_type():
...
+    np_values = np.array(values, np.int8)
+    assert tensor_proto.tensor_content == np_values.tobytes()
+    assert tensor_proto.dtype == DataType.DT_INT8
+
+def test_make_tensor_proto_valid_float():
+    values = [1.0, 2.0, 3.0]
+    tensor_proto = make_tensor_proto(shape=[3], dtype = DataType.DT_FLOAT, values = values)
+    np_values = np.array(values, np.float32)
+    assert tensor_proto.tensor_content == np_values.tobytes()
+    assert tensor_proto.dtype == DataType.DT_FLOAT
```
dkalinowski, line 120: We should be checking tensor shape in all tests as well

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,88 @@ def test_make_ndarray_invalid_no_type():
...
+def test_make_tensor_proto_valid_scalar():
+    values = 5.0
+    tensor_proto = make_tensor_proto(shape=[3], dtype = DataType.DT_FLOAT, values = values)
+    np_values = np.array(values, np.float32)
+    assert tensor_proto.tensor_content == np_values.tobytes()
+    assert tensor_proto.dtype == DataType.DT_FLOAT
+
+def test_make_tensor_proto_valid_string():
+    values = bytes([0x13, 0x00, 0x00, 0x00, 0x08, 0x00])
+    tensor_proto = make_tensor_proto(shape=[4], dtype = DataType.DT_STRING, values = [values])
```
dkalinowski, line 131: Why passing shape [4] if batch=1? Shape should be [1]. I think we should also validate if number binary images passed to make_tensor_proto matches the shape? @michalkulakowski @dtrawins @mkuczyns11

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -14,11 +14,52 @@
 # limitations under the License.
 #
 
+from tensorflow.core.framework.tensor_shape_pb2 import *
 from tensorflow.core.framework.tensor_pb2 import TensorProto
 from tensorflow.core.framework.types_pb2 import DataType
 from enum import IntEnum
```
mkuczyns11, line 20: IntEnum is not used

client/python/lib/test/test_tensors.py
dkalinowski: The same goes for non binary data, we specify shape but we should also validate data if is in same shape as specified in parameter

mkuczyns11: I agree. With that config:
`data = 5.0`
`tensor_proto = make_tensor_proto(data, shape = [3])`
In output tensor_shape is 
`dim {
  size: 3
}`, but the data provided has is a scalar.

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,135 @@ def test_make_ndarray_invalid_no_type():
...
+    exception = exception_info.value
+    assert str(exception) == "shape type should be list, but is int"
+
+def test_make_tensor_proto_invalid_dimensions_no_shape_provided():
+    values = [[1,2,3], [4,5]]
+    with pytest.raises(ValueError) as exception_info:
+        make_tensor_proto(shape=None, dtype = DataType.DT_INT8, values = values)
+
+    exception = exception_info.value
+    assert str(exception) == "setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part."
```
mkuczyns11, line 230: In make_tensor_proto with shape=None, we determine the shape ourselves and we don't check whether it matches values list. Would it be better to check
```
list(np.array(values).shape) == infered_shape
```
and raise an exception if they do not match @dkalinowski ?
In this test the comparsion would be: [2] == [2,3]

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,135 @@ def test_make_ndarray_invalid_no_type():
...
+
+def test_make_tensor_proto_invalid_dtype_as_float():
+    values = [[1.0, 2.0, 3.0], [1.0, 2.0, 3.0]]
+    with pytest.raises(TypeError) as exception_info:
+        make_tensor_proto(shape = None, dtype = int, values = values)
+
+    exception = exception_info.value
+    assert str(exception) == "<class 'int'> has type type, but expected one of: int, long"
```
mkuczyns11, line 238: Should we check if dtype provided is at least type int, long? @dkalinowski

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -106,16 +102,22 @@ def make_tensor_proto(values, dtype=None, shape=None):
     '''
     if shape is None:
         infered_shape = []
-    else:
+    elif _is_array_like(shape):
+        if shape != list(np.array(values).shape):
+            raise ValueError(f'shape provided does not match values array shape')
```
dkalinowski, line 107: Please check with `make_tensor_proto` from TF - in case we provide data in different shape than the shape from parameter what should we do. Before your adjustment we did not care about provided shape. Now we raise an error if it does not match. What TF does I think is that they reshape the data to expected shape (from parameter). @dtrawins  what is the desired behavior here?

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -55,10 +54,7 @@ def _as_numpy_dtype(tensor_dtype):
         return np.uint32
 
 def _is_array_like(value):
-    if isinstance(value, list) or isinstance(value, np.ndarray):
-        return True
-    else:
-        return False
+    return isinstance(value, list) or isinstance(value, np.ndarray)
 
 def make_tensor_proto(values, dtype=None, shape=None):
```
dkalinowski, line 59: I can't see `list` object validation. What if user passes data:
```
invalid_data = [
  [ 1, 2, 5 ],
  [ 8 ],
  [ 3, 9, 1]
]
make_tensor_proto(invalid_data)
```
?

dkalinowski: Ok I see, then converting to numpy array will fail and we handle that.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
mkuczyns11: You are right, TF does indeed reshape the data to expected shape. In that case I think we might want to remove this.

client/python/lib/test/test_tensors.py
dkalinowski: I think this test would fail for other reason. `[[1,2,3], [4,5]]` is not valid numpy array due to number of elements in each row.
But to answer your question I think it's not neccesssary check since we use numpy inside make_tensor_proto to get the shape anyway.

client/python/lib/test/test_tensors.py
dkalinowski: What is expected behavior of the library? To follow numpy behavior?

client/python/lib/test/test_tensors.py
```diff
@@ -111,42 +111,63 @@ def test_make_tensor_proto_valid_int():
...
+def test_make_tensor_proto_valid_empty_list():
+    values = []
+    tensor_proto = make_tensor_proto(shape=[0], dtype = DataType.DT_INT8, values = values)
+    np_values = np.array(values, np.int8)
+    assert tensor_proto.tensor_content == np_values.tobytes()
+    assert tensor_proto.dtype == DataType.DT_INT8
+    assert tensor_proto.tensor_shape == TensorShapeProto(dim = [TensorShapeProto.Dim(size=0)])
+
+def test_make_tensor_proto_valid_empty_list_of_empty_lists():
+    values = [[],[],[]]
```
dkalinowski, line 125: Here you add test which is similar to the one which could expose endless loop isssue. The only thing that is missing is sestting shape to None. Please add such test so we are sure we will not get hang.

client/python/lib/test/test_tensors.py
mkuczyns11: The exception we are getting is from TensorProto() constructor call. Earlier we are calling:
```
np_dtype = _as_numpy_dtype(dtype)
np_values = np.array(values, np_dtype)
```
_as_numpy_dtype(dtype) is our function and as if for now if we do not provide dtype=DataType.* the function returns np.uint32.
This cause np.array() function to work properly.
Maybe we should check if isinstance(dtype, int) in _as_numpy_dtype() and raise an exception if it is not?

client/python/lib/test/test_tensors.py
```diff
@@ -113,6 +113,14 @@ def test_make_tensor_proto_valid_int():
     assert tensor_proto.dtype == DataType.DT_INT8
     assert tensor_proto.tensor_shape == TensorShapeProto(dim = [TensorShapeProto.Dim(size=3)])
 
+def test_make_tensor_proto_valid_int_reshape():
+    values = [1, 2, 3]
+    tensor_proto = make_tensor_proto(shape=[4], dtype = DataType.DT_INT8, values = values)
+    np_values = np.array(values, np.int8)
+    assert tensor_proto.tensor_content == np_values.tobytes()
+    assert tensor_proto.dtype == DataType.DT_INT8
+    assert tensor_proto.tensor_shape == TensorShapeProto(dim = [TensorShapeProto.Dim(size=4)])
```
dkalinowski, line 122: is it possible to reshape 3 elements to [4]?

mkuczyns11: Yes, it is possible.
With values = [1,2,3] and shape = [4] the tensor_proto from output looks like this:
```
dtype: DT_INT32
tensor_shape {
  dim {
    size: 4
  }
}
tensor_content: "\001\000\000\000\002\000\000\000\003\000\000\000"
```

mzegla: `tensor_shape` is not connected to any other field either `tensor_content` or `val` containers.
Therefore we need to make sure the shape value and the actual data are adjusted throughout our processing.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -63,43 +100,70 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
+        elif isinstance(value, np.complex64):
+            dtype = DataType.DT_COMPLEX64
+        elif isinstance(value, np.complex128):
+            dtype = DataType.DT_COMPLEX128
+        elif isinstance(value, complex):
+            dtype = DataType.DT_COMPLEX64
+        elif isinstance(value, bool):
+            dtype = DataType.DT_BOOL
+        else:
+            dtype = DataType.DT_INT32
```
mzegla, line 154: Can we extract this block to a function

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,175 @@ def test_make_ndarray_invalid_no_type():
...
+
+def test_make_tensor_proto_valid_2_dims_shape():
+    values = [[1.0, 2.0, 3.0], [1.0, 2.0, 3.0]]
+    tensor_proto = make_tensor_proto(shape=[2,3], dtype = DataType.DT_FLOAT, values = values)
+    np_values = np.array(values, np.float32)
+    assert tensor_proto.tensor_content == np_values.tobytes()
+    assert tensor_proto.dtype == DataType.DT_FLOAT
+    assert tensor_proto.tensor_shape == TensorShapeProto(dim = [TensorShapeProto.Dim(size=2), TensorShapeProto.Dim(size=3)])
+    
+def test_make_tensor_proto_valid_2_dims_shape_reshape():
```
mzegla, line 204: Isn't it the same case as the one above?

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,175 @@ def test_make_ndarray_invalid_no_type():
...
+    tensor_proto = make_tensor_proto(shape=[2,3], dtype = DataType.DT_FLOAT, values = values)
+    np_values = np.array(values, np.float32)
+    assert tensor_proto.tensor_content == np_values.tobytes()
+    assert tensor_proto.dtype == DataType.DT_FLOAT
+    assert tensor_proto.tensor_shape == TensorShapeProto(dim = [TensorShapeProto.Dim(size=2), TensorShapeProto.Dim(size=3)])
+
+def test_make_tensor_proto_valid_ndarray():
+    values = [[1.0, 2.0, 3.0], [1.0, 2.0, 3.0]]
+    np_values = np.array(values, dtype = np.float32)
+    _shape = TensorShapeProto(dim = [TensorShapeProto.Dim(size=1), TensorShapeProto.Dim(size=2), TensorShapeProto.Dim(size=3)])
```
mzegla, line 215: Why shape (1,2,3) and not (2,3)?

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,175 @@ def test_make_ndarray_invalid_no_type():
...
+    
+def test_make_tensor_proto_valid_2_dims_shape_reshape():
+    values = [[1.0, 2.0, 3.0], [1.0, 2.0, 3.0]]
+    tensor_proto = make_tensor_proto(shape=[2,3], dtype = DataType.DT_FLOAT, values = values)
+    np_values = np.array(values, np.float32)
+    assert tensor_proto.tensor_content == np_values.tobytes()
+    assert tensor_proto.dtype == DataType.DT_FLOAT
+    assert tensor_proto.tensor_shape == TensorShapeProto(dim = [TensorShapeProto.Dim(size=2), TensorShapeProto.Dim(size=3)])
+
+def test_make_tensor_proto_valid_ndarray():
```
mzegla, line 212: What are we testing in this case? 
We should test either `make_tensor_proto` or `make_ndarray`. 
If one is subject to testing then the other one should be mocked and return fixed value.

If both are tested then change the name of this test to include `make_ndarray`

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,175 @@ def test_make_ndarray_invalid_no_type():
...
+    assert tensor_proto.tensor_shape == expected_shape
+    assert tensor_proto.dtype == DataType.DT_FLOAT
+
+def test_make_tensor_proto_invalid_dtype_provided():
+    values = [[1.0, 2.0, 3.0], [1.0, 2.0, 3.0]]
+    with pytest.raises(TypeError) as exception_info:
+        make_tensor_proto(shape = None, dtype = DataType.DT_STRING, values = values)
+
+    exception = exception_info.value
+    assert str(exception) == "[1.0, 2.0, 3.0] has type list, but expected one of: bytes"
```
mzegla, line 246: Can we bring this part under `with` scope?

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,175 @@ def test_make_ndarray_invalid_no_type():
...
+
+def test_make_tensor_proto_invalid_dimsions():
+    values = [[1.0, 2.0], [1.0, 2.0, 3.0]]
+    with pytest.raises(ValueError) as exception_info:
+        make_tensor_proto(shape=[2,3], dtype = DataType.DT_FLOAT, values = values)
+
+    exception = exception_info.value
+    assert str(exception) == "setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part."
```
mzegla, line 254: Can we bring this part under `with` scope?

client/python/lib/test/test_tensors.py
```diff
@@ -104,3 +104,175 @@ def test_make_ndarray_invalid_no_type():
...
+
+def test_make_tensor_proto_invalid_shape_type():
+    values = 5.0
+    with pytest.raises(TypeError) as exception_info:
+        make_tensor_proto(shape=0, dtype = DataType.DT_FLOAT, values = values)
+
+    exception = exception_info.value
+    assert str(exception) == "shape type should be list, but is int"
```
mzegla, line 262: Can we bring this part under `with` scope?

client/python/lib/test/test_tensors.py
mzegla: Can we also bring this part under `with` scope?

client/python/lib/test/test_tensors.py
mzegla: Can we bring this part under `with` scope?

mzegla (commented):
The exception messages, exception types and the tested data could be paired together as test cases for easier modification and reusability.

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -63,43 +96,70 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
-
+        raise TypeError(f'shape type should be list or tuple, but is {type(shape).__name__}')
+    
+    tensor_values = values
+    if np.isscalar(tensor_values):
+        tensor_values = [tensor_values]
+
+    if _is_array_like(tensor_values):
+        if dtype is None:
+            tensor_values = np.array(tensor_values)
```
mzegla, line 112: Is there a risk an exception would happen here?

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -63,43 +96,70 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
+                dtype = tensor_type.TensorDtype
+            elif tensor_values.dtype.type == np.bytes_:
+                dtype = DT_STRING
+            else:
+                if tensor_values.dtype == object:
+                    raise ValueError(f"The requested array has inhomogeneous shape. Detected shape was {tensor_values.shape} + inhomogeneous part")
+                else:
+                    raise TypeError(f"values provided are not valid")
+        elif dtype == DT_STRING:
+            tensor_values = np.array(tensor_values)
```
mzegla, line 124: Is there a risk an exception would happen here? 
What if there's a list of `bytes` ? Like:
[b'ababab', b'ccccccccccccccccccccc']

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
mkuczyns11: with _is_array_like we are sure that tensor_values is either type list or np.ndarray.
There are cases such as this ```values = [[(1,2,3)],[(1,2,3)],[(1,2,3)]], dtype = DataType.DT_INT64```
In that scenario everything works, however with ```values = [[(1,2,3)],[(1,2)],[(1,2,3)]]``` we are getting
```ValueError: setting an array element with a sequence.```
I think that np.array() will not raise an exception while we are sure about the type being list/np.ndarray.
However we can expect exceptions while creating np arrays with declared dtype.

Comment (mkuczyns11) [reply client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py:112]: That error was generated by ```tensor_values = tensor_values.astype(np_dtype)```

File: client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py (hunk 21 above)
Comment (mkuczyns11) [reply client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py:124]: ```[b'ababab', b'ccccccccccccccccccccc']``` does not generate any kinds of exceptions

File: client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py [hunk 22]
```diff
@@ -14,11 +14,44 @@
...
+}
+
+TENSOR_TO_NP_MAP = {v.TensorDtype: k for k, v in NP_TO_TENSOR_MAP.items()}
+
+TENSOR_DTYPE_TO_PROTOFIELD = {v.TensorDtype : v.TensorProtoField for v in NP_TO_TENSOR_MAP.values()}
+
+def _is_array_like(value):
+    return isinstance(value, (list, np.ndarray))
+
+def _is_shape_valid(shape):
```
mzegla, line 52: Perhaps we should make sure that values inside shape list or tuple are unsigned integers?

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -63,43 +96,72 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
+        inferred_shape = []
+    elif _is_shape_valid(shape):
+        inferred_shape = list(shape)
     else:
-        return np.uint32
-
+        raise TypeError(f'shape type should be list or tuple, but is {type(shape).__name__}')
+    
+    tensor_values = values
+    # scalars are packed into 1 elemt list
```
mzegla, line 107:
```suggestion
    # scalars are packed into 1-element list
```

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -63,43 +96,72 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
+    if np.isscalar(tensor_values):
+        tensor_values = [tensor_values]
+
+    # create np.ndarray from values and find its dtype if not provided
```
mzegla, line 111:
```suggestion
    # create numpy ndarray from values and find its dtype, if not provided
```

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -63,43 +96,72 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
+        # binary input
+        elif dtype == DT_STRING:
+            tensor_values = np.array(tensor_values)
+        else:
+            np_dtype = TENSOR_TO_NP_MAP.get(dtype)
+            if np_dtype is None:
+                raise TypeError(f"{dtype} is not valid dtype value")
+            
+            tensor_values = np.array(tensor_values)
+            # input was binary, but dtype was not DT_STRING
```
mzegla, line 131:
```suggestion
            # values are binary, but dtype was not DT_STRING
```

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -63,43 +96,72 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
+            if tensor_values.dtype.type == np.bytes_:
+                tensor_values = np.frombuffer(tensor_values.tobytes(), dtype=np_dtype)
+            else:
+                tensor_values = tensor_values.astype(np_dtype)
```
mzegla, line 135: You mentioned that this part may generate exceptions. Can end user get some useful information from them or should we throw our own here?

client/python/lib/ovmsclient/tfs_compat/grpc/tensors.py
```diff
@@ -63,43 +96,72 @@ def make_tensor_proto(values, dtype=None, shape=None):
...
+        tensor_values = tensor_values.reshape(inferred_shape)
+
+    dims = []
+    for d in inferred_shape:
+        dims.append(TensorShapeProto.Dim(size=d))
+    tensor_shape = TensorShapeProto(dim = dims)
+    if dtype == DT_STRING:
+        tensor_proto = TensorProto(dtype=dtype, tensor_shape=tensor_shape, string_val=tensor_values.tolist())
+    elif np.isscalar(values):
+        # picking the right field for TensorProto() call
```
mzegla, line 155:
```suggestion
        # picking the right field of TensorProto to put data in
```

client/python/lib/test/test_tensors.py
```diff
@@ -75,32 +75,258 @@ def test_make_ndarray_valid_no_content_string_with_other_type_content():
...
         make_ndarray(tensor_proto)
+        exception = exception_info.value
+        assert str(exception) == "cannot reshape array of size 4 into shape (3,)"
+
+# def test_make_ndarray_invalid_no_type():
+#     tensor_proto = TensorProto(tensor_shape=shape, tensor_content = bytes([1,2,3]))
+#     with pytest.raises(ValueError) as exception_info:
+#         make_ndarray(tensor_proto)
+#         exception = exception_info.value
+#         assert str(exception) == "Tensor data type invalid"
```
mzegla, line 102: If we do not plan to enable this case, please remove it.

client/python/lib/test/test_tensors.py
```diff
@@ -75,32 +75,258 @@ def test_make_ndarray_valid_no_content_string_with_other_type_content():
...
+
+def test_make_tensor_proto_valid_int():
```
mzegla, line 104: Can we parametrize this test or run make_tensor_proto multiple times? I'd like to see calls like:
```
make_tensor_proto(values)
make_tensor_proto(values, dtype)
make_tensor_proto(values, shape)
make_tensor_proto(values, dtype, shape)
```
With the same `values` + `dtype` and/or `shape` set to make all this calls equivalent (produce exactly the same proto). Such thing could be applied to all cases that does not require non-default `shape` and `dtype` configuration

client/python/lib/test/test_tensors.py
```diff
@@ -75,32 +75,258 @@ def test_make_ndarray_valid_no_content_string_with_other_type_content():
...
+    assert tensor_proto.tensor_content == np_values.tobytes()
+    assert tensor_proto.tensor_shape == expected_shape
+    assert tensor_proto.dtype == DataType.DT_FLOAT
+
+# def test_make_tensor_proto_invalid_dtype_provided():
+#     values = [[1.0, 2.0, 3.0], [1.0, 2.0, 3.0]]
+#     with pytest.raises(TypeError) as exception_info:
+#         make_tensor_proto(shape = None, dtype = DataType.DT_STRING, values = values)
+#         exception = exception_info.value
+#         assert str(exception) == "[1.0, 2.0, 3.0] has type list, but expected one of: bytes"
```
mzegla, line 262: If this case will not be enabled, please remove it.

client/python/lib/test/test_tensors.py
```diff
@@ -75,32 +75,258 @@ def test_make_ndarray_valid_no_content_string_with_other_type_content():
...
+        assert str(exception) == "1 has type int, but expected one of: bytes"
+
+def test_make_tensor_proto_invalid_shape_type():
+    values = 5.0
+    with pytest.raises(TypeError) as exception_info:
+        make_tensor_proto(shape=0, dtype = DataType.DT_FLOAT, values = values)
+        exception = exception_info.value
+        assert str(exception) == "shape type should be list or tuple, but is int"
+
+def test_make_tensor_proto_invalid_dimensions_no_shape_provided():
```
mzegla, line 292: The situation here is very similar to `test_make_tensor_proto_invalid_dimsions`. Why the error is different?

client/python/lib/test/test_tensors.py
```diff
@@ -75,32 +75,258 @@ def test_make_ndarray_valid_no_content_string_with_other_type_content():
...
+        assert str(exception) == "The requested array has inhomogeneous shape. Detected shape was (2,) + inhomogeneous part"
+
+def test_make_tensor_proto_invalid_values_dtype():
+    values = [np.float128(2.5)]
+    with pytest.raises(TypeError) as exception_info:
+        make_tensor_proto(shape = None, dtype = None, values = values)
+        exception = exception_info.value
+        assert str(exception) == "values provided are not valid"
+
+def test_make_tensor_proto_invalid_dtype_as_float():
```
mzegla, line 306:
```suggestion
def test_make_tensor_proto_invalid_dtype_value():
```

mzegla (commented):
I'd suggest more tests for casting between numerical types: 
ints -> floats , floats -> ints
Tests with numpy  arrays on input + downcasting and upcasting:
np.int32 -> np.uint32, np.int64 -> np.int32
Do we handle such conversions?

Also binary input case with not complete data would be nice:
values = b'\x00\x10', dtype=np.float32
values = b'\x00\x00\x10', dtype=np.float16

client/python/lib/test/test_tensors.py
```diff
@@ -237,96 +353,135 @@ def test_make_tensor_proto_valid_make_ndarray_valid():
...
+    exception = exception_info.value
+    assert str(exception) == "could not cast values to <class 'numpy.float32'>. setting an array element with a sequence."
+    
 def test_make_tensor_proto_invalid_dimensions_no_shape_provided():
     values = [[1,2,3], [4,5]]
     with pytest.raises(ValueError) as exception_info:
-        make_tensor_proto(shape=None, dtype = DataType.DT_INT8, values = values)
-        exception = exception_info.value
-        assert str(exception) == "The requested array has inhomogeneous shape. Detected shape was (2,) + inhomogeneous part"
+        make_tensor_proto(values = values, shape=None, dtype = DataType.DT_INT8)
```
mzegla, line 408: The issue here is with asarray()? not np.array() call on array creation? 
The error message does not make it clear that there's a problem with dimensions.

client/python/lib/test/test_tensors.py
```diff
@@ -237,96 +353,135 @@ def test_make_tensor_proto_valid_make_ndarray_valid():
...
+    values = bytes([0x13, 0x00, 0x00, 0x00, 0x08, 0x00])
+    with pytest.raises(ValueError) as exception_info:
+        make_tensor_proto(values = [[values, values, values],[values, values]], shape= None, dtype = DataType.DT_STRING)
+    exception = exception_info.value
+    assert str(exception) == "could not cast values to <class 'numpy.bytes_'>. setting an array element with a sequence"
 
+def test_make_tensor_proto_invalid_dimensions_2():
+    values = [[(1,2,3)],[(1,2)],[(1,2,3)]]
+    with pytest.raises(ValueError) as exception_info:
+        make_tensor_proto(values = values, shape=[2,3], dtype = DataType.DT_FLOAT)
```
mzegla, line 401: The issue here is with asarray()? not np.array() call on array creation?
The error message does not make it clear that there's a problem with dimensions.

client/python/lib/test/test_tensors.py
```diff
@@ -237,96 +353,135 @@ def test_make_tensor_proto_valid_make_ndarray_valid():
...
-def test_make_tensor_proto_invalid_shape_type():
-    values = 5.0
-    with pytest.raises(TypeError) as exception_info:
-        make_tensor_proto(shape=0, dtype = DataType.DT_FLOAT, values = values)
-        exception = exception_info.value
-        assert str(exception) == "shape type should be list or tuple, but is int"
+def test_make_tensor_proto_invalid_string_dimensions():
+    values = bytes([0x13, 0x00, 0x00, 0x00, 0x08, 0x00])
+    with pytest.raises(ValueError) as exception_info:
+        make_tensor_proto(values = [[values, values, values],[values, values]], shape= None, dtype = DataType.DT_STRING)
```
mzegla, line 394: The issue here is with asarray()? not np.array() call on array creation?
The error message does not make it clear that there's a problem with dimensions.

client/python/lib/test/test_tensors.py
```diff
@@ -237,96 +353,135 @@ def test_make_tensor_proto_valid_make_ndarray_valid():
...
+    assert str(exception) == "shape elements should be unsigned integers"
 
-def test_make_tensor_proto_invalid_dimsions():
+def test_make_tensor_proto_invalid_dimensions():
     values = [[1.0, 2.0], [1.0, 2.0, 3.0]]
     with pytest.raises(ValueError) as exception_info:
-        make_tensor_proto(shape=[2,3], dtype = DataType.DT_FLOAT, values = values)
-        exception = exception_info.value
-        assert str(exception) == "setting an array element with a sequence"
+        make_tensor_proto(values = values, shape=[2,3], dtype = DataType.DT_FLOAT)
```
mzegla, line 380: The issue here is with asarray()? not np.array() call on array creation?
The error message does not make it clear that there's a problem with dimensions.

client/python/lib/test/test_tensors.py
```diff
@@ -77,154 +77,270 @@ def test_make_ndarray_invalid_type():
...
+
+@pytest.mark.parametrize("params", [
+    {"values" : bytes([0x13, 0x00, 0x00, 0x00, 0x08])},
+    {"values" : bytes([0x13, 0x00, 0x00, 0x00, 0x08]), "shape" : [1]},
+    {"values" : bytes([0x13, 0x00, 0x00, 0x00, 0x08]), "dtype" : DataType.DT_STRING},
+    {"values" : bytes([0x13, 0x00, 0x00, 0x00, 0x08]), "dtype" : DataType.DT_INT8},
+    {"values" : bytes([0x13, 0x00, 0x00, 0x00, 0x08]), "shape" : [5], "dtype" : DataType.DT_INT8},
+    {"values" : bytes([0x13, 0x00, 0x00, 0x00, 0x08]), "shape" : [1], "dtype" : DataType.DT_STRING},
+])
+def test_make_tensor_proto_valid_string(params):
```
mzegla, line 248: We get a lot of if..else... blocks here. Is it possible to remove them? For example by creating a mapping (condition -> expected_value) instead of adding additional comparisons.

client/python/lib/test/test_tensors.py
```diff
@@ -77,154 +77,270 @@ def test_make_ndarray_invalid_type():
...
-    assert tensor_proto.tensor_shape == TensorShapeProto(dim = [TensorShapeProto.Dim(size=2), TensorShapeProto.Dim(size=3)])
+@pytest.mark.parametrize("params", [
+    {"values" : [bytes([0x13, 0x00, 0x00, 0x00, 0x08]) ,bytes([0x13, 0x00, 0x00, 0x00, 0x08])]},
+    {"values" : [bytes([0x13, 0x00, 0x00, 0x00, 0x08]) ,bytes([0x13, 0x00, 0x00, 0x00, 0x08])], "shape" : [2]},
+    {"values" : [bytes([0x13, 0x00, 0x00, 0x00, 0x08]) ,bytes([0x13, 0x00, 0x00, 0x00, 0x08])], "dtype" : DataType.DT_STRING},
+    {"values" : [bytes([0x13, 0x00, 0x00, 0x00, 0x08]) ,bytes([0x13, 0x00, 0x00, 0x00, 0x08])], "dtype" : DataType.DT_INT8},
+    {"values" : [bytes([0x13, 0x00, 0x00, 0x00, 0x08]) ,bytes([0x13, 0x00, 0x00, 0x00, 0x08])], "shape" : [10], "dtype" : DataType.DT_INT8},
+    {"values" : [bytes([0x13, 0x00, 0x00, 0x00, 0x08]) ,bytes([0x13, 0x00, 0x00, 0x00, 0x08])], "shape" : [2], "dtype" : DataType.DT_STRING},
+])
+def test_make_tensor_proto_valid_string_batch_size_2(params):
```
mzegla, line 286: We get a lot of if..else... blocks here. Is it possible to remove them? For example by creating a mapping (condition -> expected_value) instead of adding additional comparisons.

mkuczyns11 (approved):
LGTM

mzegla (approved):

PR #829, opened 2021-08-09
Client library - PredictResponse

mkuczyns11:


client/python/lib/test/test_grpc_responses.py
```diff
@@ -23,13 +25,36 @@
...
+
+    predict_response = GrpcPredictResponse(predict_raw_response)
+    response_dict = predict_response.to_dict()
+
+    assert isinstance(response_dict, dict)
+    assert len(response_dict) == len(outputs_dict)
+    raw_response = predict_response.raw_response
+    for output_name, array in response_dict.items():
+        assert output_name in raw_response.outputs.keys()
+        assert type(array) == ndarray
```
mzegla, line 55:
```suggestion
        assert type(array) is ndarray
```

mzegla (commented):
Is it possible to have some negative tests too?

mzegla (approved):

dtrawins (approved):

PR #835, opened 2021-08-10
Client library - initial make targets

mzegla:


client/python/lib/Makefile
```diff
@@ -0,0 +1,55 @@
...
+build-deps: venv
+	@echo "Building dependencies"
+	@sudo apt install -y protobuf-compiler
+	@. $(ACTIVATE); pip install setuptools
+	@. $(ACTIVATE); python setup.py build_apis
+
+install-package: venv
+	@echo "Building and installing ovmsclient package"
+	@. $(ACTIVATE); python setup.py bdist_wheel
+	@. $(ACTIVATE); pip install dist/ovmsclient-0.1-py3-none-any.whl
```
mkuczyns11, line 44: Does the library reload correctly with this commands? Can you check whether tests will work even with some minor change?

mzegla: It will not reinstall. I'll add another target for reinstallation of the same version of the package.

mkuczyns11 (approved):

client/python/lib/Makefile
```diff
@@ -0,0 +1,60 @@
...
+	@echo "Updating virtualenv dependencies in: $(VIRTUALENV_DIR)..."
+	@test -d $(VIRTUALENV_DIR) || $(VIRTUALENV_EXE) $(VIRTUALENV_DIR)
+	@. $(ACTIVATE); pip install --upgrade pip
+	@. $(ACTIVATE); pip install -r tests/requirements.txt
+	@touch $(ACTIVATE)
+
+
+build-deps: venv
+	@echo "Building dependencies"
+	@sudo apt install -y protobuf-compiler
```
dtrawins, line 37: I don't think it would be good approach to add root commands to the makefile scope

client/python/lib/README.md
```diff
@@ -3,37 +3,32 @@
...
-## Building pip package
+## Installation
 
-The library is distributed in form of pip package. To build and install the package first clone model server repository:
+The client library requires Python in version >= 3.6. Validated and recommended way of installing the library is with the use of virtual python environment.
 
-`git clone https://github.com/openvinotoolkit/model_server.git`
+### Linux
+For one-step building and installation simply run:
+`make install`
```
dtrawins, line 12: perhaps it would be easier just to generate wheel package and from that point anyone could install it in any environment?
Do we need to require make command to install the client?

client/python/lib/Makefile
mzegla: `protobuf-compiler` needs to be installed in order to build TFS API.
Do you think we should specify it as a prerequisite instead of using it with make call?

client/python/lib/README.md
mzegla: This make has been created to make one-step installation possible. I didn't want to split the installation to something like:
```
make build-package
pip install dist/ovmsclient-0.1-py3-none-any.whl
```
Do you think it's better to remove installation step from it?

dtrawins (approved):

PR #847, opened 2021-08-17
CVS-58948 Enable binary inputs for pipelines with only custom nodes as entry

dkalinowski:
* Enable binary inputs for piplines with only custom nodes as entry. This fixes the issue with rejecting requests when pipeline input is reported as ANY. Such pipelines contain only custom nodes at entry therefore cannot deduce input layout.

The fix postpones blob validation to node validation phase, and allows creating blobs out of binary inputs even if pipeline input layout is set to ANY.

* Fix issue with demultiplexing for pipelines with only custom nodes at entry. Reported effective shape was `(0)` and no requests could be processed.

https://jira.devtools.intel.com/browse/CVS-58948


src/tensorinfo.cpp
```diff
@@ -309,6 +309,7 @@ std::shared_ptr<TensorInfo> TensorInfo::createCopyWithNewShape(const shape_t& sh
 std::shared_ptr<TensorInfo> TensorInfo::createCopyWithEffectiveDimensionPrefix(size_t dim) const {
     auto copy = std::make_shared<TensorInfo>(*this);
     copy->influencedByDemultiplexer = true;
+    copy->effectiveShape = this->getEffectiveShape();
```
atobiszei, line 312: Why this is required? Shouldn't that be handled in copy constructor in L310?

src/pipelinedefinition.cpp
```diff
@@ -29,6 +29,7 @@
 #include "pipeline.hpp"
 #include "pipelinedefinitionunloadguard.hpp"
 #include "prediction_service_utils.hpp"
+#include "tensorinfo.hpp"  // to remove
```
atobiszei, line 32: To remove?

rasapala, line 32: when to remove ?

src/test/custom_nodes/node_dynamic_image.c
```diff
@@ -0,0 +1,91 @@
...
+    (*info)->dims[0] = 0;
+    (*info)->dims[1] = 0;
+    (*info)->dims[2] = 0;
+    (*info)->dims[3] = 0;
+    (*info)->precision = FP32;
+    return 0;
+}
+
+int release(void* ptr) {
+    fflush(stdout);
```
atobiszei, line 88: No need?

src/test/ensemble_tests.cpp
```diff
@@ -4083,3 +4095,155 @@ TEST_F(EnsembleFlowTestBinaryInput, EntryDemultiplexer) {
...
+// This test ensure binary inputs work for pipelines with layout ANY.
+// Such pipelines have only custom nodes as entry nodes.
+// In this case we do not reject the request but create NHWC content out of that.
+TEST_F(EnsembleFlowTestBinaryInput, BinaryInputWithPipelineInputLayoutANY) {
+    std::string fileToReload = directoryPath + "/config.json";
+    createConfigFileWithContent(pipelineWithOnlyDynamicCustomNode, fileToReload);
+    ConstructorEnabledModelManager manager;
+    std::unique_ptr<Pipeline> pipeline;
+
+    // BS=1
```
atobiszei, line 4145: Not needed comment

src/test/ensemble_tests.cpp
```diff
@@ -4083,3 +4095,155 @@ TEST_F(EnsembleFlowTestBinaryInput, EntryDemultiplexer) {
...
+    int batchSize = 1;
+    prepareBinaryRequest(imagePath, request, "pipeline_input", batchSize);
+
+    ASSERT_EQ(manager.loadConfig(fileToReload), StatusCode::OK);
+    ASSERT_EQ(manager.getPipelineFactory().create(pipeline, "my_pipeline", &request, &response, manager), StatusCode::OK);
+
+    ASSERT_EQ(pipeline->execute(), StatusCode::OK);
+    checkIncrement4DimResponse("pipeline_output", {44.0, 35.0, 245.0}, request, response, {1, 1, 1, 1, 3});
```
atobiszei, line 4228: Minor:
file is defined as jpg and its values are stored there. Then in test you rely on knowledge about values saved there rather than reading original image (and its values) and explicitly adding the same as in custom node.

src/tensorinfo.cpp
dkalinowski: No, because `getEffectiveShape` does not return always the same class member. Depending if `this.effectiveShape` is empty or not, it chooses to return `this.shape` instead.

In a scenario with custom node as entry, variable `this.effectiveShape` is empty, because effective shape is the same as `this.shape` (layout change does not work on custom libraries).

What will happen when you copy `this.effectiveShape` (which is empty) and add demultiplexer dimension? We will end up with one element - `(0)` (equal to demultiply count).

To summarize, this is required to first get effective shape from source object (which will return `this.shape`) and then add demultiplexer dimension afterwards. This way we can be sure to end up with correct effective shape.

src/pipelinedefinition.cpp
dkalinowski: now, good catch :)

src/test/custom_nodes/node_dynamic_image.c
```diff
@@ -0,0 +1,91 @@
...
+    output->dims = (uint64_t*)malloc(input->dimsCount * sizeof(uint64_t));
+    output->dimsCount = input->dimsCount;
+    memcpy((void*)output->dims, (void*)input->dims, input->dimsCount * sizeof(uint64_t));
+    output->precision = input->precision;
+
+    for (uint64_t i = 0; i < output->dataBytes; i += sizeof(float)) {
+        *(float*)(output->data + i) = *(float*)(input->data + i) + addValue;
+    }
+
+    fflush(stdout);
```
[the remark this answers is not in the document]
dkalinowski: As I remember correctly, this was to make sure that custom node logs do not clash with OVMS logs. This should work similarly to std::end in cpp?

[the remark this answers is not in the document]
dkalinowski: I see, we have no logs anymore. I have copied node_add_sub. Yes, this is not needed.

src/binaryutils.cpp
```diff
@@ -146,9 +170,17 @@ Status validateNumberOfChannels(const std::shared_ptr<TensorInfo>& tensorInfo,
     return StatusCode::OK;
 }
 
+Status validateResolutionAgainstFirstBatchImage(const cv::Mat input, cv::Mat* firstBatchImage) {
+    if (input.cols == firstBatchImage->cols && input.rows == firstBatchImage->rows) {
+        return StatusCode::OK;
+    }
+    SPDLOG_ERROR("Binary images need to have resolutions matched");
```
atobiszei, line 177:
```suggestion
    SPDLOG_DEBUG("Binary images need to have resolutions matched");
```

src/binaryutils.cpp
```diff
@@ -158,13 +190,21 @@ bool checkBatchSizeMismatch(const std::shared_ptr<TensorInfo>& tensorInfo,
     return false;
 }
 
-Status validateInput(const std::shared_ptr<TensorInfo>& tensorInfo, const cv::Mat input) {
-    return validateNumberOfChannels(tensorInfo, input);
+Status validateInput(const std::shared_ptr<TensorInfo>& tensorInfo, const cv::Mat input, cv::Mat* firstBatchImage) {
+    // Layout ANY is for pipelines with only custom node as entry.
+    // In this case we need to be sure multiple batch images are aligned to the same resolution.
```
atobiszei, line 195: Is it required for batching of all shards into one response?
Could that requirement be lifted assuming that based upon DAG configuration (ie parameters of custom node) such DAG should still adhere to such requirement?

Just asking if this is possible improvement for the future.

src/binaryutils.cpp
```diff
@@ -146,9 +170,17 @@ Status validateNumberOfChannels(const std::shared_ptr<TensorInfo>& tensorInfo,
     return StatusCode::OK;
 }
 
+Status validateResolutionAgainstFirstBatchImage(const cv::Mat input, cv::Mat* firstBatchImage) {
+    if (input.cols == firstBatchImage->cols && input.rows == firstBatchImage->rows) {
+        return StatusCode::OK;
+    }
```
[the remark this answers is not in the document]
dkalinowski: `firstBatchImage` is nullptr when we process first image

src/tensorinfo.cpp
atobiszei: You're right. It was bug from the start then.

atobiszei: Actually I see that all constructors initialize effectiveShape so we might want to redefine getEffectiveShape. Unless there is some special case

src/binaryutils.cpp
dkalinowski: This is outdated comment, will change

dtrawins (approved):

rasapala (approved):
LGTM

src/binaryutils.cpp
dkalinowski: changed

atobiszei (approved):

PR #855, opened 2021-08-20
Ovmsclient sample scripts

mkuczyns11:


client/python/samples/resnet_grpc_predict.py
```diff
@@ -0,0 +1,119 @@
...
+import argparse
+import numpy as np
+import classes
+from ovmsclient import make_grpc_client, make_grpc_predict_request
+
+parser = argparse.ArgumentParser(description='Make prediction using images in numpy format')
+parser.add_argument('--images_numpy_path', required=False, default='./utils/imgs.npy',
+                    help='path to image')
+parser.add_argument('--labels_numpy_path', required=False,
+                    help='numpy in shape [n,1] - can be used to check model accuracy')
```
mzegla, line 26: Can we drop using `.npy` files and replace those parameters with `--images_dir` that would contain path to the directory with images in JPG format? 
Also let's drop passing labels. The script would not calculate the accuracy and just print the predicted label for each JPG file what do you think?

client/python/samples/resnet_grpc_predict.py
```diff
@@ -0,0 +1,119 @@
...
+parser.add_argument('--model_name', default='resnet', help='Model name to query. default: resnet',
+                    dest='model_name')
+parser.add_argument('--model_version', default=0, type=int,
+                    help='Model version to query. Lists all versions if omitted',
+                    dest='model_version')
+parser.add_argument('--input_name', required=False, default='0',
+                    help='Specify input tensor name. default: 0')
+parser.add_argument('--output_name', required=False, default='1463',
+                    help='Specify output name. default: 1463')
+parser.add_argument('--batchsize', default=1, type=int,
```
mzegla, line 40: Let's drop that parameter. Minimal amount of options would be clearer to the user.

client/python/samples/resnet_grpc_predict.py
```diff
@@ -0,0 +1,119 @@
...
+        if output.shape[1] == 1001:
+            offset = 1
+        ma = np.argmax(single_result) - offset
+        mark_message = ""
+        if labels_numpy_path is not None:
+            if int(lb[i]) == ma:
+                mark_message = "; Correct match."
+            else:
+                mark_message = "; Incorrect match. Should be {} {}".format(lb[i], classes.imagenet_classes[int(lb[i])])
+        print(classes.imagenet_classes[ma], ma, mark_message)
```
mzegla, line 88: This could be excluded to another module to make this script more about using client library and less about pre/post processing of the data.

client/python/samples/resnet_grpc_predict.py
```diff
@@ -0,0 +1,119 @@
...
+        offset = 0
+        if output.shape[1] == 1001:
+            offset = 1
+        ma = np.argmax(single_result) - offset
+        mark_message = ""
+        if labels_numpy_path is not None:
+            if int(lb[i]) == ma:
+                mark_message = "; Correct match."
+            else:
+                mark_message = "; Incorrect match. Should be {} {}".format(lb[i], classes.imagenet_classes[int(lb[i])])
```
mzegla, line 87: Let's drop accuracy check and just print
`<image_path>: <predicted_label>`

client/python/samples/resnet_grpc_predict.py
```diff
@@ -0,0 +1,119 @@
...
+    }
+    request = make_grpc_predict_request(inputs, model_name, model_version)
+    response = client.predict(request)
+    response_dict = response.to_dict()
+
+    # output post-processing
+    if output_name not in response_dict.keys():
+        print(f"Invalid output name - {output_name}")
+    output = response_dict[output_name]
+    post_processing(output)
```
mzegla, line 119: I think for the main flow I would go like this:
```python
paths = get_all_images_paths
for path in paths:
    img = load_image(path) # load single image and preprocess it to get final numpy ndarray
    request = make_predict_request(img)
    response = client.predict(request)
    label, confidence_score = postprocess(response)
    print("Image {path} has been classified as {label} with {confidence_score}% confidence")
```

client/python/samples/resnet_grpc_predict_binary.py
```diff
@@ -0,0 +1,90 @@
...
+# limitations under the License.
+#
+
+import argparse
+import numpy as np
+import classes
+from ovmsclient import make_grpc_client, make_grpc_predict_request
+
+parser = argparse.ArgumentParser(description='Make prediction using images in binary format')
+parser.add_argument('--images_list', required=False, default='./utils/resnet_images.txt',
```
mzegla, line 23: Lets go with `--images_dir` and drop accuracy check like with non-binary version

client/python/samples/resnet_grpc_predict.py
```diff
@@ -0,0 +1,119 @@
...
+                    help='Specify port to grpc service. default: 9000')
+parser.add_argument('--model_name', default='resnet', help='Model name to query. default: resnet',
+                    dest='model_name')
+parser.add_argument('--model_version', default=0, type=int,
+                    help='Model version to query. Lists all versions if omitted',
+                    dest='model_version')
+parser.add_argument('--input_name', required=False, default='0',
+                    help='Specify input tensor name. default: 0')
+parser.add_argument('--output_name', required=False, default='1463',
+                    help='Specify output name. default: 1463')
```
mzegla, line 39: Since we run inference on resnet, in this case we can assume there will be one input and one output in the model.
Can we further simplify this call by using metadata call to get names of the input and output? (different implementations might have different names so we can't hardcode it and the default value might not apply to many versions of Resnet loaded in the OVMS)

client/python/samples/resnet_grpc_predict_binary.py
```diff
@@ -0,0 +1,90 @@
...
+    offset = 0
+    if output.shape[1] == 1001:
+        offset = 1
+    ma = np.argmax(output[0]) - offset
+    mark_message = ""
+    if int(label) == ma:
+        mark_message = "; Correct match."
+    else:
+        mark_message = "; Incorrect match. Should be {} {}".format(label, classes.imagenet_classes[int(label)])
+    print(classes.imagenet_classes[ma], ma, mark_message)
```
mzegla, line 90: Same as for numpy version. Lets drop checking match correctness.

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,91 @@
...
+parser = argparse.ArgumentParser(description='Make prediction using images in numpy format')
+parser.add_argument('--grpc_address', required=False, default='localhost',
+                    help='Specify url to grpc service. default:localhost')
+parser.add_argument('--grpc_port', required=False, default=9000,
+                    help='Specify port to grpc service. default: 9000')
+parser.add_argument('--model_name', default='vehicle-detection', help='Pipeline name to query. default: detection',
+                    dest='pipeline_name')
+parser.add_argument('--input_name', required=False, default='image',
+                    help='Specify input tensor name. default: image')
+parser.add_argument('--images_input_path', required=True,
```
mzegla, line 31: Can we go with the same name as in resnet samples?

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,91 @@
...
+from ovmsclient import make_grpc_client, make_grpc_predict_request
+
+parser = argparse.ArgumentParser(description='Make prediction using images in numpy format')
+parser.add_argument('--grpc_address', required=False, default='localhost',
+                    help='Specify url to grpc service. default:localhost')
+parser.add_argument('--grpc_port', required=False, default=9000,
+                    help='Specify port to grpc service. default: 9000')
+parser.add_argument('--model_name', default='vehicle-detection', help='Pipeline name to query. default: detection',
+                    dest='pipeline_name')
+parser.add_argument('--input_name', required=False, default='image',
```
mzegla, line 29: We can get that with metadata call and simplify the use of this script to the users

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,91 @@
...
+                    help='Specify input tensor name. default: image')
+parser.add_argument('--images_input_path', required=True,
+                    help='Input images path.')
+parser.add_argument('--output_name', required=False, default='detection_out',
```
mzegla, line 33: We can get that with metadata call and simplify the use of this script to the users

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,91 @@
...
+    width = image.shape[1]
+    height = image.shape[0]
+    for i in range(output.shape[2]):
+        if output[0, 0, i, 2] > 0.5:   # if confidence level is greater than 50%
+            x_min = int(output[0, 0, i, 3] * width)
+            y_min = int(output[0, 0, i, 4] * height)
+            x_max = int(output[0, 0, i, 5] * width)
+            y_max = int(output[0, 0, i, 6] * height)
+            image = cv2.rectangle(cv2.UMat(image), (x_min, y_min), (x_max, y_max), (0, 0, 255), 1)
+    cv2.imwrite(os.path.join(output_save_path, 'vehicle-detection' + '.jpg'), image)
```
mzegla, line 59: If you choose to extract resnet pre/post processing we can do the same here.

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,91 @@
...
+output_name = args.get('output_name')
+output_save_path = args.get('output_save_path')
+
+
+def save_image_with_detected_vehicles(image_path, output):
+    image = cv2.imread(image_path)
+    width = image.shape[1]
+    height = image.shape[0]
+    for i in range(output.shape[2]):
+        if output[0, 0, i, 2] > 0.5:   # if confidence level is greater than 50%
```
mzegla, line 53: Can we parametrize this value and pass it to this function?

client/python/samples/resnet_grpc_predict.py
```diff
@@ -0,0 +1,71 @@
...
+    "address": address,
+    "port": port
+}
+client = make_grpc_client(config)
+
+# receiving metadata from model
+input_name, output_name = get_single_metadata(client, model_name, model_version)
+
+# preparing images
+imgs = read_imgs_as_ndarray(images_dir, 224, 224)
```
mzegla, line 54: I think it might be worth passing the whole input shape here. Some ResNets might have different layout and the function to read the images should be aware of that.

client/python/samples/resnet_grpc_predict.py
```diff
@@ -0,0 +1,71 @@
...
+
+for i in range(imgs.shape[0]):
+    # reading image and its label
+    img = imgs[i:i+1]
```
mzegla, line 58: Can we just use `for img in imgs`?

client/python/samples/resnet_grpc_predict_binary.py
```diff
@@ -0,0 +1,72 @@
...
+                    help='Path to images')
+parser.add_argument('--grpc_address', required=False, default='localhost',
+                    help='Specify url to grpc service. default:localhost')
+parser.add_argument('--grpc_port', required=False, default=9000,
+                    help='Specify port to grpc service. default: 9000')
+parser.add_argument('--model_name', default='resnet', help='Model name to query. default: resnet',
+                    dest='model_name')
+parser.add_argument('--model_version', default=0, type=int, help='Model version to query. Lists all versions if omitted',
+                    dest='model_version')
+args = vars(parser.parse_args())
```
mzegla, line 34: Same suggestions as for non binary example apply here as well

client/python/samples/resnet_grpc_predict_binary.py
```diff
@@ -0,0 +1,72 @@
...
+
+# creating grpc client
+config = {
+    "address": address,
+    "port": port
+}
+client = make_grpc_client(config)
+
+# receiving metadata from model
+input_name, output_name = get_single_metadata(client, model_name, model_version)
```
mzegla, line 51: `get_model_io_names` maybe?

client/python/samples/resnet_grpc_predict_binary.py
```diff
@@ -0,0 +1,72 @@
...
+
+# preparing images
+paths = read_paths(images_dir)
```
mzegla, line 54:
```suggestion
img_paths = read_image_paths(images_dir)
```

client/python/samples/utils/resnet_utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+def resnet_postprocess(response, output_name):
+    response_dict = response.to_dict()
+    output = response_dict[output_name]
+    offset = 0
+    if output.shape[1] == 1001:
+        offset = 1
+    label = np.argmax(output[0]) - offset
+    confidence_score = output[0][label]
+    return label, confidence_score
```
mzegla, line 28: Please return label name instead of raw index

client/python/samples/resnet_grpc_predict_binary.py
```diff
@@ -0,0 +1,72 @@
...
+        input_name: img
+    }
+    request = make_grpc_predict_request(inputs, model_name, model_version)
+
+    # sending predict request and receiving response
+    response = client.predict(request)
+
+    # response post processing
+    label, confidence_score = resnet_postprocess(response, output_name)
+    print(f"Image {path} has been classified as {label} with {confidence_score}% confidence")
```
mzegla, line 72: I believe confidence level is in value 0.0-1.0, so we should multiply it to get percentage.

client/python/samples/utils/utils.py
```diff
@@ -0,0 +1,84 @@
...
+    metadata = client.get_model_metadata(metadata_request)
+    metadata_dict = metadata.to_dict()
+    version = next(iter(metadata_dict))
+    input_name = next(iter(metadata_dict[version]['inputs']))  # by default resnet has one input and one output
+    output_name = next(iter(metadata_dict[version]['outputs']))
+    return input_name, output_name
+
+
+# read .txt file with listed images into ndarray in model shape
+def read_imgs_as_ndarray(images_dir, height, width):
```
mzegla, line 44: Can we just read all jpg and png files in the `images_dir` path instead of relying on .txt file?

client/python/samples/utils/utils.py
```diff
@@ -0,0 +1,84 @@
...
+    # retrived array has BGR format and 0-255 normalization
+    # format of data is HWC
+    # add image preprocessing if needed by the model
+    img = crop_resize(img, height, width)
+    img = img.astype('float32')
+    #convert to RGB instead of BGR if required by model
+    if rgb_image:
+        img = img[:, :, [2, 1, 0]]
+    # switch from HWC to CHW and reshape to 1,3,size,size for model blob input requirements
+    img = img.transpose(2, 0, 1).reshape(1, 3, height, width)
```
mzegla, line 83: This action could depend on input shape received by the metadata call

client/python/samples/utils/vehicle_utils.py
```diff
@@ -0,0 +1,39 @@
...
+    image = cv2.imread(image_path)
+    width = image.shape[1]
+    height = image.shape[0]
+    for i in range(output.shape[2]):
+        image = draw_vehicle(output[0, 0, i, ...], image, width, height)
+    cv2.imwrite(os.path.join(output_save_path, 'vehicle-detection' + '.jpg'), image)
+
+
+def draw_vehicle(vehicle_output, image, width, height):
+    if vehicle_output[2] > 0.5:   # if confidence level is greater than 50%
```
mzegla, line 33: confidence threshold could be parametrized

client/python/samples/utils/vehicle_utils.py
```diff
@@ -0,0 +1,39 @@
...
+    output = response_dict[output_name]
+    image = cv2.imread(image_path)
+    width = image.shape[1]
+    height = image.shape[0]
+    for i in range(output.shape[2]):
+        image = draw_vehicle(output[0, 0, i, ...], image, width, height)
+    cv2.imwrite(os.path.join(output_save_path, 'vehicle-detection' + '.jpg'), image)
+
+
+def draw_vehicle(vehicle_output, image, width, height):
```
mzegla, line 32: We could change the name of this function to `draw_bounding_box`. 
`vehicle_output` could be changed to `prediction_result`
If `width` and `height` mean input image width and height, we could hardcode it here, since functions from this file are dedicated to that specific model.

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,73 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import argparse
+from ovmsclient import make_grpc_client, make_grpc_predict_request
+from utils.utils import load_image, read_paths, get_single_metadata
+from utils.vehicle_utils import vehicle_postprocess
+
+parser = argparse.ArgumentParser(description='Make prediction using images in numpy format')
```
mzegla, line 22:
```suggestion
parser = argparse.ArgumentParser(description='Make prediction using images in binary format')
```

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,73 @@
...
+                    help='Input images path.')
+parser.add_argument('--grpc_address', required=False, default='localhost',
+                    help='Specify url to grpc service. default:localhost')
+parser.add_argument('--grpc_port', required=False, default=9000,
+                    help='Specify port to grpc service. default: 9000')
+parser.add_argument('--model_name', default='vehicle-detection',
+                    help='Pipeline name to query. default: detection')
+parser.add_argument('--model_version', default=0, type=int,
+                    help='Model version to query. Lists all versions if omitted')
+parser.add_argument('--output_save_path', required=False, default='./result',
```
mzegla, line 33: Can we make it required?

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,73 @@
...
+parser.add_argument('--images_dir', required=True,
+                    help='Input images path.')
+parser.add_argument('--grpc_address', required=False, default='localhost',
+                    help='Specify url to grpc service. default:localhost')
+parser.add_argument('--grpc_port', required=False, default=9000,
+                    help='Specify port to grpc service. default: 9000')
+parser.add_argument('--model_name', default='vehicle-detection',
+                    help='Pipeline name to query. default: detection')
+parser.add_argument('--model_version', default=0, type=int,
+                    help='Model version to query. Lists all versions if omitted')
```
mzegla, line 32:
```suggestion
                    help='Model version to query. default: latest available')
```

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,73 @@
...
+
+parser = argparse.ArgumentParser(description='Make prediction using images in numpy format')
+parser.add_argument('--images_dir', required=True,
+                    help='Input images path.')
+parser.add_argument('--grpc_address', required=False, default='localhost',
+                    help='Specify url to grpc service. default:localhost')
+parser.add_argument('--grpc_port', required=False, default=9000,
+                    help='Specify port to grpc service. default: 9000')
+parser.add_argument('--model_name', default='vehicle-detection',
+                    help='Pipeline name to query. default: detection')
```
mzegla, line 30:
```suggestion
                    help='Model name to query. default: vehicle-detection')
```

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,73 @@
...
+from utils.vehicle_utils import vehicle_postprocess
+
+parser = argparse.ArgumentParser(description='Make prediction using images in numpy format')
+parser.add_argument('--images_dir', required=True,
+                    help='Input images path.')
+parser.add_argument('--grpc_address', required=False, default='localhost',
+                    help='Specify url to grpc service. default:localhost')
+parser.add_argument('--grpc_port', required=False, default=9000,
+                    help='Specify port to grpc service. default: 9000')
+parser.add_argument('--model_name', default='vehicle-detection',
```
mzegla, line 29: What exact model do we target with this script?

client/python/samples/vehicle_detection_predict_binary.py
```diff
@@ -0,0 +1,73 @@
...
+# configuration
+images_dir = args.get('images_dir')
+address = args.get('grpc_address')
+port = args.get('grpc_port')
+model_name = args.get('model_name')
+model_version = args.get('model_version')
+output_save_path = args.get('output_save_path')
+
+# preparing images
+paths = read_paths(images_dir)
```
mzegla, line 46:
```suggestion
img_paths = read_image_paths(images_dir)
```

client/python/samples/vehicle_detection_predict_binary.py
mkuczyns11: vehicle-detection model from here https://docs.openvinotoolkit.org/2019_R1/_vehicle_detection_adas_0002_description_vehicle_detection_adas_0002.html

client/python/samples/resnet_grpc_predict.py
```diff
@@ -68,4 +68,4 @@
 
     # response post processing
     label, confidence_score = resnet_postprocess(response, output_name)
-    print(f"Image {i} has been classified as {label} with {confidence_score}% confidence")
+    print(f"Image {i+1} has been classified as {imagenet_classes[label]} with {confidence_score}% confidence")
```
mzegla, line 71: why not return `imagenet_classes[label]` from `resnet_postprocess` instead of raw label?

client/python/samples/utils/vehicle_utils.py
```diff
@@ -25,15 +25,15 @@ def vehicle_postprocess(response, image_path, output_name, output_save_path):
     width = image.shape[1]
     height = image.shape[0]
     for i in range(output.shape[2]):
-        image = draw_vehicle(output[0, 0, i, ...], image, width, height)
+        image = draw_bounding_box(output[0, 0, i, ...], image, width, height, 0.5)
```
mzegla, line 28: i would still assign this `0.5` to a variable to make it more visible what this value is about

mzegla (approved):

michalkulakowski (approved):

PR #857, opened 2021-08-20
Restore building from dldt

michalkulakowski:


Makefile
```diff
@@ -71,8 +71,7 @@ DIST_OS_TAG ?= $(BASE_OS_TAG)
 ifeq ($(BASE_OS),ubuntu)
   BASE_OS_TAG=$(BASE_OS_TAG_UBUNTU)
   BASE_IMAGE=ubuntu:$(BASE_OS_TAG_UBUNTU)
-  # Temporarily build from APT
-  DLDT_PACKAGE_URL=""
```
rasapala, line 75: We cant use s3 this is internal path. For local testing it is ok but cannot be published on github.
Try removing the empty setting and double check if it is building from ENV variable on Jenkins.
# Temporarily build from APT
  DLDT_PACKAGE_URL=""

rasapala (approved):
LGTM

dkalinowski (approved):

atobiszei (approved):

PR #867, opened 2021-08-23
Task add support for gpu device in docker

jasiu86:
GPU RUN:
https://ci.iotg.sclab.intel.com/job/ovmsc/job/OVMSCOps/5362/

### DIFF
--- a/tests/functional/config.py
+++ b/tests/functional/config.py
@@ -25,9 +25,6 @@ try:
 except ImportError:
     pass
 
-"""IMAGE - docker image name which should be used to run tests"""
-image = os.environ.get("IMAGE", "openvino/model_server")
-
 """TEST_DIR -  location where models and test data should be copied from TEST_DIR_CACHE and deleted after tests"""
 test_dir = os.environ.get("TEST_DIR", "/tmp/{}".format(generate_test_object_name(prefix='ovms_models')))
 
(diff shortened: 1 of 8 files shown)


tests/functional/config.py
```diff
@@ -66,6 +63,13 @@
 """ TT_TARGET_DEVICE - one of "CPU", "GPU" """
 target_device = os.environ.get("TT_TARGET_DEVICE", "CPU")
 
+"""IMAGE - docker image name which should be used to run tests"""
+if target_device == "CPU":
```
pgierasi, line 67: in case of HDDL & NCS we would like to use the default Image (non-GPU)

pgierasi (changes requested):
In case of HDDL & NCS we would like to use the default Image (non-GPU)

tests/functional/config.py
jasiu86: Done

jasiu86:
> In case of HDDL & NCS we would like to use the default Image (non-GPU)

Fixed

tests/functional/config.py
```diff
@@ -78,3 +82,15 @@
...
+""" TT_DEFAULT_INFER_TIMEOUT - Timeout for CPU target device"""
+default_infer_timeout = get_int("TT_DEFAULT_INFER_TIMEOUT", 10)
+
+""" TT_DEFAULT_GPU_INFER_TIMEOUT - Timeout for CPU target device"""
+default_gpu_infer_timeout = get_int("TT_DEFAULT_GPU_INFER_TIMEOUT", 10*default_infer_timeout)
+
+""" INFER TIMEOUT """
+if target_device == "GPU":
+    infer_timeout = default_gpu_infer_timeout
+else:
```
pgierasi, line 95: NCS and HDDL devices are slower than CPU

pgierasi (approved):

tests/functional/config.py
jasiu86: NCS and HDDL will be changed in following tasks and PR.

krzyczar (approved):

PR #868, opened 2021-08-23
Add REST.post call validation in example clients

jasiu86:


ncybulsk (approved):

pgierasi (changes requested):
Please do not change the code of example clients. Please make the change for the client that is being used by functional tests.

PR #871, opened 2021-08-23
Client library - extend main readme + add usage section

mzegla:


README.md
```diff
@@ -78,6 +78,14 @@ Learn more about [REST API](docs/model_server_rest_api.md)
 
 Refer to the [REST API example client code](./example_client/README.md#rest-api-client-examples) to learn how to use REST API 
 
+### OVMS Python Client Library
+
+For simplified interaction with the model server API, the [Python client library](client/python/lib) has been introduced. It's a set of Python functions and objects that wrap things like:
+- setting gRPC/HTTP connection
```
mkuczyns11, line 84: Should we mention HTTP connection here even if it hasn't been implemented yet?

client/python/lib/README.md
```diff
@@ -40,3 +44,82 @@ Apart from `make build`, there are also other targets available:
...
+#       },
+#       "outputs" {
+#           "output": {
+#               "shape": (1, 1000),
+#               "dtype": DT_FLOAT32  
+#           }
+#       }
+#   }
+#}
+#
```
mkuczyns11, line 105: I think shape type should be list. Can you double check on that?

client/python/lib/README.md
```diff
@@ -40,3 +44,82 @@ Apart from `make build`, there are also other targets available:
...
+predict_response = client.predict(predict_request)
+predict_response.to_dict()
+
+# Examplary predict_response.to_dict() output:
+#
+#{
+#   "output": [[0.01, 0.03, 0.91, ... , 0.00021]]
+#}
+#
+```
```
mkuczyns11, line 123: Maybe instead of input/output we could place input_name/output_name? What do you think?

mkuczyns11 (approved):

michalkulakowski (approved):

PR #885, opened 2021-08-26
ovmsclient - building adjustments to support python 3.6+

mkuczyns11:


client/python/lib/Makefile
```diff
@@ -33,11 +33,17 @@ $(ACTIVATE):
 	@. $(ACTIVATE); pip install -r tests/requirements.txt
 	@touch $(ACTIVATE)
 
-build-deps:
+update-packages:
+	@echo "Updating crucial packages"
+	@pip install --upgrade pip
+	@pip install "setuptools==57.4.0"
+	@pip install "wheel==0.37.0"
```
mzegla, line 40: Do we need **exactly** these versions for every Python? 
Or do we need at least this high versions?

mkuczyns11: Those are one of the newest versions of that modules and we do not need exactly that versions. I can investigate what are the lowest working ones.

mzegla: No need to search for the lowest. If there has been some changes in those dependencies starting with those versions then maybe we can go with `pip install package>=0.0.1`

mkuczyns11: I checked setuptools and wheel documentation. It seems like setuptools should be >= 40.1.0 and wheel should be >= 0.23.0. I tested them and everything seems to work properly.

mzegla (approved):

PR #906, opened 2021-09-05
drop redundant sample images

dtrawins:


mzegla (commented):
I think we need to unify that. I don't like making the paths longer and more complex, but I agree we shouldn't duplicate assets. 
Could you also remove `result/` directory?

rasapala (approved):
LGTM

dkalinowski:
> I think we need to unify that. I don't like making the paths longer and more complex, but I agree we shouldn't duplicate assets.
> Could you also remove `result/` directory?

The same goes for example C++ client. This also uses ./repo/example_clients/images.

PR #910, opened 2021-09-06
Client library - make_http_predict_request

mzegla:
Implemented make_http_predict_request and http serving base. Tests has been reorganized to avoid unnecessary duplicates for testing the same flow for GRPC and REST.


client/python/lib/ovmsclient/tfs_compat/http/serving_client.py
```diff
@@ -19,6 +19,12 @@
 
 class HttpClient(ServingClient):
 
+    def __init__(self, address, port, client_key=None, server_cert=None):
+        self.address = address
+        self.port = port
+        self.client_key = client_key
+        self.server_cert = server_cert
```
mkuczyns11, line 26: Why are we storing client_key and server_cert in HttpClient?

client/python/lib/ovmsclient/tfs_compat/http/serving_client.py
```diff
@@ -105,7 +111,17 @@ def get_model_status(self, request):
...
+        address = config.get("address")
+        port = config.get("port")
+        client_cert = None
+        server_cert = None
+        if "tls_config" in config:
+            tls_config = config["tls_config"]
+            if "client_cert_path" in tls_config and "client_key_path" in tls_config:
+                client_cert = (tls_config["client_cert_path"], tls_config["client_key_path"])
+            server_cert = tls_config.get('server_cert_path', None),
+        cls(address, port, client_cert, server_cert)
```
mkuczyns11, line 124: client_cert and server_cert here are simply paths provided in config. There is a function _prepare_certs in grpc implementation, which reads provided paths. Shouldn't is be used here? If we want to pass paths, maybe we should change variable names to *_path? Also in HttpClient __init__ we are expecting client_key whilehere client_cert is provided.

client/python/lib/ovmsclient/tfs_compat/http/serving_client.py
mzegla: We plan to use requests library for sending HTTP requests. As far as I know, we need to pass paths to the final method, so they are needed in HttpClient so it can call POST or GET methods with SSL.

client/python/lib/ovmsclient/tfs_compat/http/serving_client.py
mzegla: We plan to use requests library for sending HTTP requests. The client key and cert are passed to the requests methods as a tuple (client_key, client_cert), that's why they are merged here and passed as client_cert.

mkuczyns11 (approved):
LGTM

michalkulakowski (approved):

PR #932, opened 2021-09-17
Dockerfiles with ovmsclient environment

mzegla:


mkuczyns11 (approved):

client/python/Dockerfile.redhat
```diff
@@ -0,0 +1,60 @@
...
+RUN mkdir /ovmsclient
+COPY lib /ovmsclient/lib
+
+RUN dnf install -y python38 make && \
+    cd /ovmsclient/lib && make build
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ #
+FROM $FINAL_IMAGE as final_build
+
+ENV OVMSCLIENT_DIR="/home/ovmsclient"
```
dtrawins, line 41: do we need it as an env variable?
container can be started from different security context so it will not have access to home/ovmsclient folder

client/python/Dockerfile.redhat
```diff
@@ -0,0 +1,60 @@
...
+
+RUN microdnf install -y python38 && microdnf clean all
+
+USER ovmsclient
+
+ENV PATH=${PATH}:${OVMSCLIENT_DIR}/.local/bin
+RUN pip3 install --user ${OVMSCLIENT_DIR}/ovmsclient-0.1-py3-none-any.whl && \
+    rm  ${OVMSCLIENT_DIR}/ovmsclient-0.1-py3-none-any.whl
+
+ENV PS1="$(whoami)@$(cat /etc/hostname):$(pwd)\$ "
```
dtrawins, line 59: why do we need it?

client/python/Dockerfile.ubuntu
```diff
@@ -0,0 +1,58 @@
...
+# limitations under the License.
+#
+
+ARG BASE_IMAGE=ubuntu:20.04
+ARG FINAL_IMAGE=ubuntu:20.04
+
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ #
+FROM $BASE_IMAGE as base_build
+
+RUN groupadd --gid 5000 ovms && \
```
dtrawins, line 23: why is it need in the build image?

client/python/Dockerfile.ubuntu
```diff
@@ -0,0 +1,58 @@
...
+    useradd --home-dir /home/ovms --create-home --uid 5000 --gid 5000 --shell /bin/bash --skel /dev/null ovms
+
+RUN apt-get update && apt-get install -y unzip curl && \
+    curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v3.18.0/protoc-3.18.0-linux-x86_64.zip && \
+    unzip protoc-3.18.0-linux-x86_64.zip -d /opt/protoc
+
+ENV PATH=$PATH:/opt/protoc
+
+RUN mkdir /ovmsclient
```
dtrawins, line 32: this command is redundant

client/python/Dockerfile.redhat
mzegla: We can use different directory for storing and using the client. 
Would you prefer something like: `/ovmsclient` ?

client/python/Dockerfile.redhat
mzegla: It's for better user experience. So when you enter the container in bash you have nicer prompt:
`<user>@<container_id>:<current_path>$`, for example
`ovmslient@1d34fsa154cs:/ovmsclient/samples $` 
instead of just something like:
`bash 4.4 $`

client/python/Dockerfile.ubuntu
mzegla: We don't. I'll remove it in the next commit.

dtrawins (approved):

PR #939, opened 2021-09-22
Client library - HttpClient::get_model_metadata

mzegla:


client/python/lib/ovmsclient/tfs_compat/http/serving_client.py
```diff
@@ -85,7 +87,20 @@ def get_model_metadata(self, request):
...
 
-        raise NotImplementedError
+        HttpClient._check_model_metadata_request(request)
+
+        raw_response = None
+        try:
+            raw_response = self.session.get(f"http://{self.address}:{self.port}"
+                                            f"/v1/models/{request.model_name}"
+                                            f"/versions/{request.model_version}"
+                                            f"/metadata",
```
mkuczyns11, line 97: Have you run style tests? I think that flake should point out that f string is missing placeholder.

mzegla: I did. Styles passed, so it seems okay.

mkuczyns11 (approved):

michalkulakowski (approved):

PR #946, opened 2021-09-28
Ov license refresh

rasapala:


mzegla (approved):

dtrawins:
@rasapala  The license file was to be copied to OV package instead of adding it statically.

PR #948, opened 2021-09-28
[WIP] Avoid using latest tag due to caching on nodes

Xaenalt:
Ideally this tag could get populated on a given build date, but this avoids the problem of caching the :latest tag on the nodes


Xaenalt:
It'd be great if we could have some kind of automatic mechanism in the CR to get the date or SHA256 from the built image and use that as the tag name, etc

Xaenalt:
If the notebook repo was tagged, that would help with versioning the notebooks shown to the user

PR #992, opened 2021-10-28
mapping_config.json input/output mapping fix

mkuczyns11:


src/modelinstance.cpp
```diff
@@ -82,13 +82,13 @@ Status ModelInstance::loadInputTensors(const ModelConfig& config, const DynamicM
         if (name == ANONYMOUS_INPUT_NAME) {
             continue;
         }
-        if (networkInputs.count(name) == 0) {
+        if (networkInputs.count(name) == 0 && networkInputs.count(config.getRealInputByValue(name)) == 0) {
```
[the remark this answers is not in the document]
dkalinowski: What if we do config mapping a=>b & g=>a
And set shape to be: `a: [1,3,224,224]`, should the shape be applied to original `a` or the mapped `a`?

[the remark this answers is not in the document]
mkuczyns11: I think that currently it would be applied for both, because: 
a=>b: we would find config.getShapes().count(name)
g=>a: we would find config.getShapes().count(config.getMappingInputByKey(name))
If we support such mapping then we would need to check whether input/output is mapped and if thats the case we would look only for mapped_name in config.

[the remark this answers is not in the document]
dkalinowski: From user perspective, I think it would be better to apply only to mapped names. As an user I want to map the names because those original ones are long and annoying. Once user maps to understandable names, he is able to forget about original names. @dtrawins what do you think about it?

src/model.cpp
```diff
@@ -248,13 +248,16 @@ Status Model::reloadVersions(std::shared_ptr<model_versions_t> versionsToReload,
...
+            config.setLocalPath(modelVersion->getModelConfig().getLocalPath());
+        }
         auto status = config.parseModelMapping();
         if ((!status.ok()) && (status != StatusCode::FILE_INVALID)) {
             SPDLOG_ERROR("Error while parsing model mapping for model {}", status.string());
         }
 
-        auto modelVersion = getModelInstanceByVersion(version);
-
         if (!modelVersion) {
```
dkalinowski, line 261: Why not to move this if statement to 253 line instead?

src/model.cpp
```diff
@@ -265,8 +268,6 @@ Status Model::reloadVersions(std::shared_ptr<model_versions_t> versionsToReload,
         }
         if (modelVersion->getStatus().getState() == ModelVersionState::END || modelVersion->getModelConfig().getBasePath() != config.getBasePath()) {
             downloadModels(fs, config, versionsToReload);
-        } else {
-            config.setLocalPath(modelVersion->getModelConfig().getLocalPath());
```
dkalinowski, line 269: Moving around this part may break cloud storage functionalities. I suppose unit tests pass, but it would be good to check whether regression tests pass before merge.

src/modelinstance.cpp
```diff
@@ -82,13 +82,19 @@ Status ModelInstance::loadInputTensors(const ModelConfig& config, const DynamicM
...
+            SPDLOG_WARN("Config shape - {} is mapped by {}. Changes will not apply", name, config.getMappingInputByKey(name));
+        } else if (networkInputs.count(name) == 0 && networkInputs.count(config.getRealInputNameByValue(name)) == 0) {
             SPDLOG_WARN("Config shape - {} not found in network", name);
             return StatusCode::CONFIG_SHAPE_IS_NOT_IN_NETWORK;
         }
     }
     for (const auto& [name, _] : config.getLayouts()) {
-        if (networkInputs.count(name) == 0 && network->getOutputsInfo().count(name) == 0) {
+        if (networkInputs.count(name) == 1 && config.getMappingInputByKey(name) != "") {
+            SPDLOG_WARN("Config layout - {} is mapped by {}. Changes will not apply", name, config.getMappingInputByKey(name));
```
dkalinowski, line 94: Are you sure we do not want to return an error similarly to line 99?

src/test/modelconfig_test.cpp
```diff
@@ -274,6 +274,44 @@ TEST(ModelConfig, mappingOutputs) {
     EXPECT_EQ(empty, "");
 }
 
+TEST(ModelConfig, mappingRealInputs) {
```
dkalinowski, line 277: There are missing unit tests for model loading. It would be great to have higher level test that checks whether model loads correctly with mapping and shape/layout set with mapped name.

dkalinowski: Possibly also request validation to add such end to end test?

src/modelinstance.cpp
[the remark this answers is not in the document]
atobiszei: When user defined input/output mapping we should forget about the original name everywhere outside the mapping file.
We shouldn't allow for passing both original name & mapped name as this could lead to issues with transitional mapping a mapped to b, b mapped to c. Less ambiguity the better.

src/test/modelconfig_test.cpp
atobiszei: Agree. After this MR we should have following tests (possibly some of them are already written):
-> model with mapping, trying to set shape/layout with old name -> expect FAIL
-> model with mapping, trying to set shape/layout with new mapped name -> expect PASS
-> some DAG with model mapping, using original name load should FAIL
-> some DAG with model mapping, using mapped name load should PASS

src/model.cpp
mkuczyns11: If we move this if statement to 253 line we would not get status, because `config.parseModelMapping()` returns one.

src/model.cpp
mkuczyns11: localPath property is overwritten in` downloadModels()` function, so I suppose everything should work just fine

src/modelinstance.cpp
mkuczyns11: Should we then return someting like `StatusCode::CONFIG_LAYOUT_MAPPED_BUT_USED_REAL_NAME` and same with shape everytime we find real input/output name used in config, which is mapped in mapping_config.json? Technically we would need to create 2 new StatusCodes and after every warn written  we would need to return appropriate StatusCode.

src/model.cpp
dkalinowski: Then why not moving parseModelMapping before `auto modelVersion = getModelInstanceByVersion(version);` ?

atobiszei (approved):

src/test/modelinstance_test.cpp
```diff
@@ -287,6 +287,86 @@ TEST_F(TestLoadModel, UnSuccessfulLoadWhenNireqTooHigh) {
...
+};
+
+TEST_F(TestLoadModelWithMapping, SuccessfulLoad) {
+    ovms::ModelInstance modelInstance("UNUSED_NAME", UNUSED_MODEL_VERSION, *ieCore);
+
+    ovms::ShapeInfo inputShape{ovms::FIXED, {1, 20}};
+    shapeMap["input"] = inputShape;
+    config.setShapes(shapeMap);
+
+    layouts["input"] = "LAYOUT_INPUT";
```
dkalinowski, line 317: Should it really work? `LAYOUT_INPUT` is not supported layout, only `NHWC` and `NCHW`

dkalinowski (approved):

PR #998, opened 2021-11-03
accepting numerical plugin config values and enforcing its correct format

dtrawins:


dkalinowski (changes requested):
Unit tests are missing

dkalinowski (approved):

atobiszei (approved):

PR #1003, opened 2021-11-04
simpler and more secure deployment in a container with an NCS stick

dtrawins:


rasapala (approved):
LGTM

mzegla (commented):
What about redhat image? Don't we need to also change it?

dtrawins:
@mzegla VPU is not supported on RH

mzegla (approved):

PR #1008, opened 2021-11-05
Cvs 68031 fix reporting product build number

jasiu86:
https://ci.iotg.sclab.intel.com/job/ovmsc/job/OVMSCOps/5994/


pgierasi (approved):

tests/functional/conftest.py
```diff
@@ -187,6 +187,17 @@ def get_docker_image_os_version_from_container():
     return os_distname
 
 
+def get_ov_and_ovms_version_from_container():
+    client = docker.from_env()
+    output = client.containers.run(image=image,
```
jszczepa, line 192: Why in this case you didn't use OvmsDocker?

jasiu86: It could be excessive, there is no difference, we require only simple version strings.
We do not want to make additional OvmsDocker setup.

jszczepa (approved):
Looks ok

PR #1026, opened 2021-11-17
Cvs 70944 - Relaxate requirement for DAG to require batch size equal 0 when using demultiplexer

mkuczyns11:


src/BUILD
```diff
@@ -245,6 +245,15 @@ cc_binary(
     linkshared = 1,
 )
 
+cc_binary(
```
dkalinowski, line 248: Maybe we could make use of `createLibraryMock<>` instead of making new custom node binary?

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -3033,15 +3033,15 @@ TEST_F(EnsembleConfigurationValidationWithDemultiplexer, MultipleBatchInCustomNo
         {NodeKind::ENTRY, ENTRY_NODE_NAME, "", std::nullopt, {{pipelineInputName, pipelineInputName}}},
         {NodeKind::CUSTOM, "custom_node_1", "", std::nullopt, {{"1", "out_OutputNumbers_1"}, {"2", "out_OutputNumbers_2"}}, demultiplyCount, {}, mockedLibrary,
             parameters_t{
-                {"in_InputNumbers", "3,3,10;FP32"},  // 1,3,10 is correct
```
dkalinowski, line 3036: Why do we remove this comment?

src/test/custom_nodes/node_do_nothing.cpp
```diff
@@ -0,0 +1,123 @@
...
+    (*info)->dims = (uint64_t*)malloc((*info)->dimsCount * sizeof(uint64_t));
+    for (int i = 0; i < dims.size(); i++) {
+        (*info)->dims[i] = dims.at(i);
+    }
+    (*info)->precision = FP32;
+    return 0;
+}
+
+int getOutputsInfo(struct CustomNodeTensorInfo** info, int* infoCount, const struct CustomNodeParam* params, int paramsCount, void* customNodeLibraryInternalManager) {
+    std::vector<int> dims{3, 5, 10};
```
dkalinowski, line 103: Why do we declare {3,5,10} vector if we fill it afterwards anyway?

src/test/custom_nodes/node_do_nothing.cpp
```diff
@@ -0,0 +1,123 @@
...
+                std::cout << "dims param wrong format, should be [int, ...]" << std::endl;
+                return 1;
+            }
+            dims.clear();
+            // remove '[' and ']'
+            dimsString.erase(0, 1).pop_back();
+            size_t pos = 0;
+            while ((pos = dimsString.find(',')) != std::string::npos) {
+                dims.emplace_back(atoi(dimsString.substr(0, pos).c_str()));
+                dimsString.erase(0, pos + 1);
```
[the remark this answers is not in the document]
dkalinowski: This is full required string valitation I suppose:
https://github.com/openvinotoolkit/model_server/blob/develop/src/custom_nodes/image_transformation/utils.hpp#L231

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -3033,15 +3033,15 @@ TEST_F(EnsembleConfigurationValidationWithDemultiplexer, MultipleBatchInCustomNo
...
+                {"in_InputNumbers", "3,3,10;FP32"},
                 {"out_OutputNumbers_1", "9,1,700;I32"},
```
dkalinowski, line 3037: Why do we need to create new node to pass shapes via parameters if we could use this mocked library?

src/BUILD
mkuczyns11: We can use LibraryMock instead, however I assume that we would only validate `pipelineDefinition` instead of creating the pipeline and executing. If that is fine then we can get rid of that CN and replace the tests with Mock.

src/test/custom_nodes/node_do_nothing.cpp
mkuczyns11: dropped dims declaring

src/test/custom_nodes/node_do_nothing.cpp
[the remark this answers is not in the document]
mkuczyns11: changed paramterer processing from this function to get_int_list_parameter from utils of image_transformation

src/test/ensemble_flow_custom_node_tests.cpp
mkuczyns11: created Library mock and used it in a test instead of creating new node

atobiszei (approved):

dkalinowski (approved):

PR #1031, opened 2021-11-18
2021.4.2 merge 2 develop

dtrawins:


src/custom_nodes/model_zoo_intel_object_detection/model_zoo_intel_object_detection.cpp
```diff
@@ -525,6 +527,15 @@ int getOutputsInfo(struct CustomNodeTensorInfo** info, int* infoCount, const str
     (*info)[2].dims[1] = 1;
     (*info)[2].dims[2] = 1;
     (*info)[2].precision = FP32;
+
```
dkalinowski, line 530: This is not resolved correctly. There is separate task to do it correctly: https://jira.devtools.intel.com/browse/CVS-69864
If we merge this code we will have memory leak or double free issue.

On main, the model zoo custom node does not use internal manager, therefore it does not use preallocated buffers. On develop, it uses preallocated buffers but you do not use it for info[3]. What's more, there will be segfault due to writing to unallocated memory (setting `*outputsCount = 3;` but writting to 4th element).

dkalinowski (changes requested):
Please compare this:

main:
https://github.com/openvinotoolkit/model_server/tree/main/example_client/cpp#start-the-client-1

develop:
https://github.com/openvinotoolkit/model_server/tree/develop/example_client/cpp#start-the-client-1

The --batch_size parameters differ. 1 on main 4 on develop. 1 is correct.
This change is missing in this PR. Are there maybe more doc changes skipped by mistake during conflict resolving?

rasapala (approved):
LGTM

dkalinowski (approved):

dkalinowski (approved):

PR #1033, opened 2021-11-19
Client library - changes in the readme and fixed max message size

mzegla:


rasapala (approved):
LGTM

client/python/lib/ovmsclient/tfs_compat/grpc/serving_client.py
```diff
@@ -94,6 +94,14 @@ def _build(cls, url, tls_config):
 
         ServingClient._check_url(url)
 
+        # Setting 1 GB max message size.
+        # TO DO: Make it configurable from make_grpc_client function
```
dtrawins, line 98: I would drop this todo comment

michalkulakowski (approved):

PR #1054, opened 2021-12-05
update opencv to 4.5.4 and new install package format

dtrawins:


Dockerfile.redhat
```diff
@@ -131,6 +131,10 @@ RUN if [ "$ov_use_binary" == "1" ] && [ "$DLDT_PACKAGE_URL" != "" ]; then true ;
     tar -zxf l_openvino_toolkit*.tgz -C /opt/intel && \
     ln -s /opt/intel/l_openvino_toolkit_runtime_rhel8_p* /opt/intel/openvino
 
+# install opencv
+#
+RUN cd /opt/intel/openvino && curl -s  http://s3.toolbox.iotg.sclab.intel.com/opencv-packages/centos7.tgz | tar -xzf -
```
dkalinowski, line 136: I guess we only need it with `"$ov_use_binary" == "1"` ?

src/custom_nodes/image_transformation/Dockerfile.redhat
```diff
@@ -16,12 +16,12 @@
 
 FROM registry.access.redhat.com/ubi8/ubi:8.4
 RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && yum update -d6 -y && yum install -d6 -y gcc-c++ curl xz
-RUN curl -s https://download.01.org/opencv/master/openvinotoolkit/thirdparty/linux/opencv/opencv_4.5.1-044_centos7.txz | tar --use-compress-program=xz -xvf -
+RUN curl -s http://s3.toolbox.iotg.sclab.intel.com/opencv-packages/ubuntu20.tgz | tar -xzf -
```
dkalinowski, line 19: Are we sure the package will be public once the OpenVINO release is ready? We should add to our release plan to add a commit replacing all opencv package links to public ones.

We only needed to do it for OpenVINO up till now, and now we need to do it for OpenCV in custom nodes as well.

Dockerfile.redhat
dtrawins: it turned out it is needed also for building from ov src

src/custom_nodes/image_transformation/Dockerfile.redhat
dtrawins: yes, it will be adjusted to the public version

dkalinowski (approved):

mkuczyns11 (approved):

PR #1062, opened 2021-12-07
ovms operator 0.2 for kubernetes

dtrawins:


rasapala (approved):
LGTM

mzegla (commented):
Some names still contain "ovms". Is that okay? Don't we want to get rid of "ovms" part everywhere?

mzegla (approved):

PR #1063, opened 2021-12-08
Ravikb config md5

ravikumarbhattiprolu:
While running modelserver inside SGX Encalve using gramine libOS, the stat system call is returning zeros for timestamp. This is because time stamp is not trusted in side enclave and hence not supported by Gramine. Because of this, model reloading due to config file change is failing as the config file change detection is based on time stamp. To make the config file change detection generic, have implemented it based on md5sum using openssl libraries. 

Had to modify few unit tests to load a config file with content change to trigger a file difference. Earlier tests were creating same config file again as time stamp change was considered as file change.

### DIFF
--- a/src/modelmanager.cpp
+++ b/src/modelmanager.cpp
@@ -33,6 +33,7 @@
 #include <rapidjson/prettywriter.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include "openssl/md5.h"
 
 #include "azurefilesystem.hpp"
 #include "config.hpp"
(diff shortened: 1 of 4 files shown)


dtrawins:
It looks ok. We will just run some performance tests to confirm if there is any noticeable impact on the throughput.

dtrawins (approved):

dkalinowski:
Needs to be rebased, develop has changed alot with OV API 2.0

dkalinowski (approved):

PR #1073, opened 2021-12-20
CVS-70086 internalManager ownership mechanism

mkuczyns11:


src/custom_node.hpp
```diff
@@ -46,7 +47,7 @@ class CustomNode : public Node {
         const std::unordered_map<std::string, std::string>& nodeOutputNameAlias = {},
         std::optional<uint32_t> demultiplyCount = std::nullopt,
         std::set<std::string> gatherFromNode = {},
-        void* customNodeLibraryInternalManager = nullptr);
+        std::shared_ptr<CNLIMWrapper> customNodeLibraryInternalManager = std::make_shared<CNLIMWrapper>(nullptr, [](void* ptr) -> int { return 0; }));
```
dkalinowski, line 50: Why can't it be just `nullptr` here?

src/custom_node_output_allocator.hpp
```diff
@@ -54,10 +57,10 @@ bool operator==(const CustomNodeTensor& t1, const CustomNodeTensor& t2);
 class CustomNodeOutputAllocator_2 : public ov::runtime::AllocatorImpl {
     struct CustomNodeTensor tensor;
     NodeLibrary nodeLibrary;
-    void* customNodeLibraryInternalManager;
+    std::shared_ptr<CNLIMWrapper>& customNodeLibraryInternalManager;
```
dkalinowski, line 60: Why is it reference here?

src/customnodesession.hpp
```diff
@@ -53,6 +54,6 @@ class CustomNodeSession : public NodeSession {
 
 private:
     static void releaseTensorResources(const struct CustomNodeTensor* tensor, const NodeLibrary& library, void* customNodeLibraryInternalManager);
-    Status createBlob(const struct CustomNodeTensor* tensor, std::shared_ptr<ov::runtime::Tensor>& resultBlob, const NodeLibrary& library, void* customNodeLibraryInternalManager);
+    Status createBlob(const struct CustomNodeTensor* tensor, std::shared_ptr<ov::runtime::Tensor>& resultBlob, const NodeLibrary& library, std::shared_ptr<CNLIMWrapper>& customNodeLibraryInternalManager);
```
dkalinowski, line 57: I see alot of places use shared pointer reference. Do we not want to track usage (use counter) of such places? If so, why?

src/pipelinedefinition.cpp
```diff
@@ -91,11 +91,6 @@ Status PipelineDefinition::initializeNodeResources() {
     for (const auto& nodeInfo : nodeInfos) {
         if (nodeInfo.kind == NodeKind::CUSTOM) {
             void* customNodeLibraryInternalManager = nullptr;
-            auto it = nodeResources.find(nodeInfo.nodeName);
-            bool internalManagerExists = (it != nodeResources.end());
-            if (internalManagerExists) {
```
dkalinowski, line 96: Why is it removed? What if we already have existing internal manager?

src/pipelinedefinition.cpp
```diff
@@ -1288,7 +1276,7 @@ const NodeInfo& PipelineDefinition::findNodeByName(const std::string& name) cons
     });
 }
 
-Shape PipelineDefinition::getNodeGatherShape(const NodeInfo& info) const {
+Shape PipelineDefinition::getNodeGatherShape(const NodeInfo& info) {
```
dkalinowski, line 1279: What has changed that forced us to remove `const` here?

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -4866,5 +4866,5 @@ TEST_F(EnsembleFlowCustomNodePipelineExecutionTest, SingleDeinitializeCallOnRelo
     ASSERT_EQ(factory.reloadDefinition("my_new_pipeline", std::move(info), std::move(connections), manager), StatusCode::OK);
     // Each custom node has effectively 1 internalManager initialized, because they use same library instance
     // in order to count whether deinitialize has been called expected number of times
-    ASSERT_EQ(LibraryCountDeinitialize::deinitializeCounter, 1);
+    ASSERT_EQ(LibraryCountDeinitialize::deinitializeCounter, 3);
```
dkalinowski, line 4869: Why is it 3?

src/custom_node.hpp
mkuczyns11: It can be nullptr. Will change that.

src/customnodesession.hpp
mkuczyns11: I suppose that we could pass void* instead of reference to shared_ptr in some places.

src/pipelinedefinition.cpp
mkuczyns11: Inside that function we call `getCustomNodeMetadata`. One of its arguments was void*, but it is changed to shared_ptr&. Trying to compile with const keyword causes compilation error.

dkalinowski: Could it be `const shared_ptr&` then? Or `void*` as you mentioned earlier?

src/customnodesession.hpp
dkalinowski: I mean, are we sure that we only need to keep one copy of shared_ptr in CustomNode object and pass reference in all other places?

src/test/ensemble_flow_custom_node_tests.cpp
mkuczyns11: It is 3 because as if for now, because calling reload on pipelinedefinition causes all nodeResources to be replaced/erased. Therefore, when we effectively remove 1 of 3 custom_nodes in pipeline deinitialize is called 3 times.
One deinitialize call is caused by erasing the resource allocated for the removed node.
Two deinitialize calls are caused by creating new customNodeLibraryInternalManager instances, which replace the old ones. Since there are none executions active replacing shared_ptr causes CNLIMWrapper to call its destructor.

michalkulakowski (approved):

src/pipelinedefinition.cpp
mkuczyns11: const shared_ptr& solved the problem

src/test/ensemble_flow_custom_node_tests.cpp
mkuczyns11: Mind that this test is going to be changed in `cleanup mechanism` task.

src/customnodesession.hpp
mkuczyns11: Pipeline object with CustomNode objects are destroyed after its creation and execution. Since all functions are part of that execution I think that there is no case in which that shared_ptr do not exist. Correct me if I am wrong.

dkalinowski: Agree

src/custom_node.hpp
mkuczyns11: It seems like I was wrong and it cannot be nullptr. That is because CNLIMWrapper has destructor which calls deinitialize(void*). When we pass second argument as nullptr than in destructor we try to call nullptr, which produces error(found that when unit test failed).

dkalinowski: What about:
```
std::shared_ptr<CNLIMWrapper> customNodeLibraryInternalManager = nullptr
```

dkalinowski: This way destructor is never called

src/customNodeLibraryInternalManagerWrapper.hpp
```diff
@@ -0,0 +1,40 @@
+//*****************************************************************************
```
dkalinowski, line 1: Use file naming convention

custom_node_library_interal_manager_wrapper.hpp

src/customNodeLibraryInternalManagerWrapper.cpp
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+#include "customNodeLibraryInternalManagerWrapper.hpp"
+
+namespace ovms {
+
+void* getCLIMWrapperPtr(const std::shared_ptr<CNLIMWrapper>& wrapper) {
+    if (wrapper.get() == nullptr) {
```
dkalinowski, line 21: `if (wrapper == nullptr)` would be enough

src/customNodeLibraryInternalManagerWrapper.hpp
```diff
@@ -0,0 +1,40 @@
...
+// limitations under the License.
+//*****************************************************************************
+#pragma once
+
+#include <iostream>
+#include <memory>
+
+namespace ovms {
+
+typedef int (*deinitialize_fn)(void*);
```
dkalinowski, line 23: Type re-definition.
This is already defined in another header.
This is bad, if we will ever change the `deinitialize_fn` prototype (add params or change types) we will need to do it in both places.

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -1140,6 +1140,177 @@ TEST_F(EnsembleFlowCustomNodeFactoryCreateThenExecuteTest, ParallelPipelineFacto
...
+    }
+
+    bool isPtrOwnedByManager(void* ptr) {
+        return (ptr == inputDims || ptr == outputDims || ptr == inputInfo || ptr == outputInfo || ptr == outputTensor || ptr == outputTensorData || ptr == outputTensorDims);
+    }
+};
+
+struct LibraryAddSubWithInternalManager {
+    static int initialize(void** customNodeLibraryInternalManager, const struct CustomNodeParam* params, int paramsCount) {
+        std::unique_ptr<AddSubInternalManager> internalManager = std::make_unique<AddSubInternalManager>();
```
dkalinowski, line 1169: Why use unique_ptr if we release it in L1170 anyway?

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -1140,6 +1140,177 @@ TEST_F(EnsembleFlowCustomNodeFactoryCreateThenExecuteTest, ParallelPipelineFacto
...
+        *customNodeLibraryInternalManager = internalManager.release();
+        return 0;
+    }
+    static int deinitialize(void* customNodeLibraryInternalManager) {
+        if (customNodeLibraryInternalManager != nullptr) {
+            AddSubInternalManager* internalManager = static_cast<AddSubInternalManager*>(customNodeLibraryInternalManager);
```
dkalinowski, line 1175: Just random thought, if user closes ovms with CTRL+C there is no cleanup of internal managers.

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -1140,6 +1140,177 @@ TEST_F(EnsembleFlowCustomNodeFactoryCreateThenExecuteTest, ParallelPipelineFacto
...
+        if (internalManager == nullptr)
+            return 1;
+        if (!internalManager->isPtrOwnedByManager(ptr)) {
+            free(ptr);
+        }
+        return 0;
+    }
+};
+
+TEST_F(EnsembleFlowCustomNodeFactoryCreateThenExecuteTest, PipelineFactoryCreationAndExecuteWithCustomNodeUsingInternalManager) {
```
dkalinowski, line 1274: Can we somehow check that the internal manager was in fact used?

mkuczyns11: We might implement getters and in each getter we could increment `counter` variable? What do you think?

dkalinowski: What about initializing the buffers with some data and make the execute simply return the data. This way you can check in test if the output data is the same as defined in `initialize`

src/test/ensemble_flow_custom_node_tests.cpp
mkuczyns11: I am not aware what is the expected behaviour in that situation, however I suppose that in order to cleanup internal managers pipelineDefinition and potentially CustomNode objects need to be destroyed. With introduction of cleanup mechanism additionally watcher thread will be responsible.

src/custom_node_library_internal_manager_wrapper.cpp
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+#include "custom_node_library_internal_manager_wrapper.hpp"
+
+namespace ovms {
+
+void* getCLIMWrapperPtr(const std::shared_ptr<CNLIMWrapper>& wrapper) {
+    if (wrapper == nullptr) {
```
dkalinowski, line 21: We can simplify this to `wrapper ? wrapper->ptr : nullptr`.
If so, maybe we could use this line directly in all those places and get rid of `getCLIMWrapperPtr` entirely? 😋

dkalinowski (approved):

PR #1082, opened 2021-12-29
CVS-72906/CVS-75078 API 1.0 cleanup part2

dkalinowski:
- CVS-75078 Fix stateful low accurracy issue
- bump OV package to 2022.1.0.476
- adjust ov::Function to ov::Model and ov::runtime::ExecutableNetwork to ov::runtime::CompiledModel
- adjust bazel BUILD file to support new openvino artifact names (libov_runtime.so)
- fix issue with binary inputs when uncaught exception was thrown if one of resolution parameters was -1 (any); added unit test for that
- removal of 1.0 api dependend OVMS code
- removal of inference_engine.hpp includes
- removal of getAnyValue and getFlatShape
- API 1.0 model loading removed (IE::Core/IE::CNNNetwork/IE::ExecutableNetwork) meaning API 1.0 loaded model is no longer in memory
- fix issue with changing shape but not modyfing layout
- cpu extension uses API 2.0 now
- metrics are read from API 2.0 now
- fixed tensorClone error log
- 

Not included
- tensorinfo layout enum still uses 1.0 api
- batch size in certain parts of OVMS is still considered on 1st position

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ DIST_OS_TAG ?= $(BASE_OS_TAG)
 ifeq ($(BASE_OS),ubuntu)
   BASE_OS_TAG=$(BASE_OS_TAG_UBUNTU)
   BASE_IMAGE ?= ubuntu:$(BASE_OS_TAG_UBUNTU)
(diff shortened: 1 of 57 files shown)


dtrawins (approved):

rasapala (approved):
LGTM

PR #1089, opened 2022-01-05
Fixing debug.

rasapala:


dtrawins (approved):

.bazelrc
```diff
@@ -53,6 +53,9 @@ build --apple_platform_type=macos
 build -c opt
 build --copt=-DGRPC_BAZEL_BUILD
 
+# fixing debug builds with ignore warnings
+build --copt=-O2
```
dkalinowski, line 57: Is this flag added to both debug and release builds?

If so, why O2? This is not the highest optimization flag (is it O3?)
Are we sure the performance is not affected?

rasapala: It is release and debug setting. Currently the default in release isO2 in release as I have checked so no performance affected.

dkalinowski (approved):

PR #1110, opened 2022-01-25
Updating GPU drivers.

rasapala:


docs/docker_container.md
```diff
@@ -82,12 +82,11 @@ as well as a release package (.tar.gz, with ovms binary and necessary libraries)
 Use command `make docker_build BASE_OS=redhat` or `make docker_build BASE_OS=centos`. OVMS with ubi base image doesn't support NCS and HDDL accelerators.
 
 Additionally you can set version of GPU driver used by the produced image. Currently following versions are available:
-- 19.41.14441
-- 20.35.17767
+- 21.49.21786
```
mzegla, line 85: Can we add also the version for Ubuntu and specify which one works with which system?

Makefile
```diff
@@ -41,7 +41,7 @@ BASE_OS_TAG_CLEARLINUX ?= latest
 BASE_OS_TAG_REDHAT ?= 8.4
 
 INSTALL_RPMS_FROM_URL ?=
-INSTALL_DRIVER_VERSION ?= "20.35.17767"
+INSTALL_DRIVER_VERSION ?= "21.49.21786"
```
mzegla, line 44: Can we make the version conditional? Depending on the BASE_OS we would set different default driver version.

docs/docker_container.md
```diff
@@ -82,12 +82,12 @@ as well as a release package (.tar.gz, with ovms binary and necessary libraries)
...
-- 19.41.14441
-- 20.35.17767
+- 21.49.21786 - Redhat
+- 21.48.21782 - Ubuntu
 
 Provide version from the list above as INSTALL_DRIVER_VERSION argument in make command to build image with specific version of the driver like 
-`make docker_build INSTALL_DRIVER_VERSION=19.41.14441`. 
-If not provided, version 20.35.17767 is used.
+`make docker_build INSTALL_DRIVER_VERSION=21.49.21786`.
+If not provided, version 21.49.21786 is used for Redhat os and 21.48.21782 is used for Ubuntu os.
```
mzegla, line 90:
```suggestion
If not provided, version 21.49.21786 is used for Redhat and 21.48.21782 is used for Ubuntu.
```

release_files/Dockerfile.redhat
```diff
@@ -38,43 +39,20 @@ LABEL "description"="OpenVINO(TM) Model Server is a solution for serving AI mode
 ARG INSTALL_RPMS_FROM_URL=
 ENV INSTALL_RPMS_FROM_URL=$INSTALL_RPMS_FROM_URL
 
-ARG INSTALL_DRIVER_VERSION="20.35.17767"
+ARG INSTALL_DRIVER_VERSION="21.49.21786"
```
mzegla, line 42: If we specify this argument the it should be used in this file. Currently it's defined but not used.

docs/docker_container.md
dtrawins: I think was could drop this whole section about using INSTALL_DRIVER_VERSION paramter. Only one version will be officially supported this time for a given os.

release_files/Dockerfile.redhat
rasapala: Currently we use the ocl neo driver solution that installs newest available driver.
The solution with specifying direct driver version does not work.
It might be changed in future.

mzegla (approved):

dtrawins (approved):

PR #1114, opened 2022-01-27
Bert client with dynamic shape usage

mkuczyns11:


mzegla (commented):
Please add information that this demo uses dynamic shapes to the description on main README page in demos directory.

demos/bert_question_answering/python/bert_question_answering.py
```diff
@@ -21,14 +21,47 @@
...
+        self.tokens += self.append_offset(tokens, len(self.text), len(self.text))
+        self.tokens_length = len(self.tokens)
+
+    def append_offset(self, tokens, s_offset, e_offset):
+        return list(map(lambda token: Token(token.id, token.start + s_offset, token.end + e_offset), tokens))
+    
+    def get_tokens_id(self):
+        return [token.id for token in self.tokens]
+
+    def get_tokens_se(self):
```
mzegla, line 55: To be honest I also don't like this "se" abberv. 
Maybe we could rename it to something like "range" or "position"?
"get_token_range" or "get_token_position" or something

demos/bert_question_answering/python/bert_question_answering.py
```diff
@@ -21,14 +21,47 @@
...
+        return [(token.start, token.end) for token in self.tokens]
+
+    def get_start(self, token_index):
```
mzegla, line 58: Maybe this could be Token object responsibility?
Token.get_start_position() or something like that?

demos/bert_question_answering/python/bert_question_answering.py
```diff
@@ -21,14 +21,47 @@
...
+        return self.tokens[token_index].start
+
+    def get_end(self, token_index):
```
mzegla, line 61: Same as above. Could be moved to Token maybe?

demos/bert_question_answering/python/bert_question_answering.py
mzegla: We could have get_token method that would return Token on provided index, but the start/end position might be returned by the Token itself.

demos/bert_question_answering/python/bert_question_answering.py
mkuczyns11: It seems like this method is not used anymore so I think just removing it should be okay.

mzegla: Even better 😆

demos/bert_question_answering/python/bert_question_answering.py
mkuczyns11: I agree, this functionality should not be the part of this class and should be split just like you suggest.

mzegla (approved):

dtrawins (approved):

PR #1118, opened 2022-01-27
Update redhat to ubi8.5.

rasapala:


dtrawins (approved):

mzegla:
Aren't there any other places we specify UBI version? I recall `BASE_OS_TAG_REDHAT` argument in some Makefiles or Dockerfiles. Could you check it?

mzegla (approved):

PR #1129, opened 2022-02-03
CVS-75094 Disallow layout setting incompatible with shape 

dkalinowski:
Incompatible layout with shape meaning different number of dimensions inside layout and shape.
Example of incompatible pairs:
- NCHW & (1,10)
- NC & (1,3,224,224)


rasapala (approved):
LGTM

mzegla (approved):

PR #1139, opened 2022-02-08
Allow cache with custom loaders model state LOADING,ERROR

mkuczyns11:


src/status.cpp
```diff
@@ -271,6 +272,7 @@ const std::unordered_map<const StatusCode, grpc::StatusCode> Status::grpcStatusM
     {StatusCode::MODEL_SPEC_MISSING, grpc::StatusCode::INVALID_ARGUMENT},
     {StatusCode::INVALID_SIGNATURE_DEF, grpc::StatusCode::INVALID_ARGUMENT},
     {StatusCode::PIPELINE_DEMULTIPLEXER_NO_RESULTS, grpc::StatusCode::ABORTED},
+    {StatusCode::ALLOW_CACHE_WITH_CUSTOM_LOADER, grpc::StatusCode::INVALID_ARGUMENT},
```
dkalinowski, line 275: Why is this translated to http/grpc? This status is never going to be exposed over the network, right? Is this only internal status we see in logs?

src/modelconfig.hpp
```diff
@@ -116,6 +116,11 @@ class ModelConfig {
          */
     bool disabledCaching = false;
 
+    /**
+         * @brief Cache allowing parameter
```
dkalinowski, line 120: Isn't `disabledCaching` and `allowCache` too much of configuration? By that I mean this might be misleading to have 2 flags

src/status.cpp
mkuczyns11: It seems like it is only status that we see in logs. I will remove the mappings.

src/modelconfig.hpp
mkuczyns11: allowCache was added to be able to determine if such parameter was in config file. With disabledCaching we are not aware if `allow_cache` was set in config, therefore we could not determine if the model should be loaded or not.

mkuczyns11: I suppose that it could work.

rasapala (approved):
LGTM

src/modelconfig.cpp
```diff
@@ -614,12 +614,12 @@ Status ModelConfig::parseNode(const rapidjson::Value& v) {
 
     // Model Cache options
     if (anyShapeSetToAuto())
-        this->setDisableCaching(true);
+        this->setModelCacheState(ModelCacheState::FORCE_CACHE_OFF);
```
dkalinowski, line 617: As stated in documentation: https://github.com/openvinotoolkit/model_server/blob/develop/docs/model_cache.md#enabling-the-cache-functionality
there are 2 scenarios where we disable cache by default:
- shape/batching auto
- w/ custom loader

There is only one scenario wehere we forcibly turn it off, without posibility to turn it on anyway. Doing so should throw an error. And this is w/ custom loader scenario.

This is why I'm worried of using FORCE_CACHE_OFF enum here. My suggestion is:
```
FORCE_CACHE_OFF - when user enforced cache but using custom loader
CACHE_OFF - when cache is disabled (default value)
CACHE_ON - when cache is enabled and not using custom loader
```

dkalinowski (approved):

michalkulakowski (approved):

PR #1144, opened 2022-02-10
model_cache docs fix

ficol:


dkalinowski (approved):

dkalinowski (changes requested):
Isn't it used in more places? Everywhere we use GPU device in example. For example here: https://github.com/openvinotoolkit/model_server/blob/develop/docs/docker_container.md#running-openvino-model-server-with-ai-accelerators-ncs-hddl-and-gpu-

Please find remaining usages

dkalinowski (approved):

atobiszei (approved):

PR #1146, opened 2022-02-11
Cvs 78676 opencv installation via download_opencv.sh script

dtrawins:


Dockerfile.redhat
```diff
@@ -131,8 +131,8 @@ RUN if [ "$ov_use_binary" == "1" ] && [ "$DLDT_PACKAGE_URL" != "" ]; then true ;
     ln -s /opt/intel/l_openvino_toolkit_runtime_rhel8_p* /opt/intel/openvino
 
 # install opencv
-#
-RUN cd /opt/intel/openvino && curl -s  http://s3.toolbox.iotg.sclab.intel.com/opencv-packages/ubuntu18.tgz | tar -xzf -
+ARG OPENVINO_OPENCV_DOWNLOAD_SERVER=""
```
dkalinowski, line 134: Shouldn't args be at the top of dockerfile for readability purposes?

dtrawins: for performance reasons it is not at the begining. Otherwise each change in the param would trigger refreshing all the layers

dkalinowski (approved):

atobiszei (approved):

PR #1159, opened 2022-02-16
update ov package to 592 and update opencv in custom nodes

dtrawins:


dkalinowski (approved):

mzegla (approved):
Consider changing `OPENCV_DOWNLOAD_FOLDER` to `OPENCV_DOWNLOAD_SOURCE`

`OPENCV_DOWNLOAD_FOLDER` looks like it describes the path to the catalog we are going to download opencv to.

PR #1170, opened 2022-02-23
Face_blur CN with demo using face-detection

mkuczyns11:
This PR introduces new custom node. Face_blur CN is responsible for blurring the image provided as one of the inputs in spots detected by object detection model from OMZ.  Additionally we are introducing a simple demo, which shows the usage of such a CN with the usage of face-detection-retail-0004 model.


docs/custom_node_development.md
```diff
@@ -118,6 +118,7 @@ Fully functional custom nodes are available here:
 - [east-resnet50 OCR custom node](https://github.com/openvinotoolkit/model_server/tree/develop/src/custom_nodes/east_ocr)
 - [model zoo intel object detection custom node](https://github.com/openvinotoolkit/model_server/tree/develop/src/custom_nodes/model_zoo_intel_object_detection)
 - [image transformation custom node](https://github.com/openvinotoolkit/model_server/tree/develop/src/custom_nodes/image_transformation)
+- [face blur custom node](https://github.com/openvinotoolkit/model_server/tree/develop/src/custom_nodes/face_blur)
```
mkuczyns11, line 121: This file does not include horizontal_ocr CN

demos/face_blur/python/face_blur.py
```diff
@@ -0,0 +1,102 @@
...
+    img = cv2.imread(path).astype(np.float32)  # BGR color format, shape HWC
+    img = cv2.resize(img, (resize_to_shape[1], resize_to_shape[0]))
+    target_shape = (img.shape[0], img.shape[1])
+    img = img.transpose(2,0,1).reshape(1,3,target_shape[0],target_shape[1])
+    request.inputs[name].CopyFrom(make_tensor_proto(img, shape=img.shape))
+
+def prepare_img_input_in_nhwc_format(request, name, path, resize_to_shape):
+    img = cv2.imread(path).astype(np.float32)  # BGR color format, shape HWC
+    img = cv2.resize(img, (resize_to_shape[1], resize_to_shape[0]))
+    target_shape = (img.shape[0], img.shape[1])
```
mzegla, line 50: Same as above

src/custom_nodes/face_blur/config.json
```diff
@@ -0,0 +1,90 @@
+{
```
mzegla, line 1: Isn't it a duplicate of what you added to demos? If so maybe remove this and link demo in the custom node readme?

src/custom_nodes/face_blur/face_blur.cpp
```diff
@@ -0,0 +1,259 @@
+//*****************************************************************************
+// Copyright 2021 Intel Corporation
```
mzegla, line 2:
```suggestion
// Copyright 2022 Intel Corporation
```

src/custom_nodes/face_blur/face_blur.cpp
```diff
@@ -0,0 +1,259 @@
...
+            boxes.emplace_back(box);
+            if (debugMode) {
+                std::cout << "Detection:\nImageID: " << imageId << "; LabelID:" << labelId << "; Confidence:" << confidence << "; Box:" << box << std::endl;
+            }
+        }
+    }
+
+    // Applying blur on detected areas
+    for (const auto& box : boxes) {
+        cv::GaussianBlur(image(box), image(box), cv::Size(51, 51), 0);
```
mzegla, line 134: What are these 51 s?

demos/face_blur/python/face_blur.py
```diff
@@ -0,0 +1,102 @@
...
+parser.add_argument('--image_layout', required=False, default='NHWC', choices=['NCHW', 'NHWC', 'BINARY'], help='Pipeline input image layout. default: NCHW')
+parser.add_argument('--image_output_name', required=False, default='image', help='Pipeline output name for output with image. default: image')
+parser.add_argument('--blurred_image_save_path', required=False, default='', help='If specified, people images will be saved to disk.')
+
+args = vars(parser.parse_args())
+
+def prepare_img_input_in_nchw_format(request, name, path, resize_to_shape):
+    img = cv2.imread(path).astype(np.float32)  # BGR color format, shape HWC
+    img = cv2.resize(img, (resize_to_shape[1], resize_to_shape[0]))
+    target_shape = (img.shape[0], img.shape[1])
```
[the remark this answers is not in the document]
mkuczyns11: Changed params to height,width instead of target_size/resize_to_shape since I also think that these two values would not be diffrent.

src/custom_nodes/face_blur/config.json
mkuczyns11: Yes it is, however each custom node has such a config.json file in their folder. So technically each CN in our repo has duplicate of config.json in demo. Is this something that we should address? Should I remove that file in such a case?

src/custom_nodes/face_blur/face_blur.cpp
mkuczyns11: `cv::Size(51,51)` parameter is kernel size. The bigger the kernel size more pixels from surrounding area are included in median computation. That said blur effect with `cv::Size(5,5)` would be weaker than with `cv::Size(51,51)`. Each pixel with gaussian blur is replaced with the median from surrounding pixels determined by kernel size.

mzegla: Sounds like a parameter we could expose to users. Why not let users set this value in config?
What do you think?

src/custom_nodes/face_blur/config.json
mzegla: I would have one config file in the demos and link to the demo in CN documentation, if you're okay with that.

mzegla (approved):

rasapala (approved):
LGTM

PR #1184, opened 2022-03-03
CVS-80230 mapping layout fix

ficol:


atobiszei (approved):

src/test/modelinstance_test.cpp
```diff
@@ -642,8 +657,8 @@ TEST_F(TestReloadModelWithMapping, SuccessfulReload) {
     shapeMap["input"] = inputShape;
     config.setShapes(shapeMap);
 
-    layouts["input"] = "LAYOUT_INPUT";
-    layouts["output"] = "LAYOUT_OUTPUT";
+    layouts["input"] = "NC";
```
dkalinowski, line 660: Why this change?

ficol: Because openvino doesn't accept this when creating ov::Layout here:
https://github.com/openvinotoolkit/model_server/blob/ad24df4970a131235f2421ef44dc71ad44df0697/src/modelinstance.cpp#L208-L209

dkalinowski: How comes this test was passing before?

ficol: Because before change, code fell into last else branch instead of into:
else if (config.getLayouts().count(mapped_name) > 0)

dkalinowski (approved):

PR #1188, opened 2022-03-07
model cache hddl documented

ficol:


docs/model_cache.md
```diff
@@ -42,6 +42,8 @@ In case there are valid reasons to enable the model cache also for models with a
 
 > IMPORTANT: Models imported via the custom loaders never create or use any cache.
 
+> IMPORTANT: Model cache can't be used with HDDL devices.
```
dkalinowski, line 45: Model cache can be used with HDDL, but doing so causes service crash, is that correct? There is also "Troubleshooting" document for such cases. @dtrawins

dtrawins: indeed we could add a small section in troubleshooting doc about caching. It could include several potential issues: caching is not effective because cache folder is not mounted or has wrong permissions, one cpu it is not giving performance gain and hddl is not working.

docs/troubleshooting.md
```diff
@@ -99,3 +100,9 @@ HTTP_proxy
 
 ## Cannot compile model into target device: `can't protect` in server logs <a name="load-network-issue"></a>
 - Since this is known bug, please refer OpenVINO&trade; [release notes](https://software.intel.com/content/www/us/en/develop/articles/openvino-relnotes.html).
+
+## Model Cache Issues <a name="model-cache"></a>
+
+- Cache folder should be mounted into docker container and have read-write access.
+- Using model cache on CPU doesn't give performance gain.
+- Using model cache with HDDL device is not working and results with deserialization error during inference.
```
dtrawins, line 108:
```suggestion
- Currently using model cache is not supported with HDDL target device. Do not enable model cache while using HDDL cards.
```

dtrawins (approved):

dkalinowski (approved):

PR #1200, opened 2022-03-15
venv and sdl-check targets update

mkuczyns11:
- Removal of PY_VERSION parameter since pip3 should be enough given creating venv with python3 link
- COPYRIGHT check in lib_search.py(sdl-check runs this script) changed. We should check for the word "Copyright", followed by the year of the first publication of the work and the name of the copyright holder. Several years may be noted if the work has gone through substantial revisions.
New check should check the following:
Copyright (single year|range|years devided by ',') (license holder)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2020-2021 Intel Corporation
+# Copyright (c) 2020-2022 Intel Corporation
 #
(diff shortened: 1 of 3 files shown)


rasapala (approved):
LGTM

dtrawins (approved):

jszczepa (approved):
Looks ok

PR #1203, opened 2022-03-16
v22 driver gpu ubuntu

dkalinowski:


docs/accelerators.md
```diff
@@ -87,13 +87,17 @@ docker run --rm -it  --device=/dev/dri --group-add=$(stat -c "%g" /dev/dri/rende
 ```
 
 >Note:
-> The public docker image includes the OpenCL drivers for GPU in version 21.38.21026.
-To use an older version, build the image with the following parameter: 
+> The public docker image includes the OpenCL drivers for GPU in version 21.38.21026 (RedHat) and 21.48.21782 (Ubuntu).
 
+To use an older version, build the image with the following parameter:
```
dtrawins, line 92: I think we could drop from readme the old version. there is no know reason to use it now

docs/accelerators.md
```diff
@@ -87,13 +87,17 @@ docker run --rm -it  --device=/dev/dri --group-add=$(stat -c "%g" /dev/dri/rende
...
-To use an older version, build the image with the following parameter: 
+> The public docker image includes the OpenCL drivers for GPU in version 21.38.21026 (RedHat) and 21.48.21782 (Ubuntu).
 
+To use an older version, build the image with the following parameter: 
 ```bash
-make docker_build INSTALL_DRIVER_VERSION=19.41.14441
+make docker_build INSTALL_DRIVER_VERSION=20.35.17767
 ```
 
+To use newer version on Ubuntu (for dGPU), build image with the following parameter:
```
dtrawins, line 97:
```suggestion
Support for [Intel Arc](https://www.intel.com/content/www/us/en/architecture-and-technology/visual-technology/arc-discrete-graphics.html), which is in preview now, requires newer driver version `22.10.22597`. You can build OpenVINO Model server with ubuntu base image and that driver using the command below:
```

docs/accelerators.md
dkalinowski: removed

dtrawins (approved):

rasapala (approved):
LGTM

atobiszei (changes requested):
What about RH?

dkalinowski:
> What about RH?

> What about RH?

https://download.copr.fedorainfracloud.org/results/jdanecki/intel-opencl/centos-stream-8-x86_64/
It doesn't look like there is version for redhat

dtrawins (approved):

atobiszei (approved):

PR #1221, opened 2022-03-24
ovmsclient 2022.1 changes

mzegla:


client/python/ovmsclient/lib/ovmsclient/tfs_compat/base/serving_client.py
```diff
@@ -140,32 +139,10 @@ def _open_private_key(cls, key_path):
...
     def _check_url(cls, url):
-        if isinstance(url, str):
-            try:
-                [address, port] = url.split(":", 1)
-            except ValueError:
-                raise ValueError("url must be a string in format <address>:<port>")
-        else:
-            raise TypeError("url must be a string in format <address>:<port>")
-        cls._check_address(address)
-        cls._check_port(port)
```
mkuczyns11, line 168: Make sure to update all demos and documents to use only "<address>:<port>".

client/python/ovmsclient/lib/ovmsclient/tfs_compat/http/serving_client.py
```diff
@@ -26,7 +27,8 @@
                                                   HttpModelMetadataResponse,
                                                   HttpPredictResponse)
 
-from ovmsclient.tfs_compat.base.errors import BadResponseError, ModelServerError, raise_from_http
+from ovmsclient.tfs_compat.base.errors import (BadResponseError, ModelServerError,
+                                               raise_from_http, raise_from_http_response)
```
mkuczyns11, line 31: This imports seems weird to me on github. Is there any indentation problem here?

client/python/ovmsclient/lib/ovmsclient/tfs_compat/http/serving_client.py
```diff
@@ -56,6 +58,10 @@ def predict(self, inputs, model_name, model_version=0, timeout=10.0):
         except ModelServerError as model_server_error:
             raise model_server_error
         except (JSONDecodeError, ValueError) as parsing_error:
+            if raw_response.status_code is not HTTPStatus.OK:
+                error_code = HTTPStatus(raw_response.status_code)
+                error_message = f"{error_code.value} {error_code.phrase}"
+                raise_from_http_response(error_code, error_message)
```
mkuczyns11, line 64: I suppose that this solves the issue with too big http request. I wonder if there could be better solution than trying to parse the response to json and then after it fails validating if this response was even valid.

client/python/ovmsclient/lib/setup.py
```diff
@@ -49,17 +49,17 @@ def run(self):
...
      long_description=long_description,
      long_description_content_type="text/markdown",
-     url="https://github.com/openvinotoolkit/model_server/tree/main/client/python/lib",
+     url="https://github.com/openvinotoolkit/model_server/tree/main/client/python/ovmsclient/lib",
      cmdclass={
         "build_apis": BuildApis,
     },
      packages=setuptools.find_namespace_packages(include=["ovmsclient*", "tensorflow*", "tensorflow_serving*"]),
-     install_requires=["grpcio>=1.21", "protobuf>=3.8", "numpy>=1.16.4", "validators>=0.18.2", "requests >= 2.26.0"],
+     install_requires=["grpcio==1.44", "protobuf==3.19.4", "numpy==1.19.5", "requests==2.27.1"],
```
mkuczyns11, line 64: Is there a reason to numpy version being 1.19.5? I suppose that this requirement is via our python >=3.6 requirement?

client/python/ovmsclient/lib/tests/tfs_compat_http/config.py
```diff
@@ -161,13 +161,13 @@
         "url": f"cluster.cloud.iotg.intel.com:{2**16-1}",
     }, {"_check_url": CallCount.ONE, "_check_tls_config": CallCount.ZERO}, None, None),
     ({
-        "url": "localhost:9000",
+        "url": "localhost:-90001111111111",
```
mkuczyns11, line 164: Should -<number> port be valid?

mkuczyns11 (approved):

client/python/ovmsclient/lib/ovmsclient/tfs_compat/http/serving_client.py
mkuczyns11: Now I see it was github :P. Resolved.

client/python/ovmsclient/lib/ovmsclient/tfs_compat/http/serving_client.py
mzegla: Yes, I think this is something we could refactor to have cleaner flow, but I'd do this for 2022.2 release.

client/python/ovmsclient/lib/setup.py
mzegla: Exactly, I picked the highest versions that support 3.6 and made them fixed to avoid missing some issues with different packages version for different python versions.

dkalinowski (changes requested):
This should not be commited before tagging v2022.1

dkalinowski (approved):

PR #1259, opened 2022-04-29
POC: Do not copy data during demultiplication (CVS-82879)

dkalinowski:
This PR shows posibility to not copy data during demultiplication. The original tensor can be saved (life time can as long as lifetime of shards). CVS-82879


dkalinowski:
Implemented here: https://github.com/openvinotoolkit/model_server/pull/1265

PR #1260, opened 2022-05-04
benchmark_client: fix timing & Docker build

ngaloppo:
Replace all time.time() calls with time.perf_count(), since the former
is not very accurate, and more importantly not a monotonic counter.
That means that we can end up with negative time deltas, causing all
kinds of trouble.


krzyczar:
> Replace all time.time() calls with time.perf_count(), since the former is not a very accurate, and more importantly not a monotonic counter. That means that we can end up with negative time deltas, causing all kinds of trouble.

I accept the proposition.

krzyczar (approved):

mzegla (approved):

michalkulakowski (approved):

PR #1299, opened 2022-06-15
fix version detection for apt package

dtrawins:


jasiu86 (approved):
Little tricky, but looks good.

PR #1312, opened 2022-06-30
Custom NEO runtime installation for Ubuntu 20.04

mzegla:


release_files/Dockerfile.ubuntu
```diff
@@ -76,6 +76,8 @@ ENV DEBIAN_FRONTEND=noninteractive
 SHELL ["/bin/bash", "-c"]
 WORKDIR /
 
+COPY drivers/dg2 /drivers/dg2
```
dkalinowski, line 79: This line will fail if user has no `drivers/dg2` directory. User usually have no such directory because he does not build with DG2 support.

dkalinowski (approved):

dtrawins (approved):

PR #1313, opened 2022-07-04
Paddlepaddle support

bstrzele:


demos/segmentation_using_paddlepaddle_model/python/requirements.txt
```diff
@@ -0,0 +1,3 @@
+ovmsclient
+paddlepaddle
```
dtrawins, line 2: I don't see it used on the example client

dtrawins (changes requested):

mzegla (approved):

PR #1357, opened 2022-08-19
Mkulakow/bug fixing

michalkulakowski:


demos/bert_question_answering/python/requirements.txt
```diff
@@ -1,2 +1,3 @@
 tensorflow-serving-api
```
dkalinowski, line 1: We should fix tensorflow-serving-api to 2.7.0 for example to match the protobuf,
check it here: https://github.com/openvinotoolkit/model_server/commit/d2179eb9752f54f7fa4425f93ef0150834473d65#diff-27c5febcc9ce93e8e5341f263b7d68962733b8a2a86c583723d025a1db39d7a3L8

mzegla (approved):

dkalinowski (approved):

PR #1392, opened 2022-09-08
Explicit paddlepaddle installation in the demo steps

mzegla:


dtrawins (approved):

dkalinowski (approved):

demos/segmentation_using_paddlepaddle_model/python/README.md
```diff
@@ -20,6 +18,13 @@ git clone https://github.com/openvinotoolkit/model_server.git
 cd model_server/demos/segmentation_using_paddlepaddle_model/python
 ```
 
+Install [paddlepaddle](https://pypi.org/project/paddlepaddle/) and other packages required to export the model to deployable format
+```bash
+pip3 install paddlepaddle
+pip3 install pyyaml
+pip3 install scipy
```
jasiu86, line 25: Can we use venv mechanism like in many other examples?

image

jasiu86: image

dtrawins: we are not enforcing the use of virtual env in other samples, that would be out of scope for this PR

ngrozae: can we enclose it in requirements file and add versions to it? this way we could be sure that there would be no problems     with wrong versions in the future

PR #1398, opened 2022-09-09
CVS-91621_test_commands_ovms_ocr_update

ngrozae:


dtrawins (approved):

demos/optical_character_recognition/python/README.md
```diff
@@ -122,7 +122,7 @@ Copy this `lib` folder to the same location with `text-recognition` and `east_ic
 
 ```bash
 mkdir -p OCR/east_fp32 OCR/lib
-cp -R model_server/src/custom_nodes/lib/ubuntu/libcustom_node_east_ocr.so OCR/lib/
+cp -R model_server/src/custom_nodes/lib/${BASE_OS}/libcustom_node_east_ocr.so OCR/lib/
```
dtrawins, line 125: because BASE_OS is not defined earlier in the readme, it might be confusing what are the options for the value. It would be better to add earlier:
export BASE_OS=ubuntu  # replace to 'redhat` if using UBI base image

dtrawins (changes requested):

demos/optical_character_recognition/python/README.md
jasiu86: added

dkalinowski (approved):

dtrawins (approved):

PR #1399, opened 2022-09-12
Cvs 91634 metric family

rasapala:
Enabling metrics with family names and changing bool handling to sets.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -151,11 +151,13 @@ Config& Config::parse(int argc, char** argv) {
                 cxxopts::value<bool>()->default_value("false"),
                 "STATEFUL")
             ("metrics_enabled",
-                "Flag indicating if metrics are enabled",
+                "Flag indicating if metrics are enabled.",
                 cxxopts::value<bool>()->default_value("false"),
                 "METRICS")
             ("metrics_list",
-                "List of metrics to be enabled. Comma separated.",
+                R"(List of metrics families to be enabled. If empty the default metrics are enabled:
+                ovms_requests_success, ovms_requests_fail, ovms_request_time_us, ovms_streams, ovms_inference_time_us, ovms_wait_for_infer_req_time_us.
+                Additional families: ovms_infer_req_queue_size, ovms_infer_req_active.)",
                 cxxopts::value<std::string>()->default_value(""),
                 "METRICS_LIST")
             ("idle_sequence_cleanup",
--- a/src/metric_config.hpp
+++ b/src/metric_config.hpp
@@ -30,6 +30,21 @@ public:
     bool metricsEnabled;
     std::string endpointsPath;
 
+    std::unordered_map<std::string, std::unordered_set<std::string>> additionalMetricFamilies = {
+        "ovms_streams" : {},
+        "ovms_infer_req_queue_size" : {},
+        "ovms_infer_req_active" : {},
+        "ovms_current_requests" : {}
+    };
+
+    std::unordered_map<std::string, std::unordered_set<std::string>> defaultMetricFamilies = {
+        "ovms_requests_success" : {"requestSuccessGrpcPredict"},
+        "ovms_requests_fail" : {" requestFailGrpcPredict "},
+        "ovms_request_time_us" : {},
+        "ovms_inference_time_us" : {},
+        "ovms_wait_for_infer_req_time_us" : {}
+    };
+
     // Request Success/Fail
     // TFS
     bool requestSuccessGrpcPredict;
--- a/src/metric_config.hpp
+++ b/src/metric_config.hpp
@@ -32,9 +32,7 @@ public:
     bool metricsEnabled;
     std::string endpointsPath;
 
-    Status parseMetricsArray(const rapidjson::Value& v);
     Status parseMetricsConfig(const rapidjson::Value& v);
-    bool validateEndpointPath(const std::string& endpoint);
     bool isFamilyEnabled(const std::string& family) const;
 
     void setDefaultMetricsTo(bool enabled);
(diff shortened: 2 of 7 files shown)


src/metric_config.hpp
```diff
@@ -30,73 +32,64 @@ class MetricConfig {
...
+            {"ovms_requests_success_grpc_predict",
+                "ovms_requests_success_grpc_getmodelmetadata",
+                "ovms_requests_success_grpc_getmodelstatus",
+                "ovms_requests_success_grpc_modelinfer",
+                "ovms_requests_success_grpc_modelmetadata",
+                "ovms_requests_success_grpc_modelready",
+                "ovms_requests_success_rest_modelinfer",
+                "ovms_requests_success_rest_predict",
+                "ovms_requests_success_rest_get_modelmetadata",
+                "ovms_requests_success_rest_get_modelstatus",
```
dkalinowski, line 54: same with `get_modelstatus`

src/model_metric_reporter.cpp
```diff
@@ -42,171 +42,176 @@ ServableMetricReporter::ServableMetricReporter(const MetricConfig* metricConfig,
         this->buckets.emplace_back(floor(BUCKET_MULTIPLIER * pow(BUCKET_POWER_BASE, i)));
     }
 
-    auto family = registry->createFamily<MetricCounter>("ovms_requests_success",
+    auto familyName = "ovms_requests_success";
```
dkalinowski, line 45:
```suggestion
    std::string familyName = "ovms_requests_success";
```
Using `auto` in this context might be std::string or const char*. Probably const char. Why wouldnt we define what type we want?

src/metric_config.hpp
```diff
@@ -30,73 +32,64 @@ class MetricConfig {
...
+                "ovms_requests_success_rest_modelmetadata",
+                "ovms_requests_success_rest_modelready"}},
+        {"ovms_requests_fail",
+            {"ovms_requests_fail_grpc_predict",
+                "ovms_requests_fail_grpc_getmodelmetadata",
```
dkalinowski, line 59: Why do we have those `"ovms_requests_fail_grpc_getmodelmetadata",...` metrics? We never use it in model metric reporter as far I can see.

dtrawins:
with invalid metric name, server fails with incorrect error message:
[servablemanagermodule.cpp:50] ovms::ModelManager::Start() Error: Configuration file not found or cannot open
It should mention invalid metric name instead

src/metric_config.hpp
```diff
@@ -30,73 +32,64 @@ class MetricConfig {
...
+        {"ovms_requests_success",
+            {"ovms_requests_success_grpc_predict",
+                "ovms_requests_success_grpc_getmodelmetadata",
+                "ovms_requests_success_grpc_getmodelstatus",
+                "ovms_requests_success_grpc_modelinfer",
+                "ovms_requests_success_grpc_modelmetadata",
+                "ovms_requests_success_grpc_modelready",
+                "ovms_requests_success_rest_modelinfer",
+                "ovms_requests_success_rest_predict",
+                "ovms_requests_success_rest_get_modelmetadata",
```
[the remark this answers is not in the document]
rasapala: Fixed

src/metric_config.hpp
rasapala: Fixed

src/metric_config.hpp
dkalinowski: We can add it in the future when we need it. The format of those single metric filters will differ anyway (will not be a single string but json object instead)

dtrawins (approved):

src/metric_config.hpp
rasapala: Ok

src/test/metric_config_test.cpp
```diff
@@ -121,9 +120,44 @@ TEST_F(MetricsConfigTest, ChangedValues) {
     ASSERT_EQ(metricConfig.metricsEnabled, true);
 
     // ASSERT_EQ(metricConfig.endpointsPath, "/newmetrics");
-    ASSERT_EQ(metricConfig.requestSuccessGrpcPredict, true);
-    ASSERT_EQ(metricConfig.requestFailRestModelReady, true);
-    ASSERT_EQ(metricConfig.requestSuccessGrpcModelReady, false);
+    ASSERT_TRUE(metricConfig.enabledFamiliesList.find("ovms_requests_success") != metricConfig.enabledFamiliesList.end());
```
dkalinowski, line 123: `enabledFamiliesList` could be private field

dkalinowski: Use `isFamilyEnabled`

rasapala: Done.

dkalinowski: All MetricConfig fields and methods are still public

src/config.cpp
```diff
@@ -151,11 +151,13 @@ Config& Config::parse(int argc, char** argv) {
...
             ("metrics_enabled",
-                "Flag indicating if metrics are enabled",
+                "Flag indicating if metrics are enabled.",
                 cxxopts::value<bool>()->default_value("false"),
                 "METRICS")
             ("metrics_list",
-                "List of metrics to be enabled. Comma separated.",
+                R"(List of metrics families to be enabled. If empty the default metrics are enabled:
+                ovms_requests_success, ovms_requests_fail, ovms_request_time_us, ovms_streams, ovms_inference_time_us, ovms_wait_for_infer_req_time_us.
+                Additional families: ovms_infer_req_queue_size, ovms_infer_req_active.)",
```
dkalinowski, line 160:
```suggestion
                R"(List of metric families to be enabled. If empty the default metrics are enabled:
                ovms_requests_success, ovms_requests_fail, ovms_request_time_us, ovms_streams, ovms_inference_time_us, ovms_wait_for_infer_req_time_us.
                Additional families: ovms_infer_req_queue_size, ovms_infer_req_active.)",
```

dkalinowski (approved):

PR #1404, opened 2022-09-14
Rest port fix

rasapala:


src/metric_config.cpp
```diff
@@ -107,7 +107,12 @@ void MetricConfig::setDefaultMetricsTo(bool enabled) {
     }
 }
 
-Status MetricConfig::loadFromCLIString(bool isEnabled, const std::string& metricsList) {
+Status MetricConfig::loadFromCLIString(bool isEnabled, const std::string& metricsList, uint64_t restPort) {
+    if (isEnabled && restPort == 0) {
+        SPDLOG_LOGGER_ERROR(modelmanager_logger, "Cli rest_port setting is missing, metrics are enabled on rest port.");
```
dtrawins, line 112:
```suggestion
        SPDLOG_LOGGER_ERROR(modelmanager_logger, "Parameter rest_port is not defined. It must be set to enable metrics on the REST interface");
```

src/test/metric_config_test.cpp
```diff
@@ -433,7 +433,10 @@ TEST_F(ModelMetricReporterTest, MetricReporterConstructorTest) {
     ASSERT_TRUE(reporter5.requestFailGrpcGetModelMetadata != nullptr);
 }
 
-class MetricsCli : public ::testing::Test {};
+class MetricsCli : public ::testing::Test {
+public:
+    uint64_t restPort = 100000;
```
dtrawins, line 438: allowed ports are in a range 1024-65536. It might be better to use lower port number in tests.

rasapala: Ok.

dtrawins (approved):

src/metric_config.cpp
```diff
@@ -107,7 +107,12 @@ void MetricConfig::setDefaultMetricsTo(bool enabled) {
     }
 }
 
-Status MetricConfig::loadFromCLIString(bool isEnabled, const std::string& metricsList) {
+Status MetricConfig::loadFromCLIString(bool isEnabled, const std::string& metricsList, uint64_t restPort) {
+    if (isEnabled && restPort == 0) {
```
dkalinowski, line 111: I have tested this manually and this does not work. Running config.json with monitoring enabled and missing `--rest_port` still starts OVMS normally.
What's the point of adding this parameter here instead of validating it in ovms::Config class?

src/test/metric_config_test.cpp
```diff
@@ -52,9 +56,18 @@ class MetricsConfigTest : public TestWithTempDir {
     }
     void SetUp() override {
         TestWithTempDir::SetUp();
-        // Prepare manager
-        modelPath = directoryPath + "/dummy/";
-        configFilePath = directoryPath + "/ovms_config.json";
+        char* n_argv[] = {"ovms", "--model_path", "/path/to/model", "--model_name", "some_name", "--rest_port", "8080"};
+        int arg_count = 9;
+        static bool parseOnce = [&arg_count, &n_argv]() {
```
dkalinowski, line 61:
```suggestion
        static void parseOnce = [&arg_count, &n_argv]() {
```

src/modelmanager.cpp
```diff
@@ -586,7 +586,7 @@ Status ModelManager::loadMetricsConfig(rapidjson::Document& configJson) {
     } else {
         const auto& metrics = itr2->value.GetObject();
         SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Parsing monitoring metrics config settings.");
-        firstErrorStatus = this->metricConfig.parseMetricsConfig(metrics);
+        firstErrorStatus = this->metricConfig.parseMetricsConfig(metrics, ovms::Config::instance().restPort());
```
dkalinowski, line 589: Why do we need to pass port to `parseMetricsConfig`? We pass down the responsibility to parser method. We chould keep this responsibility here in ModelManager by just returning error in line 587 if rest port is 0.

src/test/metric_config_test.cpp
dkalinowski, line 61:
```suggestion
        static void parseOnce = [&arg_count, &n_argv]() {
```

src/test/metric_config_test.cpp
```diff
@@ -52,9 +56,18 @@ class MetricsConfigTest : public TestWithTempDir {
...
+            ovms::Config::instance().parse(arg_count, n_argv);
+            return true;
+        }();
+
+        if (parseOnce) {
```
dkalinowski, line 66: Am I right that parseOnce is always true?

dkalinowski, line 66:
```suggestion
(void)parseOnce;
```

dkalinowski (approved):

PR #1405, opened 2022-09-14
KServe binary inputs docs and samples update

mzegla:


client/python/kserve-api/samples/README.md
```diff
@@ -569,13 +558,146 @@ optional arguments:
...
+time percentile 50: 191.00 ms; speed percentile 50: 5.24 fps
+time standard deviation: 9.58
+time variance: 91.76
+Classification accuracy: 100.00
+```
+
+
+### Run the Client to perform inference with binary encoded image <a name="http-model-infer-binary"></a>
+
+Using binary inputs feautre requires model to accept input in layout NHWC. The model used in other clients has native layout NCWH, therefore it must be adjusted on model server start up. For samples that use binary inputs, please start docker container with `--layout NHWC:NCHW` parameter:
```
dtrawins, line 618: this requirement is only for jpeg/png encoded input. With other types of data, the layout might be arbitrary.

docs/binary_input.md
```diff
@@ -10,16 +10,8 @@
...
-
-OpenVINO Model Server API allows for sending the request data in a variety of formats inside the TensorProto objects.
-Array data is passed inside the tensor_content field, which represent the input data buffer.
-
-When the data is sent in the `string_val` field, it is interpreted as a binary format of the input data.
-
-Note, that while the model metadata reports the inputs shape with layout NHWC, the binary data must be sent with 
-shape: [N] with dtype: DT_STRING. Where N represents number of images converted to string bytes.
+While OpenVINO models don't have the ability to process images directly in their binary format, the model server can accept them and convert
+automatically from JPEG/PNG to OpenVINO friendly format using OpenCV library.
```
dtrawins, line 14:
```suggestion
automatically from JPEG/PNG to OpenVINO friendly format using built-in OpenCV library.
```

client/python/kserve-api/samples/http_infer_binary_resnet.py
```diff
@@ -0,0 +1,139 @@
+#
```
dtrawins, line 1: perhaps we should have a version showing how to send binary data with jpeg content and a separate sample for other types using the REST extension?

docs/binary_input_kfs.md
```diff
@@ -0,0 +1,322 @@
...
+<7889 bytes of the third image in the batch for model_input tensor>
+```
+
+On the server side, the binary encoded data is loaded using OpenCV which then converts it to OpenVINO-friendly data format for inference.
+
+The structure of the response is specified [Inference Response specification](https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#inference-response-json-object).
+
+### Raw data
+
+Above section described how to send JPEG/PNG encoded image via REST interface. Data sent like this is processed by OpenCV to convert it to OpenVINO-friendly format. Many times, data is already available in OpenVINO-friendly format and all you want to do is to send it and get the prediction.
```
michalkulakowski, line 106: Is this comma after Many times needed?

michalkulakowski (approved):

dtrawins (approved):

PR #1409, opened 2022-09-14
Fail build script immediately when custom node image building fails

dkalinowski:
Previously it was possible to end script with exit code 0 when custom node building failed.
In rare scenarios it was possible to fail image building process and retrieve old custom node artifacts from old images.


rasapala (approved):
LGTM

atobiszei (approved):

mzegla (approved):

PR #1434, opened 2022-09-23
Kserve c++ samples with vector input format

bstrzele:


client/cpp/kserve-api/samples/grpc_infer_dummy.cc
```diff
@@ -0,0 +1,182 @@
...
+    input_ptr.reset(input);
+
+    FAIL_IF_ERR(
+        input_ptr->AppendRaw(
+            reinterpret_cast<uint8_t*>(&input_data[0]),
+            input_data.size() * sizeof(float)),
+        "unable to set data for input");
+
+    // Generate the outputs to be requested.
+    tc::InferRequestedOutput* output;
```
mzegla, line 121: I think, `outputs` parameter in KServe Inference API is optional and since model has only one output, we are interested in all of the outputs. Therefore maybe we can drop whole part with `outputs` ? Would it work?

dtrawins (commented):
looks good but change the file extension to cpp

client/cpp/kserve-api/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Run the Client to get server liveness <a name="grpc-server-live"></a>
+
+- Command
+
+```Bash
+./grpc_server_live --help
+Sends requests via KServe gRPC API to check if server is alive.
+Usage:
+  grpc_infer_dummy [OPTION...]
```
mzegla, line 28: grpc_infer_dummy?

client/cpp/kserve-api/README.md
```diff
@@ -0,0 +1,94 @@
...
+- Usage Example 
+
+```Bash
+./grpc_server_live --grpc_port 9000 --grpc_address localhost
+Server Live: True
+```
+
+## GRPC Examples with Dummy Model
+
+This section demonstrates request to simple model, which increments each provided value.
```
mzegla, line 46:
```suggestion
This section demonstrates inference on a simple model, which increments each provided value. 
```

client/cpp/kserve-api/README.md
```diff
@@ -0,0 +1,94 @@
...
+Server Live: True
+```
+
+## GRPC Examples with Dummy Model
+
+This section demonstrates request to simple model, which increments each provided value. 
+
+### Start the Model Server Container with Dummy Model
+```Bash
+docker run --rm -d -v $(pwd)/src/test/dummy:/models -p 9000:9000 openvino/model_server:latest --model_name dummy --model_path /models --port 9000
```
mzegla, line 50: `$pwd` is expected to be a specific value here as this command will work properly only when run from top level directory. Currently validation will not be able to handle that part. 
@ngrozae do you have any recommendation on how to handle this?

mzegla: Also, let's move this part before liveness check sample is described, so we have a consistent flow:
`building lib and samples -> starting OVMS -> checking liveness -> running inference`

client/cpp/kserve-api/README.md
```diff
@@ -0,0 +1,94 @@
...
+5 => 6
+6 => 7
+7 => 8
+8 => 9
+9 => 10
+======Client Statistics======
+completed_request_count 1
+Cumulative total request time 6.70928 ms
+Cumulative send time 0.039253 ms
+Cumulative receive time 0.005686 ms
```
mzegla, line 93: I don't know how to interpret those statistics. I think more friendly message could look like:
```
Number of requests: X1
Total processing time: X2
Latency: X3
Requests per second: X4
```

client/cpp/kserve-api/samples/grpc_infer_dummy.cpp
```diff
@@ -0,0 +1,175 @@
...
+    std::shared_ptr<tc::InferInput> input_ptr;
+    input_ptr.reset(input);
+
+    FAIL_IF_ERR(
+        input_ptr->AppendRaw(
+            reinterpret_cast<uint8_t*>(&input_data[0]),
+            input_data.size() * sizeof(float)),
+        "unable to set data for input");
+
+    // Generate the outputs to be requested.
```
mzegla, line 120: Is this comment still relevant?

client/cpp/kserve-api/README.md
ngrozae: To use ${pwd} here we should be in model_server directory. So you can either use $(pwd)../../../src/test/dummy here or change order of commands to:
git clone https://github.com/openvinotoolkit/model_server.git
cd model_server
docker run --rm -d -v $(pwd)/src/test/dummy:/models -p 9000:9000 openvino/model_server:latest --model_name dummy --model_path /models --port 9000 
cd client/cpp/kserve-api
cmake . && make

client/cpp/kserve-api/README.md
```diff
@@ -0,0 +1,97 @@
+# KServe API usage samples
+
+OpenVINO Model Server 2022.2 release introduced support for [KServe API](https://github.com/kserve/kserve/tree/master/docs/predict-api/v2).
+
+## Before you run the samples
+
+### Clone OpenVINO&trade; Model Server GitHub repository and go to the directory with the samples.
```
mzegla, line 7:
```suggestion
### Clone OpenVINO&trade; Model Server GitHub repository and go to the top directory.
```

client/cpp/kserve-api/README.md
```diff
@@ -0,0 +1,97 @@
...
+```Bash
+git clone https://github.com/openvinotoolkit/model_server.git
+cd model_server/client/cpp/kserve-api
```
mzegla, line 10:
```suggestion
cd model_server
```

mzegla (approved):

dtrawins (approved):

PR #1438, opened 2022-09-26
Check coverage

rasapala:


check_coverage.bat
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+#Ubuntu
+#MIN_LINES_COV=74.9
+#MIN_FUNCTION_COV=86.5
```
mzegla, line 5: Why these lines are commented, but RHELs aren't?

Dockerfile.redhat
```diff
@@ -179,6 +183,14 @@ RUN cd /ovms/src/example/SampleCpuExtension/ && make
 
 RUN bazel test ${debug_bazel_flags} --jobs $JOBS --test_summary=detailed --test_output=streamed  //src:ovms_test
 
+COPY check_coverage.bat /ovms/
+# Test Coverage
+RUN if [ "$CHECK_COVERAGE" == "1" ] ; then true ; else exit 0 ; fi ; wget https://github.com/linux-test-project/lcov/releases/download/v1.16/lcov-1.16-1.noarch.rpm &&\
+    yum install -y lcov-1.16-1.noarch.rpm &&\
+    bazel coverage --combined_report=lcov //src:ovms_test &&\
+    genhtml --output genhtml "$(bazel info output_path)/_coverage/_coverage_report.dat" &&\
+    ./check_coverage.bat && exit $?
```
mzegla, line 192: I guess it not possible to run coverage check before main building process, so we could fail faster?

rasapala: I can try it.

check_coverage.bat
rasapala: The rhel ones are lower thus we can use it. Currently onCommit tests rhel.

Dockerfile.redhat
rasapala: Done.

dtrawins (approved):

mzegla (approved):

PR #1485, opened 2022-10-24
[WIP] Documentation structure and start page 

tsavina:


docs/home.md
```diff
@@ -7,113 +7,67 @@
...
 
 ![OVMS picture](ovms.png)
 
-OpenVINO&trade; Model Server (OVMS) is a high-performance system for serving machine learning models. It is based on C++ for high scalability 
-and optimized for Intel solutions, so that you can take advantage of all the power of the Intel® Xeon® processor or Intel’s AI accelerators 
-and expose it over a network interface. OVMS uses the same architecture and API as [TensorFlow Serving](https://github.com/tensorflow/serving), 
-while applying OpenVINO for inference execution. Inference service is provided via gRPC or REST API, making it easy to deploy new algorithms and AI experiments.
+The key advantage of a model server is its ability to process several client requests simultaneously so that the model uses less device memory. 
+As a result, you can create a lightweight client with only the necessary functions to perform API calls for edge or cloud deployments.
```
tsavina, line 33: There is a note from JIRA ticker:
Most importantly we want to say why serving should be used, i.e. if you deploy in cluster or just want to disconnect app from inference, perhaps do model upgrades, etc.
It would be cool to get your comments on this info, should we include it and in which format.

docs/home.md
```diff
@@ -7,113 +7,67 @@
...
-Google Cloud Storage (GCS), Amazon S3, or Azure Blob Storage. 
+## OpenVINO Model Server Key Features: 
 
-Read [release notes](https://github.com/openvinotoolkit/model_server/releases) to find out what’s new.
+The models used by OpenVINO Model Server need to be stored locally or hosted remotely by object storage services. For more details, refer to [Preparing Model Storage](./models_repository.md) documentation.  
 
-Review the [Architecture concept](architecture.md) document for more details.
+OpenVINO&trade; Model Server works with [Bare Metal Hosts](host.md) as well as [Docker containers](docker_container.md). It is also suitable for landing in the [Kubernetes environment](../deploy/README.md).
 
-Key features:
```
tsavina, line 45: It's good to make some structure in presenting features here, also we could probably add API Reference Guide?
My first guess for structure is Basic/Advanced feature, but maybe you have a better idea (eg Model/Server/Client... or according to what users would need to use first, etc)

docs/home.md
```diff
@@ -7,113 +7,66 @@
...
-- [serving stateful models](stateful_models.md) - models that operate on sequences of data and maintain their state between inference requests
 - [binary format of the input data](binary_input.md) - data can be sent in JPEG or PNG formats to reduce traffic and offload the client applications
+- [serving stateful models](stateful_models.md) - models that operate on sequences of data and maintain their state between inference requests
+- online deployment of new [model versions](model_version_policy.md)
+- [configuration updates in runtime](online_config_changes.md) 
+- [performance tuning](performance_tuning.md)
 - [model caching](model_cache.md) - cache the models on first load and re-use models from cache on subsequent loads
 - [metrics](metrics.md) - metrics compatible with Prometheus standard
 
-**Note:** OVMS has been tested on RedHat, CentOS, and Ubuntu. The latest publicly released docker images are based on Ubuntu and UBI.
```
tsavina, line 62: can we remove this note or place it somewhere else? looks misplaced on the first page

PR #1489, opened 2022-10-26
Fix debug build with -O0 optimization level

dkalinowski:
With -O0 optimization level compiler had issues with passing LAYOUT_CONFIGURATION_DELIMETER by reference to std::count in `LayoutConfiguration::fromString`
Observed only with -O0


rasapala (approved):
LGTM

atobiszei (approved):

PR #1505, opened 2022-10-28
Add makefile debug target

michalkulakowski:


jasiu86:
Looks good, but let dev team verify correctness of this Makefile change and give approves.

mzegla (approved):

rasapala (approved):
LGTM

mzegla:
@michalkulakowski We will have debug images with that change, but I'm thinking about having a possibility to build an image with everything, so a build version with gdb, valgrind and GPU drivers, so it could be used to automate debuggers scans also with GPU enabled. Do we have a task to do that?

PR #1512, opened 2022-11-04
test version with the dGPU public drivers

dtrawins:


Makefile
```diff
@@ -72,7 +72,7 @@ DIST_OS_TAG ?= $(BASE_OS_TAG)
 ifeq ($(BASE_OS),ubuntu)
   BASE_OS_TAG=$(BASE_OS_TAG_UBUNTU)
   BASE_IMAGE ?= ubuntu:$(BASE_OS_TAG_UBUNTU)
-  INSTALL_DRIVER_VERSION ?= "21.48.21782"
+  INSTALL_DRIVER_VERSION ?= "22.35.24055"
```
mzegla, line 75: Shouldn't we do the same for redhat?

dtrawins: the latest rh package is now 22.28

mzegla: Yes, but in the section below (line 82 of this file) we have INSTALL_DRIVER_VERSION ?= "21.38.21026". 
Should we update it to 22.28?

mzegla (approved):

DockerfileMakePackage
```diff
@@ -35,7 +35,7 @@ RUN mkdir -vp /ovms_release/lib/custom_nodes
 
 RUN if [ -d /ovms/src/custom_nodes/lib/${BASE_OS} ] ; then true ; else exit 0 ; fi ; cp /ovms/src/custom_nodes/lib/${BASE_OS}/*.so /ovms_release/lib/custom_nodes/
 RUN cp /ovms/metadata.json /ovms_release/
-RUN if [ "$ov_use_binary" == "0" ] ; then true ; else exit 0 ; fi ; cp -v /openvino/bin/intel64/Release/lib/plugins.xml /ovms_release/lib/
+RUN if [ "$ov_use_binary" == "0" ] ; then true ; else exit 0 ; fi ; cp -v /openvino/bin/intel64/Release/plugins.xml /ovms_release/lib/
```
dkalinowski, line 38: Don't we need more than just plugins.xml? How about all the libs when building from src?

PR #1531, opened 2022-11-17
update OV backend to 17th Nov version

dtrawins:


src/test/model_cache_test.cpp
```diff
@@ -250,7 +250,7 @@ TEST_F(ModelCacheTest, PluginConfigChangeDoesImpactCache) {
 
     size_t lastCachedFileCount = currentCacheFileCount;
     currentCacheFileCount = this->getCachedFileCount();
-    ASSERT_GT(currentCacheFileCount, lastCachedFileCount);
+    ASSERT_GE(currentCacheFileCount, lastCachedFileCount);
```
dkalinowski, line 253: By changing only the assert from GT to GE the test name is wrong:
Plugin Config Change **Does** Impact Cache
I suggest to change it to ***DoesNot** and also change it to ASSERT_EQ to make sure it will never generate new files.

dkalinowski (changes requested):

dkalinowski (approved):

Dockerfile.redhat
```diff
@@ -73,6 +73,14 @@ RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.n
             vim \
             xz && \
             yum clean all
+
+# Build and install pugixml
+RUN git clone -b v1.13 https://github.com/zeux/pugixml && \
```
mzegla, line 78: Do we want to add some cleanup at the end of this layer?

Makefile
```diff
@@ -73,21 +73,21 @@ ifeq ($(BASE_OS),ubuntu)
   BASE_OS_TAG=$(BASE_OS_TAG_UBUNTU)
   BASE_IMAGE ?= ubuntu:$(BASE_OS_TAG_UBUNTU)
   INSTALL_DRIVER_VERSION ?= "21.48.21782"
-  DLDT_PACKAGE_URL ?= https://storage.openvinotoolkit.org/repositories/openvino/packages/2022.2/linux/l_openvino_toolkit_ubuntu20_2022.2.0.7713.af16ea1d79a_x86_64.tgz
+  DLDT_PACKAGE_URL ?= http://ov-share-03.sclab.intel.com/openvino_ci/private_builds/dldt/master/commit/0f647e6b75a87e334f44e0451f0f7ef9c66141c3/swf_drop/packages/releases/l_openvino_toolkit_ubuntu20_2022.3.0.8693.0f647e6b75a_x86_64.tgz
```
mzegla, line 76: It's our internal storage. Should we use it in public Makefile that can be used by people outside Intel?

Dockerfile.redhat
dtrawins: this is a build image and the repo is very small

Makefile
dtrawins: that is a development branch so it is ok to use internal package, it would be possible to use a param DLDT_PACKAGE_URL with a public prerelease but we need to use in tests the latest internal version

mzegla (approved):

PR #1537, opened 2022-11-21
Fix grpc_infer_binary_resnet.py

michalkulakowski:


client/python/kserve-api/samples/grpc_infer_binary_resnet.py
```diff
@@ -20,12 +20,47 @@
...
+        int_contents        "INT32"
+        uint64_contents     "UINT64"
+        uint_contents       "UINT32"
+    """
+    if datatype == DataType.INT32.name:
+        return "int_contents"
+    elif datatype == DataType.UINT32.name:
+        return "uint_contents"
+    else:
+        return f"{datatype.lower()}_contents"    # all other types
```
mzegla, line 63: Hmm, what if DataType instead of Enum was a map of pairs <dtype, field_name>. We would avoid those ifs. What do you think?

mzegla (approved):

PR #1555, opened 2022-11-30
CVS-97889 C-API error handling

dkalinowski:


src/test/c_api_tests.cpp
```diff
@@ -196,14 +196,43 @@ TEST(CApiStartTest, StartFlow) {
     ASSERT_NE(OVMS_ServerStartFromConfigurationFile(srv, go, mmo), nullptr);
 
     // Fix and expect ok
-    ASSERT_EQ(OVMS_ServerGeneralOptionsSetRestPort(go, 6666), nullptr);  // Different port
+    ASSERT_EQ(OVMS_ServerGeneralOptionsSetRestPort(go, 6666), nullptr);  // Different por
```
rasapala, line 199: port*

src/pocapi.cpp
```diff
@@ -35,6 +35,24 @@ using ovms::InferenceTensor;
...
 
+void OVMS_StatusDelete(OVMS_Status* status) {
+    delete reinterpret_cast<ovms::Status*>(status);
+}
+
+OVMS_Status* OVMS_StatusGetCode(OVMS_Status* status,
+    uint32_t* code) {
+    ovms::Status* sts = reinterpret_cast<ovms::Status*>(status);
+    *code = static_cast<uint32_t>(sts->getCode());
+    return nullptr;
```
atobiszei, line 46: here I wonder if the return of uint32_t wouldn't be better.

dkalinowski: I should add a check if status is nullptr. If we remove possibility of returning status then we cannot return an error

src/test/c_api_tests.cpp
dkalinowski: done

src/pocapi.cpp
dkalinowski: done

src/status.cpp
```diff
@@ -280,5 +280,17 @@ const std::unordered_map<const StatusCode, const std::string> Status::statusMess
     {StatusCode::NONEXISTENT_TENSOR_FOR_SET_BUFFER, "Tried to set buffer for nonexisting tensor"},
     {StatusCode::NONEXISTENT_TENSOR_FOR_REMOVE_BUFFER, "Tried to remove buffer for nonexisting tensor"},
     {StatusCode::NONEXISTENT_TENSOR_FOR_REMOVAL, "Tried to remove nonexisting tensor"},
+    {StatusCode::NONEXISTENT_BUFFER_FOR_REMOVAL, "Tried to remove nonexisting buffer"},
```
dkalinowski, line 283: Duplicated

src/pocapi.cpp
```diff
@@ -49,6 +49,34 @@ using ovms::StatusCode;
...
+}
+
+OVMS_Status* OVMS_StatusGetDetails(OVMS_Status* status,
+    const char** details) {
+    if (status == nullptr)
+        return reinterpret_cast<OVMS_Status*>(new Status(StatusCode::NONEXISTENT_STATUS));
+    if (details == nullptr)
+        return reinterpret_cast<OVMS_Status*>(new Status(StatusCode::NONEXISTENT_STATUS_DETAILS_PTR));
```
atobiszei, line 74: and here use NONEXISTENT_STRING?

src/pocapi.cpp
```diff
@@ -49,6 +49,34 @@ using ovms::StatusCode;
...
+    ovms::Status* sts = reinterpret_cast<ovms::Status*>(status);
+    *details = sts->string().c_str();
```
atobiszei, line 76: This line is correct as long as passed status is not ok - and it is probably issue on our side if someone here does have status ok - we don't return such statuses from C-API. However it could be good to add here additional safety check -> if status is ok then return INTERNAL_ERROR.

This is because status==OK does not have underlying message ptr so this could segfault in that case.

src/test/c_api_tests.cpp
```diff
@@ -41,14 +44,54 @@ static void testDefaultSingleModelOptions(MultiModelOptionsImpl* mmo) {
...
 
+#define ASSERT_CAPI_STATUS_NULL(code)         \
+    {                                         \
+        auto* err = code;                     \
+        if (err != nullptr) {                 \
+            const char* msg = nullptr;        \
+            OVMS_StatusGetDetails(err, &msg); \
+            std::string smsg(msg);            \
+            OVMS_StatusDelete(err);           \
+            ASSERT_TRUE(false) << smsg;       \
```
atobiszei, line 55:
```suggestion
            ASSERT_EQ(err, nullptr) << smsg;       \
```

src/test/c_api_tests.cpp
```diff
@@ -41,14 +44,54 @@ static void testDefaultSingleModelOptions(MultiModelOptionsImpl* mmo) {
     EXPECT_EQ(mmo->idleSequenceCleanup, std::nullopt);
 }
 
+#define ASSERT_CAPI_STATUS_NULL(code)         \
```
atobiszei, line 47: Rename code -> error or ovmsError. Code could be mistaken for StatusCode

src/test/c_api_tests.cpp
```diff
@@ -41,14 +44,54 @@ static void testDefaultSingleModelOptions(MultiModelOptionsImpl* mmo) {
...
+        }                                     \
+    }
+
+#define ASSERT_CAPI_STATUS_NOT_NULL(code) \
+    {                                     \
+        auto* err = code;                 \
+        if (err != nullptr) {             \
+            OVMS_StatusDelete(err);       \
+        } else {                          \
+            ASSERT_TRUE(false);           \
```
atobiszei, line 65:
```suggestion
            ASSERT_NE(err, nullptr);           \
```

src/test/c_api_tests.cpp
```diff
@@ -175,90 +219,110 @@ TEST(CApiConfigTest, MultiModelConfiguration) {
...
+    ASSERT_CAPI_STATUS_NOT_NULL_EXPECT_CODE(OVMS_StatusGetCode(sts, nullptr), StatusCode::NONEXISTENT_STATUS_CODE_PTR);
+    ASSERT_CAPI_STATUS_NULL(OVMS_StatusGetCode(sts, &code));
+    const char* details = nullptr;
+    ASSERT_CAPI_STATUS_NOT_NULL_EXPECT_CODE(OVMS_StatusGetDetails(nullptr, &details), StatusCode::NONEXISTENT_STATUS);
+    ASSERT_CAPI_STATUS_NOT_NULL_EXPECT_CODE(OVMS_StatusGetDetails(sts, nullptr), StatusCode::NONEXISTENT_STATUS_DETAILS_PTR);
+    ASSERT_CAPI_STATUS_NULL(OVMS_StatusGetDetails(sts, &details));
+    std::stringstream ss;
+    ss << Status(StatusCode::INTERNAL_ERROR).string() << " - custom message";
+    EXPECT_EQ(std::string(details), ss.str());
+    OVMS_StatusDelete(reinterpret_cast<OVMS_Status*>(s.release()));
```
atobiszei, line 302: Add test for trying to read Status(StatusCode::OK)

src/pocapi.cpp
```diff
@@ -49,6 +49,34 @@ using ovms::StatusCode;
...
+        return;
+    delete reinterpret_cast<ovms::Status*>(status);
+}
+
+OVMS_Status* OVMS_StatusGetCode(OVMS_Status* status,
+    uint32_t* code) {
+    if (status == nullptr)
+        return reinterpret_cast<OVMS_Status*>(new Status(StatusCode::NONEXISTENT_STATUS));
+    if (code == nullptr)
+        return reinterpret_cast<OVMS_Status*>(new Status(StatusCode::NONEXISTENT_STATUS_CODE_PTR));
```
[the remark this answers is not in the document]
dkalinowski: NONEXISTENT_NUMBER seem off since it is the pointer that does not exist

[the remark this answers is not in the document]
dkalinowski: but I guess we need to stick to NONEXISTENT_NUMBER to be consistent with other errrors

src/test/c_api_tests.cpp
dkalinowski: You cannot assert for such equality, err is not nullptr. It was valid ptr prior line 54.

src/test/c_api_tests.cpp
dkalinowski: This assert will never work, at this point err is always nullptr

src/pocapi.cpp
dkalinowski: There will never be segfault in case of OK status, the OK status has a string and it is empty string:
https://github.com/openvinotoolkit/model_server/blob/develop/src/status.cpp#L22

Added test for that.

src/pocapi.cpp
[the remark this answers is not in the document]
dkalinowski: done

src/pocapi.cpp
dkalinowski: done

src/test/c_api_tests.cpp
dkalinowski: This in fact is a code which get called inside the macro to get the code. Should I really rename it to ovmsError?

src/test/c_api_tests.cpp
dkalinowski: done

src/pocapi.cpp
atobiszei: Check this line:
https://github.com/openvinotoolkit/model_server/blob/develop/src/status.hpp#L296

src/test/status_test.cpp
```diff
@@ -36,4 +39,20 @@ TEST(StatusCodeTest, AllStatusCodesMapped) {
         ASSERT_NE(status.string(), "Undefined error");
     }
 }
+
+TEST(StatusCodeTest, CApi) {
+    for (auto statusCode = StatusCode::OK; statusCode != StatusCode::STATUS_CODE_END; ++statusCode) {
+        Status status = Status(statusCode);
```
atobiszei, line 45: You are using here copy constructor in L46. Please check what happens if you try to get details from status object from L45

atobiszei: why not use &status with reinterpret_cast? Previous comment is invalid

atobiszei (approved):
Please fix discussed comments

src/test/status_test.cpp
dkalinowski: done

src/test/c_api_tests.cpp
dkalinowski: done

src/test/c_api_tests.cpp
dkalinowski: done

PR #1559, opened 2022-12-02
set tfs-api import on version 2.11.0

dtrawins:


rasapala (approved):
LGTM

demos/bert_question_answering/python/requirements.txt
```diff
@@ -1,3 +1,2 @@
-tensorflow-serving-api==2.7.0
+tensorflow-serving-api==2.11.0
 validators
-protobuf>=3.18,<4.0.0
```
mzegla, line 3: Why do we delete this? I recall we added that specifically due to build issues. Don't we need it anymore? New TFS package comes with working version?

dtrawins: now tfs-api is installing a proper version of protobuf 3.19.6

mzegla (approved):

PR #1568, opened 2022-12-06
Calculate KFS REST binary inputs size from shape

michalkulakowski:


src/http_rest_api_handler.cpp
```diff
@@ -424,6 +383,56 @@ static Status validateContentFieldsEmptiness(KFSTensorInputProto* input) {
     return StatusCode::OK;
 }
 
+static bool isInputEmpty(::KFSRequest::InferInputTensor* input) {
```
dkalinowski, line 386: This function acts in a way that input can never be nullptr (there is no check if input is nullptr before proceeding).
```suggestion
static bool isInputEmpty(const ::KFSRequest::InferInputTensor& input) {
```
Therefore I'd suggest to change pointer to reference. 
I'd also add `const` since you never change the underlying data.

src/http_rest_api_handler.cpp
```diff
@@ -424,6 +383,56 @@ static Status validateContentFieldsEmptiness(KFSTensorInputProto* input) {
...
+    if (input->datatype() == "UINT8")
+        return input->contents().uint_contents_size() == 0;
+    if (input->datatype() == "FP64")
+        return input->contents().fp64_contents_size() == 0;
+    if (input->datatype() == "BYTES")
+        return input->contents().bytes_contents_size() == 0;
+    return true;
+}
+
+static Status handleBinaryInput(const int& binary_input_size, size_t& binary_input_offset, const size_t& binary_inputs_size, const char* binary_inputs, ::KFSRequest::InferInputTensor* input) {
```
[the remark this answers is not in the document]
dkalinowski: Same goes for `binary_inputs_size`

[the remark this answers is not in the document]
dkalinowski: The names are close to each other, what is the difference?

dkalinowski, line 412: Pass input by reference if you do not handle nullptr anyway

dkalinowski: Do that in all other places if possible

src/http_rest_api_handler.cpp
```diff
@@ -424,6 +383,56 @@ static Status validateContentFieldsEmptiness(KFSTensorInputProto* input) {
...
+        return status;
+    }
+    binary_input_offset += binary_input_size;
+    return StatusCode::OK;
+}
+
+static Status calculateBinaryDataSize(::KFSRequest::InferInputTensor* input, size_t& binary_data_size) {
+    auto element_size = KFSDataTypeSize(input->datatype());
+    size_t elements_number = 1;
+    for (auto dim : input->shape()) {
```
dkalinowski, line 429: You can use stl to calculate elements number in one line: https://github.com/openvinotoolkit/model_server/blob/main/src/customnodesession.cpp#L197

src/http_rest_api_handler.cpp
```diff
@@ -424,6 +383,56 @@ static Status validateContentFieldsEmptiness(KFSTensorInputProto* input) {
...
+        elements_number *= dim;
+    }
+    binary_data_size = elements_number * element_size;
+    return StatusCode::OK;
```
dkalinowski, line 433: Here we always return OK, why not to make the function void?

dkalinowski: Or even better - return binary_data_size directly from function

src/rest_parser.cpp
```diff
@@ -654,7 +659,7 @@ Status KFSRestParser::parseInputs(rapidjson::Value& node) {
     }
     requestProto.mutable_inputs()->Clear();
     for (auto& input : node.GetArray()) {
-        auto status = parseInput(input);
+        auto status = parseInput(input, (node.GetArray().Size() == 1));
```
dkalinowski, line 662: parseInput would be capable to deduce if there is only one input itself. Move `(node.GetArray().Size() == 1)` to the parseInput method

src/test/kfs_rest_parser_test.cpp
```diff
@@ -319,6 +319,98 @@ TEST_F(KFSRestParserTest, parseValidRequestBYTES) {
...
+TEST_F(KFSRestParserTest, parseValidRequestBYTES_noBinaryDataSizeParameter) {
+    std::string request = R"({
+    "inputs" : [
+        {
+        "name" : "input0",
+        "shape" : [ 2, 2 ],
+        "datatype" : "BYTES"
+        }
+    ]
+    })";
```
dkalinowski, line 331: Is this correct behavior? Sending binary tensor with shape [2,2] and no binary data?

src/test/kfs_rest_test.cpp
```diff
@@ -363,6 +363,24 @@ TEST_F(HttpRestApiHandlerTest, binaryInputsINT8) {
     ASSERT_EQ(i, 4);
 }
 
+static void assertSingleBinaryInput(const std::string& modelName, const std::optional<uint64_t>& modelVersion, ::KFSRequest& grpc_request) {
+    ASSERT_EQ(grpc_request.inputs_size(), 1);
+    ASSERT_EQ(grpc_request.model_name(), modelName);
+    ASSERT_EQ(grpc_request.model_version(), std::to_string(modelVersion.value()));
+
+    ASSERT_EQ(grpc_request.inputs()[0].name(), "b");
```
dkalinowski, line 371: assert size first to be sure it equals 1, otherwise we may end up with segfault in tests

src/http_rest_api_handler.cpp
```diff
@@ -472,6 +474,22 @@ static Status handleBinaryInputs(::KFSRequest& grpc_request, const std::string&
                 SPDLOG_DEBUG("binary_data_size parameter type should be int64 or string");
                 return StatusCode::REST_BINARY_DATA_SIZE_PARAMETER_INVALID;
             }
+        } else {
+            if (!isInputEmpty(input))
```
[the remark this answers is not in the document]
michalkulakowski: No but this method checks if input is empty after parsing request json and should be filled with binary data from request

PR #1572, opened 2022-12-07
CVS-95536 Make all unsupported http methods defined in RFC2616 return 406

dkalinowski:
Adding similar handling for CONNECT & TRACE as for other unsupported methods: HEAD/PUT/DELETE/OPTIONS/TRACE - HTTP status 406.
Other methods not specified in [RFC2616 ](https://www.rfc-editor.org/rfc/rfc2616#section-5.1.1) are still falling back to TFS handling, returning 501.
http://www.wangafu.net/~nickm/libevent-2.0/doxygen/html/http_8h.html#ac858319d667267f9fc848c2bb6931aa3

### DIFF
--- a/external/net_http.patch
+++ b/external/net_http.patch
@@ -1,9 +1,9 @@
 diff --git a/tensorflow_serving/util/BUILD b/tensorflow_serving/util/BUILD
-index 5beb2fb..e7b166c 100644
+index 5beb2fb1..e7b166c1 100644
 --- a/tensorflow_serving/util/BUILD
 +++ b/tensorflow_serving/util/BUILD
 @@ -4,7 +4,7 @@ load("//tensorflow_serving:serving.bzl", "serving_proto_library")
-
+ 
  package(
      default_visibility = [
 -        "//tensorflow_serving:internal",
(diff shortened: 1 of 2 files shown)


external/net_http.patch
```diff
@@ -27,12 +27,20 @@ index 36c925a..86a5ba9 100644
    // By default libevents only allow GET, POST, HEAD, PUT, DELETE request
    // we have to manually turn OPTIONS and PATCH flag on documentation:
    // (http://www.wangafu.net/~nickm/libevent-2.0/doxygen/html/http_8h.html)
+   evhttp_set_allowed_methods(
+       ev_http_, EVHTTP_REQ_GET | EVHTTP_REQ_POST | EVHTTP_REQ_HEAD |
+                     EVHTTP_REQ_PUT | EVHTTP_REQ_DELETE | EVHTTP_REQ_OPTIONS |
+-                    EVHTTP_REQ_PATCH);
```
dtrawins, line 33: why patch is removed?

dkalinowski: Patch is not removed, check final file after the change

dtrawins (approved):

atobiszei (approved):

PR #1439, opened 2022-09-27
KSereve c++ samples with resnet

bstrzele:

PR #1573, opened 2022-12-07
CVS-93520_doc_fixes_for_test

ngrozae:


ngrozae:
changes added to https://github.com/openvinotoolkit/model_server/pull/1439

PR #1593, opened 2022-12-13
Hadolint 

dkalinowski:
- remove boost for HDDL on ubuntu due to HDDL being not part of the 2022.3 scope

PR #1596, opened 2022-12-14
Add include

rasapala:


DockerfileMakePackage
```diff
@@ -53,11 +53,15 @@ RUN rm -f /ovms_release/lib/libcustom_node*
...
-liblcms.so libmanagement_agent.so libmanagement_ext.so libmanagement.so libmlib_image.so libnet.so libnio.so \
-libprefs.so librmi.so libsaproc.so libsctp.so libsplashscreen.so libsunec.so libsystemconf.so libunpack.so libverify.so libzip.so
+	rm -rf  libatk-wrapper.so libattach.so libawt_headless.so libawt.so libawt_xawt.so libdt_socket.so \
+	libextnet.so libfontmanager.so libinstrument.so libj2gss.so libj2pcsc.so libj2pkcs11.so libjaas.so \
+	libjavajpeg.so libjava.so libjawt.so libjdwp.so libjimage.so libjli.so libjsig.so libjsound.so libjvm.so \
+	liblcms.so libmanagement_agent.so libmanagement_ext.so libmanagement.so libmlib_image.so libnet.so libnio.so \
+	libprefs.so librmi.so libsaproc.so libsctp.so libsplashscreen.so libsunec.so libsystemconf.so libunpack.so libverify.so libzip.so
+
+# Remove capi temp libraries
+RUN if [ -f /ovms_release/lib/libsrc_Slibovms_Ushared.so ] ; then true ; else exit 0 ; fi ;cd /ovms_release/lib/ &&\
```
dkalinowski, line 63: Hadolint will error here due to using `cd` instead of `WORKDIR`. Instead of using `cd` use rm -rf with absolute path:
```
rm -rf /ovms_release/lib/libsrc_Slibovms_Ushared.so
```

DockerfileMakePackage
```diff
@@ -78,6 +82,7 @@ RUN find /ovms_release/lib/ -iname '*.so*' -exec patchelf --debug --set-rpath '$
 WORKDIR /ovms
 RUN cp -v /ovms/release_files/LICENSE /ovms_release/
 RUN cp -rv /ovms/release_files/thirdparty-licenses /ovms_release/
+RUN mkdir -vp /ovms_release/include && cp /ovms/src/pocapi.h /ovms_release/include
```
dkalinowski, line 85: Do we have source files in release image? /ovms/src/

rasapala: We use build image: BUILD_IMAGE=$(OVMS_CPP_DOCKER_IMAGE)-build

DockerfileMakePackage
rasapala: This is not a release docker image it's ok to use cd.

dkalinowski: SDL task forces us to fix issues or get approval from PSE: https://github.com/openvinotoolkit/model_server/pull/1593

PR #1605, opened 2022-12-15
Remove libs

rasapala:


release_files/Dockerfile.redhat
```diff
@@ -117,6 +117,11 @@ RUN     set -e ; \
         useradd --home-dir /home/ovms --create-home --uid 5000 --gid 5000 --groups 39,44 --shell /bin/bash --skel /dev/null ovms
 
 COPY --from=base_build /ovms /ovms
+
+# Remove capi library
```
dtrawins, line 121: move it to the first stage of the dockerfile

release_files/Dockerfile.ubuntu
```diff
@@ -153,6 +153,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends libxml2 && rm -
 COPY --from=base_build /usr/local/lib/ /ovms/lib/
 COPY --from=base_build /ovms /ovms
 
+# Remove capi library
+RUN if [ -f /ovms/lib/libovms_shared.so ] ; then true ; else exit 0 ; fi ;\
```
dtrawins, line 157: move it to the first stage of the build and simplify the command. else is not needed here

release_files/Dockerfile.redhat
rasapala: Why?

atobiszei (approved):

release_files/Dockerfile.redhat
dtrawins: to reduce the final image size

release_files/Dockerfile.ubuntu
rasapala: Cant move earlier , it just has been copied from /ovms.

release_files/Dockerfile.redhat
rasapala: Cant move earlier , it just has been copied from /ovms.

dkalinowski (approved):

release_files/Dockerfile.ubuntu
dtrawins: you should do it in between line 62 and 63

dtrawins (changes requested):
changes in release dockerfiles are needed.
When the unnecessary files will be removed in the first stage of the release build, a smaller package will be copied to the final image. Otherwise the image size will be the same in size the not needed files will be in the intermittent layers

rasapala:
Ok removed earlier in base image.

release_files/Dockerfile.ubuntu
```diff
@@ -64,6 +64,8 @@ RUN groupadd --gid 5000 ovms && useradd --home-dir /home/ovms --create-home --ui
              --gid 5000 --shell /bin/bash --skel /dev/null ovms && \
     chown -R ovms:ovms /ovms
 
+# Remove capi library
+RUN if [ -f /ovms/lib/libovms_shared.so ] ; then rm -rf  /ovms/lib/libovms_shared.so ; else exit 0 ; fi ;
```
dtrawins, line 68: looks good now but I guess `else` part could be dropped - it will return 0 anyway
```suggestion
RUN if [ -f /ovms/lib/libovms_shared.so ] ; then rm -rf  /ovms/lib/libovms_shared.so ; fi
```

dtrawins (approved):

dkalinowski (changes requested):
This might need rebase since hadolint fixes for dockefiles were merged

dkalinowski (approved):

PR #1626, opened 2022-12-22
use correct libcutensor version compatible with nvdia plugin

dtrawins:


michalkulakowski (approved):

dkalinowski (changes requested):
Are you sure we want to merge it to main after the release?

Dockerfile.ubuntu
```diff
@@ -103,8 +103,8 @@ RUN if [ "$NVIDIA" == "1" ] ; then true ; else exit 0 ; fi ; \
       cuda-11-8 \
```
jasiu86, line 103: Could we add explicit, currently used versions for other cuda packages (prevent problems in future)

jasiu86 (approved):
Looks good, consider specifying exact versions of cuda packages

PR #1641, opened 2023-01-02
Async kserve samples

bstrzele:


client/cpp/kserve-api/samples/grpc_async_infer_resnet.cpp
```diff
@@ -0,0 +1,204 @@
...
+                        acc++;
+                    }
+                    std::cout << std::endl;
+                }
+                cv.notify_all();
+            }, options, inputs);
+        input->Reset();
+    }
+    {
+        std::unique_lock<std::mutex> lk(mtx);
```
dkalinowski, line 180: Fix indentation, possibly use make style? cppstyle and cpplint.
To me, it looks like L180-L187 is in one indent

client/cpp/kserve-api/samples/grpc_async_infer_resnet.cpp
```diff
@@ -0,0 +1,204 @@
...
+        cv.wait(lk, [&]() {
+        if (c >= imgs.size()) {
```
dkalinowski, line 182: Missing indent (tab) for the entire if and else

dkalinowski: Is it possible to run `make style` for those sources? cpplint/cppstyle

bstrzele: Doesn't look like it

client/cpp/kserve-api/README.md
```diff
@@ -492,7 +496,55 @@ Latency: 9.6651 ms
...
+../../../../demos/common/static/images/golden_retriever.jpeg classified as 207 golden retriever 
+../../../../demos/common/static/images/airliner.jpeg classified as 404 airliner 
+../../../../demos/common/static/images/pelican.jpeg classified as 144 pelican 
+../../../../demos/common/static/images/snail.jpeg classified as 113 snail 
+../../../../demos/common/static/images/zebra.jpeg classified as 340 zebra 
+../../../../demos/common/static/images/arctic-fox.jpeg classified as 279 Arctic fox, white fox, Alopex lagopus 
+Accuracy 100%
+======Client Statistics======
+Number of requests: 10
+Total processing time: 283.336 ms
```
dkalinowski, line 542: Why is async slower than sync? 283ms vs 96ms

client/python/kserve-api/samples/grpc_async_infer_resnet.py
```diff
@@ -0,0 +1,203 @@
...
+            inputs[0].set_data_from_numpy(img)
+            triton_client.async_infer(
+                model_name=args.get(
+                    'pipeline_name') if is_pipeline_request else args.get('model_name'),
+                callback=partial(callback, data, x),
+                inputs=inputs,
+                outputs=outputs)
+
+        timeout = int(args.get('timeout'))
+        while ((len(data) != iterations) and timeout > 0):
```
dkalinowski, line 162: Do we want to support `--timeout 0` meaning endless wait?

client/python/kserve-api/samples/grpc_async_infer_resnet.py
```diff
@@ -0,0 +1,203 @@
+#
```
dtrawins, line 1: we should preserve original header if this sample is based on the triton samples

client/python/kserve-api/samples/http_async_infer_resnet.py
```diff
@@ -0,0 +1,191 @@
...
+                    args['input_name'], img.shape, "FP32"))
+            outputs = []
+            inputs[0].set_data_from_numpy(img)
+            data.append(triton_client.async_infer(
+                model_name=args.get(
+                    'pipeline_name') if is_pipeline_request else args.get('model_name'),
+                inputs=inputs,
+                outputs=outputs))
+
+            output = data[x].get_result().as_numpy(args['output_name'])
```
mzegla, line 160: Why for http there's a different approach than with grpc sample? This part calls async_infer method, but it looks like effectively it's synchronous as we call get_result() right after inference is scheduled.

bstrzele: There doesn't seem to be a callback equivalent in the http client https://github.com/triton-inference-server/client/blob/1a98fe0cd6b18ded726ada052f1bab06ab07ceb1/src/python/library/tritonclient/http/__init__.py#L1494

mzegla: You're right. Let's stick to this version then.

mzegla (approved):

dtrawins (approved):

dkalinowski (approved):

PR #1643, opened 2023-01-02
ovmsclient 2022.3 

mzegla:
- update links in PyPi overview 
- update base image for RedHat Dockerfile to UBI 8.7
- remove upper bounds for dependencies
- make adjustments so it can work with >= 1.24 numpy version and unify behavior for pre and post 1.24 for inhomogeneous array creation attempt
- add Makefile argument for also testing with numpy==1.23
- changed location of TF protos to stay under ovmsclient namespace - that way it can reside in the same Python environment with tensorflow without overwriting one another

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -391,7 +391,7 @@ test_functional: venv
 test_client_lib:
 	@cd client/python/ovmsclient/lib && \
 		make style || exit 1 && \
-		. .venv/bin/activate; make build || exit 1 && \
+		. .venv-ovmsclient/bin/activate; make build || exit 1 && \
 		make test TEST_TYPE=FULL || exit 1 && \
 		make clean
 
--- a/Makefile
+++ b/Makefile
@@ -391,7 +391,7 @@ test_functional: venv
 test_client_lib:
 	@cd client/python/ovmsclient/lib && \
 		make style || exit 1 && \
-		. .venv-ovmsclient/bin/activate; pip3 install protobuf==3.19.4; make build || exit 1 && \
+		. .venv-ovmsclient/bin/activate; make build || exit 1 && \
 		make test TEST_TYPE=FULL || exit 1 && \
 		make clean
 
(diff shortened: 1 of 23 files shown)


Makefile
```diff
@@ -391,8 +391,8 @@ test_functional: venv
 test_client_lib:
 	@cd client/python/ovmsclient/lib && \
 		make style || exit 1 && \
-		. .venv/bin/activate; make build || exit 1 && \
-		make test || exit 1 && \
+		. .venv-ovmsclient/bin/activate; make build || exit 1 && \
```
dkalinowski, line 394: test_client_lib has no target dependency, how can we ensure that .venv-ovmsclient will be present?

client/python/ovmsclient/lib/Makefile
```diff
@@ -15,10 +15,11 @@
 #
 
 VIRTUALENV_EXE := python3 -m virtualenv -p python3
-VIRTUALENV_DIR := .venv
+VIRTUALENV_DIR := .venv-ovmsclient
 ACTIVATE="$(VIRTUALENV_DIR)/bin/activate"
 
-PACKAGE_PATH := dist/ovmsclient-2022.2-py3-none-any.whl
+PACKAGE_PATH := dist/ovmsclient-2022.3-py3-none-any.whl
+TEST_TYPE := BASIC
```
dkalinowski, line 22: What's the difference between basic and full? Is it explained in some doc?

Makefile
mzegla: .venv_ovmsclient is created in Makefile in ovmsclient dir, so make style creates and configures such venv if it doesn't exist.

client/python/ovmsclient/lib/Makefile
mzegla: Basic runs with latest (default) versions dependencies for the environment. Full also reruns tests but with reinstalled versions of dependencies that we know are significantly different than latest (different in a way that forces us to change the code to support both). For now it's for numpy 1.23.

dkalinowski (approved):

michalkulakowski (approved):

mzegla:
This change will be merged to develop branch and get to main with the next release.

PR #1658, opened 2023-01-11
Override config metrics enable with cli

bstrzele:
Allows using metrics_enable param in CLI even with config. CLI will overwrite config setting if defined. Added warning in the logs.
JIRA:CVS-99880

### DIFF
--- a/src/cli_parser.cpp
+++ b/src/cli_parser.cpp
@@ -90,6 +90,14 @@ void CLIParser::parse(int argc, char** argv) {
                 "Overrides model cache directory. By default cache files are saved into /opt/cache if the directory is present. When enabled, first model load will produce cache files.",
                 cxxopts::value<std::string>(),
                 "CACHE_DIR")
+            ("metrics_enable",
+                "Flag enabling metrics endpoint on rest_port.",
+                cxxopts::value<bool>()->default_value("false"),
+                "METRICS")
+            ("metrics_list",
+                "Comma separated list of metrics. If unset, only default metrics will be enabled. Default metrics: ovms_requests_success, ovms_requests_fail, ovms_request_time_us, ovms_streams, ovms_inference_time_us, ovms_wait_for_infer_req_time_us. When set, only the listed metrics will be enabled. Optional metrics: ovms_infer_req_queue_size, ovms_infer_req_active.",
+                cxxopts::value<std::string>()->default_value(""),
+                "METRICS_LIST")
             ("cpu_extension",
                 "A path to shared library containing custom CPU layer implementation. Default: empty.",
                 cxxopts::value<std::string>()->default_value(""),
@@ -141,14 +149,6 @@ void CLIParser::parse(int argc, char** argv) {
                 "Flag indicating model is stateful",
                 cxxopts::value<bool>()->default_value("false"),
                 "STATEFUL")
-            ("metrics_enable",
-                "Flag enabling metrics endpoint on rest_port.",
-                cxxopts::value<bool>()->default_value("false"),
-                "METRICS")
-            ("metrics_list",
-                "Comma separated list of metrics. If unset, only default metrics will be enabled. Default metrics: ovms_requests_success, ovms_requests_fail, ovms_request_time_us, ovms_streams, ovms_inference_time_us, ovms_wait_for_infer_req_time_us. When set, only the listed metrics will be enabled. Optional metrics: ovms_infer_req_queue_size, ovms_infer_req_active.",
-                cxxopts::value<std::string>()->default_value(""),
-                "METRICS_LIST")
             ("idle_sequence_cleanup",
                 "Flag indicating if model is subject to sequence cleaner scans",
                 cxxopts::value<bool>()->default_value("true"),
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -125,11 +125,6 @@ bool Config::validate() {
         return false;
     }
 
-    // metrics on rest port
-    if ((metricsEnabled() || !metricsList().empty()) && !configPath().empty()) {
-        SPDLOG_WARN("When used cli param config.json cannot contain metrics configuration");
-    }
-
     // metrics_list without metrics_enable
     if (!metricsEnabled() && !metricsList().empty()) {
         std::cerr << "metrics_enable setting is missing, required when metrics_list is provided" << std::endl;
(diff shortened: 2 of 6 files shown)


src/config.cpp
```diff
@@ -126,8 +127,7 @@ bool Config::validate() {
 
     // metrics on rest port
     if ((metricsEnabled() || !metricsList().empty()) && !configPath().empty()) {
-        std::cerr << "metrics_enable or metrics_list and config_path cant be used together. Use json config file to enable metrics when using config_path." << std::endl;
-        return false;
+        SPDLOG_WARN("metrics_enable will override config contents");
```
dtrawins, line 130: metrics related parameters in CLI will overwrite metrics configuration from the config file.

dtrawins:
check also ovms --help output and make sure metrics params are in server settings section.

dkalinowski:
Missing ticket ID and description

src/config.cpp
```diff
@@ -126,8 +127,7 @@ bool Config::validate() {
 
     // metrics on rest port
     if ((metricsEnabled() || !metricsList().empty()) && !configPath().empty()) {
-        std::cerr << "metrics_enable or metrics_list and config_path cant be used together. Use json config file to enable metrics when using config_path." << std::endl;
```
dkalinowski, line 129: We should add **error** in case someone has CLI metric configuration and config.json metric configuration.
We should also have **error** in case somebody adds metric configuration into config.json at runtime while having CLI metric configuration from the beginning.
We should not pass this information through `startFromFile/loadConfig` calls.

src/modelmanager.cpp
```diff
@@ -757,6 +757,20 @@ Status ModelManager::loadConfig(const std::string& jsonFilename) {
 
     // Reading metric config only once per server start
     if (!this->metricConfigLoadedOnce) {
```
dkalinowski, line 759: It is not maintainable to edit loaded config.json into memory in the before applying the configuration. This can lead to hard investigate bugs when rapidjson object has different shape than file located on disk.

To address this issue, we could use existing `this->metricConfig.loadFromCLIString(config.metricsEnabled(), config.metricsList());` calls to modify `ModelManager` before loading config.json at all. Then, we can adjust applying the configuration to error in case metric config is already set. This way we can leave `ModelManager::startFromFile` and `ModelManager::loadConfig` without additional parameters such as `overrideMetricsEnable`.

src/test/modelmanager_test.cpp
```diff
@@ -270,6 +270,28 @@ static const char* modelMetricsMissingPort = R"(
...
+                "nireq": 100,
+                "shape": {"b": "(1,10) "}
+            }
+        }
+    ],
+    "monitoring":
+        {
+            "metrics":
+            {
+                "enable" : false
```
dkalinowski, line 291: It seems like we only handle `--metrics_enable`. Metrics list should also be accepted via CLI.

src/cli_parser.cpp
```diff
@@ -89,6 +89,14 @@ void CLIParser::parse(int argc, char** argv) {
                 "Overrides model cache directory. By default cache files are saved into /opt/cache if the directory is present. When enabled, first model load will produce cache files.",
                 cxxopts::value<std::string>(),
                 "CACHE_DIR")
+            ("metrics_enable",
+                "Flag enabling metrics endpoint on rest_port.",
+                cxxopts::value<bool>()->default_value("false"),
+                "METRICS")
+            ("metrics_list",
```
dkalinowski, line 96: Should we include information that when used, config.json cannot contain metric configuration? @dtrawins

src/config.cpp
```diff
@@ -126,8 +127,7 @@ bool Config::validate() {
...
-        return false;
+        SPDLOG_WARN("When used cli param config.json cannot contain metrics configuration");
```
dtrawins, line 130: This warning is not needed here because config is not parsed yet

src/modelmanager.cpp
```diff
@@ -585,10 +585,18 @@ Status ModelManager::loadCustomLoadersConfig(rapidjson::Document& configJson) {
...
+    auto& config = ovms::Config::instance();
     if (itr2 == configJson.MemberEnd() || !itr2->value.IsObject()) {
+        if (config.metricsEnabled()) {
+            return this->metricConfig.loadFromCLIString(true, config.metricsList());
+        }
         SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Configuration file doesn't have monitoring property.");
         return StatusCode::OK;
     } else {
+        if (config.metricsEnabled()) {
+            SPDLOG_LOGGER_ERROR(modelmanager_logger, "When used cli param config.json cannot contain metrics configuration");
```
dtrawins, line 597:
```suggestion
            SPDLOG_LOGGER_ERROR(modelmanager_logger, "Metrics configuration duplicated in configuration file and CLI parameters");
```

src/test/ovmsconfig_test.cpp
```diff
@@ -129,6 +129,13 @@ TEST_F(OvmsConfigDeathTest, negativeConfigPathWithTargetDevice) {
     EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(EX_USAGE), "Model parameters in CLI are exclusive with the config file");
 }
 
+TEST_F(OvmsConfigDeathTest, metricListInCli) {
```
dtrawins, line 132: Use more meaningful test name like OVMSDuplicatedMetricsConfig

dkalinowski (approved):

src/modelmanager.cpp
```diff
@@ -585,10 +585,18 @@ Status ModelManager::loadCustomLoadersConfig(rapidjson::Document& configJson) {
 
 Status ModelManager::loadMetricsConfig(rapidjson::Document& configJson) {
     const auto itr2 = configJson.FindMember("monitoring");
+    auto& config = ovms::Config::instance();
     if (itr2 == configJson.MemberEnd() || !itr2->value.IsObject()) {
+        if (config.metricsEnabled()) {
```
dtrawins, line 590: metricsList should be included, it also can not be used when the config.json has metrics settings

bstrzele: Does it matter, since using metric_list without metrics_enabled yields following error "metrics_enable setting is missing, required when metrics_list is provided"?

src/cli_parser.cpp
dtrawins: I would keep it short but indeed documentation should be adjusted in https://github.com/openvinotoolkit/model_server/blob/develop/docs/parameters.md 
A note in https://github.com/openvinotoolkit/model_server/blob/main/docs/metrics.md could be also useful that configuration can not be duplicated.

dtrawins (approved):

PR #1661, opened 2023-01-13
Loaded from cache log

bstrzele:


src/modelinstance.cpp
```diff
@@ -854,6 +854,7 @@ Status ModelInstance::loadModelImpl(const ModelConfig& config, const DynamicMode
         this->status.setLoading(ModelVersionStatusErrorCode::UNKNOWN);
         return StatusCode::MODEL_NOT_LOADED;
     }
+    SPDLOG_DEBUG("Model loaded from cache: {}", compiledModel->get_property(ov::loaded_from_cache));
```
dkalinowski, line 857: What is the type of returned compiledModel->get_property(ov::loaded_from_cache)? Are we sure all values are convertible to format required by SPDLOG?
To ensure, I would split the call to two, hardcoding the type. Is it bool? If so:
```
bool IsModelLoadedFromCache = compiledModel->get_property(ov::loaded_from_cache);
SPDLOG_DEBUG("Is model loaded from cache: {}", isModelLoadedFromCache);
```

dtrawins: does this method return some exception when this property is missing for some plugin different from CPU? I wonder if we should catch such exception

src/modelinstance.cpp
```diff
@@ -854,6 +854,12 @@ Status ModelInstance::loadModelImpl(const ModelConfig& config, const DynamicMode
         this->status.setLoading(ModelVersionStatusErrorCode::UNKNOWN);
         return StatusCode::MODEL_NOT_LOADED;
     }
+    try {
+        bool isModelLoadedFromCache = compiledModel->get_property(ov::loaded_from_cache);
+        SPDLOG_DEBUG("Is model loaded from cache: {}", isModelLoadedFromCache);
```
dkalinowski, line 859: Use `SPDLOG_LOGGER_DEBUG(modelmanager_logger` here as well

dtrawins (approved):

dkalinowski (approved):

PR #1675, opened 2023-01-20
Mkulakow/auto plugin num streams workaround

michalkulakowski:


src/modelinstance.cpp
```diff
@@ -695,14 +695,7 @@ Status ModelInstance::loadOVCompiledModel(const ModelConfig& config) {
...
     uint32_t numberOfStreams = 0;
-    try {
-        numberOfStreams = getNumOfStreams();
-        SPDLOG_LOGGER_INFO(modelmanager_logger, "Number of OpenVINO streams: {}", numberOfStreams);
-    } catch (ov::Exception& e) {
-        SPDLOG_LOGGER_WARN(modelmanager_logger, "Unable to get information about number of streams with error: {}; model: {}; version: {}; device: {}", e.what(), getName(), getVersion(), config.getTargetDevice());
-    } catch (...) {
-        SPDLOG_LOGGER_WARN(modelmanager_logger, "Unable to get information about number of streams; model: {}; version: {}; device: {}", getName(), getVersion(), config.getTargetDevice());
-    }
+    getNumOfStreams(numberOfStreams);
```
dkalinowski, line 698: Wouldnt it be better to change return type to `uint32_t` and return 0 in case of errors? I think it would be more clean

dkalinowski (approved):

src/modelinstance.cpp
```diff
@@ -1251,8 +1243,32 @@ const size_t ModelInstance::getBatchSizeIndex() const {
...
+}
+
 uint32_t ModelInstance::getNumOfStreams() const {
-    return compiledModel->get_property(ov::num_streams);
+    try {
+        uint32_t numStreams = compiledModel->get_property(ov::num_streams);
+        SPDLOG_LOGGER_INFO(modelmanager_logger, "Number of OpenVINO streams: {}", numStreams);
+        return numStreams;
+    } catch (ov::Exception& e) {
+        SPDLOG_LOGGER_WARN(modelmanager_logger, "Unable to get information about number of streams with error: {}; model: {}; version: {}; device: {}", e.what(), getName(), getVersion(), config.getTargetDevice());
```
dtrawins, line 1265: That is not a reason for a warrning - it will be a normal situation by design for virtual plugins and gpu.
It suggest a debug log instead with a message Unable to get property ov::num_streams for device ...

src/modelinstance.cpp
```diff
@@ -1251,8 +1243,32 @@ const size_t ModelInstance::getBatchSizeIndex() const {
...
+        SPDLOG_LOGGER_WARN(modelmanager_logger, "Number of streams will be set to number of optimal infer requests.");
+    } catch (...) {
+        SPDLOG_LOGGER_WARN(modelmanager_logger, "Unable to get information about number of streams; model: {}; version: {}; device: {}", getName(), getVersion(), config.getTargetDevice());
```
dtrawins, line 1268: like in previous comment

src/modelinstance.cpp
```diff
@@ -1251,8 +1243,32 @@ const size_t ModelInstance::getBatchSizeIndex() const {
...
+
 uint32_t ModelInstance::getNumOfStreams() const {
-    return compiledModel->get_property(ov::num_streams);
+    try {
+        uint32_t numStreams = compiledModel->get_property(ov::num_streams);
+        SPDLOG_LOGGER_INFO(modelmanager_logger, "Number of OpenVINO streams: {}", numStreams);
+        return numStreams;
+    } catch (ov::Exception& e) {
+        SPDLOG_LOGGER_WARN(modelmanager_logger, "Unable to get information about number of streams with error: {}; model: {}; version: {}; device: {}", e.what(), getName(), getVersion(), config.getTargetDevice());
+        SPDLOG_LOGGER_WARN(modelmanager_logger, "Number of streams will be set to number of optimal infer requests.");
```
dtrawins, line 1266: change to debug

src/modelinstance.cpp
```diff
@@ -1251,8 +1243,32 @@ const size_t ModelInstance::getBatchSizeIndex() const {
...
 }
 
+uint32_t ModelInstance::getOptimalNumberOfInferRequests() const {
+    try {
+        uint32_t numOptimalInferRequests = compiledModel->get_property(ov::optimal_number_of_infer_requests);
+        SPDLOG_LOGGER_INFO(modelmanager_logger, "Number of OpenVINO streams: {}", numOptimalInferRequests);
+        return numOptimalInferRequests;
+    } catch (ov::Exception& e) {
+        SPDLOG_LOGGER_WARN(modelmanager_logger, "Unable to get information about number of optimal infer requests with error: {}; model: {}; version: {}; device: {}", e.what(), getName(), getVersion(), config.getTargetDevice());
+    } catch (...) {
```
dtrawins, line 1253: what could be the exception other than ov::Exception? why this is handled separately but exactly the same way

dtrawins (approved):

PR #1679, opened 2023-01-24
Return dag name in kserve api response

michalkulakowski:


dkalinowski (changes requested):
Tests are missing

dkalinowski:
Missing task ID

src/dags/exit_node.hpp
```diff
@@ -35,13 +35,15 @@ class ExitNode : public Node {
     ResponseType* response;
     const tensor_map_t outputsInfo;
     bool useSharedOutputContent;
+    const std::string pipelineName;
 
 public:
-    ExitNode(ResponseType* response, const tensor_map_t& outputsInfo, std::set<std::string> gatherFromNode = {}, bool useSharedOutputContent = true) :
+    ExitNode(ResponseType* response, const tensor_map_t& outputsInfo, std::set<std::string> gatherFromNode = {}, bool useSharedOutputContent = true, const std::string pipelineName = "") :
```
dkalinowski, line 41: With `const std::string pipelineName = ""` there is high chance of perf drop since we copy the name each time we create pipeline (each time new request is received)

src/test/test_utils.cpp
```diff
@@ -145,7 +145,7 @@ ovms::tensor_map_t prepareTensors(
 
 void checkDummyResponse(const std::string outputName,
     const std::vector<float>& requestData,
-    PredictRequest& request, PredictResponse& response, int seriesLength, int batchSize) {
+    PredictRequest& request, PredictResponse& response, int seriesLength, int batchSize, std::string servableName) {
```
dkalinowski, line 148: Again, we do not need to copy here

src/test/test_utils.cpp
```diff
@@ -166,7 +166,8 @@ void checkDummyResponse(const std::string outputName,
 
 void checkDummyResponse(const std::string outputName,
     const std::vector<float>& requestData,
-    ::KFSRequest& request, ::KFSResponse& response, int seriesLength, int batchSize) {
+    ::KFSRequest& request, ::KFSResponse& response, int seriesLength, int batchSize, std::string servableName) {
```
dkalinowski, line 169: Copy

src/test/ensemble_tests.cpp
```diff
@@ -88,7 +88,7 @@ class EnsembleFlowBothApiTest : public TestWithTempDir {
         prepareKFSInferInputTensor(request, customPipelineInputName, std::make_tuple(shape, ovmsPrecisionToKFSPrecision(ovms::Precision::FP32)), requestData);
     }
 
-    void checkDummyResponse(int seriesLength, int batchSize = 1, std::string servableName = "") {
+    void checkDummyResponse(int seriesLength, int batchSize = 1, std::string& servableName = "") {
```
dkalinowski, line 91: `checkDummyResponse` should not change `servableName` parameter inside its implementation, therefore I suggest changing to const, even though its only unit test

dkalinowski (approved):

PR #1682, opened 2023-01-25
Document metric overhead

dkalinowski:


docs/metrics.md
```diff
@@ -205,6 +205,9 @@ curl http://localhost:8000/metrics
 ```
 [Example metrics output](https://raw.githubusercontent.com/openvinotoolkit/model_server/v2022.2/docs/metrics_output.out)
 
+## Performance considerations
+The metrics have minimal-to-zero performance overhead when used with standard models. However when used with lightweight, fast models which inference time is very small, the metric incrementation can take significant amount of CPU time compared to actual inference. Consider disabling metrics for such models.
```
dtrawins, line 209:
```suggestion
Collecting metrics has negligible performance overhead when used with typical models. However when used with very lightweight, fast models which inference time is very short, the metric incrementation can take noticeable proportion of the processing time. Consider it while enabling metrics for such models.
```

docs/performance_tuning.md
```diff
@@ -150,6 +150,7 @@ The default value is 1 second which ensures prompt response to creating new mode
   or even disable it. For example, with cloud storage, it could cause a cost for API calls to the storage cloud provider. Detecting new versions 
   can be disabled with a value `0`.
 
+- While metrics have minimal-to-zero performance overhead when used with standard models, when used with lightweight, fast models, the metric incrementation can take significant amount of CPU time compared to actual inference. Consider disabling metrics for such models.
```
dtrawins, line 153:
```suggestion
- While collecting metrics has negligible performance overhead when used with typical models, when used with lightweight, fast models, the metric incrementation can consume noticeable proportion of CPU time compared to actual inference. Take in into account while enabled metrics for such models. 
It applies also for the DEBUG log level which could increase the load for a very high throughout use cases.
```

docs/metrics.md
```diff
@@ -205,6 +205,9 @@ curl http://localhost:8000/metrics
 ```
 [Example metrics output](https://raw.githubusercontent.com/openvinotoolkit/model_server/v2022.2/docs/metrics_output.out)
 
+## Performance considerations
+Collecting metrics has negligible performance overhead when used with typical models. However when used with very lightweight, fast models which inference time is very short, the metric incrementation can take noticeable proportion of the processing time. Consider it while enabling metrics for such models.
```
mzegla, line 209: Maybe instead of "typical models" we could be more precise? Like replace it with something like:
"models of average size and complexity". It would fit nicely with the next sentence when you mention "very lightweight, fast models"

docs/performance_tuning.md
```diff
@@ -150,6 +150,7 @@ The default value is 1 second which ensures prompt response to creating new mode
   or even disable it. For example, with cloud storage, it could cause a cost for API calls to the storage cloud provider. Detecting new versions 
   can be disabled with a value `0`.
 
+- While collecting metrics has negligible performance overhead when used with typical models, when used with lightweight, fast models, the metric incrementation can consume noticeable proportion of CPU time compared to actual inference. Take it into account while enabled metrics for such models. It applies also for the DEBUG log level which could increase the load for a very high throughout use cases.
```
mzegla, line 153: Here we could also be more precise about "typical models". Maybe we could also mention `metrics_enable` parameter, so each bullet point in this section addresses the use specific parameter? 
I would also split `log_level` and put it as a separate point so it's more visible.

mzegla (approved):

docs/metrics.md
dkalinowski: fixed

docs/performance_tuning.md
dkalinowski: fixed

dtrawins (approved):

PR #1687, opened 2023-01-27
CVS-102160 Fix incorrect mapping_config.json behavior during reload

dkalinowski:
There was issue when model is reloaded or added back to config.json. mapping_config.json file was not applied for first version in set of version to load due to wrong path to mapping_config.json.


rasapala (approved):
LGTM

dtrawins (approved):

PR #1696, opened 2023-02-09
Add bytes type to tensorinfo

michalkulakowski:


src/tensorinfo.hpp
```diff
@@ -38,6 +38,9 @@ class TensorInfo;
 
 using tensor_map_t = std::map<std::string, std::shared_ptr<TensorInfo>>;
 
+enum BytesType { encoded,
```
dkalinowski, line 41: Use `enum class`. This is cpp way of declaring enums, this way you do not bloat ovms namespace with raw/encoded, and you can use `ovms::BytesType::encoded` instead of `ovms::encoded`.

dkalinowski: We use uppercase convention for enums:
https://github.com/openvinotoolkit/model_server/blob/main/src/pipelinedefinitionstatus.hpp#L30

src/tensorinfo.cpp
```diff
@@ -30,42 +30,38 @@ namespace ovms {
...
     const Layout& layout) :
-    name(name),
-    mapping(""),
-    precision(precision),
-    shape(shape),
-    layout(layout) {
+    TensorInfo(name, "", precision, shape, layout) {}
+
+template <class T>
+static BytesType getBytesTypeFromShape(std::vector<T> shape) {
```
dkalinowski, line 59: Each `getBytesTypeFromShape` call creates copy of `shape`. Use `const std::vector<T>& shape` to avoid that

PR #1709, opened 2023-02-21
Fixed OVMS respond with 500 after get status with very long model version

bstrzele:


src/http_rest_api_handler.cpp
```diff
@@ -123,6 +123,8 @@ Status HttpRestApiHandler::parseModelVersion(std::string& model_version_str, std
     if (!model_version_str.empty()) {
         try {
             model_version = std::stoll(model_version_str.c_str());
+        } catch (std::out_of_range const& ex) {
```
dtrawins, line 126: can we add a unit test for that scenario?

dkalinowski (approved):

PR #1710, opened 2023-02-21
Fixed OVMS respond with 500 after GET with model starting with '..'

bstrzele:


src/filesystem.hpp
```diff
@@ -142,7 +143,7 @@ class FileSystem {
     }
 
     static bool isPathEscaped(const std::string& path) {
-        return std::string::npos != path.find("../") || std::string::npos != path.find("/..");
+        return std::regex_match(path, std::regex("(/../)|(^../)|(/..$)"));
```
dtrawins, line 146: can we add unit tests?

dtrawins (approved):

rasapala (commented):
Please add unit test for the problematic  scanario.

rasapala (approved):

PR #1712, opened 2023-02-21
Updated bert output

bstrzele:


dtrawins (approved):

bstrzele:
Caused by update in wiki https://en.wikipedia.org/wiki/BERT_(language_model)

PR #1717, opened 2023-02-22
[CVS-103137, CVS-103142] Add string inputs support in kserve grpc

michalkulakowski:


dkalinowski:
Missing ticket id in PR name

src/binaryutils.cpp
```diff
@@ -461,6 +461,35 @@ static Status convertNativeFileFormatRequestTensorToOVTensor(const TensorType& s
     return StatusCode::OK;
 }
 
+Status convertStringTensorToOVTensor(
```
dkalinowski, line 464:
```suggestion
Status convertStringProtoToOVTensor(
```
What do you think?

src/binaryutils.cpp
```diff
@@ -461,6 +461,35 @@ static Status convertNativeFileFormatRequestTensorToOVTensor(const TensorType& s
...
+    const KFSTensorInputProto& src,
+    ov::Tensor& tensor,
+    const std::shared_ptr<TensorInfo>& tensorInfo,
+    const std::string* buffer) {
+    if (nullptr != buffer) {
+        SPDLOG_ERROR("STRING input should be located in bytes_contents field.");
```
dkalinowski, line 470:
```suggestion
        SPDLOG_DEBUG("STRING input should be located in bytes_contents field.");
```
We should not bloat console with user level errors

src/binaryutils.cpp
```diff
@@ -461,6 +461,35 @@ static Status convertNativeFileFormatRequestTensorToOVTensor(const TensorType& s
...
+        i++;
+    }
+
+    return StatusCode::OK;
+}
+
 template Status convertNativeFileFormatRequestTensorToOVTensor<tensorflow::TensorProto>(const tensorflow::TensorProto& src, ov::Tensor& tensor, const std::shared_ptr<TensorInfo>& tensorInfo, const std::string* buffer);
 template Status convertNativeFileFormatRequestTensorToOVTensor<::KFSRequest::InferInputTensor>(const ::KFSRequest::InferInputTensor& src, ov::Tensor& tensor, const std::shared_ptr<TensorInfo>& tensorInfo, const std::string* buffer);
+
+// template Status convertStringTensorToOVTensor<tensorflow::TensorProto>(const tensorflow::TensorProto& src, ov::Tensor& tensor, const TensorInfo& tensorInfo, const std::string* buffer);
```
dkalinowski, line 493: This will not be needed since convertStringTensorToOVTensor is not templatized

src/binaryutils.hpp
```diff
@@ -20,8 +20,14 @@
 
 #include "tensorinfo.hpp"
 
+namespace inference {
+class ModelInferRequest_InferInputTensor;
```
dkalinowski, line 24: The type is `inference::ModelInferRequest::InferInputTensor;`
therefore it should be
```
namespace inference {
  namespace ModelInferRequest {
    class InferInputTensor;
  }
```

src/predict_request_validation_utils.cpp
```diff
@@ -130,7 +130,10 @@ class RequestValidator {
     Status validateTensorContent(const InputTensorType& proto, ovms::Precision expectedPrecision, size_t bufferId) const;
     Status validateNumberOfShapeDimensions(const ovms::TensorInfo& inputInfo, const InputTensorType& proto) const;
     Status validatePrecision(const ovms::TensorInfo& inputInfo, const InputTensorType& proto) const;
-    bool checkIfNativeFileFormatUsed(const InputTensorType& proto, const std::string inputName) const;
+    bool checkIfNativeFileFormatUsed(const InputTensorType& proto, const std::string inputName, std::shared_ptr<TensorInfo> inputInfo) const;
```
dkalinowski, line 135: We can avoid copying shared_ptr by accepting `TensorInfo&` and passing dereference inputInfo to the function. You can look up my PR.

src/predict_request_validation_utils.cpp
```diff
@@ -483,6 +486,72 @@ Status RequestValidator<RequestType, InputTensorType, IteratorType, ShapeType>::
...
+}
+
+template <typename RequestType, typename InputTensorType, typename IteratorType, typename ShapeType>
+Status RequestValidator<RequestType, InputTensorType, IteratorType, ShapeType>::checkStringShapeMismatch(const InputTensorType& proto, const ovms::TensorInfo& inputInfo, Status& finalStatus, Mode batchingMode, Mode shapeMode) const {
+    const auto& shape = inputInfo.getShape();
+    bool mismatch = false;
+    size_t width = getStringInputWidth(proto);
+    size_t batchSize = getStringBatchSize(proto);
+    if (batchingMode == AUTO) {  // Skip batch dimension
+        if (!shape[1].match(static_cast<dimension_value_t>(width))) {
```
dkalinowski, line 530: I see that we only run `checkStringShapeMismatch` if processing hint is STRING, therefore we are sure **now** that shape[1] is not out of bounds.
I'm afraid that we would change the processing hint in the future to accept single dimension as well, making it possible to crash the service here.

My suggestion is to check for number of dimensions before working with dimension indices

src/predict_request_validation_utils.cpp
```diff
@@ -483,6 +486,72 @@ Status RequestValidator<RequestType, InputTensorType, IteratorType, ShapeType>::
...
+            mismatch = true;
+        }
+    } else {  // Do not skip batch dimension
+        if ((!shape[0].match(static_cast<dimension_value_t>(batchSize))) || (!shape[1].match(static_cast<dimension_value_t>(width)))) {
```
dkalinowski, line 534: It would be cleaner to match entire shape. There should be matchers available.
Something like:
```
if (!shape.match(Shape{batchSize, width})) {
...
}
```

src/predict_request_validation_utils.cpp
```diff
@@ -483,6 +486,72 @@ Status RequestValidator<RequestType, InputTensorType, IteratorType, ShapeType>::
...
+            mismatch = true;
+        }
+    }
+    if (!mismatch) {
+        return StatusCode::OK;
```
dkalinowski, line 539: This will not work and simple shape=auto + string tests should show that.
Scenario: shape=auto, model shape [2, -1], request string with shape [3] ["A","B","C"]

`mismatch` will be false, and we will return StatusCode::OK instead of BATCHSIZE CHANGE REQUIRED

src/predict_request_validation_utils.cpp
```diff
@@ -864,6 +976,21 @@ Status RequestValidator<RequestType, InputTensorType, IteratorType, ShapeType>::
             continue;
         }
 
+        if (checkIfStringFormatUsed(proto, name, inputInfo)) {
```
dkalinowski, line 981: In deserialization the order is to first check if this is string processing **then** if this is nativefile format processing. We should have the same logic in validation

src/tfs_frontend/tfs_utils.cpp
```diff
@@ -147,4 +148,8 @@ bool isNativeFileFormatUsed(const TFSInputTensorType& proto) {
     return proto.dtype() == TFSDataType::DT_STRING;
     // return request.string_val_size() > 0;
 }
+
+bool isStringFormatUsed(const TFSInputTensorType& proto, const TensorInfo& tensorInfo) {
+    return proto.dtype() == tensorflow::DataType::DT_STRING && tensorInfo.getPrecision() == ovms::Precision::U8 && tensorInfo.getShape().size() == 2;
```
dkalinowski, line 153: It should probably be
```
return proto.dtype() == tensorflow::DataType::DT_STRING && tensorInfo.getProcessingHint() == TensorInfo::ProcessingHint::STRING;
```

dkalinowski (approved):

PR #1728, opened 2023-03-01
Mkulakow/grpc string inputs support uts

michalkulakowski:


src/test/predict_validation_test.cpp
```diff
@@ -1328,4 +1328,123 @@ INSTANTIATE_TEST_SUITE_P(
...
+}
+
+template <typename TensorType>
+class PredictValidationStringTest : public ::testing::Test {
+protected:
+    TensorType request;
+    const char* tensorName = DUMMY_MODEL_INPUT_NAME;
+    ovms::tensor_map_t mockedInputsInfo;
+    void SetUp() override {
+        mockedInputsInfo[tensorName] = std::make_shared<ovms::TensorInfo>(tensorName, ovms::Precision::U8, ovms::Shape{-1, -1}, ovms::Layout{"NC"});
```
dkalinowski, line 1404: Missing cases for static shape, partly static and ranges

src/test/tensor_conversion_test.cpp
```diff
@@ -15,19 +15,20 @@
 //*****************************************************************************
 
 #include <fstream>
+#include <iostream>
```
dkalinowski, line 18: Why do we need iostream?

src/test/tensor_conversion_test.cpp
```diff
@@ -626,20 +627,105 @@ TEST_F(BinaryUtilsTestKFSRawInputsContents, Positive) {
...
+    size_t maxStringLength = 0;
+    for (const auto& input : inputStrings) {
+        maxStringLength = std::max(maxStringLength, input.length());
+    }
+    size_t width = maxStringLength + 1;
+    size_t i = 0;
+    ASSERT_EQ(width * inputStrings.size(), tensor.get_size());
+    for (const auto& input : inputStrings) {
+        for (size_t j = 0; j < input.size(); j++) {
+            ASSERT_EQ(tensor.data<uint8_t>()[i * width + j], reinterpret_cast<const uint8_t*>(input.data())[j]);
```
dkalinowski, line 682: You can display proper error by using stream operator:

```
ASSERT_EQ(2, 1) << "why";
```

You can use it to indicate what is exactly wrong, which `i` and `j` position cause the error

src/test/tensor_conversion_test.cpp
```diff
@@ -626,20 +627,105 @@ TEST_F(BinaryUtilsTestKFSRawInputsContents, Positive) {
...
+TYPED_TEST_SUITE(StringInputsConversionTest, MyTypes);
+
+static void assertOutputTensorMatchExpectations(const ov::Tensor& tensor, std::vector<std::string> inputStrings) {
+    size_t maxStringLength = 0;
+    for (const auto& input : inputStrings) {
+        maxStringLength = std::max(maxStringLength, input.length());
+    }
+    size_t width = maxStringLength + 1;
+    size_t i = 0;
+    ASSERT_EQ(width * inputStrings.size(), tensor.get_size());
```
dkalinowski, line 679: You can also validate the shape here?

src/test/tensor_conversion_test.cpp
```diff
@@ -626,20 +627,105 @@ TEST_F(BinaryUtilsTestKFSRawInputsContents, Positive) {
...
+        for (size_t j = 0; j < input.size(); j++) {
+            ASSERT_EQ(tensor.data<uint8_t>()[i * width + j], reinterpret_cast<const uint8_t*>(input.data())[j]);
+        }
+        ASSERT_EQ(tensor.data<uint8_t>()[i * width + input.size()], 0);
+        i++;
+    }
+}
+
+TYPED_TEST(StringInputsConversionTest, positive) {
+    std::vector<std::string> inputStrings = {"String_123"};
```
dkalinowski, line 690: You can rename it to `expectedStrings` to avoid confusion

src/test/tensor_conversion_test.cpp
```diff
@@ -626,20 +627,105 @@ TEST_F(BinaryUtilsTestKFSRawInputsContents, Positive) {
...
+        }
+        tensor.mutable_shape()->Add(inputStrings.size());
+        tensor.set_datatype("BYTES");
+    }
+};
+
+using MyTypes = ::testing::Types<tensorflow::TensorProto, ::KFSRequest::InferInputTensor>;
+TYPED_TEST_SUITE(StringInputsConversionTest, MyTypes);
+
+static void assertOutputTensorMatchExpectations(const ov::Tensor& tensor, std::vector<std::string> inputStrings) {
```
dkalinowski, line 672: Could you add comment here noting that vector of string is expanded into 2D U8 array and assertion checks for correctness against that?

src/test/tensor_conversion_test.cpp
dkalinowski, line 679: We could assert shape of the generated ov vector as well

dkalinowski: Should be (inputStrings.size(), width)

src/test/tensor_conversion_test.cpp
```diff
@@ -626,20 +627,105 @@ TEST_F(BinaryUtilsTestKFSRawInputsContents, Positive) {
...
+TYPED_TEST(StringInputsConversionTest, positive_batch_size_3_one_string_empty) {
+    std::vector<std::string> inputStrings = {"String_123", "zebra", ""};
+    this->prepareStringTensor(this->requestTensor, inputStrings);
+    ov::Tensor tensor;
+    ASSERT_EQ(convertStringRequestTensorToOVTensor(this->requestTensor, tensor, nullptr), ovms::StatusCode::OK);
+    assertOutputTensorMatchExpectations(tensor, inputStrings);
+}
+
+TYPED_TEST(StringInputsConversionTest, positive_empty_inputs) {
+    std::vector<std::string> inputStrings = {};
```
dkalinowski, line 714: What is the ov::Tensor shape in this case and how is it positive?

src/test/prediction_service_test.cpp
```diff
@@ -1572,4 +1572,21 @@ TYPED_TEST(TestPredict, InferenceWithNegativeShapeDynamicParameter) {
...
+    prepareInferStringRequest(request, PASSTHROUGH_MODEL_INPUT_NAME, inputStrings);
+    ovms::ModelConfig config = PASSTHROUGH_MODEL_CONFIG;
+    config.setBatchingParams("auto");
+    ASSERT_EQ(this->manager.reloadModelWithVersions(config), ovms::StatusCode::OK_RELOADED);
+    std::shared_ptr<ovms::ModelInstance> modelInstance;
+    std::unique_ptr<ovms::ModelInstanceUnloadGuard> modelInstanceUnloadGuard;
+    ASSERT_EQ(this->manager.getModelInstance(config.getName(), config.getVersion(), modelInstance, modelInstanceUnloadGuard), ovms::StatusCode::OK);
+    typename TypeParam::second_type response;
+    ASSERT_EQ(modelInstance->infer(&request, &response, modelInstanceUnloadGuard), ovms::StatusCode::OK);
+    this->checkOutputShape(response, {2,11}, PASSTHROUGH_MODEL_OUTPUT_NAME);
```
dkalinowski, line 1589: Can we assert proto data value as well?

src/test/prediction_service_test.cpp
```diff
@@ -1572,4 +1572,21 @@ TYPED_TEST(TestPredict, InferenceWithNegativeShapeDynamicParameter) {
...
 }
 
+TYPED_TEST(TestPredict, InferenceWithStringInputs_positive) {
+    if (typeid(typename TypeParam::first_type) == typeid(ovms::InferenceRequest))
+        GTEST_SKIP() << "String inputs not supported for C-API";
+    typename TypeParam::first_type request;
+    std::vector<std::string> inputStrings = {"String_123", "String"};
+    prepareInferStringRequest(request, PASSTHROUGH_MODEL_INPUT_NAME, inputStrings);
+    ovms::ModelConfig config = PASSTHROUGH_MODEL_CONFIG;
+    config.setBatchingParams("auto");
```
dkalinowski, line 1582: We can perform the same test for shape `"(-1, -1)"`, static `"(2, 11)"` and range `"(1:3,9:15)"`.
We can perform negative test for static `"(3,12)"` and range `"(3:6,20:30)"`

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -5259,3 +5259,137 @@ TEST_F(EnsembleFlowCustomNodePipelineExecutionTest, ReloadPipelineWithoutNodeDei
...
+    const std::string customNodeName = "passthrough";
+    static constexpr const char* pipelineInputName = "pipeline_input";
+    const std::string pipelineOutputName = "pipeline_output";
+};
+
+using MyTypes = ::testing::Types<TFSInterface, KFSInterface>;
+TYPED_TEST_SUITE(EnsembleFlowStringInput, MyTypes);
+
+TYPED_TEST(EnsembleFlowStringInput, positive) {
+    // Most basic configuration, just process single add-sub custom node pipeline request
```
dkalinowski, line 5364: Comment does not match, there is no add-sub custom node in this dag

src/test/ensemble_flow_custom_node_tests.cpp
```diff
@@ -5259,3 +5259,137 @@ TEST_F(EnsembleFlowCustomNodePipelineExecutionTest, ReloadPipelineWithoutNodeDei
     // in order to count whether deinitialize has been called expected number of times
     ASSERT_EQ(LibraryCountDeinitialize::deinitializeCounter, 3);
 }
+
+
+static constexpr const char* INPUT_TENSOR_NAME = "input_string";
+static constexpr const char* OUTPUT_TENSOR_NAME = "output_string";
+
+struct Passthrough {
```
dkalinowski, line 5267: Rename to indicate what is the difference from other mocked nodes
- precision u8
- 2 dimensions

My suggestion is `Passthrough_2D_U8`

dkalinowski (approved):

PR #1742, opened 2023-03-12
Updates for CUDA build on UBI

Xaenalt:
Here's the cumulative updates, I added in a few more bits like using CentOS repos for the stuff not in UBI if entitlement isn't enabled


Xaenalt:
I'll need to rebase it onto main, but here's where we're at @dtrawins thanks for the help on this

dtrawins:
@Xaenalt my observation is that without the entitlement and subscription it is not possible to install via yum the package cuda-11-8:
`requires cuda-runtime-11-8 >= 11.8.0, but none of the providers can be installed`
It seems to my that adding the entitlement is the only way to successfully build the plugin. Did I miss some step?

Xaenalt:
Hmmm, due to the build system, we're currently using this without entitlements, and cuda-runtime should be in the cuda repos. Let me dig into it

dtrawins:
@Xaenalt here are my steps to reproduce: on branch https://github.com/openvinotoolkit/model_server/tree/cuda-ubi-1 I'm building on ubuntu20 host with a command: make docker_build BASE_OS=redhat NVIDIA=1 OV_USE_BINARY=0
It fails on step `yum install -y cuda-11-8`. When the entitlement is mounted, there is no problem with installation. Without it, there is an error:
```
+ yum install -y cuda-11-8
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 0:04:03 ago on Wed Mar 29 14:41:48 2023.
Error:
 Problem: package cuda-11-8-11.8.0-1.x86_64 requires cuda-runtime-11-8 >= 11.8.0, but none of the providers can be installed
  - package cuda-runtime-11-8-11.8.0-1.x86_64 requires cuda-drivers >= 520.61.05, but none of the providers can be installed
  - package cuda-drivers-530.30.02-1.x86_64 requires nvidia-settings >= 3:530.30.02, but none of the providers can be installed
  - conflicting requests
  - package cuda-drivers-520.61.05-1.x86_64 is filtered out by modular filtering
  - package cuda-drivers-525.60.13-1.x86_64 is filtered out by modular filtering
  - package cuda-drivers-525.85.12-1.x86_64 is filtered out by modular filtering
  - nothing provides libvdpau(x86-64) >= 0.9 needed by nvidia-settings-3:530.30.02-1.el8.x86_64
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

Do you have some idea what could be the culprit?

Xaenalt:
Oof, not off the top of my head, libvdpau should be available, but modular filtering sometimes breaks stuff -_-

Let me put together an environment with ubuntu20 to play with, maybe we're hitting some edge case somewhere

dtrawins:
The changes are incorporated in the main branch. Thank you for your contribution.

PR #1749, opened 2023-03-14
2D U8 String serialization to gRPC TFS/KServe

dkalinowski:
With tensor name suffix _string_2d_u8 hint

### DIFF
--- a/src/deserialization.hpp
+++ b/src/deserialization.hpp
@@ -356,7 +356,7 @@ Status deserializePredictRequest(
             ov::Tensor tensor;
 
             if (requiresProcessing(requestInput)) {
-                switch (tensorInfo->getProcessingHint()) {
+                switch (tensorInfo->getPreProcessingHint()) {
                 case TensorInfo::ProcessingHint::STRING_1D_U8:
                     SPDLOG_DEBUG("Request contains input in 1D string format: {}", name);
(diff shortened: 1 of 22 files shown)


src/tensorinfo.cpp
```diff
@@ -23,9 +23,12 @@
 
 #include "kfs_frontend/kfs_grpc_inference_service.hpp"
 #include "logging.hpp"
+#include "stringutils.hpp"
 
 namespace ovms {
 
+const std::string STRING_SERIALIZATION_HINT_NAME_SUFFIX = "_string_2d_u8";
```
dkalinowski, line 30: change to `_string`

atobiszei (approved):

src/deserialization.hpp
```diff
@@ -356,7 +356,7 @@ Status deserializePredictRequest(
             ov::Tensor tensor;
 
             if (requiresProcessing(requestInput)) {
```
michalkulakowski, line 358: If you changed getProcessingHint to getPreProcessingHint maybe change also requiresProcessing to requiresPreProcessing

michalkulakowski (approved):

src/tensorinfo.cpp
dkalinowski: done

src/deserialization.hpp
dkalinowski: done

PR #1780, opened 2023-04-07
Fix binary extension bug

michalkulakowski:


src/http_rest_api_handler.cpp
```diff
@@ -286,7 +286,7 @@ static size_t calculateBinaryDataSize(::KFSRequest::InferInputTensor& input) {
 }
 
 static Status handleBinaryInputs(::KFSRequest& grpc_request, const std::string& request_body, size_t endOfJson) {
-    const char* binary_inputs = &(request_body[endOfJson]);
+    const char* binary_inputs_buffer = &(request_body[endOfJson]);
```
dkalinowski, line 289: No change in unit tests means we are missing such scenario in unit test

src/test/kfs_rest_test.cpp
```diff
@@ -377,6 +377,45 @@ TEST_F(HttpRestApiHandlerTest, binaryInputsINT8) {
...
+
+    ::KFSRequest grpc_request;
+    int inferenceHeaderContentLength = (request_body.size() - binaryData.size());
+    ASSERT_EQ(HttpRestApiHandler::prepareGrpcRequest(modelName, modelVersion, request_body, grpc_request, inferenceHeaderContentLength), ovms::StatusCode::OK);
+
+    ASSERT_EQ(grpc_request.inputs_size(), 2);
+    ASSERT_EQ(grpc_request.raw_input_contents_size(), 2);
+    ASSERT_EQ(grpc_request.model_name(), modelName);
+    ASSERT_EQ(grpc_request.model_version(), std::to_string(modelVersion.value()));
+    auto params = grpc_request.parameters();
```
dkalinowski, line 393: not used

atobiszei (approved):

PR #1801, opened 2023-04-21
Fix tokenizer node readme

dkalinowski:
JIRA: CVS-107730

### DIFF
--- a/src/custom_nodes/tokenizer/README.md
+++ b/src/custom_nodes/tokenizer/README.md
@@ -23,17 +23,24 @@ Known models with such format:
 
 # Building custom node library
 
-You can build the shared library of the custom node simply by running cmake (minimum 3.10):
+You can build the shared library of the custom node simply by running command in the context of this document:
 ```bash
 git clone https://github.com/openvinotoolkit/model_server && cd model_server/src/custom_nodes/tokenizer
-mkdir build && cd build
-cmake .. && make -j`nproc`
+make
 ```
-It will compile the libraries:
-```
-ls src/*.so
+It will compile the library inside a docker container and save the results in `lib/<OS>/` folder.
 
-src/libdetokenizer.so  src/libtokenizer.so
+You can also select base OS between RH 8.5 (redhat) and Ubuntu 20.04 (ubuntu) by setting `BASE_OS` environment variable.
+```bash
+make BASE_OS=redhat
+```
+
+It will compile the libraries:
+```bash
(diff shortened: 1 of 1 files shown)


src/custom_nodes/tokenizer/README.md
```diff
@@ -23,17 +23,24 @@ Known models with such format:
...
+You can also select base OS between RH 8.5 (redhat) and Ubuntu 20.04 (ubuntu) by setting `BASE_OS` environment variable.
+```bash
+make BASE_OS=redhat
 ```
-ls src/*.so
 
-src/libdetokenizer.so  src/libtokenizer.so
+It will compile the libraries:
+```bash
+ls lib/redhat -A1
```
ngrozae, line 40:
```suggestion
ls lib/${BASE_OS} -A1
```

src/custom_nodes/tokenizer/README.md
```diff
@@ -23,17 +23,24 @@ Known models with such format:
 
 # Building custom node library
 
-You can build the shared library of the custom node simply by running cmake (minimum 3.10):
+You can build the shared library of the custom node simply by running command in the context of this document:
```
dtrawins, line 26:
```suggestion
You can build the shared library of the custom node simply by running the following commands:
```

src/custom_nodes/tokenizer/README.md
```diff
@@ -23,17 +23,24 @@ Known models with such format:
...
 ```bash
 git clone https://github.com/openvinotoolkit/model_server && cd model_server/src/custom_nodes/tokenizer
-mkdir build && cd build
-cmake .. && make -j`nproc`
+make
 ```
-It will compile the libraries:
+It will compile the library inside a docker container and save the results in `lib/<OS>/` folder.
+
+You can also select base OS between RH 8.5 (redhat) and Ubuntu 20.04 (ubuntu) by setting `BASE_OS` environment variable.
```
dtrawins, line 33:
```suggestion
You can also select base OS between UBI 8.6 (redhat) and Ubuntu 20.04 (ubuntu) by setting `BASE_OS` environment variable.
```

michalkulakowski (approved):

src/custom_nodes/tokenizer/README.md
dkalinowski: done

dtrawins (approved):

PR #1807, opened 2023-04-24
Revert logic for raw/non-raw placement of data in KServe

bstrzele:


src/test/test_utils.cpp
```diff
@@ -427,10 +427,20 @@ void assertStringResponse(const tensorflow::serving::PredictResponse& proto, con
 void assertStringResponse(const ::KFSResponse& proto, const std::vector<std::string>& expectedStrings, const std::string& outputName) {
     ASSERT_EQ(proto.outputs_size(), 1);
     ASSERT_EQ(proto.outputs(0).name(), outputName);
-    ASSERT_EQ(proto.outputs(0).datatype(), "BYTES");
-    ASSERT_EQ(proto.outputs(0).shape_size(), 1);
+    ASSERT_EQ(proto.outputs(0).datatype(), "UINT8");
```
dkalinowski, line 430: This assert should still check for BYTES datatype but `assertStringOutputProto(KFS)` should be changed to check for correct data in raw field.
Correct raw field meaning each batch is prepended with 4 bytes of unsigned integer which represents its length. Check other fixtures, the assert is probably already implemented

src/serialization.cpp
```diff
@@ -85,6 +86,10 @@ static Status serializePrecision(
     case ovms::Precision::U32:
     case ovms::Precision::U16:
     case ovms::Precision::U8:
+        if (servableOutput->getPostProcessingHint() == TensorInfo::ProcessingHint::STRING_2D_U8) {
```
dkalinowski, line 89: This code will be triggered for all types which is overhead.
Can we just set precision to BYTES before switch if processing hint is specified?

src/serialization.cpp
```diff
@@ -132,6 +137,10 @@ static Status serializeShape(
     ov::Tensor& tensor) {
     OVMS_PROFILE_FUNCTION();
     responseOutput.clear_shape();
+    if (servableOutput->getPostProcessingHint() == TensorInfo::ProcessingHint::STRING_2D_U8) {
+        responseOutput.add_shape(tensor.get_shape()[0]);
+        return StatusCode::OK;
+    }
     auto& effectiveNetworkOutputShape = servableOutput->getShape();
     ov::Shape actualTensorShape = tensor.get_shape();
     if (effectiveNetworkOutputShape.size() != actualTensorShape.size()) {
```
dkalinowski, line 146: This check is important, we should not omit it in case of processing hint 2d_u8. I think this new check should be placed in line 150-151

src/serialization.cpp
```diff
@@ -160,6 +169,24 @@ static void serializeContent(std::string* content, ov::Tensor& tensor) {
...
+static void serializeStringContent(std::string* content, ov::Tensor& tensor) {
+    OVMS_PROFILE_FUNCTION();
+    // We only fill if the content is not already filled.
+    // It can be filled in gather exit node handler.
+    if (content->size() == 0) {
+        std::string data = "";
+        size_t j = 0;
+        for (size_t i = 0; i < tensor.get_shape()[0]; i++, j += tensor.get_shape()[1]) {
+            std::string tmp((char*)tensor.data() + j);
+            int size = tmp.size();
```
dkalinowski, line 181: We have implicit downcast here - size_t (64bit) to int (32bit)

dkalinowski: also signed/unsigned mismatch

src/serialization.cpp
```diff
@@ -160,6 +169,24 @@ static void serializeContent(std::string* content, ov::Tensor& tensor) {
     }
 }
 
+static void serializeStringContent(std::string* content, ov::Tensor& tensor) {
```
dkalinowski, line 172:
```suggestion
static void serializeStringContent(std::string* content, ov::Tensor& tensor) {
    OVMS_PROFILE_FUNCTION();
    // We only fill if the content is not already filled.
    // It can be filled in gather exit node handler.
    if (!content->empty()) {
        return;
    }

    size_t batchSize = tensor.get_shape()[0];
    size_t maxStringLen = tensor.get_shape()[1];
    for (size_t i = 0; i < batchSize; i++) {
        // std::string tmp((char*)tensor.data() + i * tensor.get_shape()[1]);
        uint32_t strLen = strnlen((char*)tensor.data() + i * maxStringLen, maxStringLen);
        content->append(reinterpret_cast<const char*>(&strLen), sizeof(strLen));
        content->append((char*)tensor.data() + i * maxStringLen, strLen);
    }
}
```

This way we can avoid unneccessary string copy

michalkulakowski (approved):

PR #1863, opened 2023-05-19
Fixed demos/horizontal_text_detection/horizontal_text_detection.py fail with mp4 for OCR pipeline part

bstrzele:


src/custom_nodes/horizontal_ocr/horizontal_ocr.cpp
```diff
@@ -198,7 +198,6 @@ int execute(const struct CustomNodeTensor* inputs, int inputsCount, struct Custo
 
     uint64_t _numDetections = boxesTensor->dims[0];
     uint64_t _numItems = boxesTensor->dims[1];
-    NODE_ASSERT(boxesTensor->dims[0] == 100, "boxes has dim 0 not equal to 100");
```
dkalinowski, line 201: Validate for at least 1 box, otherwise we will produce output with batch=0 which is undefined behavior

dkalinowski (approved):

PR #1868, opened 2023-05-22
Check size

rasapala:


dkalinowski (approved):

michalkulakowski (approved):

src/test/localfilesystem_test.cpp
```diff
@@ -166,3 +166,50 @@ TEST(FileSystem, CheckIfPathIsEscaped) {
...
+    ASSERT_TRUE(ovms::FileSystem::isLocalFilesystem("o_O$az://"));
+    ASSERT_TRUE(ovms::FileSystem::isLocalFilesystem("../"));
+    ASSERT_TRUE(ovms::FileSystem::isLocalFilesystem("/localfilesystem"));
+    ASSERT_TRUE(ovms::FileSystem::isLocalFilesystem("/long/local/filesystem"));
+}
+
+TEST(FileSystem, setRootDirectoryPath) {
+    std::string rootPath = "";
+    std::string givenPath = "/givenpath";
+    auto fs = std::make_shared<ovms::LocalFileSystem>();
```
dkalinowski, line 187: We should not be required to create LocalFileSystem class instance just to call `joinPath`. The function does not use any internal field, therefore it should be static.

What's more, joinPath exists in 2 classes - in LocalFileSystem and in AzureStorageAdapter - both methods are identical (and both non static).

src/modelmanager.cpp
```diff
@@ -1311,7 +1311,7 @@ const std::string ModelManager::getFullPath(const std::string& pathToCheck) cons
     if (!FileSystem::isLocalFilesystem(pathToCheck)) {
         // Cloud filesystem
         return pathToCheck;
-    } else if (pathToCheck.at(0) == '/') {
+    } else if (pathToCheck.size() > 0 && pathToCheck.at(0) == '/') {
```
dkalinowski, line 1314: Missing unit test

src/filesystem.hpp
```diff
@@ -177,7 +177,7 @@ class FileSystem {
         if (!FileSystem::isLocalFilesystem(givenPath)) {
             // Cloud filesystem
             path = givenPath;
-        } else if (givenPath.at(0) == '/') {
+        } else if (givenPath.size() > 0 && givenPath.at(0) == '/') {
```
dkalinowski, line 180: Missing test

rasapala: Added

PR #1870, opened 2023-05-22
Cvs 110308 model config api mediapipe reload and status

rasapala:


src/model_service.cpp
```diff
@@ -198,6 +198,20 @@ Status GetModelStatusImpl::getAllModelsStatuses(std::map<std::string, tensorflow
         modelsStatusesTmp.insert({pipelineName, response});
     }
 
+    const std::vector<std::string>& mediapipeNames = manager.getMediapipeFactory().getPipelinesNames();
```
mzegla, line 201: I'd unify naming as I feel like we are mixing a little bit here. The mediapipe servables will be refered to as "pipelines", "graphs" or something else? Once we make a choice I'd go with for example:
-  mediapipePipelinesNames = manager.getMediapipeFactory().getPipelinesNames();
- mediapipeGraphsNames = manager.getMediapipeFactory().getGraphsNames();

or something else, but let's not mix terms.

mzegla (commented):
Can we also have a test for unloading the model by removing it from the configuration?

src/model_service.cpp
rasapala: The pipeline functionality is implemented with mediapipe solution. The DAG pipeline is another implementation of the pipieline so I would stay with the name.

src/test/http_rest_api_handler_test.cpp
```diff
@@ -625,6 +626,82 @@ TEST_F(ConfigReload, StartWith1DummyThenReloadToAddPipeline) {
     EXPECT_EQ(status, ovms::StatusCode::OK_RELOADED);
 }
 
+TEST_F(ConfigReload, StartWith1DummyThenReloadToMediapipe) {
+    ovms::Server& ovmsServer = ovms::Server::instance();
+    TestHelper1 t(*this, configWith1Dummy);
+    auto handler = ovms::HttpRestApiHandler(ovmsServer, 10);
+
+    std::this_thread::sleep_for(std::chrono::seconds(1));
```
jasiu86, line 634: What we expect to happen in  1 sec? Some kind of initial load ?
What will happen if machine will be overloaded and 1sec will be not enough ?

src/model_service.cpp
```diff
@@ -198,6 +198,20 @@ Status GetModelStatusImpl::getAllModelsStatuses(std::map<std::string, tensorflow
...
 
+    const std::vector<std::string>& mediapipePipelineNames = manager.getMediapipeFactory().getMediapipePipelinesNames();
+    for (auto const& mediapipePipelineName : mediapipePipelineNames) {
+        std::optional<int64_t> noValueModelVersion;
+        tensorflow::serving::GetModelStatusRequest request;
+        GetModelStatusImpl::createGrpcRequest(mediapipePipelineName, noValueModelVersion, &request);
+        tensorflow::serving::GetModelStatusResponse response;
+        auto status = GetModelStatusImpl::getModelStatus(&request, &response, manager, context);
+        if (status != StatusCode::OK) {
+            // Same situation like with models.
```
jasiu86, line 209: Shouldn't we do something? 
Is it expected ?

jasiu86 (approved):

src/model_service.cpp
rasapala: Expected.

src/test/http_rest_api_handler_test.cpp
rasapala: Standard timeout for models configs reload used in tests.

PR #1846, opened 2023-05-11
[WIP] Sdl fixes

dkalinowski:

PR #1924, opened 2023-06-20
Mediapipe with two outputs test

michalkulakowski:


src/test/mediapipeflow_test.cpp
```diff
@@ -237,6 +242,55 @@ TEST_F(MediapipeFlowDummySubconfigTest, Infer) {
...
+    ::KFSRequest request;
+    ::KFSResponse response;
+    const std::string modelName = "mediapipeDummyTwoOutputs";
+    performMediapipeInfer(server, request, response, precision, modelName);
+
+    std::vector<float> requestData{0., 0., 0, 0., 0., 0., 0., 0, 0., 0.};
+    ASSERT_EQ(response.model_name(), modelName);
+    ASSERT_EQ(response.outputs_size(), 2);
+    ASSERT_EQ(response.raw_output_contents_size(), 2);
+    ASSERT_EQ(response.outputs().begin()->name(), "out_2");
```
dkalinowski, line 262: Are we sure about the order of outputs? Why is `out_2` first?

dkalinowski (approved):

atobiszei (commented):
I don't think this test case covers this fix:
https://github.com/openvinotoolkit/model_server/pull/1846/files#diff-4970675caf5aa4de72a9ded3d0444a2f4088fe2c1eff686cc31769b798f4398cL153

atobiszei (approved):

PR #1932, opened 2023-06-26
RTSP client demo

bstrzele:


demos/horizontal_text_detection/python/README.md
```diff
@@ -182,3 +182,41 @@ ThreadID:   2; Current FPS:    29.41; Average FPS:    25.70; Average latency:
...
+options:
+  -h, --help            show this help message and exit
+  --grpc_address GRPC_ADDRESS
+                        Specify url to grpc service
+  --grpc_port GRPC_PORT
+                        Specify port to grpc service
+  --input_stream INPUT_STREAM
+                        Url of input rtsp stream
+  --output_stream OUTPUT_STREAM
+                        Url of output rtsp stream
```
mzegla, line 213: I think we should mention in the readme that having such two streams is a prerequisite and is not part of the app - as in the application we read from one stream and write to another.

demos/horizontal_text_detection/python/rtsp_client.py
```diff
@@ -0,0 +1,133 @@
...
+parser.add_argument('--output_stream', required=False, default="rtsp://localhost:8080/channel2", type=str, help='Url of output rtsp stream')
+parser.add_argument('--fps', required=False, default=30, type=int, help='Framerate of input stream')
+args = parser.parse_args()
+output_stream = args.output_stream
+
+address = "{}:{}".format(args.grpc_address, args.grpc_port)
+triton_client = grpcclient.InferenceServerClient( url=address, verbose=False)
+
+WIDTH = 704
+HEIGHT = 704
```
mzegla, line 38: Don't we want to make it input parameters instead of const in the script?

demos/horizontal_text_detection/python/rtsp_client.py
```diff
@@ -0,0 +1,133 @@
...
+cap = cv2.VideoCapture(args.input_stream, cv2.CAP_FFMPEG)
+force_exit = False
+
+def decode(text):
+    word = ''
+    last_character = None
+    for character in text:
+        if character == last_character:
+            continue
+        elif character == '#':
```
mzegla, line 48: What kind of special character is that? End of the word?

demos/horizontal_text_detection/python/rtsp_client.py
```diff
@@ -0,0 +1,133 @@
...
+            last_character = character
+            word += character
+    return word
+
+def get_text(output):
+    alphabet = '#1234567890abcdefghijklmnopqrstuvwxyz'
+    preds = output.argmax(2)
+    word = ''
+    for i in range(preds.shape[0]):
+        word += alphabet[preds[i,0]]
```
mzegla, line 60: Maybe we can rephrase that to for each i.e.
```
for pred in preds:
    word += alphabet[pred[0]]
```

Also I think adding a comment at the beginning of this function with information what kind of structure you expect as the output argument would be nice.

demos/horizontal_text_detection/python/rtsp_client.py
```diff
@@ -0,0 +1,133 @@
...
+        finish()
+        return None
+
+    # crop and resize if original image is too small or too big for the model
+    if frame.shape[0] > HEIGHT and frame.shape[1] > WIDTH:
+        frame = frame[0:HEIGHT, 0:WIDTH]
+    else:
+        res = min(frame.shape[0], frame.shape[1])
+        frame = frame[0:res, 0:res]
+        frame = cv2.resize(frame, (HEIGHT, WIDTH), interpolation=cv2.INTER_AREA)
```
mzegla, line 80: I don't like this whole block. I think we shouldn't crop at all and only resize the frames that come in from the input stream to the size required by the model. That size should be specified by HEIGHT,WIDTH and we should accept that those two values can be different (not all models work with N x N sizes). 

We could also check if input frame requires any processing and leave if not (and encourage users to provide data that matches the model as resizing most likely will impact accuracy)

demos/horizontal_text_detection/python/rtsp_client.py
```diff
@@ -0,0 +1,133 @@
...
+    return frame
+
+def postprocess(frame, result):
```
mzegla, line 83: Can you put result structure in the comment so it's easier to follow?

demos/horizontal_text_detection/python/rtsp_client.py
```diff
@@ -0,0 +1,133 @@
...
+    if result is not None:
+        texts = result.as_numpy("texts")
+        text_coordinates = result.as_numpy('text_coordinates')
+        for i in range(len(texts)):
+            text = get_text(texts[i])
+            x_min = text_coordinates[i][0][0]
```
mzegla, line 89: This `text_coordinates[i][0]` could be a variable

demos/horizontal_text_detection/python/rtsp_client.py
bstrzele: # - special blank character for CTC decoding algorithm.

demos/horizontal_text_detection/python/rtsp_client.py
```diff
@@ -0,0 +1,136 @@
...
+    success, frame = cap.read()
+    if not success:
+        print("[WARNING] No Input frame")
+        finish()
+        return None
+
+    if frame.shape[0] > HEIGHT and frame.shape[1] > WIDTH:
+        frame = cv2.resize(frame, (HEIGHT, WIDTH), interpolation=cv2.INTER_LINEAR)
+    else:
+        frame = cv2.resize(frame, (HEIGHT, WIDTH), interpolation=cv2.INTER_AREA)
```
mzegla, line 80: I think it should be reversed - INTER_LINEAR when we scale the frame up (it's smaller than HEIGHT and WIDTH) and INTER_AREA when we scale down.

demos/horizontal_text_detection/python/rtsp_client.py
```diff
@@ -0,0 +1,136 @@
...
+        f"-f rtsp {output_stream}"
+    ).split()
+    return subprocess.Popen(args, stdin=subprocess.PIPE) #nosec
+
+ffmpeg_process = open_ffmpeg_stream_process()
+
+if grab_frame(cap) is None:
+    force_exit = True
+while not force_exit:
+    time.sleep(1/fps)
```
mzegla, line 122: Do we still need it?

dkalinowski (approved):
LGTM

demos/horizontal_text_detection/python/README.md
```diff
@@ -182,3 +182,54 @@ ThreadID:   2; Current FPS:    29.41; Average FPS:    25.70; Average latency:
...
+  --model_name MODEL_NAME
+                        Name of the model
+  --width WIDTH         Width of model's input image
+  --height HEIGHT       Height of model's input image
+  --verbose VERBOSE     Should client dump debug information
+  --input_name INPUT_NAME
+                        Name of the model's input
+```
+
+- Usage example
```
dtrawins, line 225: it would be helpful to add instruction how to create this input stream and how to display the output stream

demos/horizontal_text_detection/python/README.md
```diff
@@ -182,3 +182,54 @@ ThreadID:   2; Current FPS:    29.41; Average FPS:    25.70; Average latency:
...
+
+```bash
+docker run rtsp_client --grpc_address localhost:9000 --input_stream 'rtsp://localhost:8080/channel1' --output_stream 'rtsp://localhost:8080/channel2'
+```
+
+One might as well use prerecorded video and schedule it for inference.
+
+```bash
```
dtrawins, line 233: for testing purposes it would be helpful to include expected output

demos/common/python/stream_client.py
```diff
@@ -0,0 +1,130 @@
...
+import cv2
+from functools import partial
+import queue
+import threading
+import tritonclient.grpc as grpcclient
+import subprocess #nosec
+from enum import Enum
+import numpy as np
+
+class StreamClient:
```
dtrawins, line 26: this class might be reused in many examples but it doesn't have any documentation or even comments

dtrawins (commented):
rebase with main to fix image building errors

PR #1942, opened 2023-06-30
Added C-API Server Status (Live/Ready)

bstrzele:


atobiszei (approved):

src/test/c_api_tests.cpp
```diff
@@ -805,13 +805,17 @@ class CAPIState : public ::testing::Test {
     };
     class MockServer : public Server {
     public:
-        MockServer() {
+        void setReady() {
+            msmm = new MockServableManagerModule(*this);
```
dkalinowski, line 809: Memory leak

src/test/c_api_tests.cpp
```diff
@@ -824,9 +828,35 @@ class CAPIState : public ::testing::Test {
     };
 };
 
+TEST_F(CAPIState, ServerLive) {
```
dkalinowski, line 831: We should have a test checking that server is not live and not ready before calling OVMS_StartFromConfigurationFile, and that it is in fact live and ready after the call. We could also check what happens in case of corrupted config file.

src/test/c_api_tests.cpp
bstrzele: destruction managed by unique_ptr one line later

dkalinowski (approved):

PR #1974, opened 2023-08-02
report error when invalid specific version policy

dtrawins:


src/modelconfig.cpp
```diff
@@ -252,6 +252,11 @@ Status ModelConfig::parseModelVersionPolicy(std::string command) {
         }
         m = specific.FindMember("versions");
         if (m == specific.MemberEnd()) {
+            SPDLOG_WARN("Model policy is invalid. 'specific' policy should include 'versions' item with a list of number as a value");
+            return StatusCode::MODEL_VERSION_POLICY_WRONG_FORMAT;
+        }
+        if (!m->value.IsArray()) {
```
dkalinowski, line 258: Add unit test

src/modelconfig.cpp
```diff
@@ -252,6 +252,11 @@ Status ModelConfig::parseModelVersionPolicy(std::string command) {
...
+            SPDLOG_WARN("Model policy is invalid. 'versions' item should have a list of numbers as a value, for example [1,2]");
```
dkalinowski, line 259: Maybe it would be better to add validation on schema level?

dkalinowski (approved):

mzegla (approved):

PR #1984, opened 2023-08-16
Accept requests with 0-value dimension in shape, allow server configurations with 0 dimension (part1 - single model / grpc)

dkalinowski:
+ unit tests

Missing REST endpoints (both tfs and kserve)
Missing support in DAG
Missing support in MediaPipe


src/test/prediction_service_test.cpp
```diff
@@ -880,6 +880,112 @@ TYPED_TEST(TestPredict, SuccesfullInferenceOnModelWithScalar) {
     this->checkOutputShape(response, {}, SCALAR_MODEL_OUTPUT_NAME);
 }
 
+TYPED_TEST(TestPredict, Succesfull0DimInferenceOnModelWithDynamicDim) {
+    ovms::ModelConfig config = DUMMY_MODEL_CONFIG;
+    config.setBatchingParams("");
+    config.parseShapeParameter("(1,-1)");
```
dkalinowski, line 886: Test `(-1,1)`

dtrawins (approved):

src/test/prediction_service_test.cpp
dkalinowski: Added to next task

michalkulakowski (approved):

PR #1989, opened 2023-08-22
Introducing Python in model server build system

mzegla:


rasapala (commented):
Add default bazel flags in .bazelrc

rasapala (approved):
LGTM

install_python_ubuntu.sh
```diff
@@ -0,0 +1,38 @@
...
+        --enable-loadable-sqlite-extensions \
+		--enable-optimizations \
+		--enable-option-checking=fatal \
+		--enable-shared \
+		--with-lto \
+		--with-system-expat
+make -j ${nproc} 
+make install 
+ldconfig
+cd .. && rm -rf Python-3.${PYTHON3_VERSION_MAJOR}.${PYTHON3_VERSION_MINOR}.tgz Python-3.${PYTHON3_VERSION_MAJOR}.${PYTHON3_VERSION_MINOR}
```
jasiu86, line 38: If any error occures previoulsy, we will not delete tar.gz nor Python3. dir.

Please take a peak into this article:

https://stackoverflow.com/questions/15656492/writing-try-catch-finally-in-shell

Dockerfile.ubuntu
```diff
@@ -371,6 +367,11 @@ RUN python3 -c "import json; m={'PROJECT_VERSION':'${PROJECT_VERSION}','OPENVINO
     'BAZEL_BUILD_FLAGS':'${debug_bazel_flags}${minitrace_flags}','BASE_IMAGE':'${BASE_IMAGE}' }; \
     print(json.dumps(m,indent=4))" > /ovms/release_files/metadata.json
 
+## Installing utils for debugging
```
jasiu86, line 370: Do we need it in official image?

install_python_ubuntu.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
jasiu86, line 1:
```suggestion
#!/bin/bash -e
```

Then you can get rid of '&&' operators

rasapala (approved):
LGTM

Makefile
```diff
@@ -51,6 +51,7 @@ NVIDIA ?=0
 GPU ?= 0
 BUILD_NGINX ?= 0
 MEDIAPIPE_DISABLE ?= 0
+PYTHON_DISABLE ?= 1
```
dkalinowski, line 54: Should we check for incorrect setting such as:
`PYTHON_DISABLE=0 MEDIAPIPE_DISABLE=1`
?

mzegla: I think we could skip such check for now. I'd like to discuss if we even want to have such flag. Perhaps we could always build with Python enabled.

dkalinowski (approved):

PR #2008, opened 2023-09-04
Mkulakow/cvs 112046

michalkulakowski:


rasapala (approved):
LGTM

dtrawins (changes requested):
this change is breaking the functionality when OV is built from src because of incorrect tbb version in the release image

dtrawins (approved):

PR #2123, opened 2023-11-03
Added a test to validate OVMS client with all supported python version

bstrzele:


client/python/ovmsclient/samples/requirements.txt
```diff
@@ -1,2 +1,2 @@
 ovmsclient
-opencv-python==4.5.3.56
```
dtrawins, line 2: that should be changed to a higher version

michalkulakowski (approved):

dtrawins (approved):

PR #2131, opened 2023-11-07
correct copying artefacts for baseos ubuntu22

dtrawins:


atobiszei (approved):
I am ok with this as long as we will keep only one way to decide which version to use:
1 variable:
ubuntu20/ubuntu22
and stop support for 
ubuntu & separate set image tag variable

dkalinowski (approved):

PR #2158, opened 2023-11-17
Add openvino/python to PYTHONPATH

michalkulakowski:


Dockerfile.redhat
```diff
@@ -306,13 +306,17 @@ RUN cp /ovms/src/custom_node_interface.h /ovms/src/custom_nodes/tokenizer/ && \
 WORKDIR /ovms/src/example/SampleCpuExtension/
 RUN make
 
+WORKDIR /opt/intel/openvino/python/openvino-2023.2.dist-info/
```
mzegla, line 309: Is this line needed?

Dockerfile.redhat
```diff
@@ -409,6 +413,10 @@ LABEL supported-devices="CPU=1 GNA=1 GPU=${GPU} NVIDIA=${NVIDIA}"
 ARG RELEASE_BASE_IMAGE
 LABEL base-image=${RELEASE_BASE_IMAGE}
 
+WORKDIR /opt/intel/openvino/python
```
mzegla, line 416: Is this line needed?

Dockerfile.ubuntu
```diff
@@ -318,13 +318,17 @@ RUN cp /ovms/src/custom_node_interface.h /ovms/src/custom_nodes/tokenizer/ && \
 WORKDIR /ovms/src/example/SampleCpuExtension/
 RUN make
 
+WORKDIR /opt/intel/openvino/python/openvino-2023.2.dist-info/
```
mzegla, line 321: Is this line needed?

Dockerfile.ubuntu
```diff
@@ -416,6 +421,11 @@ LABEL supported-devices="CPU=1 GNA=1 GPU=${GPU} NVIDIA=${NVIDIA}"
 ARG BASE_IMAGE
 LABEL base-image=${BASE_IMAGE}
 SHELL ["/bin/bash", "-c"]
+
+WORKDIR /opt/intel/openvino/python
```
mzegla, line 425: Is this line needed?

Dockerfile.redhat
```diff
@@ -409,6 +413,10 @@ LABEL supported-devices="CPU=1 GNA=1 GPU=${GPU} NVIDIA=${NVIDIA}"
...
+COPY --from=build /opt/intel/openvino/python /opt/intel/openvino/python
```
mzegla, line 417: I wonder if we can make that copy depending on PYTHON_DISABLE setting

Dockerfile.ubuntu
```diff
@@ -416,6 +421,11 @@ LABEL supported-devices="CPU=1 GNA=1 GPU=${GPU} NVIDIA=${NVIDIA}"
...
+COPY --from=build /opt/intel/openvino/python /opt/intel/openvino/python
```
mzegla, line 426: I wonder if we can make that copy depending on PYTHON_DISABLE setting

Dockerfile.redhat
```diff
@@ -306,13 +306,17 @@ RUN cp /ovms/src/custom_node_interface.h /ovms/src/custom_nodes/tokenizer/ && \
 WORKDIR /ovms/src/example/SampleCpuExtension/
 RUN make
 
+WORKDIR /opt/intel/openvino/python/openvino-2023.2.dist-info/
+RUN touch /opt/intel/openvino/python/openvino-2023.2.dist-info/METADATA && echo $'Metadata-Version: 1.0\nName: openvino\nVersion: 2023.2' > /opt/intel/openvino/python/openvino-2023.2.dist-info/METADATA
+ENV PYTHONPATH "${PYTHONPATH}:/opt/intel/openvino/python"
```
dtrawins, line 311: pythonpath is initially empty so not need for ${PYTHONPATH}

Dockerfile.redhat
```diff
@@ -306,13 +306,17 @@ RUN cp /ovms/src/custom_node_interface.h /ovms/src/custom_nodes/tokenizer/ && \
...
+
 WORKDIR /ovms
 
 ARG PROJECT_VERSION="2023.2"
 ARG PROJECT_NAME="OpenVINO Model Server"
 LABEL description=${PROJECT_NAME}
 ARG minitrace_flags
-ENV PYTHONPATH=/ovms/bazel-bin/src
+ENV PYTHONPATH=${PYTHONPATH}:/ovms/bazel-bin/src
```
dtrawins, line 319: why it is set second time?

Dockerfile.redhat
dtrawins: we should add those python bindings to the binary package depending on the PYTHON_DISABLE
It should be added during creating binary package instead of here
all ovms related binaries should be in /ovms/lib

Dockerfile.redhat
```diff
@@ -409,6 +413,10 @@ LABEL supported-devices="CPU=1 GNA=1 GPU=${GPU} NVIDIA=${NVIDIA}"
 ARG RELEASE_BASE_IMAGE
 LABEL base-image=${RELEASE_BASE_IMAGE}
 
+WORKDIR /opt/intel/openvino/python
+COPY --from=build /opt/intel/openvino/python /opt/intel/openvino/python
+ENV PYTHONPATH "${PYTHONPATH}:/opt/intel/openvino/python:/ovms/lib"
```
dtrawins, line 418: "${PYTHONPATH}:/opt/intel/openvino/python should not be needed

dtrawins (changes requested):

dtrawins (approved):

PR #2159, opened 2023-11-17
Python finalize implementation

rasapala:


src/mediapipe_internal/pythonnoderesource.cpp
```diff
@@ -37,6 +37,32 @@ PythonNodeResource::PythonNodeResource() {
...
+Status PythonNodeResource::finalize() {
+    if (this->nodeResourceObject) {
+        py::gil_scoped_acquire acquire;
+        try {
+            if (!py::hasattr(*nodeResourceObject.get(), "finalize")) {
+                SPDLOG_DEBUG("Python node resource does not have a finalize method. Python node path {} ", this->pythonNodeFilePath);
+                return StatusCode::OK;
+            }
+
+            py::bool_ success = nodeResourceObject.get()->attr("finalize")();
```
mzegla, line 49: I'm not sure if we should expect any value from finalize(). Unloading happens anyway so OVMS will not do anything with such information except for logging it. Whether finalize will complete successfully, return true, false or None, the node will get destroyed and won't be functional anymore.

src/mediapipe_internal/pythonnoderesource.cpp
```diff
@@ -37,6 +37,32 @@ PythonNodeResource::PythonNodeResource() {
     this->nodeResourceObject = nullptr;
 }
 
+Status PythonNodeResource::finalize() {
```
mzegla, line 40: I don't think we need to return status. This could be void as we will not do anything with that status. Destruction will proceed regardless of finalize() operation success or failure. We would just log if any error is encountered.

src/mediapipe_internal/pythonnoderesource.cpp
```diff
@@ -37,6 +37,32 @@ PythonNodeResource::PythonNodeResource() {
...
+    if (this->nodeResourceObject) {
```
mzegla, line 41: Is it possible that in this part of execution nodeResourceObject would be nullptr?

dkalinowski: We should inform administrator since this is abnormal behavior

src/mediapipe_internal/pythonnoderesource.cpp
```diff
@@ -38,8 +38,30 @@ PythonNodeResource::PythonNodeResource(PythonBackend* pythonBackend) {
...
+        } catch (const pybind11::error_already_set& e) {
+            SPDLOG_ERROR("Failed to process python node finalize method. {}  Python node path {} ", e.what(), this->pythonNodeFilePath);
+            return;
+        } catch (...) {
+            SPDLOG_ERROR("Failed to process python node finalize method. Python node path {} ", this->pythonNodeFilePath);
+            return;
+        }
+    }
+
+    return;
```
dkalinowski, line 60: not needed

src/test/mediapipe/python/pythonNodeFinalizeMissingPass.py
```diff
@@ -0,0 +1,22 @@
+#*****************************************************************************
```
dkalinowski, line 1: There already is a directory for python scripts inside `src/test/mediapipe/python/scripts` directory. Also please use script name convention used in this directory made by @mzegla

src/test/pythonnode_test.cpp
```diff
@@ -844,3 +844,127 @@ TEST_F(PythonFlowTest, PythonCalculatorTestSingleInSingleOutMultiRunWithErrors)
...
+                }
+            }
+        }
+    )"};
+    ::mediapipe::CalculatorGraphConfig config;
+    ASSERT_TRUE(::google::protobuf::TextFormat::ParseFromString(pbTxt, &config));
+
+    std::shared_ptr<PythonNodeResource> nodeResouce = nullptr;
+    ASSERT_EQ(PythonNodeResource::createPythonNodeResource(nodeResouce, config.node(0).node_options(0)), StatusCode::OK);
+    nodeResouce->finalize();
```
dkalinowski, line 969:
```suggestion
    nodeResource->finalize();
```

src/mediapipe_internal/pythonnoderesource.cpp
```diff
@@ -38,8 +38,30 @@ PythonNodeResource::PythonNodeResource(PythonBackend* pythonBackend) {
...
+}
+
+Status PythonNodeResource::createPythonNodeResource(std::shared_ptr<PythonNodeResource>& nodeResource, const google::protobuf::Any& nodeOptions) {
+    mediapipe::PythonBackendCalculatorOptions options;
```
mzegla, line 64:
```suggestion
    mediapipe::PythonBackendCalculatorOptions nodeOptions;
```

src/mediapipe_internal/pythonnoderesource.cpp
```diff
@@ -68,6 +90,7 @@ Status PythonNodeResource::createPythonNodeResource(std::shared_ptr<PythonNodeRe
 
         nodeResource = std::make_shared<PythonNodeResource>(pythonBackend);
         nodeResource->nodeResourceObject = std::make_unique<py::object>(pythonModel);
+        nodeResource->pythonNodeFilePath = options.handler_path();
```
mzegla, line 93:
```suggestion
        nodeResource->pythonNodeFilePath = nodeOptions.handler_path();
```

src/test/mediapipe/python/pythonNodeFinalizeException.py
```diff
@@ -0,0 +1,25 @@
...
+
+class OvmsPythonModel:
+    def initialize(self, kwargs: dict):
+        return True
+
+    def execute(self, inputs: list, kwargs: dict) -> list:
+        return None
+    
+    def finalize(self):
+        return NotExistingException
```
mzegla, line 25: I know it doesn't matter since it throws anyway, but it's still a bit confusing. Please change to
`return not_existing_variable` or `raise NotExistingException()`

src/test/mediapipe/python/pythonNodeFinalizePass.py
```diff
@@ -0,0 +1,25 @@
...
+
+class OvmsPythonModel:
+    def initialize(self, kwargs: dict):
+        return True
+
+    def execute(self, inputs: list, kwargs: dict) -> list:
+        return None
+
+    def finalize(self):
+        return True
```
mzegla, line 25: Can we `pass` here?

src/mediapipe_internal/pythonnoderesource.cpp
rasapala: Ok

src/mediapipe_internal/pythonnoderesource.cpp
rasapala: Those are other options.

src/mediapipe_internal/pythonnoderesource.cpp
rasapala: nodeOptions are passed as function argument.

src/mediapipe_internal/pythonnoderesource.cpp
```diff
@@ -36,16 +36,40 @@ namespace ovms {
...
+            SPDLOG_ERROR("Failed to process python node finalize method. Python node path {} ", this->pythonNodeFilePath);
+            return;
+        }
+    } else {
+        SPDLOG_ERROR("nodeResourceObject is not initialized. Python node path {} ", this->pythonNodeFilePath);
+        throw std::exception();
+    }
+}
+
+Status PythonNodeResource::createPythonNodeResource(std::shared_ptr<PythonNodeResource>& nodeResource, const google::protobuf::Any& nodeInput, PythonBackend* pythonBackend) {
```
mzegla, line 65:
```suggestion
Status PythonNodeResource::createPythonNodeResource(std::shared_ptr<PythonNodeResource>& nodeResource, const google::protobuf::Any& nodeConfig, PythonBackend* pythonBackend) {
```

src/mediapipe_internal/pythonnoderesource.cpp
```diff
@@ -36,16 +36,40 @@ namespace ovms {
...
+    mediapipe::PythonExecutorCalculatorOptions nodeOptions;
+    nodeInput.UnpackTo(&nodeOptions);
```
mzegla, line 67:
```suggestion
    nodeConfig.UnpackTo(&nodeOptions);
```

mzegla (approved):

PR #2162, opened 2023-11-21
Align timing routines, fix average inference calculation

ngaloppo:
The callback function records inference time based on the timestamp sent by the client using `self.get_timestamp()`. The `display()` function needs to use the same timing routine to get consistent time deltas.


ngaloppo:
@dtrawins FYI

rasapala (approved):
LGTM

ngaloppo:
@dtrawins can you merge?

PR #2187, opened 2023-12-05
Deinitialize correctly initialized nodes upon single loading procedure, when one node initialization fails

dkalinowski:
Part of CVS-121956

### DIFF
--- a/src/mediapipe_internal/mediapipegraphdefinition.cpp
+++ b/src/mediapipe_internal/mediapipegraphdefinition.cpp
@@ -438,22 +438,38 @@ std::pair<std::string, mediapipe_packet_type_enum> MediapipeGraphDefinition::get
     return {"", mediapipe_packet_type_enum::UNKNOWN};
 }
 
+struct PythonResourceInitializationCleaningGuard {
+    bool shouldCleanup{true};
+    std::unordered_map<std::string, std::shared_ptr<PythonNodeResource>>& resource;
+    PythonResourceInitializationCleaningGuard(std::unordered_map<std::string, std::shared_ptr<PythonNodeResource>>& resource) :
+        resource(resource) {}
+    ~PythonResourceInitializationCleaningGuard() {
+        if (shouldCleanup) {
+            resource.clear();
+        }
+    }
+    void markSuccess() {
+        shouldCleanup = false;
+    }
+};
+
 Status MediapipeGraphDefinition::initializeNodes() {
 #if (PYTHON_DISABLE == 0)
+    PythonResourceInitializationCleaningGuard initializationCleaningGuard(this->pythonNodeResources);
     SPDLOG_INFO("MediapipeGraphDefinition initializing graph nodes");
     for (int i = 0; i < config.node().size(); i++) {
         if (config.node(i).calculator() == PYTHON_NODE_CALCULATOR_NAME) {
             if (!config.node(i).node_options().size()) {
-                SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Python node missing options in graph: {}. ", this->name);
+                SPDLOG_LOGGER_ERROR(modelmanager_logger, "Python node missing options in graph: {}. ", this->name);
                 return StatusCode::PYTHON_NODE_MISSING_OPTIONS;
             }
             if (config.node(i).name().empty()) {
-                SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Python node name is missing in graph: {}. ", this->name);
+                SPDLOG_LOGGER_ERROR(modelmanager_logger, "Python node name is missing in graph: {}. ", this->name);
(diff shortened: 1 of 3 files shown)


rasapala (approved):

src/mediapipe_internal/mediapipegraphdefinition.cpp
```diff
@@ -438,22 +438,38 @@ std::pair<std::string, mediapipe_packet_type_enum> MediapipeGraphDefinition::get
...
+    ~PythonResourceInitializationCleaningGuard() {
+        if (shouldCleanup) {
+            resource.clear();
+        }
+    }
+    void markSuccess() {
+        shouldCleanup = false;
+    }
+};
+
```
mzegla, line 455: resource -> resources

src/mediapipe_internal/mediapipegraphdefinition.cpp
```diff
@@ -438,22 +438,38 @@ std::pair<std::string, mediapipe_packet_type_enum> MediapipeGraphDefinition::get
...
+    bool shouldCleanup{true};
+    std::unordered_map<std::string, std::shared_ptr<PythonNodeResource>>& resource;
+    PythonResourceInitializationCleaningGuard(std::unordered_map<std::string, std::shared_ptr<PythonNodeResource>>& resource) :
+        resource(resource) {}
+    ~PythonResourceInitializationCleaningGuard() {
+        if (shouldCleanup) {
+            resource.clear();
+        }
+    }
+    void markSuccess() {
```
mzegla, line 451: What about naming this method to something like: `doNotClean`, `cleanupNotNeeded` or `abandonCleanup`, so it more directly tells about the consequences.

src/mediapipe_internal/mediapipegraphdefinition.cpp
```diff
@@ -438,22 +438,38 @@ std::pair<std::string, mediapipe_packet_type_enum> MediapipeGraphDefinition::get
     return {"", mediapipe_packet_type_enum::UNKNOWN};
 }
 
+struct PythonResourceInitializationCleaningGuard {
```
mzegla, line 441: Maybe we can shorten this: 
PythonResourceInitializationCleaningGuard -> PythonResourcesCleaningGuard

src/mediapipe_internal/mediapipegraphdefinition.cpp
```diff
@@ -438,22 +438,38 @@ std::pair<std::string, mediapipe_packet_type_enum> MediapipeGraphDefinition::get
...
+        }
+    }
+    void markSuccess() {
+        shouldCleanup = false;
+    }
+};
+
 Status MediapipeGraphDefinition::initializeNodes() {
 #if (PYTHON_DISABLE == 0)
+    PythonResourceInitializationCleaningGuard initializationCleaningGuard(this->pythonNodeResources);
```
mzegla, line 458: pythonResourcesCleaningGuard?
more directly says what the cleanup is about

src/test/pythonnode_test.cpp
```diff
@@ -1026,3 +1026,42 @@ TEST_F(PythonFlowTest, ReloadWithDifferentScriptName) {
...
+        }
+        node {
+            name: "pythonNode2"
+            calculator: "PythonExecutorCalculator"
+            input_side_packet: "PYTHON_NODE_RESOURCES:py"
+            input_stream: "INPUT:inter"
+            output_stream: "OUTPUT:out"
+            node_options: {
+                [type.googleapis.com / mediapipe.PythonExecutorCalculatorOptions]: {
+                    handler_path: "/ovms/src/test/mediapipe/python/scripts/bad_initialize_return_false.py"
```
mzegla, line 1055: Does this script simply return false? Could you use script that throws exception in initialize?
We don't want to return status in initialize, so this may get invalidated soon.

src/mediapipe_internal/mediapipegraphdefinition.cpp
dkalinowski: done

src/mediapipe_internal/mediapipegraphdefinition.cpp
dkalinowski: changed

src/mediapipe_internal/mediapipegraphdefinition.cpp
dkalinowski: done

src/mediapipe_internal/mediapipegraphdefinition.cpp
dkalinowski: done

src/test/pythonnode_test.cpp
dkalinowski: done

PR #2191, opened 2023-12-07
Extend ovms::Config class with information about Python startup

michalkulakowski:


src/config.hpp
```diff
@@ -47,6 +47,8 @@ class Config {
     ModelsSettingsImpl modelsSettings;
 
 public:
+     bool withPython = true;
```
mzegla, line 50: I'd rather see it as a member of ServerSettingsImpl object like: `runWithPythonInterpreterModule` with default value `true`. 
Then in CAPI tests you'd probably need to modify that value to `false` before calling start() with ServerSettingsImpl object.

PR #2224, opened 2023-12-21
changes to support nvidia plugin on ubuntu22

dtrawins:


tests/sdl/whitelists.py
```diff
@@ -114,6 +114,7 @@ class OvmsBaseImageType(Enum):
         'libtbb2',
         'libxml2',
         'openssl',
+        'libbrotli1'
```
ngrozae, line 117: 'libtrotli1' should be included in packages[OvmsBaseImageType.UBUNTU_NGINX] only

dtrawins: @ngrozae that would not be correct. libbrotli1 is a dependency of curl which are are adding the the release image regardless of the nginx

rasapala (approved):
LGTM

mzegla (approved):

ngrozae (approved):

PR #2248, opened 2024-01-09
Mkulakow/fix cleaner thread test sporadic

michalkulakowski:


src/test/modelmanager_test.cpp
```diff
@@ -25,6 +25,7 @@
 #include "../model.hpp"
 #include "../modelmanager.hpp"
 #include "../prediction_service_utils.hpp"
+#include "absl/synchronization/notification.h"
```
dkalinowski, line 28: Since you are using absl, please check build without MediaPipe (isnt absl dependency of MP? might be missing)

dkalinowski (approved):

atobiszei (approved):

PR #2284, opened 2024-01-22
Add torch

rasapala:


mzegla:
Don't we need the same change for main?

dtrawins (approved):

PR #2313, opened 2024-01-26
Fixed module tensorflow failed in client/python/tensorflow-serving-api/samples/README.md

bstrzele:


client/python/tensorflow-serving-api/samples/requirements.txt
```diff
@@ -1,2 +1,3 @@
 tensorflow-serving-api<=2.12.2,>=2.10.1
 protobuf<=3.20.3
+tensorflow==2.13.1
```
mzegla, line 3: Why this particular version? 
Don't we want to also change upper bound for `tensorflow-serving-api` so in default scenario they match?

bstrzele: This is the latest version that does not cause any conflicts.
Bumped tensorflow-serving-api to 2.13.1

mzegla (approved):

dkalinowski (approved):

PR #2321, opened 2024-02-08
Mem leak and test fix

rasapala:


atobiszei (approved):
Note this is so far only positive path fix. We need to write custom packet holder. Given time constraints I'm okay for this quick fix but we need to address it with proper fix.

PR #2338, opened 2024-02-26
Fixed stream client

bstrzele:
This pr addresses the bug where packet contains error does not release the load limiting semaphore

### DIFF
--- a/demos/common/stream_client/stream_client.py
+++ b/demos/common/stream_client/stream_client.py
@@ -119,7 +119,7 @@ class StreamClient:
         self.exact = exact
         self.benchmark = benchmark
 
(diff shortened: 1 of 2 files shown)


demos/common/stream_client/stream_client.py
```diff
@@ -143,6 +143,7 @@ def callback(self, frame, i, timestamp, result, error):
                 self.dropped_frames += 1
             if self.verbose:
                 print(error)
+            self.req_q.get()
```
dtrawins, line 146: maybe we should move getting from queue before checking the error status?

rasapala (approved):
LGTM

dtrawins (approved):

PR #2344, opened 2024-03-01
Resolve python node todo's

michalkulakowski:


src/python/python_executor_calculator.cc
```diff
@@ -212,23 +215,21 @@ class PythonExecutorCalculator : public CalculatorBase {
             }
         } catch (const UnexpectedOutputTensorError& e) {
             LOG(INFO) << "Error occurred during node " << cc->NodeName() << " execution: " << e.what();
-            return absl::Status(absl::StatusCode::kInternal, "Python execute function returned unexpected output");
+            RETURN_EXECUTION_FAILED_STATUS();
         } catch (const UnexpectedPythonObjectError& e) {
-            // TODO: maybe some more descriptive information where to seek the issue.
-            LOG(INFO) << "Error occurred during node " << cc->NodeName() << " execution. Wrong object on execute input or output: " << e.what();
-            return absl::Status(absl::StatusCode::kInternal, "Python execute function received or returned bad value");
+            LOG(INFO) << "Error occurred during node " << cc->NodeName() << " execution. Wrong object on execute output: " << e.what();
```
mzegla, line 220: I think this error is thrown in validateOvmsPyTensor which is called also in prepareInputs function here in the calculator. To avoid ambiguity I think we should make a change, so error in prepareInputs does not end up in this block.

mzegla (approved):

PR #2353, opened 2024-03-05
use latest onnx version

dtrawins:


demos/gptj_causal_lm/python/requirements.txt
```diff
@@ -1,7 +1,7 @@
 accelerate==0.20.3
 numpy==1.23
 torch==1.13.1
-onnx==1.15.0
+onnx
```
mzegla, line 4: Don't we risk incompatibility in time? Upgrading to the current latest with a specific number would give us more control.

mzegla (approved):

dkalinowski (approved):

PR #2365, opened 2024-03-11
Log conversion

rasapala:


src/mediapipe_internal/mediapipegraphexecutor.cpp
```diff
@@ -473,6 +473,7 @@ static Status deserializeTensor(const std::string& requestedName, const KFSReque
         }
 
         auto formatIt = datatypeToBufferFormat.find(requestInputItr->datatype());
+        SPDLOG_DEBUG("Converting datatype {} to datatype {} for tensor {}", requestInputItr->datatype(), formatIt != datatypeToBufferFormat.end() ? formatIt : "RAW BINARY UINT8", requestedName);
```
mzegla, line 476: I don't get the purpose of this change. There is no conversion going on - `datatype` and `format` are separate values for two different contexts (`datatype` in user and API and `format` in Python internals).

PR #2366, opened 2024-03-12
Python improvements and bugfixes

mzegla:
Summary of changes:
- validate if `execute` method is implemented in node initialization phase
- trigger finalize on graph definition retire
- add optional arguments `shape` and `datatype` to `pyovms.Tensor` constructor
- tests and docs for above
- print with flush=True in python demos


rasapala (approved):
LGTM

demos/python_demos/llm_text_generation/servable_stream/model.py
```diff
@@ -132,17 +132,17 @@ def serialize_completions(batch_size, result):
...
     def initialize(self, kwargs: dict):
-        print("-------- Running initialize")
+        print("-------- Running initialize", flush=True)
         self.ov_model = OVModelForCausalLM.from_pretrained(
             MODEL_PATH,
             device="AUTO",
             ov_config=OV_CONFIG,
             config=AutoConfig.from_pretrained(MODEL_PATH, trust_remote_code=True))
-        print("-------- Model loaded")
+        print("-------- Model loaded", flush=True)
```
dkalinowski, line 141: Maybe we should create logger? Probably out of scope for this task

dkalinowski (approved):

demos/python_demos/llm_text_generation/servable_stream/model.py
mzegla: I agree. It would be a separate feature that we could add along with some custom exception type that user could throw and error message would be returned in the server response etc.

PR #2372, opened 2024-03-19
ov::string serialization for single models (+ single model DAG)

dkalinowski:
CVS-132111

### DIFF
--- a/src/kfs_frontend/kfs_utils.cpp
+++ b/src/kfs_frontend/kfs_utils.cpp
@@ -189,8 +189,11 @@ Status getRawInputContentsBatchSizeAndWidth(const std::string& buffer, int32_t& 
         offset += (sizeof(uint32_t) + inputSize);
         tmpBatchSize++;
     }
-    if (offset != buffer.size()) {
-        SPDLOG_DEBUG("Raw input contents invalid format. Every input need to be preceded by four bytes of its size.");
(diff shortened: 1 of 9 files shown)


atobiszei (approved):

src/kfs_frontend/kfs_utils.cpp
```diff
@@ -189,8 +189,11 @@ Status getRawInputContentsBatchSizeAndWidth(const std::string& buffer, int32_t&
         offset += (sizeof(uint32_t) + inputSize);
         tmpBatchSize++;
     }
-    if (offset != buffer.size()) {
-        SPDLOG_DEBUG("Raw input contents invalid format. Every input need to be preceded by four bytes of its size.");
+    if (offset > buffer.size()) {
+        SPDLOG_DEBUG("Raw input contents invalid format. Every input need to be preceded by four bytes of its size. Buffer exceeded by {} bytes", offset - buffer.size());
```
dtrawins, line 193: Is it the case also for BS1?

dtrawins (approved):

src/kfs_frontend/kfs_utils.cpp
dkalinowski: If we want to follow triton's binary extension, we need to include 4 bytes length even for BS=1

PR #2373, opened 2024-03-19
Updated llm demo to report number of generated tokens in the last response

bstrzele:
CVS-135176


demos/python_demos/llm_text_generation/servable_stream/model.py
```diff
@@ -170,13 +170,16 @@ def execute(self, inputs: list):
...
+            self.token_count += len([1 for x in result.numpy().flatten() if x not in tokenizer.convert_tokens_to_ids(tokenizer.all_special_tokens)])
 
         t1 = threading.Thread(target=generate)
         t1.start()
 
         for partial_result in streamer:
             yield serialize_completions(batch_size, partial_result)
-        yield [Tensor("end_signal", "".encode())]
+        t1.join()
+        yield [Tensor("end_signal", f"{str(self.token_count)}".encode())]
```
dkalinowski, line 184: Wouldn't it be better to return numpy array of uint32_t/uint64_t of size 1?
```
yield [Tensor(
  "end_signal",
  np.array([self.token_count], dtype=np.int])
)]
```

And then receive it on the client side:
```
num_of_tokens = result.as_numpy('end_signal')[0]
```

demos/python_demos/llm_text_generation/servable_stream/model.py
```diff
@@ -170,13 +170,16 @@ def execute(self, inputs: list):
             generate_kwargs["stopping_criteria"] = StoppingCriteriaList(stop_tokens)
 
         ov_model_exec = self.ov_model.clone()
+        self.token_count = 0
         def generate():
-            ov_model_exec.generate(**tokens, **generate_kwargs)
+            result = ov_model_exec.generate(**tokens, **generate_kwargs)
```
dkalinowski, line 175: What's the reason we do not include special tokens? What are those special tokens?

demos/python_demos/llm_text_generation/servable_unary/model.py
```diff
@@ -184,5 +188,6 @@ def generate():
             else:
                 completions = [a + b for a, b in zip(completions, partial_result)]
         print('end', flush=True)
-
-        return serialize_completions(batch_size, completions)
+        t1.join()
+        completions.append(str(self.token_count))
```
dkalinowski, line 192: I don't think it is clean solution, it might show us that adding separate output is not bad idea. @dtrawins 
The same could be for stream endpoint. 
Suggestion - leave end_signal for what it is and add separate one.

demos/python_demos/llm_text_generation/servable_unary/model.py
```diff
@@ -171,7 +173,9 @@ def execute(self, inputs: list):
 
         ov_model_exec = self.ov_model.clone()
         def generate():
-            ov_model_exec.generate(**tokens, **generate_kwargs)
+            result = ov_model_exec.generate(**tokens, **generate_kwargs)
+            self.token_count = len([1 for x in result.numpy().flatten() if x not in tokenizer.convert_tokens_to_ids(tokenizer.all_special_tokens)])
```
dtrawins, line 177: self.token_count can be overwritten by different thread from other client.

dtrawins:
include output changes in readme

demos/python_demos/llm_text_generation/servable_stream/model.py
bstrzele: In this case special tokens are used for left/right padding. They are stripped to assure accurate token counting when batch_size > 1.

demos/python_demos/llm_text_generation/servable_unary/model.py
```diff
@@ -170,8 +170,11 @@ def execute(self, inputs: list):
             generate_kwargs["stopping_criteria"] = StoppingCriteriaList(stop_tokens)
 
         ov_model_exec = self.ov_model.clone()
+        token_count: List[int]= []
         def generate():
-            ov_model_exec.generate(**tokens, **generate_kwargs)
+            result = ov_model_exec.generate(**tokens, **generate_kwargs)
+            token_count.append(len([1 for x in result.numpy().flatten() if x not in tokenizer.convert_tokens_to_ids(tokenizer.all_special_tokens)]))
```
dtrawins, line 176: token count includes tokens sent to the server in prompt. We should remove from the result the number of received tokens to track only generated tokens.

demos/python_demos/llm_text_generation/servable_stream/model.py
```diff
@@ -181,6 +181,7 @@ def generate():
         for partial_result in streamer:
             yield serialize_completions(batch_size, partial_result)
         t1.join()
+        token_count[0] -= len(tokens["input_ids"].flatten())
         yield [Tensor("token_count", np.array([token_count[0]], dtype=np.int32))]
```
dtrawins, line 185: why only 1st element of token_count? Was it tested with bs>1 with variable response size?

bstrzele: First element of token_count represents final number of tokens. This snipped uses list instead of single variable due to the mutable property of python's list type.

dkalinowski: I think @dtrawins misunderstood token_count variable - it is **always** 1 element array in order to return it as numpy array.

But @dtrawins makes good point here - @bstrzele should we track token count *per batch* or just sum of tokens in all of batches?

demos/python_demos/llm_text_generation/servable_unary/model.py
```diff
@@ -184,5 +187,5 @@ def generate():
             else:
                 completions = [a + b for a, b in zip(completions, partial_result)]
         print('end', flush=True)
-
-        return serialize_completions(batch_size, completions)
+        t1.join()
+        return serialize_completions(batch_size, completions, token_count[0])
```
dkalinowski, line 191: I'm not sure why are working on a numpy array, just to return a number in line 191. Then, in serialize_completions you create numpy array anyway `Tensor("token_count", np.array([token_count], dtype=np.int32))` @bstrzele

demos/python_demos/llm_text_generation/README.md
```diff
@@ -222,15 +225,18 @@ python3 client_stream.py --url localhost:9000 \
...
 
 ==== Prompt: Who is Albert Einstein? ====
- Albert Einstein was an American-born theoretical physicist who revolutionized the fields of physics, cosmology, and relativity by proposing an explanation for gravity that has since become known as special relativity. His work paved the way for the development of quantum mechanics and led to the field of astrophysics."
+    Albert Einstein was a German-born physicist who made significant contributions to the field of physics, particularly relating to his discovery of General Relativity. He is recognized as one of the most influential scientists of the 20th century along with John Wyclif and Copernicus (who he also famously argued against). Over 150 years after he gave some of the most famous and fundamental scientific insights of all time, Einstein's theories continue to impact everyday life from how we understand gravity and space-time to our understanding of the universe from the small scale up to its grand cosmological scale. His legacy continues to inspire and shape future discoveries in these fields and beyond.
+
+
 
+Number of tokens  747
```
dtrawins, line 235: Can we add also a metric "average time per generated token" and "generated tokens per second"?

dtrawins (approved):

dtrawins (approved):

PR #2384, opened 2024-03-28
Added handling of relative paths

bstrzele:


demos/python_demos/llm_text_generation/servable_unary/python_model/model.py
```diff
@@ -137,11 +129,19 @@ def serialize_completions(batch_size, result, token_count):
 class OvmsPythonModel:
     def initialize(self, kwargs: dict):
         print("-------- Running initialize", flush=True)
+        self.stop_tokens = model_configuration.get("stop_tokens")
```
mzegla, line 132: I'd suggest:
- rename `MODEL_PATH` to `MODEL_RELATIVE_PATH`
- here in the initialize create a variable `model_absolute_path = os.path.join(kwargs['base_path'], MODEL_RELATIVE_PATH)`

And then use `model_absolute_path`. More readible imho.

src/mediapipe_internal/mediapipegraphdefinition.cpp
```diff
@@ -422,7 +422,7 @@ Status MediapipeGraphDefinition::initializeNodes() {
             }
 
             std::shared_ptr<PythonNodeResources> nodeResources = nullptr;
-            Status status = PythonNodeResources::createPythonNodeResources(nodeResources, config.node(i), pythonBackend);
+            Status status = PythonNodeResources::createPythonNodeResources(nodeResources, config.node(i), pythonBackend, mgconfig.getBasePath());
```
mzegla, line 425: Is the base_path an absolute path to the directory with config.json or graph.pbtxt?

src/test/pythonnode_test.cpp
```diff
@@ -2260,7 +2260,7 @@ TEST_F(PythonFlowTest, FinalizePassTest) {
     ASSERT_TRUE(::google::protobuf::TextFormat::ParseFromString(pbTxt, &config));
 
     std::shared_ptr<PythonNodeResources> nodeResources = nullptr;
-    ASSERT_EQ(PythonNodeResources::createPythonNodeResources(nodeResources, config.node(0), getPythonBackend()), StatusCode::OK);
+    ASSERT_EQ(PythonNodeResources::createPythonNodeResources(nodeResources, config.node(0), getPythonBackend(), ""), StatusCode::OK);
```
mzegla, line 2263: Is empty string a possible value? If createPythonNodeResources is not called with empty string as a basePath in any real scenario then this setting does not provide valueable information and might be misleading.

mzegla (commented):
Please update kwargs spec in the documentation:
https://github.com/openvinotoolkit/model_server/blob/main/docs/python_support/reference.md#parameters-and-return-value

src/mediapipe_internal/mediapipegraphdefinition.cpp
bstrzele: mgconfig.getBasePath() return path to graph.pbtxt

dtrawins (approved):

src/test/pythonnode_test.cpp
bstrzele: It is not possible in regular use case when pbtxt is located on file system. In case of these tests every value is equally valid, since there is no actual basePath in those scenarios.

src/python/pythonnoderesources.hpp
```diff
@@ -48,7 +48,7 @@ struct PythonNodeResources {
     void finalize();
 
 private:
-    static py::dict preparePythonNodeInitializeArguments(const ::mediapipe::CalculatorGraphConfig::Node& graphNodeConfig, std::string basePath);
+    static py::dict preparePythonNodeInitializeArguments(const ::mediapipe::CalculatorGraphConfig::Node& graphNodeConfig, std::string graphPath);
```
mzegla, line 51: Actually here it would be `basePath`

mzegla (approved):

PR #2385, opened 2024-03-29
Fix kfs rest model ready response parsing

michalkulakowski:


dkalinowski (approved):
Please add unit test if possible

PR #2386, opened 2024-03-29
Allow flag injection to pugixml

stevegrubb:
This commit contains a patch that adds the variables for the CXX and linker flags to the CMakeLists.txt file. It then uses the patch during build so that later we can inject build flags on the cmake command.


dtrawins:
Changes applied in https://github.com/openvinotoolkit/model_server/pull/2390

PR #2392, opened 2024-04-03
Upgraded gradio

bstrzele:
CVS-136553


dkalinowski (approved):

dkalinowski:
@bstrzele please ensure with validation that the demo still works, they are using gradio client to test it, they might need to upgrade it as well 
@ngrozae

dtrawins (approved):

PR #2426, opened 2024-04-23
Control build log verbosity from argument

stevegrubb:
This patch changes the log verbosity of the redhat Dockerfile to be controlled by an
argument. It will default to off to maintain the original behavior. It can be overridden 
by --build-arg VERBOSE_LOGS=ON or similar depending on how containers are built.

### DIFF
--- a/Dockerfile.redhat
+++ b/Dockerfile.redhat
@@ -27,6 +27,7 @@ LABEL version="1.0.0"
 SHELL ["/bin/bash", "-xo", "pipefail", "-c"]
 
 ARG JOBS=8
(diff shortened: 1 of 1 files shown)


stevegrubb:
I'd like to start experimenting with moving this to a rhel9 ubi image, but we'd like to start with this  patch already in tree. It would cleanup some of the things in the build because rhel9 has newer packages.

dtrawins:
I moved this change to https://github.com/openvinotoolkit/model_server/pull/2510
It is rebased and slightly updated to pass hadolint tests

PR #2433, opened 2024-04-26
universal_and_benchmark_documentation_updates

pgladkows:


demos/benchmark/python/README.md
```diff
@@ -323,7 +325,6 @@ NO_PROXY=localhost no_proxy=localhost python3 /ovms_benchmark_client/main.py -a
           XI worker: --> dim: 300
           XI worker: Generated data shape: (1, 3, 300, 300)
           XI worker: start workload...
-...
```
ngrozae, line 326: "..." means that there may be more output - it's not checked by tests so I don't think we need to remove it

pgladkows: done

dtrawins (approved):

PR #2442, opened 2024-05-20
Add LLM execution thread

mzegla:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.
This is just a partial change. Next PR that completes this task will follow.

### 🧪 Checklist

- [ ] Unit tests added.
- [ ] <s> The documentation updated.</s>  Documentation does not exist yet and will be created in a separate PR
``


src/llm/llm_executor.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <continuous_batching_pipeline.hpp>
+
+class LLMExecutor {
+    std::atomic<bool> end;
+    std::shared_ptr<ContinuousBatchingPipeline> pipe;
+    std::thread llmExecutorThread;
+
+    static void run(std::shared_ptr<ContinuousBatchingPipeline> pipe, std::atomic<bool>* end) {
+        while (!(*end)) {
+            pipe->step();
```
dkalinowski, line 32: Is this step blocking if there are no iterations to make? (0 requests processing)
If so, we may have a deadlock on line 45 (waiting for the thread to join will be forever)

mzegla: It's not blocking. But perhaps we could add early skip inside step() to avoid running anything when we don't have active or awaiting requests.

src/llm/BUILD
```diff
@@ -73,7 +73,7 @@ mediapipe_proto_library(
 
 cc_library(
     name = "llmnoderesources",
-    hdrs = ["llmnoderesources.hpp",],
+    hdrs = ["llmnoderesources.hpp", "llm_executor.hpp",],
```
rasapala, line 76: Please add as separate library - faster compilation and less dependencies.

src/llm/llm_executor.hpp
```diff
@@ -0,0 +1,47 @@
...
+#pragma once
+
+#include <atomic>
+#include <memory>
+#include <thread>
+
+#include <continuous_batching_pipeline.hpp>
+
+class LLMExecutor {
+    std::atomic<bool> end;
```
rasapala, line 26: Please provide better member description than end so that it is self explanatory. Is threadEnd what its for ?

src/llm/llm_executor.hpp
dkalinowski: @mzegla with some sleep in between skips? otherwise we are burning cpu cycles

mzegla: With simple sleep we would risk underutilization. Some notification solution would be better.

src/llm/BUILD
mzegla: I'm not sure it makes sense. LLM executor is tightly connected with the resources and is not really used outside of it. And at this point it's really not a lot of code.

dkalinowski (approved):

rasapala (approved):
LGTM

PR #2447, opened 2024-05-23
LLM execution thread notifications and error handling 

mzegla:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] <s>Unit tests added.</s> Tests will be added after passing sampling params is supported 
- [ ] <s>The documentation updated.</s> Documentation is not in place yet
``

### DIFF
--- a/src/llm/llm_executor.hpp
+++ b/src/llm/llm_executor.hpp
@@ -23,41 +23,72 @@
 #include <thread>
 
 #include <continuous_batching_pipeline.hpp>
+
+#include "../logging.hpp"
+
 namespace ovms {
-class LLMExecutor {
+struct LLMExecutor {
+    // For logging purposes we could have more information about graph and node here
     std::mutex mutex;
     std::condition_variable cv;
-    std::atomic<bool> finishExecutorThread;
-    std::shared_ptr<ContinuousBatchingPipeline> pipe;
-    std::thread llmExecutorThread;
+    std::shared_ptr<ContinuousBatchingPipeline> pipe = nullptr;
 
-    static void run(std::shared_ptr<ContinuousBatchingPipeline> pipe, std::atomic<bool>* finishExecutorThread) {
-        while (!(*finishExecutorThread)) {
-            if (pipe->has_awaiting_requests() || pipe->has_running_requests()) {
-                pipe->step();
-            }
-            else {
-                std::unique_lock<std::mutex> lock(mutex);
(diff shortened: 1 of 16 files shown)


src/llm/llm_executor.hpp
```diff
@@ -17,32 +17,78 @@
...
-    static void run(std::shared_ptr<ContinuousBatchingPipeline> pipe, std::atomic<bool>* end) {
-        while (!(*end)) {
-            pipe->step();
+    static void run(LLMExecutor* llmExecutor, std::atomic<bool>* receivedEndSignal) {
+        while (!(*receivedEndSignal)) {
+            try {
+                if (llmExecutor->hasRequests()) {
+                    llmExecutor->step();
+                } else {
+                    llmExecutor->waitForRequests(receivedEndSignal);
```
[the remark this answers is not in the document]
dkalinowski: @mzegla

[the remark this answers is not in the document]
mzegla: No, see the predicate in the wait() call for conditional variable inside waitForRequests method. 
We pass `receivedEndSignal` exactly for that purpose. When we want to finish the thread we set that flag and notify CV so it breaks out of this wait.

rasapala (approved):
LGTM

src/test/llmnode_test.cpp
```diff
@@ -122,7 +131,7 @@ TEST_F(LLMFlowKfsTest, LLMNodeNameMissing) {
             output_stream: "RESPONSE:out2"
             node_options: {
                 [type.googleapis.com / mediapipe.LLMCalculatorOptions]: {
-                    workspace_path: "/workspace/"
+                    models_path: "/workspace/"
```
dtrawins, line 134: later, also relative path should be supported...

dtrawins (approved):

src/test/llmnode_test.cpp
mzegla: @rasapala I think when you added node initialization you take that into account. Can you confirm it should work?

dkalinowski (approved):

src/test/llmnode_test.cpp
rasapala: Yeah it should work, but I have not tested it yet. Maybe with the model tests we can add test for relative? Or next PR.

PR #2452, opened 2024-05-23
LLM calculator compliant with /v3/chat/completions

dkalinowski:
### 🛠 Summary

CVS-139231/CVS-139233

This introduces LLM calculator that accepts HTTP OpenAI /v3/chat/completion requests and produces compliant responses.
Working in both - unary and streaming modes.
Bunch of parameters are still marked as TODO, but should be enough to perform benchmarks.

Minimal demo description how to run.

### 🧪 Checklist

- [ ] Unit tests added.

### DIFF
--- a/demos/continous_batching/README.md
+++ b/demos/continous_batching/README.md
@@ -8,30 +8,17 @@ PIP_EXTRA_INDEX_URL="https://download.pytorch.org/whl/cpu" python3 -m pip instal
 
 Run optimum-cli to download and quantize the model:
 ```bash
-optimum-cli export openvino --model meta-llama/Llama-2-7b-chat-hf --weight-format int8 ov_model
+optimum-cli export openvino --model meta-llama/Llama-2-7b-chat-hf --weight-format int8 meta-llama/Llama-2-7b-chat-hf
 ```
-
-## Server configuration
-Prepare config.json:
-```
-{
-    "model_config_list": [],
-    "mediapipe_config_list": [
-        {
-            "name": "llama",
-            "graph_path": "graph.pbtxt"
-        }
-    ]
-}
-```
-Prepare graph.pbtxt:
-```
+Copy the graph to the model folder. The same graph can be used for a range of LLM models.
+```bash
+cat graph.pbtxt
 input_stream: "HTTP_REQUEST_PAYLOAD:input"
 output_stream: "HTTP_RESPONSE_PAYLOAD:output"
 
 node: {
   name: "LLMExecutor"
-  calculator: "OpenAIChatCompletionsCalculator"
+  calculator: "HttpLLMCalculator"
   input_stream: "LOOPBACK:loopback"
   input_stream: "HTTP_REQUEST_PAYLOAD:input"
   input_side_packet: "LLM_NODE_RESOURCES:llm"
@@ -43,7 +30,7 @@ node: {
   }
   node_options: {
       [type.googleapis.com / mediapipe.LLMCalculatorOptions]: {
-          workspace_path: "/model"
+          models_path: "./"
       }
   }
   input_stream_handler {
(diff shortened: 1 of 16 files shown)


demos/continous_batching/README.md
```diff
@@ -0,0 +1,155 @@
...
+    stream=True,
+)
+for chunk in stream:
+    if chunk.choices[0].delta.content is not None:
+        print(chunk.choices[0].delta.content, end="")
+```
+
+Output:
+```
+This is a test.
```
mzegla, line 152: I wonder if it makes sense to put a whole output here while the sample is showing streaming. Maybe a small GIF would be useful?

src/llm/http_llm_calculator.cc
```diff
@@ -0,0 +1,523 @@
...
+#include "http_payload.hpp"
+
+using namespace rapidjson;
+
+namespace mediapipe {
+
+using chat_entry_t = std::unordered_map<std::string, std::string>;
+using chat_t = std::vector<chat_entry_t>;
+
+class OpenAIChatCompletionsRequest {
```
mzegla, line 47: I think we should move it outside of this file in case we potentially want to use this calculator with other kinds of requests.

src/llm/http_llm_calculator.cc
```diff
@@ -0,0 +1,523 @@
...
+    Document& doc;
+
+    chat_t messages;
+    bool stream{false};
+    std::string model;
+    float frequencyPenalty{0.0f};
+    std::optional<int64_t> maxTokens{std::nullopt};
+    float presencePenalty{0.0f};
+    float temperature{1.0f};
+    float topP{1.0f};
```
mzegla, line 57: Are these defaults from OpenAI?

src/llm/http_llm_calculator.cc
```diff
@@ -0,0 +1,523 @@
...
+        // functions TODO (deprecated)
+        return true;
+    }
+};
+
+// TODO: To be moved to CB library.
+class TextStreamer {
+    std::shared_ptr<Tokenizer> tokenizer;
+    std::vector<int64_t> tokenCache;
+    size_t printLen;
```
mzegla, line 185: I would rename that field. In our case it's not about printing. It's more like a `lastReadIndex`, what do you think?

src/llm/http_llm_calculator.cc
```diff
@@ -0,0 +1,523 @@
...
+            return chunk;
+        }
+        return std::nullopt;
+    }
+};
+
+using InputDataType = ovms::HttpPayload;
+using OutputDataType = std::string;
+
+const std::string LLM_SESSION_SIDE_PACKET_TAG = "LLM_NODE_RESOURCES";
```
mzegla, line 213: Wonder if we could import that from another file. I think node resources or mediapipe graph definition also use that name.

src/llm/http_llm_calculator.cc
```diff
@@ -0,0 +1,523 @@
...
+        }
+
+        RET_CHECK(this->generationHandle != nullptr);
+        RET_CHECK(this->request != nullptr);
+        RET_CHECK(this->streamer != nullptr);
+
+        // Unary scenario
+        if (!this->request->isStream()) {
+            std::vector<GenerationOutput> generationOutput = this->generationHandle->read_all();
+            RET_CHECK(generationOutput.size() == 1);
```
mzegla, line 322: Note that this is temporary assumption.

src/llm/http_llm_calculator.cc
```diff
@@ -0,0 +1,523 @@
...
+        }
+
+        timestamp = timestamp.NextAllowedInStream();
+
+        return absl::OkStatus();
+    }
+};
+
+// TODO: Different samplings may produce multiple generations.
+// Consider switching from string to vector<string>?
```
mzegla, line 368: 👍

demos/continous_batching/README.md
```diff
@@ -0,0 +1,155 @@
...
+## Client code
+
+Both unary and streaming calls should be available via the same servable:
+
+### Unary:
+```bash
+curl http://localhost:8000/v3/chat/completions \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "llama",
```
dtrawins, line 77: change to meta-llama/Llama-2-7b-chat-hf

demos/continous_batching/README.md
```diff
@@ -0,0 +1,155 @@
...
+    "messages": [
+      {
+        "role": "system",
```
dtrawins, line 80: let's start with a basic call with just user role to make is work with all models

demos/continous_batching/README.md
```diff
@@ -0,0 +1,155 @@
...
+    if chunk.choices[0].delta.content is not None:
+        print(chunk.choices[0].delta.content, end="")
+```
+
+Output:
+```
+This is a test.
+```
+
+Refer to [OpenAI streaming documentation](https://platform.openai.com/docs/api-reference/streaming).
```
dtrawins, line 155: benchmark_Serving.py should be added to the demo scope

demos/continous_batching/README.md
```diff
@@ -0,0 +1,188 @@
+# How to serve LLM models with Continous Batching via OpenAI API
```
dtrawins, line 1:
```suggestion
# How to serve LLM models with Continous Batching via OpenAI API {#ovms_demos_continous_batching}
```

external/partial.patch
```diff
@@ -21,10 +38,29 @@ index c8d0501b..e5bb9700 100644
...
 -  NET_LOG(FATAL, "PartialReplyWithStatus not implemented.");
 +  PartialReplyWithStatus(HTTPStatusCode::OK);
++}
++
++void EvHTTPRequest::PartialReplyEx(std::string data) {
++  bool result =
++      server_->EventLoopSchedule([this, data]() { EvPartialSendReply(data); });
++  
++  // server_->EventLoopSchedule([this]() {
++  //   std::cout << "Sleep Start" << std::endl;
```
dtrawins, line 48: cleanup needed

demos/continous_batching/README.md
```diff
@@ -0,0 +1,188 @@
...
+    if chunk.choices[0].delta.content is not None:
+        print(chunk.choices[0].delta.content, end="")
+```
+
+Output:
+```
+This is a test.
+```
+
+## Continuous batching
```
dtrawins, line 176:
```suggestion
## Benchmarking text generation with high concurrency
```

src/llm/http_llm_calculator.cc
dkalinowski: yes

src/llm/http_llm_calculator.cc
dkalinowski: Can we move it outside later, once we need it?

src/llm/http_llm_calculator.cc
dkalinowski: It is only used in 2 places - 2 calculators (grpc and http), I really do not want to create another header for that and use global var

src/llm/http_llm_calculator.cc
dkalinowski: It is copied from gen AI PR. It will be moved to CB lib anyway.

demos/continous_batching/README.md
dkalinowski: done

external/partial.patch
dkalinowski: done

demos/continous_batching/README.md
dkalinowski: done

dtrawins (approved):

mzegla (approved):
Approving conditionally. There are more changes that need to follow this.

PR #2462, opened 2024-05-29
Tokenizers path fix

rasapala:
### 🛠 Summary

JIRA CVS-142381
Adding option to pass tokenizer path at runtime and preserving compile time path setup.

Searching for libopenvino_tokenizers.so based on input paths
directory path or full file path can be provided
Search order:
backup_tokenizer_lib_path - passed during runtime - default ""
input_path - passed during compile time
LD_PRELOAD
LD_LIBRARY_PATH

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/create_package.sh
+++ b/create_package.sh
@@ -22,8 +22,9 @@ mkdir -vp /ovms_release/lib
 mkdir -vp /ovms_release/lib/hddl/config
 mkdir -vp /ovms_release/lib/custom_nodes
 
-if [ -f /openvino_tokenizers/build/src/libopenvino_tokenizers.so ]; then cp -v /openvino_tokenizers/build/src/libopenvino_tokenizers.so /ovms_release/lib/ ; fi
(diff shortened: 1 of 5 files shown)


create_package.sh
```diff
@@ -22,8 +22,9 @@ mkdir -vp /ovms_release/lib
 mkdir -vp /ovms_release/lib/hddl/config
 mkdir -vp /ovms_release/lib/custom_nodes
 
-if [ -f /openvino_tokenizers/build/src/libopenvino_tokenizers.so ]; then cp -v /openvino_tokenizers/build/src/libopenvino_tokenizers.so /ovms_release/lib/ ; fi
-if [ -f /openvino_tokenizers/build/src/libcore_tokenizers.so ]; then cp -v /openvino_tokenizers/build/src/libcore_tokenizers.so /ovms_release/lib/ ; fi
+tokenizers_path=/ovms/bazel-out/k8-opt/bin/external/llm_engine/llm_engine_cmake_ubuntu/lib/
```
dkalinowski, line 25: This way we cannot use tokenizers for images without mediapipe @dtrawins @rasapala @mzegla 
It means users cannot load single models (tokenizers) with OVMS without mediapipe

PR #2468, opened 2024-06-04
Add handling for missing OpenAI parameters + profiling traces

dkalinowski:
### 🛠 Summary

CVS-142768

Forwards beam search and multinomial sampling parameters to CB library - this enables returning more than 1 completion for beam search (only for unary)
Adds profiling traces (minitrace)

### 🧪 Checklist

- [x] Unit tests added. - only functional tests are required
- [x] The documentation updated. (in separate PR https://github.com/openvinotoolkit/model_server/pull/2471)


src/llm/http_llm_calculator.cc
```diff
@@ -46,29 +47,75 @@ class OpenAIChatCompletionsRequest {
...
+
+        // Multinomial specific
+        if (temperature.has_value())
+            config.temperature = temperature.value();
+        if (topK.has_value())
+            config.top_k = topK.value();
+        if (topP.has_value())
+            config.top_p = topP.value();
+        config.do_sample = config.temperature > 0.0f && config.group_size == 1;
+        // TODO: rng_seed = ?
```
mzegla, line 103: Should this be `seed` from the request?

src/llm/http_llm_calculator.cc
```diff
@@ -46,29 +47,75 @@ class OpenAIChatCompletionsRequest {
...
+
+        // TODO: bos_token_id = ?
+        // TODO: pad_token_id = ?
+        // TODO: eos_token_id = ?
```
mzegla, line 107: This can we read from tokenizer config. Not a generation config parameter.

src/llm/http_llm_calculator.cc
```diff
@@ -131,17 +168,71 @@ class OpenAIChatCompletionsRequest {
...
+        // repetition_penalty: float; optional - defaults to 1.0
+        // Extension, unsupported by OpenAI API, however supported by vLLM and CB lib
+        it = this->doc.FindMember("repetition_penalty");
         if (it != this->doc.MemberEnd()) {
             if (!it->value.IsDouble())
                 return false;
-            this->presencePenalty = it->value.GetDouble();
-            if (this->presencePenalty < -2.0f || this->presencePenalty > 2.0f)
+            this->repetitionPenalty = it->value.GetDouble();
+            // TODO: Validate?
```
mzegla, line 202: I don't think so. For now we should probably simply catch errors from cb lib calls and eventually such validation should be done on the library side I think (non-OVMS users can also specify bad generation parameters when using it as a standalone)

mzegla (commented):
So in general we pass parameters from the request and if user doesn't pass specific params, we use defaults from CB lib right?

dtrawins (approved):

michalkulakowski (approved):

src/llm/http_llm_calculator.cc
dkalinowski: added

src/llm/http_llm_calculator.cc
dkalinowski: removed

src/llm/http_llm_calculator.cc
dkalinowski: Validation/error handling is on separate branch of @michalkulakowski

mzegla (approved):

PR #2478, opened 2024-06-10
completion endpoint

dtrawins:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/http_llm_calculator.cc
+++ b/src/llm/http_llm_calculator.cc
@@ -115,10 +115,10 @@ public:
         return config;
     }
 
-    chat_t getMessages() const { return this->messages;};
+    chat_t getMessages() const { return this->messages; };
     Endpoint getEndpoint() const { return this->endpoint; }
     void setEndpoint(Endpoint endpoint) { this->endpoint = endpoint; }
-    std::string getPrompt() const { 
+    std::string getPrompt() const {
         if (this->prompt.has_value()) {
             return this->prompt.value();
         } else {
@@ -130,7 +130,7 @@ public:
     std::string getModel() const { return this->model; }
 
     // TODO: Use exceptions to sneak error messages into response
-    bool parse( Endpoint endpoint) {
+    bool parse(Endpoint endpoint) {
         OVMS_PROFILE_FUNCTION();
         // stream: bool; optional
         if (!this->doc.IsObject())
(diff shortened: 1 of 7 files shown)


src/llm/http_llm_calculator.cc
```diff
@@ -108,11 +116,21 @@ class OpenAIChatCompletionsRequest {
...
+        }
+    }
+
     bool isStream() const { return this->stream; }
     std::string getModel() const { return this->model; }
 
-    // TODO: Use exceptions to sneak error mesages into response
-    bool parse() {
+    // TODO: Use exceptions to sneak error messages into response
+    bool parse(Endpoint endpoint) {
```
dkalinowski, line 133: There is no need to pass endpoint to `parse`, it is already a member of `OpenAIChatCompletionsRequest ` class

src/llm/http_llm_calculator.cc
```diff
@@ -442,6 +488,7 @@ class HttpLLMCalculator : public CalculatorBase {
             nodeResources->notifyExecutorThread();
             this->streamer = std::make_shared<TextStreamer>(
                 nodeResources->cbPipe->get_tokenizer());
+            this->request->setEndpoint(endpoint);
```
dkalinowski, line 491: I think it would be better to move setEndpoint to line 458/460 so that we can use it in `parse` later on. What is more, for undefined endpoint (line 462) I would RET_CHECK it early, even before calling parse.

src/llm/http_llm_calculator.cc
```diff
@@ -422,16 +452,32 @@ class HttpLLMCalculator : public CalculatorBase {
...
             InputDataType payload = cc->Inputs().Tag(INPUT_TAG_NAME).Get<InputDataType>();
             LOG(INFO) << "Request body: " << payload.body;
+            LOG(INFO) << "Request uri: " << payload.uri;
+            Endpoint endpoint;
+            if (payload.uri == "/v3/chat/completions") {
+                endpoint = CHAT_COMPLETIONS;
+            } else if (payload.uri == "/v3/completions") {
+                endpoint = COMPLETIONS;
+            } else {
+                endpoint = UNDEFINED;
```
dkalinowski, line 462: RET_CHECK and return early.

dkalinowski: Also, check with @michalkulakowski for new error handling method

dkalinowski (approved):

PR #2496, opened 2024-06-14
documentation improvements for llm demo

dtrawins:
### 🧪 Checklist

- [ ] Unit tests added.
- [X] The documentation updated.
``

### DIFF
--- a/demos/continuous_batching/README.md
+++ b/demos/continuous_batching/README.md
@@ -93,10 +93,15 @@ Meta-Llama-3-8B-Instruct
 ├── tokenizer.json
 └── tokenizer.model
 
+
 ```
+
+
 The default configuration of the `LLMExecutor` should work in most cases but the parameters can be tunned inside the `node_options` section in the `graph.pbtxt` file. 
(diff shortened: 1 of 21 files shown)


demos/continuous_batching/README.md
```diff
@@ -53,7 +59,8 @@ node: {
   }
   node_options: {
       [type.googleapis.com / mediapipe.LLMCalculatorOptions]: {
-          models_path: "./"
+          models_path: "./",
+          cache_size: 50
```
mzegla, line 63: Should it be this big?

dtrawins: This value was used in benchmarking script and the results are included. Default value might be to low. Which value do you suggest?

mzegla: 16 maybe? For someone with 32 GBs RAM to be able to comfortably load it.

dtrawins: I added a not to reduce it if there is not enough RAM. I would expect on hosts with Xeon at least 128GB on board. 
Do you think this is good enough?

dkalinowski (approved):

mzegla (approved):

PR #2497, opened 2024-06-14
Serving LLMs - documentation update

mzegla:
### 🧪 Checklist

- [ ] Unit tests added.
``


dkalinowski (approved):

docs/llm_serving.md
```diff
@@ -0,0 +1,278 @@
...
+
+With rapid development of generative AI, new techniques and algorithms for performance optimization and better resource utilization are introduced to make best use of the hardware and provide best generation performance. OpenVINO implements those state of the art methods in it's [GenAI Library](https://github.com/ilya-lavrenov/openvino.genai/tree/ct-beam-search/text_generation/causal_lm/cpp/continuous_batching/library) like:
+  - Continuous Batching
+  - Paged Attention
+  - Dynamic Split Fuse 
+  - *and more...*
+
+ It is now integrated into OpenVINO Model Server providing efficient way to run generative workloads.
+
+## Quickstart
```
dtrawins, line 15: I suggest a separate page with a quick start guide. It should use fairly small model without authorization like tinyllama.
It should be very simple and short.

docs/features.md
```diff
@@ -6,6 +6,7 @@ maxdepth: 1
 hidden:
 ---
 
+ovms_docs_llm_serving
```
dtrawins, line 9: incorrect reference

docs/llm/quickstart.md
```diff
@@ -0,0 +1,125 @@
...
+      },
+      {
+        "role": "user",
+        "content": "What is OpenVINO?"
+      }
+    ]
+  }'| jq .
+```
+```json
+XXX
```
dtrawins, line 116: missing example?

docs/llm/reference.md
```diff
@@ -0,0 +1,147 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+Above node configuration should be used as a template since user is not expected to change most of it's content. Fields that can be safely changed are:
+ - `name`
+ - `input_stream: "HTTP_REQUEST_PAYLOAD:input"` - in case you want to change input name
```
dtrawins, line 78: why would it be changed?

docs/llm/quickstart.md
mzegla: Yes, will be added once we execute the pipeline with this model

docs/llm/reference.md
```diff
@@ -0,0 +1,147 @@
...
+
+If there are errors in loading or reading files or fields (they exist but are wrong) no template is loaded and servable will not respond to `/chat/completions` calls. 
+
+If no chat template has been specified, default template is applied. The template looks as follows:
+```
+"{% if messages|length > 1 %} {{ raise_exception('This servable accepts only single message requests') }}{% endif %}{{ messages[0]['content'] }}"
+```
+
+When default template is loaded, servable accepts `/chat/completions` calls when `messages` list contains only single element (otherwise returns error) and treats `content` value of that single message as an input prompt for the model.
+
```
dtrawins, line 131: Add a note that the template is applicable only for chat/completions endpoint. completions is not using the template and the prompt need to be fully created by the client.

docs/llm/reference.md
mzegla: I'm just listing parameters that can be changed without affecting functionality. 
There's no particular reason to change this one, but if someone wants to go with for example "request" instead of "input" they can.

docs/llm/reference.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Limitations
+
+As it's in preview, this feature has set of limitations:
+
+- Limited support for [API parameters](./model_server_rest_api_chat.md#request),
+- Only completions and chat completions endpoints supported,
```
dtrawins, line 138: We don't claim to have all openai endpoints - only text generation. Other endpoints are not related to the preview. I would drop it.

docs/llm/reference.md
mzegla: It's mentioned above:

> Chat template is used only on `/chat/completions` endpoint. Template is not applied for calls to `/completions`, so it doesn't have to exist, if you plan to work only with `/completions`.

docs/llm/reference.md
```diff
@@ -0,0 +1,147 @@
...
+- Only one node with LLM calculator can be deployed at once,
+- Lack of metrics,
+- Not all performance optimizations yet applied,
```
dtrawins, line 141: This is not a functional limitation ;-) there will never be all possible optimizations. I would drop it

docs/llm/reference.md
```diff
@@ -0,0 +1,147 @@
...
+
+
+## Limitations
+
+As it's in preview, this feature has set of limitations:
+
+- Limited support for [API parameters](./model_server_rest_api_chat.md#request),
+- Only completions and chat completions endpoints supported,
+- Only one node with LLM calculator can be deployed at once,
+- Lack of metrics,
```
dtrawins, line 140:
```suggestion
- Metrics related to text generation - they are planned to be added later
```

docs/llm/reference.md
```diff
@@ -0,0 +1,147 @@
...
+- Not all performance optimizations yet applied,
+- Improvements in stability and recovery mechanisms are also expected,
```
dtrawins: Maybe it would be a place to mention about tested models? 
As a preview feature it is made available for customers to evaluate and provide feedback.

docs/llm/reference.md
dtrawins: Maybe this note should be in the template section?
In the model_dir section just mention that template.jinja should be added only if it is not provided by the HF model.

docs/llm/reference.md
dtrawins: I see but I think we are complicating things more than needed here. Users should just change the node options.

dtrawins (approved):

PR #2510, opened 2024-06-18
Add verbosity flag for image build

dtrawins:
### 🛠 Summary

This patch changes the log verbosity of the redhat Dockerfile to be controlled by an
argument. It will default to off to maintain the original behavior. It can be overridden
by --build-arg VERBOSE_LOGS=ON or similar depending on how containers are built.

### 🧪 Checklist

- [ ] Unit tests added.
``


atobiszei (approved):

mzegla (approved):
Do we want the same kind of change for Ubuntu?

PR #2512, opened 2024-06-19
Provide LLM testing pipeline with a model

mzegla:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``


prepare_llm_models.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+echo "Downloading LLM testing models to directory $1"
+python3 -m venv .venv
+. .venv/bin/activate
+pip3 install -U pip
+export PIP_EXTRA_INDEX_URL="https://download.pytorch.org/whl/cpu https://storage.openvinotoolkit.org/simple/wheels/nightly"
+pip3 install --pre "optimum-intel[nncf,openvino]"@git+https://github.com/huggingface/optimum-intel.git openvino-tokenizers
+optimum-cli export openvino --disable-convert-tokenizer --model facebook/opt-125m --weight-format int8 $1/facebook/opt-125m
+convert_tokenizer -o $1/facebook/opt-125m --with-detokenizer --skip-special-tokens --streaming-detokenizer --not-add-special-tokens facebook/opt-125m
```
ngrozae, line 30: $1 would be full path to model already. from Mekefile: ./prepare_llm_models.sh ${TEST_LLM_PATH}/facebook/opt-125m

ci/build_test_OnCommit.groovy
```diff
@@ -65,7 +65,7 @@ pipeline {
           parallel {
             stage("Run unit tests") {
               steps {
-                  sh "make run_unit_tests BASE_OS=redhat OVMS_CPP_IMAGE_TAG=${shortCommit}"
+                  sh "make run_unit_tests TEST_LLM_PATH=/opt/home/k8sworker/ovms_models BASE_OS=redhat OVMS_CPP_IMAGE_TAG=${shortCommit}"
```
ngrozae, line 68: is it ok to set here full path? maybes use: TEST_LLM_PATH=~/ovms_models

Makefile
```diff
@@ -75,6 +75,7 @@ OV_SOURCE_ORG ?= openvinotoolkit
 OV_CONTRIB_ORG ?= openvinotoolkit
 
 TOKENIZERS ?= 1
+TEST_LLM_PATH ?= "llm_testing"
```
dtrawins, line 78: could use use /tmp/llm_testing instead

prepare_llm_models.sh
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+#
+
+if [ -d "$1" ]; then
+  echo "Models directory $1 exists. Skipping downloading models."
+  exit 0
+fi
+
+echo "Downloading LLM testing models to directory $1"
+python3 -m venv .venv
```
dtrawins, line 24: It might fail if the venv is already created. Maybe we should add existing venv target dependency instead in the Makefile?

Makefile
```diff
@@ -648,6 +649,7 @@ cpu_extension:
 	docker cp $$(docker create --rm sample_cpu_extension:latest):/workspace/libcustom_relu_cpu_extension.so ./lib/${OS}
 
 run_unit_tests:
-	docker run -e https_proxy=${https_proxy} -e RUN_TESTS=1 -e JOBS=$(JOBS) -e debug_bazel_flags=${BAZEL_DEBUG_FLAGS} $(OVMS_CPP_DOCKER_IMAGE)-build:$(OVMS_CPP_IMAGE_TAG)$(IMAGE_TAG_SUFFIX) ./rununittest.sh > test.log 2>&1 ; exit_status=$?
+	./prepare_llm_models.sh ${TEST_LLM_PATH}/facebook/opt-125m
+	docker run -v $(realpath ${TEST_LLM_PATH})/facebook/opt-125m:/llm_testing/facebook/opt-125m -e https_proxy=${https_proxy} -e RUN_TESTS=1 -e JOBS=$(JOBS) -e debug_bazel_flags=${BAZEL_DEBUG_FLAGS} $(OVMS_CPP_DOCKER_IMAGE)-build:$(OVMS_CPP_IMAGE_TAG)$(IMAGE_TAG_SUFFIX) ./rununittest.sh > test.log 2>&1 ; exit_status=$?
```
dtrawins, line 653: maybe whole ${TEST_LLM_PATH} should be mounted in case more models will be added later?

src/test/llmnode_test.cpp
```diff
@@ -184,7 +183,7 @@ TEST_F(LLMOptionsKfsTest, DISABLED_LLMNodeOptionsCheckDefault) {
             output_stream: "RESPONSE:out1"
             node_options: {
                 [type.googleapis.com / mediapipe.LLMCalculatorOptions]: {
-                    models_path: "/workspace/"
+                    models_path: "/llm_testing/facebook/opt-125m"
```
dtrawins, line 186: It would be the best if it works also in interactive sessions based on developer guide. Right now it will not.
option1 - update docker run command in developer guide to add mounting /llm_testing
option2 - use default path in the git repo and use /ovms/llm_testing path instead.

prepare_llm_models.sh
mzegla: It doesn't fail, I tested running it few times in a row with different paths and it worked. I don't know about using same venv as we use for everything else. This one is pretty heavy and keeping it separate gives as easier option to split LLM test execution if we want to do that in the future.

prepare_llm_models.sh
mzegla: changed Makefile

src/test/llmnode_test.cpp
mzegla: going with option 2, so we don't expand container setup

ngrozae (approved):

dtrawins (approved):

PR #2516, opened 2024-06-20
Detect OS for cmake

rasapala:
### 🛠 Summary

JIRA CVS-140004
Detect os in _impl of cmake dependencies instead of defining two cmake implementations - redhat,ubuntu.

### 🧪 Checklist

- [x] Unit tests added.
``

### DIFF
--- a/third_party/aws-sdk-cpp/aws-sdk-cpp.bzl
+++ b/third_party/aws-sdk-cpp/aws-sdk-cpp.bzl
@@ -32,6 +32,16 @@ def aws_sdk_cpp():
 def _impl(repository_ctx):
     http_proxy = repository_ctx.os.environ.get("http_proxy", "")
     https_proxy = repository_ctx.os.environ.get("https_proxy", "")
+
+    result = repository_ctx.execute(["cat","/etc/os-release"],quiet=False)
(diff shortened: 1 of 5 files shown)


third_party/aws-sdk-cpp/aws-sdk-cpp.bzl
```diff
@@ -114,78 +124,14 @@ cmake(
 )
 
 cc_library(
-    name = "aws-sdk-cpp_ubuntu",
+    name = "aws-sdk-cpp",
     deps = [
-        ":aws-sdk-cpp_cmake_ubuntu",
```
mzegla, line 119: That's a big block removed. Are you sure we can do that safely? 
All that building options and libraries...

mzegla: I see it's also in former ubuntu target. If this is all the same and we just remove duplicates then I'm okay with it.

rasapala: We remove duplicates.

mzegla (approved):

dtrawins (approved):

PR #2525, opened 2024-06-27
validate if the end of decoded string is a valid uft8 content

dtrawins:
### 🛠 Summary

CVS-145315
LLM stream was handling incorrectly incomplete Unicode bytes. With this change the streamer is validating if the end of detokenized string consists of a valid utf8 encoding. When invalid content is detected, streamer will wait for more content to send the chunk.

### 🧪 Checklist

- [x] Unit tests added.
``

### DIFF
--- a/src/llm/http_llm_calculator.cc
+++ b/src/llm/http_llm_calculator.cc
@@ -374,7 +374,7 @@ public:
             tokenCache.clear();
             printLen = 0;
             return chunk;
-        } else if (isValidUtf8(text) != true) {
+        } else if (!isValidUtf8(text)) {
             return std::nullopt;
         } else if (text.size() > printLen) {
             // The chunk is ready if the new text in the cache contains space.
--- a/src/llm/http_llm_calculator.cc
+++ b/src/llm/http_llm_calculator.cc
@@ -32,6 +32,7 @@
 #include <rapidjson/writer.h>
 
 #include "../profiler.hpp"
+#include "../stringutils.hpp"
 #include "http_payload.hpp"
 #include "llmnoderesources.hpp"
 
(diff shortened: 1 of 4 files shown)


src/stringutils.cpp
```diff
@@ -152,4 +152,36 @@ std::optional<int64_t> stoi64(const std::string& str) {
...
+        } else if ((x >> 4) == 0b1110) {  // first byte of 3 byte sequence
+            if (byte_counter + 1 == 3)
+                return true;
+            else
+                return false;
+        } else if ((x >> 3) == 0b11110) {  // first byte of 3 byte sequence
+            if (byte_counter + 1 == 4)
+                return true;
+            else
+                return false;
```
mzegla, line 179: We could make this block shorter if conditionals are refactored to for example:
```
else if ((x >> <shift> == <val>) && (byte_counter + 1 != <val>))
	return false;
```
Then conditions only check bad values and if it's not detected we get to the end of that function so the `return true` at the end is effective.

mzegla (approved):

src/llm/http_llm_calculator.cc
```diff
@@ -391,7 +392,8 @@ class TextStreamer {
     }
 };
 
-static bool applyChatTemplate(TextProcessor& textProcessor, std::string modelsPath, std::string& requestBody, std::string& output) {
+static bool
+applyChatTemplate(TextProcessor& textProcessor, std::string modelsPath, std::string& requestBody, std::string& output) {
```
michalkulakowski, line 396:
```suggestion
static bool applyChatTemplate(TextProcessor& textProcessor, std::string modelsPath, std::string& requestBody, std::string& output) {
```

michalkulakowski (approved):

PR #2527, opened 2024-06-27
Add UTs for LLM parameters validation

michalkulakowski:
### 🛠 Summary

https://jira.devtools.intel.com/browse/CVS-144460
https://jira.devtools.intel.com/browse/CVS-144458

Integration tests for LLM + for LLM parameters validation

### 🧪 Checklist

- [x] Unit tests added.
- ~[ ] The documentation updated.~
- ~[ ] Change follows security best practices.~
``

### DIFF
--- a/prepare_llm_models.sh
+++ b/prepare_llm_models.sh
@@ -27,5 +27,5 @@ pip3 install -U pip
 export PIP_EXTRA_INDEX_URL="https://download.pytorch.org/whl/cpu https://storage.openvinotoolkit.org/simple/wheels/nightly"
 pip3 install --pre "optimum-intel[nncf,openvino]"@git+https://github.com/huggingface/optimum-intel.git openvino-tokenizers
 optimum-cli export openvino --disable-convert-tokenizer --model facebook/opt-125m --weight-format int8 $1/facebook/opt-125m
-convert_tokenizer -o $1/facebook/opt-125m --with-detokenizer --skip-special-tokens --streaming-detokenizer --not_add-special-tokens facebook/opt-125m
+convert_tokenizer -o $1/facebook/opt-125m --with-detokenizer --skip-special-tokens --streaming-detokenizer --not-add-special-tokens facebook/opt-125m
 
--- a/prepare_llm_models.sh
+++ b/prepare_llm_models.sh
@@ -27,5 +27,5 @@ pip3 install -U pip
 export PIP_EXTRA_INDEX_URL="https://download.pytorch.org/whl/cpu https://storage.openvinotoolkit.org/simple/wheels/nightly"
 pip3 install --pre "optimum-intel[nncf,openvino]"@git+https://github.com/huggingface/optimum-intel.git openvino-tokenizers
 optimum-cli export openvino --disable-convert-tokenizer --model facebook/opt-125m --weight-format int8 $1/facebook/opt-125m
-convert_tokenizer -o $1/facebook/opt-125m --with-detokenizer --skip-special-tokens --streaming-detokenizer --not-add-special-tokens facebook/opt-125m
+convert_tokenizer -o $1/facebook/opt-125m --with-detokenizer --skip-special-tokens --streaming-detokenizer --not_add-special-tokens facebook/opt-125m
 
--- a/src/test/llmnode_test.cpp
+++ b/src/test/llmnode_test.cpp
@@ -151,15 +151,15 @@ TEST_F(LLMFlowHttpTest, inferChatCompletionsUnary) {
         }
     )";
 
-    std::time_t t = std::time(0);
     ASSERT_EQ(
         handler->dispatchToProcessor(endpointChatCompletions, requestBody, &response, comp, responseComponents, &writer),
         ovms::StatusCode::OK);
-    std::stringstream ss;
-    ss << t;
-    std::string time = ss.str();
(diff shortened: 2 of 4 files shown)


prepare_llm_models.sh
```diff
@@ -27,5 +27,5 @@ pip3 install -U pip
 export PIP_EXTRA_INDEX_URL="https://download.pytorch.org/whl/cpu https://storage.openvinotoolkit.org/simple/wheels/nightly"
 pip3 install --pre "optimum-intel[nncf,openvino]"@git+https://github.com/huggingface/optimum-intel.git openvino-tokenizers
 optimum-cli export openvino --disable-convert-tokenizer --model facebook/opt-125m --weight-format int8 $1/facebook/opt-125m
-convert_tokenizer -o $1/facebook/opt-125m --with-detokenizer --skip-special-tokens --streaming-detokenizer --not-add-special-tokens facebook/opt-125m
+convert_tokenizer -o $1/facebook/opt-125m --with-detokenizer --skip-special-tokens --streaming-detokenizer --not_add-special-tokens facebook/opt-125m
```
mzegla, line 30: That looks awkward, are you sure it's correct change? 
I see that continuous batching demo uses `--not-add...`

src/test/llmnode_test.cpp
```diff
@@ -59,39 +61,71 @@
...
+    std::string response;
+    ovms::HttpResponseComponents responseComponents;
+
+    static void SetUpTestSuite() {
+        std::string port = "9173";
+        ovms::Server& server = ovms::Server::instance();
+        ::SetUpServer(t, server, port, "/ovms/src/test/llm/config_llm_dummy_kfs.json");
+        auto start = std::chrono::high_resolution_clock::now();
+        while ((server.getModuleState(ovms::SERVABLE_MANAGER_MODULE_NAME) != ovms::ModuleState::INITIALIZED) &&
+               (std::chrono::duration_cast<std::chrono::seconds>(std::chrono::high_resolution_clock::now() - start).count() < 5)) {
```
mzegla, line 107: What is this 5? Can we move it to the const var?

src/test/llmnode_test.cpp
```diff
@@ -101,32 +135,524 @@ class LLMOptionsHttpTest : public ::testing::TestWithParam<std::string> {
...
+
+    ASSERT_EQ(
+        handler->dispatchToProcessor(endpointChatCompletions, requestBody, &response, comp, responseComponents, &writer),
+        ovms::StatusCode::OK);
+    // Assertion split in two parts to avoid timestamp missmatch
+    const size_t timestampLength = 10;
+    std::string expectedResponsePart1 = R"({"choices":[{"finish_reason":"stop","index":0,"logprobs":null,"message":{"content":"\nOpenVINO is","role":"assistant"}}],"created":)";
+    std::string expectedResponsePart2 = R"(,"model":"llmDummyKFS","object":"chat.completion"})";
+    ASSERT_EQ(response.compare(0, expectedResponsePart1.length(), expectedResponsePart1), 0);
+    ASSERT_EQ(response.compare(expectedResponsePart1.length() + timestampLength, expectedResponsePart2.length(), expectedResponsePart2), 0);
```
mzegla, line 162: I think it's risky to compare output of a real model with fixed text in oncommit tests. This can fail of different hardware due to some internal math computation differences. I suppose we do that in more places.

src/test/llmnode_test.cpp
```diff
@@ -101,32 +135,524 @@ class LLMOptionsHttpTest : public ::testing::TestWithParam<std::string> {
...
+        }
+    )";
+
+    EXPECT_CALL(writer, PartialReplyEnd()).Times(1);
+    EXPECT_CALL(writer, PartialReply(::testing::_)).Times(3);
+    EXPECT_CALL(writer, WriteResponseString(::testing::_)).Times(0);
+    ASSERT_EQ(
+        handler->dispatchToProcessor(endpointChatCompletions, requestBody, &response, comp, responseComponents, &writer),
+        ovms::StatusCode::PARTIAL_END);
+    ASSERT_EQ(response, "");
```
mzegla, line 209: I guess number of partial replies may also differ depending on what model generates. Assuming that it always generates the same configuration of tokens is dangerous.

src/test/llmnode_test.cpp
```diff
@@ -59,39 +61,71 @@
...
-    void SetUp() override {
+    void SetUp() {
+        ovms::Server& server = ovms::Server::instance();
+        handler = std::make_unique<ovms::HttpRestApiHandler>(server, 5);
+        ASSERT_EQ(handler->parseRequestComponents(comp, "POST", endpointChatCompletions, headers), ovms::StatusCode::OK);
     }
-    void TearDown() {
+
+    static void TearDownTestSuite() {
+        ovms::Server& server = ovms::Server::instance();
```
rasapala, line 118: Why setting server in tear down ?

rasapala (approved):

prepare_llm_models.sh
dtrawins: probably wrong version of convert_tokenizer

src/test/llmnode_test.cpp
dtrawins: with INFERENCE_PRECISION_HINT the results should be deterministic

mzegla (approved):

PR #2549, opened 2024-07-11
Fix number of outputs is based on best_of parameter instead of n

bstrzele:
### 🛠 Summary

JIRA/Issue if applicable.
CVS-144461
CVS-144380

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/http_llm_calculator.cc
+++ b/src/llm/http_llm_calculator.cc
@@ -730,10 +730,8 @@ std::string HttpLLMCalculator::serializeStreamingChunk(const std::string& chunkR
         }
         writer.EndObject();  // }
     } else if (endpoint == Endpoint::COMPLETIONS) {
-        if (!stop) {
-            writer.String("text");
-            writer.String(chunkResponse.c_str());
-        }
+        writer.String("text");
+        writer.String(chunkResponse.c_str());
     }
     // TODO: tools_call
     // TODO: function_call (deprecated)
--- a/src/llm/http_llm_calculator.cc
+++ b/src/llm/http_llm_calculator.cc
@@ -127,6 +127,7 @@ public:
     chat_t getMessages() const { return this->messages; }
     Endpoint getEndpoint() const { return this->endpoint; }
     std::optional<std::string> getPrompt() const { return this->prompt; }
+    std::optional<int> getNumReturnSequences() const { return this->numReturnSequences; }
 
     bool isStream() const { return this->stream; }
     std::string getModel() const { return this->model; }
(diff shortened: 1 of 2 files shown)


src/llm/http_llm_calculator.cc
```diff
@@ -730,10 +734,8 @@ std::string HttpLLMCalculator::serializeStreamingChunk(const std::string& chunkR
         }
         writer.EndObject();  // }
     } else if (endpoint == Endpoint::COMPLETIONS) {
-        if (!stop) {
-            writer.String("text");
-            writer.String(chunkResponse.c_str());
-        }
+        writer.String("text");
+        writer.String(chunkResponse.c_str());
```
mzegla, line 738: Why do you make this change? If it's needed here isn't it also needed in chat completions endpoint?

src/llm/http_llm_calculator.cc
```diff
@@ -616,7 +617,10 @@ std::string HttpLLMCalculator::serializeUnaryResponse(const std::vector<std::str
     writer.String("choices");
     writer.StartArray();  // [
     int i = 0;
+    int n = this->request->getNumReturnSequences().value_or(1);
     for (const std::string& completeResponse : completeResponses) {
+        if (i >= n)
+            break;
```
mzegla, line 623: This is not enough. When applying `n` filter we must first sort `best_of` results and then pick `n` with highest score. 
See: https://github.com/mzegla/openvino.genai/blob/master/src/cpp/src/continuous_batching_pipeline.cpp#L259
I think we could do that after `read_all()` is called to create a subset of `n` best results.

src/llm/http_llm_calculator.cc
bstrzele: This was a change to prevent crashes while using [vllm benchmarking script](https://github.com/vllm-project/vllm/blob/main/benchmarks/benchmark_serving.py). I haven't found an example where this would be necessary for the other endpoint, but I guess the other one might also be adjusted for consistency.

src/llm/http_llm_calculator.cc
```diff
@@ -536,6 +537,9 @@ class HttpLLMCalculator : public CalculatorBase {
             if (!this->request->isStream()) {
                 OVMS_PROFILE_SCOPE("Unary generation cycle");
                 std::vector<GenerationOutput> generationOutput = this->generationHandle->read_all();
+                std::sort(generationOutput.begin(), generationOutput.end(), [=](GenerationOutput& r1, GenerationOutput& r2) {
+                    return r1.score > r2.score;
+                });
                 RET_CHECK(generationOutput.size() >= 1);
```
mzegla, line 543: Maybe let's move this check above sorting. Functionally it doesn't make a difference, but it looks weird when we first use and then check the vector.

src/llm/http_llm_calculator.cc
mzegla: @dkalinowski do you recall why this condition was added in the first place?

dkalinowski: For chat completions OpenAI API specifies that we should return `delta: {}`, we should not remove that
For completions it was in fact a bug that we returned empty message

mzegla: Okay @bstrzele in such case this implementation is good.

rasapala (approved):

PR #2550, opened 2024-07-12
Fix non python build

rasapala:
### 🛠 Summary

JIRA CVS-146648


mzegla (commented):
Those files are in LLM directory and this entire functionality is not supposed to work without Python. Can we handle this in BUILD files instead of using if defs?
Without Python enabled there's no point in building this entire catalog.

src/llm/chat_template.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <string>
+
+// Python execution for template processing
+#include <pybind11/embed.h>  // everything needed for embedding
+#include <pybind11/stl.h>
+
+namespace ovms {
+class TextProcessor;
+
+bool applyChatTemplate(TextProcessor& textProcessor, std::string modelsPath, std::string& requestBody, std::string& output);
```
mzegla, line 27: If we introduce a new file maybe we could name it text_processor.cpp / hpp and move entire TextProcessor class here. Then we could also make applyChatTemplate and loadTextProcessor methods of such class and have it all encapsulated. What do you think?

atobiszei (approved):

michalkulakowski (approved):

PR #2551, opened 2024-07-15
Support specifying max_tokens and best_of limits in graph config

michalkulakowski:
### 🛠 Summary
https://jira.devtools.intel.com/browse/CVS-144381

### 🧪 Checklist

- [x] Unit tests added.
``

### DIFF
--- a/src/llm/llm_calculator.proto
+++ b/src/llm/llm_calculator.proto
@@ -41,4 +41,8 @@ message LLMCalculatorOptions {
     optional string device = 8 [default = "CPU"];
 
     optional string plugin_config = 9 [default = ""];
+
+    optional uint32 best_of_limit = 10 [default = 30];
+
+    optional uint32 max_tokens_limit = 11 [default = 800];
 }
--- a/src/llm/llm_calculator.proto
+++ b/src/llm/llm_calculator.proto
@@ -42,7 +42,7 @@ message LLMCalculatorOptions {
 
     optional string plugin_config = 9 [default = ""];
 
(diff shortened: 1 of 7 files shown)


src/llm/llmnoderesources.cpp
```diff
@@ -166,6 +166,9 @@ Status LLMNodeResources::createLLMNodeResources(std::shared_ptr<LLMNodeResources
 
     nodeResources->initiateGeneration();
 
+    nodeResources->maxTokensLimit = nodeOptions.max_tokens_limit();
+    nodeResources->bestOfLimit = nodeOptions.best_of_limit();
```
mzegla, line 170: Please move this above `initiateGeneration()` call. 
Also can we create some structure or map with `requestLimits` instead of separate fields? It would be easier to extend in the future. Then you could pass `nodeResource->requestLimits` to `parse` method.

src/llm/llm_calculator.proto
```diff
@@ -41,4 +41,8 @@ message LLMCalculatorOptions {
     optional string device = 8 [default = "CPU"];
 
     optional string plugin_config = 9 [default = ""];
+
+    optional uint32 best_of_limit = 10 [default = 20];
+
+    optional uint32 max_tokens_limit = 11 [default = 4096];
```
mzegla, line 47: Does mediapipe allow nested structure in options? Maybe we could get it under some `limits` map?

dtrawins (approved):

mzegla (approved):

PR #2562, opened 2024-07-17
More frequent token cache flush when streaming

dkalinowski:
### 🛠 Summary

Changing text streamer to flush token cache more frequently

### 🧪 Checklist

 Unit tests added. - not needed
 The documentation updated. - not needed
 Change follows security best practices.


dkalinowski:
Closing since it does not improve performance

PR #2563, opened 2024-07-18
Add text streamer tests

michalkulakowski:
### 🛠 Summary
Unit tests for text streamer
https://jira.devtools.intel.com/browse/CVS-144455

### 🧪 Checklist

- [x] Unit tests added.
- ~[ ] The documentation updated.~
- ~[ ] Change follows security best practices.~
``


src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+        streamer = std::make_shared<ovms::TextStreamer>(tokenizer);
+    }
+    void TearDown() {
+        nodeResources.reset();
+        py::finalize_interpreter();
+    }
+};
+
+TEST_F(TextStreamerTest, noValueReturned) {
+    auto tokens = tokenizer->encode("TEST");
```
dkalinowski, line 54: The test is not descriptive enough - why is the value expected not to be returned? Please explain that it is due to the fact that tokenized string does not contain space nor new line

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+    for (size_t i = 0; i < tokens.get_size(); i++) {
```
dkalinowski, line 55: We have no control over the tokenizer, if at some point in the future the tokenizer will change its behavior to return empty vector, this test will **pass**, but in fact it will not test your intended use case.

To protect against that I would suggest to provide tokens yourself instead of tokenizing the text. (You can use the tokenizer to get the tokens, but assert that those tokens are exactly the same as you expect before running the test)

The same comment can be applied for all the unit tests in this PR
Please let me know your opinion about this

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+        EXPECT_FALSE(rval.has_value());
+    }
+}
+
+TEST_F(TextStreamerTest, valueReturned) {
+    std::string testPrompt = "TEST\n";
+    auto tokens = tokenizer->encode(testPrompt);
+    for (size_t i = 0; i < tokens.get_size(); i++) {
+        std::optional<std::string> rval = this->streamer->put(static_cast<int64_t*>(tokens.data())[i]);
+        if (i < tokens.get_size() - 1) {
```
dkalinowski, line 66: Please add comment that this test case is for all tokens other than the last, and the else is for the last token, it will be easier for the reader in the future

src/test/text_streamer_test.cpp
dkalinowski: Once we have that in test it will be easier to read it, otherwise the reader never knows how many times the for loop is expected to run

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+}
+
+TEST_F(TextStreamerTest, valueReturnedCacheNotCleared) {
+    std::string testPrompt1 = "TEST";
+    auto tokens = tokenizer->encode(testPrompt1);
+    for (size_t i = 0; i < tokens.get_size(); i++) {
+        std::optional<std::string> rval = this->streamer->put(static_cast<int64_t*>(tokens.data())[i]);
+        EXPECT_FALSE(rval.has_value());
```
dkalinowski, line 80: The expected behavior is that even if there is no space or no new line, the user will get remaining text anyway. I think the test showcases an error in implementation (differences between unary and streaming)

dkalinowski: I think we are missing this: https://github.com/huggingface/transformers/blob/main/src/transformers/generation/streamers.py#L116

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+}
+
+TEST_F(TextStreamerTest, valueReturnedTextWithSpaces) {
+    std::string testPrompt = "TEST TEST TEST TEST";
+    auto tokens = tokenizer->encode(testPrompt);
+    for (size_t i = 0; i < tokens.get_size(); i++) {
+        std::optional<std::string> rval = this->streamer->put(static_cast<int64_t*>(tokens.data())[i]);
+        size_t numberOfTokensBeforeValueReturned = 2;
```
dkalinowski, line 124: The same comment as above - it is not clear to me which tokens we talk about, what are the steps in this test. Once we have it I will be able to understand what we test here

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+    }
+    void assertTokensValues(ov::Tensor generatedTokens, std::vector<int64_t> expectedTokens) {
+        ASSERT_EQ(generatedTokens.get_size(), expectedTokens.size());
+        for (size_t i = 0; i < generatedTokens.get_size(); i++) {
+            ASSERT_EQ(static_cast<int64_t*>(generatedTokens.data())[i], expectedTokens[i]);
+        }
+    }
+};
+
+TEST_F(TextStreamerTest, noValueReturneStringWitoutNewLineOrSpace) {
```
dkalinowski, line 59:
```suggestion
TEST_F(TextStreamerTest, noValueReturnedStringWitoutNewLineOrSpace) {
```

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+        if (i < tokens.get_size() - 1) {
+            EXPECT_FALSE(rval.has_value());
+        } else {
+            EXPECT_TRUE(rval.has_value());
+            EXPECT_EQ(rval.value().compare(testPrompt), 0);
+        }
+    }
+}
+
+TEST_F(TextStreamerTest, valueReturnedTextWithSpaces) {
```
dkalinowski, line 135: We are missing the tests:
- what if adding the token adds new line in the middle of the text? example: `abc\ndef`, will it flush?

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+            EXPECT_EQ(rval.value().compare(testPrompt1 + testPrompt2), 0);
+        }
+    }
+}
+
+TEST_F(TextStreamerTest, valueReturnedCacheCleared) {
+    std::string testPrompt = "TEST\n";
+    auto tokens = tokenizer->encode(testPrompt);
+    std::vector<int64_t> expectedTokens = {565, 4923, 50118};
+    assertTokensValues(tokens, expectedTokens);
```
dkalinowski, line 113:
```suggestion
    assertTokensValues(tokens, {565, 4923, 50118});
```

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+    for (size_t i = 0; i < tokens.get_size(); i++) {
+        std::optional<std::string> rval = this->streamer->put(static_cast<int64_t*>(tokens.data())[i]);
```
dkalinowski, line 115:
```suggestion
        std::optional<std::string> partialResponseText = this->streamer->put(static_cast<int64_t*>(tokens.data())[i]);
```

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+        if (i < tokens.get_size() - 1) {
```
dkalinowski, line 116:
```suggestion
        if (i < tokens.get_size() - 1) {  // all iterations besides than the last should not return partial response text
```

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+            EXPECT_FALSE(rval.has_value());
+        } else {
```
dkalinowski, line 118:
```suggestion
        } else {  // the last iteration should only teturn partial response text
```

src/test/text_streamer_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+            EXPECT_TRUE(rval.has_value());
+            EXPECT_EQ(rval.value().compare(testPrompt), 0);
+        }
+    }
+    tokens = tokenizer->encode(testPrompt);
+    for (size_t i = 0; i < tokens.get_size(); i++) {
+        std::optional<std::string> rval = this->streamer->put(static_cast<int64_t*>(tokens.data())[i]);
```
dkalinowski, line 125: It can probably be simplified:
```suggestion
        std::optional<std::string> rval = this->streamer->put(tokens.as<int64_t>()[i]);
```

dkalinowski (approved):

src/test/text_streamer_test.cpp
rasapala: Spelling :// the last iteration should only return partial response text

rasapala (approved):

PR #2583, opened 2024-07-26
Fix loading Qwen chat template

dkalinowski:
### 🛠 Summary

There was an issue when loading bos_token and eos_token from tokenizer_config.json.
OVMS assumed that if it exists, it can only be string, however, Qwen has null value.
It causes chat template loading to abort.

The fix to this issue is to override bos/eos tokens when it is not a string in config file.

CVS-147900

### 🧪 Checklist

- [x] Unit tests added.

### DIFF
--- a/src/llm/llmnoderesources.cpp
+++ b/src/llm/llmnoderesources.cpp
@@ -87,7 +87,9 @@ void LLMNodeResources::loadTextProcessor(std::shared_ptr<LLMNodeResources>& node
                 f = open(templates_directory + "/tokenizer_config.json")
                 data = json.load(f)
                 bos_token = data.get("bos_token", "")
+                bos_token = bos_token if isinstance(bos_token, str) else ""  # tokenizer_config.json allows for different types than string
                 eos_token = data.get("eos_token", "")
+                eos_token = eos_token if isinstance(eos_token, str) else ""  # tokenizer_config.json allows for different types than string
                 chat_template = data.get("chat_template", default_chat_template)
 
             if template is None:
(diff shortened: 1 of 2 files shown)


rasapala (approved):

src/llm/llmnoderesources.cpp
```diff
@@ -104,6 +106,12 @@ void LLMNodeResources::loadTextProcessor(std::shared_ptr<LLMNodeResources>& node
     } catch (const pybind11::error_already_set& e) {
         SPDLOG_INFO(CHAT_TEMPLATE_WARNING_MESSAGE);
         SPDLOG_DEBUG("Chat template loading failed with error: {}", e.what());
+    } catch (const pybind11::cast_error& e) {
+        SPDLOG_INFO(CHAT_TEMPLATE_WARNING_MESSAGE);
+        SPDLOG_DEBUG("Chat template loading failed with error: {}", e.what());
+    } catch (const pybind11::key_error& e) {
+        SPDLOG_INFO(CHAT_TEMPLATE_WARNING_MESSAGE);
+        SPDLOG_DEBUG("Chat template loading failed with error: {}", e.what());
```
mzegla, line 114: Aren't cast and key error under error_already_set in the hierarchy? Weren't they caught before?
I wonder if we need that block if we are not changing error messages.

dkalinowski: They were caught in last catch block with no specified exception type. It turns out that not all errors fall into error_already_set error. Catching other types allow for easier debug since the error with wrong type (str/null) was not in the logs, now it will be @mzegla

dkalinowski: It changes error message - in case of this error we have `e.what()`

mzegla (commented):
Can we have a test for that?

dkalinowski:
@rasapala @mzegla  I don't think so, we have agreed during test planning that we will add `template.jinja` feature to test templates outside of `tokenizer_config.json`

mzegla:
There's a dedicated test file for template:
https://github.com/openvinotoolkit/model_server/blob/main/src/test/llmtemplate_test.cpp
Can't we have a separate case there with non string bos and eos?

dkalinowski:
@mzegla you are right, added the tests

mzegla (approved):

PR #2611, opened 2024-08-02
Microdnf issue

stevegrubb:
The current demo is coded to use microdnf. This fixes it to use either dnf or microdnf whichever is on the image.


mzegla (approved):

stevegrubb:
Actually, this needs one more thing:  "-y" Let me add that before merging.

stevegrubb:
Now it's ready.

But at the same time, I am wondering if the requirements should be installed from the main Dockerfile so that the main container image has everything to do the llm demos. I really think that is needed.

PR #2630, opened 2024-08-19
Fix release image content validation

rasapala:
### 🛠 Summary
Update the list of whitelisted libraries for image release and change testing flow to be outside of image building.

``


dtrawins:
I see a few problems with running such test inside docker build. It will fail with:
nvidia plugin NVIDIA=1 build because of extra library
without tokenizer TOKENIZER=1
with different OV version via DLDT_PACKAGE_URL or src commit
I suggest to remove this step from docker build step entirely. Instead there could be an extra test step in CI to run in on release image after mounting this test folder. It will ensure we are testing the version with correct parameters. 
Beside, the lib_files and lib_files_python are almost identical except python folder.

Makefile
```diff
@@ -662,3 +662,6 @@ cpu_extension:
 run_unit_tests:
 	./prepare_llm_models.sh ${TEST_LLM_PATH}
 	docker run -v $(realpath ${TEST_LLM_PATH}):/ovms/llm_testing:ro -e https_proxy=${https_proxy} -e RUN_TESTS=1 -e JOBS=$(JOBS) -e debug_bazel_flags=${BAZEL_DEBUG_FLAGS} $(OVMS_CPP_DOCKER_IMAGE)-build:$(OVMS_CPP_IMAGE_TAG)$(IMAGE_TAG_SUFFIX) ./rununittest.sh > test.log 2>&1 ; exit_status=$$? ; tail -200 test.log ; exit $$exit_status
+
+run_lib_files_test:
```
dtrawins, line 666: why not to run it on the release image?

tests/file_lists/test_release_files.sh
```diff
@@ -17,15 +17,15 @@
 # This script should be used inside the release image to check expected file contents
 
 if [ "$#" -ne 1 ]; then
-    printf --  "ERROR: Missing script argument PYTHON_DISABLE. Please pass 0 or 1 to the script.\n"
+    printf --  "ERROR: Missing script argument debug_bazel_flags. Please pass it to the script.\n"
```
mzegla, line 20: Not sure if this help is actually helpful. What is `debug_bazel_flags`? What are the options and in what format I should pass it?

ci/build_test_OnCommit.groovy
```diff
@@ -56,6 +56,7 @@ pipeline {
           when { expression { image_build_needed == "true" } }
           steps {
                 sh "make ovms_builder_image RUN_TESTS=0 OV_USE_BINARY=1 BASE_OS=redhat OVMS_CPP_IMAGE_TAG=${shortCommit}"
+                sh "make targz_package RUN_TESTS=0 OV_USE_BINARY=1 BASE_OS=redhat OVMS_CPP_IMAGE_TAG=${shortCommit}"
```
dtrawins, line 59: if the test is run on the release image, that step is not needed.

dtrawins (approved):

tests/file_lists/test_release_files.sh
```diff
@@ -17,18 +17,18 @@
 # This script should be used inside the release image to check expected file contents
 
 if [ "$#" -ne 1 ]; then
-    printf --  "ERROR: Missing script argument PYTHON_DISABLE. Please pass 0 or 1 to the script.\n"
+    printf --  "ERROR: Missing script argument debug_bazel_flags. Please pass it to the script to determine PYTHON_DISABLE value or trigger it via `make run_lib_files_test` target..\n"
```
mzegla, line 20:
```suggestion
    printf --  "ERROR: Missing script argument debug bazel flags. Please pass it to the script to configure building options for example to build without Python support run: ./test_release_files.sh PYTHON_DISABLE=1 .\n"
```

mzegla (approved):

Issue #2626, opened 2024-08-15
[FEAT] Add "Usage" field to Non-streaming OpenAI-API-Compatible LLM Response


tjtanaa:
Will there be plans to complete the support of OpenAI-API-Compatible Endpoints?
There are many existing LLM libraries and application would depend on the "Usage" field of the response when the `stream` argument is set to false.

Currently the following are the response model from the chat-completion endpoint:

```
{
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "logprobs": null,
      "message": {
        "content": "OpenVINO is an open-source project developed by Microsoft. It's designed to help users develop, train, and deploy deep learning models",
        "role": "assistant"
      }
    }
  ],
  "created": 1723712912,
  "model": "microsoft/Phi-3-mini-128k-instruct",
  "object": "chat.completion"
}
```

mzegla:
Yes, we plan to gradually increase API support for OpenAI endpoints in the future. 
As for `usage`, it is planned for the next release.

mzegla:
Support for `usage` has been merged to main branch. You can use it now if you build the docker image (run `make docker_build` from the root repo directory) or wait for 2024.4 release when the image will be published.

tjtanaa:
> Support for `usage` has been merged to main branch. You can use it now if you build the docker image (run `make docker_build` from the root repo directory) or wait for 2024.4 release when the image will be published.

Thank you very much for the updates. I will test out the new feature within these few days. 😁

PR #2634, opened 2024-08-21
Usage field support in OpenAI API - unary

mzegla:
### 🛠 Summary

Addressing: https://github.com/openvinotoolkit/model_server/issues/2626

- Adding `usage` field to completions and chat completions endpoints
- Removing old, obsolete calculator

### 🧪 Checklist

- [x] Unit tests added.
- [ ] The documentation updated. // Will update when streaming part is also added
``


src/llm/http_llm_calculator.cc
```diff
@@ -51,6 +51,12 @@ enum class Endpoint {
     COMPLETIONS,
 };
 
+struct CompletionUsageStatistics {
+    size_t promptTokens = 0;
+    size_t completionTokens = 0;
+    size_t totalTokens = 0;
```
dkalinowski, line 57: this field is unused, remove?

dkalinowski: i would suggest adding method "getTotalTokens" that simply adds prompt tokens and completion tokens. this way we could re-use this code in serialization in multiple places

src/test/llmnode_test.cpp
```diff
@@ -145,6 +145,12 @@ TEST_F(LLMFlowHttpTest, unaryCompletionsJson) {
     }
     ASSERT_EQ(d["model"], "llmDummyKFS");
     ASSERT_EQ(d["object"], "text_completion");
+
+    ASSERT_TRUE(d["usage"].IsObject());
+    ASSERT_TRUE(d["usage"].GetObject()["prompt_tokens"].IsInt());
+    ASSERT_TRUE(d["usage"].GetObject()["completion_tokens"].IsInt());
+    ASSERT_TRUE(d["usage"].GetObject()["total_tokens"].IsInt());
+    ASSERT_EQ(d["usage"].GetObject()["completion_tokens"].GetInt(), 5 /* max_tokens */);
```
dkalinowski, line 153: can we really assert the number? the model output can change

src/llm/http_llm_calculator.cc
mzegla: I created it to make this structure in-line with OpenAI `usage` field spec, but sure I can apply your suggestion.

src/test/llmnode_test.cpp
mzegla: I think it's pretty safe. We set max_tokens to very low number and it's very unlikely that model will naturally stop generating after less than 5 tokens.

rasapala (approved):

src/llm/http_llm_calculator.cc
```diff
@@ -51,6 +51,15 @@ enum class Endpoint {
     COMPLETIONS,
 };
 
+struct CompletionUsageStatistics {
+    size_t promptTokens = 0;
+    size_t completionTokens = 0;
+
+    size_t calculateTotalTokens() {
```
dkalinowski, line 58:
```suggestion
    size_t calculateTotalTokens() const {
```
can be const

dkalinowski (approved):

PR #2637, opened 2024-08-22
Log prompt tokens

rasapala:
### 🛠 Summary

JIRA CVS-148748.
Add input tokens logging in trace mode.

### 🧪 Checklist

``


src/llm/http_llm_calculator.cc
```diff
@@ -389,6 +389,25 @@ static std::string packIntoServerSideEventMessage(const std::string& message) {
     return ss.str();
 }
 
+static std::string packPromptTokens(const ov::Tensor& input) {
```
dtrawins, line 392: is it covered in some tests?

src/llm/http_llm_calculator.cc
```diff
@@ -389,6 +389,25 @@ static std::string packIntoServerSideEventMessage(const std::string& message) {
...
+    std::stringstream ss;
+    ss << "prompt_token_ids: [";
```
mzegla, line 394: I think this line should be moved below element type check

src/llm/http_llm_calculator.cc
```diff
@@ -389,6 +389,25 @@ static std::string packIntoServerSideEventMessage(const std::string& message) {
...
+    if (input.get_element_type() != ov::element::i64) {
+        ss << "Warning: expected ov::element::i64 and got " << input.get_element_type() << " for input tokens.]";
```
mzegla, line 396: That's quite misleading. We will print such warning, but what happens with the execution? And user will get it on trace level only. I think we should either support other types and make more generic method for printing values or say more explicitly like `"Cannot print input tokens - expected input type ov::element::i64 and got " << input.get_element_type()"`

src/llm/http_llm_calculator.cc
```diff
@@ -389,6 +389,25 @@ static std::string packIntoServerSideEventMessage(const std::string& message) {
...
+        return ss.str();
+    }
+
+    int64_t* data_ptr = input.data<int64_t>();
```
mzegla, line 400: Let's stick to the camelCase

src/llm/http_llm_calculator.cc
mzegla, line 392: Maybe we could move it outside of the calculator to a separate utils file?

dtrawins (approved):

src/llm/text_processor.hpp
```diff
@@ -35,4 +38,39 @@ class TextProcessor {
...
+        return packPromptTokens(tensor.data<int>(), tensor.get_size());
+    } else if (tensor.get_element_type() == ov::element::i16) {
+        return packPromptTokens(tensor.data<int16_t>(), tensor.get_size());
+    } else if (tensor.get_element_type() == ov::element::i64) {
+        return packPromptTokens(tensor.data<int64_t>(), tensor.get_size());
+    } else if (tensor.get_element_type() == ov::element::f32) {
+        return packPromptTokens(tensor.data<float>(), tensor.get_size());
+    } else if (tensor.get_element_type() == ov::element::f64) {
+        return packPromptTokens(tensor.data<double>(), tensor.get_size());
+    }
```
mzegla, line 68: Wouldn't it look better with switch...case syntax?

src/llm/text_processor.hpp
```diff
@@ -35,4 +38,39 @@ class TextProcessor {
...
+
+    std::stringstream ss;
+    ss << "Warning: unsupported ov::element::Type - got " << tensor.get_element_type() << " for input tokens.";
```
mzegla, line 71:
```suggestion
    ss << "Could not print input tokens for element type: " << tensor.get_element_type();
```

src/llm/text_processor.hpp
```diff
@@ -35,4 +38,42 @@ class TextProcessor {
...
+        return packPromptTokens(tensor.data<int16_t>(), size);
+    case ov::element::i64:
+        return packPromptTokens(tensor.data<int64_t>(), size);
+    case ov::element::f32:
+        return packPromptTokens(tensor.data<float>(), size);
+    case ov::element::f64:
+        return packPromptTokens(tensor.data<double>(), size);
+    default: {
+        std::stringstream ss;
+        ss << "Could not print input tokens for element type: " << tensor.get_element_type();
```
mzegla, line 72:
```suggestion
        ss << "Could not pack input tokens for element type: " << tensor.get_element_type();
```

PR #2640, opened 2024-08-23
Usage field support in OpenAI API - streaming

mzegla:
### 🛠 Summary

Addressing: https://github.com/openvinotoolkit/model_server/issues/2626

- Adding `streaming_options` with `include_usage` property to supported request parameters
- Adding `usage` field to completions and chat completions endpoints on streaming endpoint
- Updating documentation and demo

### 🧪 Checklist

- [x] Unit tests added.
``


src/llm/http_llm_calculator.cc
```diff
@@ -776,23 +812,64 @@ std::string HttpLLMCalculator::serializeStreamingChunk(const std::string& chunkR
...
+
+std::string HttpLLMCalculator::prepareStreamingUsageChunk(Endpoint endpoint, CompletionUsageStatistics usage) {
+    OVMS_PROFILE_FUNCTION();
+    StringBuffer buffer;
+    Writer<StringBuffer> writer(buffer);
+
+    writer.StartObject();  // {
+
+    // choices: array of size N, where N is related to n request parameter
+    // Can also be empty for the last chunk if you set stream_options: {"include_usage": true} TODO
```
mzegla, line 837:
```suggestion
```

docs/model_server_rest_api_chat.md
```diff
@@ -60,6 +65,8 @@ curl http://localhost/v3/chat/completions \
 |-----|----------|----------|----------|---------|-----|
 | model | ✅ | ✅ | ✅ | string (required) | Name of the model to use. From administrator point of view it is the name assigned to a MediaPipe graph configured to schedule generation using desired model.  |
 | stream | ✅ | ✅ | ✅ | bool (optional, default: `false`) | If set to true, partial message deltas will be sent to the client. The generation chunks will be sent as data-only [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format) as they become available, with the stream terminated by a `data: [DONE]` message. [Example Python code](clients_openai.md) |
+| stream_options | ✅ | ✅ | ✅ | object (optional) | Options for streaming response. Only set this when you set stream: true |
+| stream_options.include_usage | ✅ | ✅ | ✅ | bool (optional) | Streaming option.If set, an additional chunk will be streamed before the data: [DONE] message. The usage field on this chunk shows the token usage statistics for the entire request, and the choices field will always be an empty array. All other chunks will also include a usage field, but with a null value. |
```
dkalinowski, line 69:
```suggestion
| stream_options.include_usage | ✅ | ✅ | ✅ | bool (optional) | Streaming option. If set, an additional chunk will be streamed before the data: [DONE] message. The usage field in this chunk shows the token usage statistics for the entire request, and the choices field will always be an empty array. All other chunks will also include a usage field, but with a null value. |
```

docs/model_server_rest_api_completions.md
```diff
@@ -48,6 +53,8 @@ curl http://localhost/v3/completions \
 |-----|----------|----------|----------|---------|-----|
 | model | ✅ | ✅ | ✅ | string (required) | Name of the model to use. From administrator point of view it is the name assigned to a MediaPipe graph configured to schedule generation using desired model.  |
 | stream | ✅ | ✅ | ✅ | bool (optional, default: `false`) | If set to true, partial message deltas will be sent to the client. The generation chunks will be sent as data-only [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format) as they become available, with the stream terminated by a `data: [DONE]` message. [Example Python code](clients_openai.md) |
+| stream_options | ✅ | ✅ | ✅ | object (optional) | Options for streaming response. Only set this when you set stream: true |
+| stream_options.include_usage | ✅ | ✅ | ✅ | bool (optional) | Streaming option.If set, an additional chunk will be streamed before the data: [DONE] message. The usage field on this chunk shows the token usage statistics for the entire request, and the choices field will always be an empty array. All other chunks will also include a usage field, but with a null value. |
```
dkalinowski, line 57:
```suggestion
| stream_options.include_usage | ✅ | ✅ | ✅ | bool (optional) | Streaming option. If set, an additional chunk will be streamed before the data: [DONE] message. The usage field in this chunk shows the token usage statistics for the entire request, and the choices field will always be an empty array. All other chunks will also include a usage field, but with a null value. |
```

src/llm/http_llm_calculator.cc
```diff
@@ -162,6 +168,22 @@ class OpenAIChatCompletionsRequest {
...
 
+        it = this->doc.FindMember("stream_options");
+        if (it != this->doc.MemberEnd()) {
+            if (!this->stream)
+                return absl::InvalidArgumentError("stream_options provided, but stream not set to true");
+            if (!it->value.IsObject())
+                return absl::InvalidArgumentError("stream_options is not an object");
+            auto streamOptionsObj = it->value.GetObject();
+
+            it = streamOptionsObj.FindMember("include_usage");
```
dkalinowski, line 179: Should we validate that there are unexpected keys in the map?

src/llm/http_llm_calculator.cc
```diff
@@ -412,6 +434,7 @@ class HttpLLMCalculator : public CalculatorBase {
 
     std::string serializeUnaryResponse(const std::vector<ov::genai::GenerationOutput>& generationOutputs, Endpoint endpoint, CompletionUsageStatistics usage);
     std::string serializeStreamingChunk(const std::string& chunkResponse, ov::genai::GenerationFinishReason finishReason, Endpoint endpoint);
+    std::string prepareStreamingUsageChunk(Endpoint endpoint, CompletionUsageStatistics usage);
```
dkalinowski, line 437:
```suggestion
    std::string serializeStreamingUsageChunk(Endpoint endpoint, CompletionUsageStatistics usage);
```

src/llm/http_llm_calculator.cc
```diff
@@ -580,11 +604,25 @@ class HttpLLMCalculator : public CalculatorBase {
                     } else {  // finish generation
                         OVMS_PROFILE_SCOPE("Generation of last streaming response");
                         std::string response = packIntoServerSideEventMessage(serializeStreamingChunk(this->streamer->end(), finishReason, this->request->getEndpoint()));
-                        response += packIntoServerSideEventMessage("[DONE]");
-                        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Generated complete streaming response: {}", response);
-                        // Produce last message, but do not produce loopback packets anymore so this is last Process() call
+
+                        // If usage statistics are not supposed to be included, then this is the last chunk
+                        if (!this->request->getStreamOptions().includeUsage)
```
dkalinowski, line 609: If I read it correctly, we never send [DONE] in case there are usage statistics requested? Is it following the specs? I think vllm benchmarking script waits for that message

src/llm/http_llm_calculator.cc
```diff
@@ -580,11 +604,25 @@ class HttpLLMCalculator : public CalculatorBase {
...
+                            response += packIntoServerSideEventMessage("[DONE]");
+
                         cc->Outputs().Tag(OUTPUT_TAG_NAME).Add(new OutputDataType{std::move(response)}, timestamp);
+                        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Generated complete streaming response: {}", response);
+
+                        // If usage statisctics are included, we trigger next iteration to send additional chunk with usage information
+                        if (this->request->getStreamOptions().includeUsage)
```
dkalinowski, line 616: Why do we need to do it in next iteration? Cant we do it the same as we did with [DONE]? In the same iteration, but pushing multiple events in single MP packet.

src/llm/http_llm_calculator.cc
rasapala: Yeah it would be good to log options we do not support.

rasapala (approved):

src/llm/http_llm_calculator.cc
mzegla: If `include_usage` is set, then [DONE] is set along with the usage chunk since in that case usage chunk is the last one. 
Otherwise we send [DONE] with last chunk with the content as we've been doing so far.

src/llm/http_llm_calculator.cc
dkalinowski: I have abstracted all that chunk logic (related to serving LLMs) on calculator level so that net_http server does not need to know anything about SSE/OpenAI. It simply takes std::string and sends it to the client.

It's up to calculator to encode the openai format + SSE protocol into std::string. It does not matter if it is 1 or more chunks. Server will send everything you prepare to the client.

Look up the [DONE] chunk, it is done in the same iteration as last text chunk (or it was back when I made the first version of this calculator)

src/llm/http_llm_calculator.cc
```diff
@@ -176,11 +176,17 @@ class OpenAIChatCompletionsRequest {
...
             if (it != streamOptionsObj.MemberEnd()) {
                 if (!it->value.IsBool())
                     return absl::InvalidArgumentError("stream_options.include_usage is not a boolean");
                 this->streamOptions.includeUsage = it->value.GetBool();
+                streamOptionsFound++;
+            }
+
+            if (streamOptionsObj.MemberCount() > streamOptionsFound) {
+                return absl::InvalidArgumentError("Found unexpected stream options. Properties accepted in stream_options: include_usage");
             }
```
dkalinowski, line 190:
```c++
it = streamOptionsObj.FindMember("include_usage");
if (it != streamOptionsObj.MemberEnd())
  return absl::InvalidArgumentError("Found unexpected stream options. Properties accepted in stream_options: include_usage");
if (!it->value.IsBool())
  return absl::InvalidArgumentError("stream_options.include_usage is not a boolean");
this->streamOptions.includeUsage = it->value.GetBool()
```

I think it is shorter and cleaner as we avoid multi-line indent

dkalinowski (approved):

src/llm/http_llm_calculator.cc
mzegla: Let's stick with this current approach as it's more extensible in case there are new options we want to add

PR #2643, opened 2024-08-26
Return proper status when stream fails

michalkulakowski:
### 🛠 Summary

https://jira.devtools.intel.com/browse/CVS-148629

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/external/partial.patch
+++ b/external/partial.patch
@@ -1,5 +1,5 @@
 diff --git a/tensorflow_serving/util/net_http/server/internal/evhttp_request.cc b/tensorflow_serving/util/net_http/server/internal/evhttp_request.cc
-index c8d0501b..c09331aa 100644
+index c8d0501b..ecddbac3 100644
 --- a/tensorflow_serving/util/net_http/server/internal/evhttp_request.cc
 +++ b/tensorflow_serving/util/net_http/server/internal/evhttp_request.cc
 @@ -17,6 +17,8 @@ limitations under the License.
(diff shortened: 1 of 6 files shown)


external/partial.patch
```diff
@@ -69,33 +69,39 @@ index c8d0501b..3c0a3df6 100644
...
-   NET_LOG(FATAL, "PartialReplyWithStatus not implemented.");
+@@ -338,12 +371,18 @@ void EvHTTPRequest::AppendResponseHeader(absl::string_view header,
+   }
  }
  
--void EvHTTPRequest::PartialReply() {
+-void EvHTTPRequest::PartialReplyWithStatus(HTTPStatusCode status) {
 -  NET_LOG(FATAL, "PartialReplyWithStatus not implemented.");
-+void EvHTTPRequest::PartialReply(std::string data) {
++void EvHTTPRequest::PartialReplyWithStatus(std::string data, int code) {
```
dkalinowski, line 78: why not using HTTPStatusCode but int?

src/http_frontend/http_graph_executor_impl.cpp
```diff
@@ -117,6 +117,14 @@ Status sendErrorImpl(
     return StatusCode::OK;
 }
 
+Status sendErrorImplWithCode(
```
dkalinowski, line 120:
```suggestion
Status sendErrorWithCodeImpl(
```

src/http_frontend/http_graph_executor_impl.cpp
```diff
@@ -117,6 +117,14 @@ Status sendErrorImpl(
...
+    const std::string& message,
+    const int code,
```
dkalinowski, line 122: unused param, why int?

src/http_rest_api_handler.cpp
```diff
@@ -499,7 +499,7 @@ Status HttpRestApiHandler::processV3(const std::string_view uri, const HttpReque
         serverReaderWriter->OverwriteResponseHeader("Connection", "keep-alive");
         auto status = executor->inferStream(request, *serverReaderWriter);
         if (!status.ok()) {
-            sendErrorImpl(status.string(), *serverReaderWriter);
+            sendErrorImplWithCode(status.string(), 400, *serverReaderWriter);
```
dkalinowski, line 502: I dont think we need to put sendErrorImplWithCode in http_graph_executor_impl.hpp. This file is supposed to contain implementations for templated methods used in graph executor (http and grpc). I would move the implementation here

dkalinowski: + use enum instead of 400

dkalinowski, line 502: There are multiple places where we send an error using `sendErrorImpl` - why do we only return with status here? Please check if we need to also return with status in all other cases (check all usages of sendErrorImpl)

rasapala (approved):

dkalinowski (approved):

PR #2647, opened 2024-08-27
[release branch] Fix running vllm benchmarking scripts

dkalinowski:
### 🛠 Summary
CVS-149701

vllm cpu requirements cannot be installed in demo's venv without this env

### 🧪 Checklist

- [x] Unit tests added.

### DIFF
--- a/demos/continuous_batching/.gitignore
+++ b/demos/continuous_batching/.gitignore
@@ -1,0 +1,3 @@
+vllm
+Meta-Llama*
+venv
--- a/demos/continuous_batching/README.md
+++ b/demos/continuous_batching/README.md
@@ -31,6 +31,7 @@ Here, we will also define the LLM engine parameters inside the `graph.pbtxt`.
 
 Install python dependencies for the conversion script:
 ```bash
+export PIP_EXTRA_INDEX_URL="https://download.pytorch.org/whl/cpu"
 pip3 install -r https://raw.githubusercontent.com/openvinotoolkit/model_server/releases/2024/3/demos/continuous_batching/requirements.txt
 ```
 
@@ -232,7 +233,7 @@ pip3 install openai
 from openai import OpenAI
 
 client = OpenAI(
-  base_url="http://localhost:8000/v3",
+  base_url="http://localhost:11401/v3",
   api_key="unused"
 )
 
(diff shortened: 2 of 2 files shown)


demos/continuous_batching/README.md
```diff
@@ -31,6 +31,7 @@ Here, we will also define the LLM engine parameters inside the `graph.pbtxt`.
 
 Install python dependencies for the conversion script:
 ```bash
+export PIP_EXTRA_INDEX_URL="https://download.pytorch.org/whl/cpu"
```
dtrawins, line 34: this is already included in requirements.txt

demos/continuous_batching/README.md
```diff
@@ -232,7 +233,7 @@ pip3 install openai
 from openai import OpenAI
 
 client = OpenAI(
-  base_url="http://localhost:8000/v3",
+  base_url="http://localhost:11401/v3",
```
dtrawins, line 236: why this port?

demos/continuous_batching/.gitignore
```diff
@@ -0,0 +1,3 @@
+vllm
```
dtrawins, line 1: why this is included here?

dtrawins (changes requested):

mzegla (approved):

demos/continuous_batching/README.md
dkalinowski: It is not. Once you scroll down to the last paragraph of the demo, there is vllm benchmarking script tutorial. It requires to install requirements from vllm repo. It has no extra index, that is why it fails. I do put this export here because it is the same on main branch. We should have it unified.

demos/continuous_batching/README.md
dkalinowski: This is leftover from debugging, will remove it

demos/continuous_batching/.gitignore
dkalinowski: We rarely use gitignores in this repository. I think it is useful to keep directories and files in gitignores so that when you run the demo, the files generated in the demo (model files, venvs, configs, graphs) are ignored by git. @dtrawins

dtrawins (approved):

PR #2705, opened 2024-09-20
Use dnf in dockerfile.redhat

dtrawins:
### 🧪 Checklist

- [ ] Unit tests added.
``


Dockerfile.redhat
```diff
@@ -420,40 +419,23 @@ WORKDIR /
...
-            mkdir /tmp_ovms ; \
-            cd /tmp_ovms ; \
-            curl -L --fail -o deps.tar.xz "$INSTALL_RPMS_FROM_URL" ; \
-            tar -xf deps.tar.xz ; \
-            ls -Rahl . ; \
-            rpm -vi pkg/bin/*.rpm ; \
-            cd / ; \
-            rm -rf /tmp_ovms ; \
-            $DNF_TOOL remove -y tar gzip; \
-    fi ; \
```
mzegla, line 442: Don't we need this part anymore?

Dockerfile.redhat
```diff
@@ -420,40 +419,23 @@ WORKDIR /
...
     # For image with Python enabled install Python library
     if ! [[ $debug_bazel_flags == *"PYTHON_DISABLE=1"* ]]; then \
-        $DNF_TOOL install -y python39-libs --setopt=install_weak_deps=0; \
-        mkdir -p /ovms/lib/openvino-4.dist-info && \
-        echo $'Metadata-Version: 1.0\nName: openvino\nVersion: 2024.4' > /ovms/lib/openvino-2024.4.dist-info/METADATA ; \
```
mzegla, line 447: Why do you remove these lines? I suppose we will loose capability of using openvino package in Python nodes if dist info is not created.

dtrawins: we are setting it in https://github.com/openvinotoolkit/model_server/blob/main/Dockerfile.redhat#L305 
This folder is not used

atobiszei (approved):

mzegla (approved):

PR #2708, opened 2024-09-23
fixing logprobs in greedy search

dtrawins:
### 🛠 Summary

CVS-153148
Add support for logprobs in greedy search 

### 🧪 Checklist

- [x] Unit tests added.
``


Makefile
```diff
@@ -73,9 +73,9 @@ FUZZER_BUILD ?= 0
 # NOTE: when changing any value below, you'll need to adjust WORKSPACE file by hand:
 #         - uncomment source build section, comment binary section
 #         - adjust binary version path - version variable is not passed to WORKSPACE file!
-OV_SOURCE_BRANCH ?= c3152d32c9c7df71397e5a3aba1d935c49eec598  # releases/2024/4 2024-09-05
+OV_SOURCE_BRANCH ?= 97a71c298c1b5aeb5df76830508b2c404977ed89  # releases/2024/4 2024-09-05
 OV_CONTRIB_BRANCH ?= d850b779e17c396c718f3b7dad63bcb7c08c1226  # releases/2024/4 2024-08-30
-OV_TOKENIZERS_BRANCH ?= 3dde884b735e7f20db8a90664e3156facf48b3f0  # releases/2024/4 2024-09-09
+OV_TOKENIZERS_BRANCH ?= 81c067c557d48011e6879a42d4a25147060eaeff  # releases/2024/4 2024-09-09
```
mzegla, line 78: Please update comments

src/llm/apis/openai_completions.cpp
```diff
@@ -712,10 +708,10 @@ std::string OpenAIChatCompletionsHandler::serializeStreamingUsageChunk() {
 }
 
 void OpenAIChatCompletionsHandler::writeLogprob(Writer<StringBuffer>& writer, float logprob) {
-    // genai returns probs values which should be in the range of 0-1
+    // genai returns probs values which should be in the range of -inf-0
```
mzegla, line 711: This comment is not valid anymore with latest changes in GenAI right?

src/llm/apis/openai_completions.cpp
```diff
@@ -712,10 +708,10 @@ std::string OpenAIChatCompletionsHandler::serializeStreamingUsageChunk() {
...
     // other values could be potentially invalid and should be treated as such
-    if (logprob > 0.0 && logprob <= 1.0)
-        writer.Double(std::log10(logprob));
+    if (logprob <= 0.0)
+        writer.Double(logprob);
     else
         writer.Null();
```
mzegla, line 716: Is it possible to get positive value here?

src/test/llmnode_test.cpp
```diff
@@ -513,7 +513,7 @@ TEST_F(LLMFlowHttpTest, unaryChatCompletionsJsonLogprobs) {
         ASSERT_TRUE(choice["logprobs"]["content"][0]["bytes"].IsArray());
         ASSERT_TRUE(choice["logprobs"]["content"][0]["bytes"][0].IsInt());
         ASSERT_TRUE(choice["logprobs"]["content"][0]["top_logprobs"].IsArray());
-        ASSERT_TRUE(choice["logprobs"]["content"][0]["top_logprobs"][0].IsObject());
+        // ASSERT_TRUE(choice["logprobs"]["content"][0]["top_logprobs"][0].IsObject());
```
mzegla, line 516: Why is this commented?

third_party/llm_engine/llm_engine.bzl
```diff
@@ -19,8 +19,8 @@ def llm_engine():
     llm_engine_repository(name="_llm_engine")
     new_git_repository(
         name = "llm_engine",
-        remote = "https://github.com/openvinotoolkit/openvino.genai",
-        commit = "42398469746fca326fbdb8e195b572f3f483331c", # releases/2024/4 / 17 Sep 2024
+        remote = "https://github.com/popovaan/openvino.genai",
+        commit = "a63a005fd9ffafc44d1cb0ab58c4c1d12c97030a", # master / 19 Sep 2024
```
mzegla, line 23: Since fix has been merged please switch to main repo

src/test/llmnode_test.cpp
dtrawins: top_logprobs returns object only if top_logprobs is >1. With top_logprobs =1 comparing with vllm, there is empty array

src/llm/apis/openai_completions.cpp
dtrawins: it wouldn't be a valid value. per definition, logprobs should return negative values or 0

src/llm/apis/openai_completions.cpp
dtrawins: looks valid to me, that is the expected range

src/llm/apis/openai_completions.cpp
mzegla: In such case shouldn't we return some internal error instead of just writing null, if we get unexpected value?

src/llm/apis/openai_completions.cpp
mzegla: Oh the part `returns probs` is misleading then. Maybe we can be more explicit?
```suggestion
    // genai returns logarithms of probability per token which should be in the range of -inf-0
```

src/llm/apis/openai_completions.cpp
```diff
@@ -454,7 +448,9 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const std::vect
                     // top_logprobs are currently hardcoded to return only one logprob to comply with the API
                     // for full support significant changes on GenAI side are required
```
mzegla, line 449: With change below this comment is no longer valid, please adjust it to inform it's always empty array now

src/test/llmnode_test.cpp
mzegla: Can you change this assertion to check if we indeed have empty array?

mzegla (approved):

PR #2710, opened 2024-09-24
Callback with reload2

rasapala:
JIRA CVS-151523
Current Async implementation does not handle online modifications - we need to move unlaod guard into callback in model instance async implementation.

### 🧪 Checklist

- [x] Unit tests added.
``


Makefile
```diff
@@ -672,7 +672,7 @@ cpu_extension:
 run_unit_tests:
 	./prepare_llm_models.sh ${TEST_LLM_PATH}
 	./prepare_gpu_models.sh ${GPU_MODEL_PATH}
-	docker run -v $(shell realpath ./rununittests.sh):/ovms/./rununittests.sh -v $(shell realpath ${GPU_MODEL_PATH}):/ovms/src/test/face_detection_adas:ro -v $(shell realpath ${TEST_LLM_PATH}):/ovms/src/test/llm_testing:ro -e https_proxy=${https_proxy} -e RUN_TESTS=1 -e JOBS=$(JOBS) -e debug_bazel_flags=${BAZEL_DEBUG_FLAGS} $(OVMS_CPP_DOCKER_IMAGE)-build:$(OVMS_CPP_IMAGE_TAG)$(IMAGE_TAG_SUFFIX) ./rununittest.sh > test.log 2>&1 ; exit_status=$$? ; tail -200 test.log ; exit $$exit_status
+	docker run -v $(shell realpath ./rununittests.sh):/ovms/./rununittests.sh:ro -v $(shell realpath ${GPU_MODEL_PATH}):/ovms/src/test/face_detection_adas:ro -v $(shell realpath ${TEST_LLM_PATH}):/ovms/src/test/llm_testing:ro -e https_proxy=${https_proxy} -e RUN_TESTS=1 -e JOBS=$(JOBS) -e debug_bazel_flags=${BAZEL_DEBUG_FLAGS} $(OVMS_CPP_DOCKER_IMAGE)-build:$(OVMS_CPP_IMAGE_TAG)$(IMAGE_TAG_SUFFIX) ./rununittest.sh > test.log 2>&1 ; exit_status=$$? ; tail -200 test.log ; exit $$exit_status
```
dkalinowski, line 675: Why are you adding `ro` here? This parameter does not work at all, I have noticed it in my task. `rununittests.sh` does not exist. Lookup the end of the command:
```
... ./rununittest.sh > test.log ...
```

ci/cppclean.sh
```diff
@@ -54,7 +54,7 @@ fi
 if [ ${NO_WARNINGS_TEST_NOTUSED} -gt 0 ]; then
     errors+="Failed probably due to unnecessary forward includes: ${NO_WARNINGS_TEST_NOTUSED}"$'\n'
 fi
-if [ ${NO_WARNINGS} -gt  166 ]; then
+if [ ${NO_WARNINGS} -gt  174 ]; then
```
dkalinowski, line 57: Do we need to increase number of code smells?

src/modelinstance.cpp
```diff
@@ -1459,43 +1459,58 @@ Status ModelInstance::inferAsync(const RequestType* requestProto,
     // set callback
     // TODO check if there is callback in async
     OVMS_InferenceRequestCompletionCallback_t userCallback = requestProto->getResponseCompleteCallback();
+    if (userCallback == nullptr) {
```
dkalinowski, line 1462: @atobiszei should we enforce the callback? does OV enforce setting callback? It might be useful to not set it for warmups?

src/modelinstance.cpp
```diff
@@ -1459,43 +1459,58 @@ Status ModelInstance::inferAsync(const RequestType* requestProto,
...
+            // here call user set callback
+            // here will go OVMS C-API serialization code
+            std::unique_ptr<ResponseType> res(new ResponseType(this->getName(), this->getVersion()));
+            OutputGetter<ov::InferRequest&> outputGetter(inferRequest);
             try {
-                std::rethrow_exception(exception);
-            } catch (const std::exception& e) {
-                SPDLOG_DEBUG("got exception in ov::InferRequest callback: {}", e.what());
+                // TODO created filter based on what is in request, then perform casual serialization for what was NOT in request, and rewrite tensors from request to response for those that were
+                auto status = serializePredictResponse(outputGetter, getName(), getVersion(), getOutputsInfo(), requestProto, res.get(), getTensorInfoName, useSharedOutputContentFn(requestProto));  // TODO FIXME handle status
```
dkalinowski, line 1490: @rasapala should we return error when serialization fails? why is it ignored?

rasapala: // TODO FIXME handle status - this will be done in another task.

dkalinowski (approved):

PR #2717, opened 2024-09-27
Introduce echo parameter for text generation

mzegla:
### 🛠 Summary

Introducing echo parameter, but also change to completions endpoint prompt encoding. Using dynamic runtime configuration for encoding, BOS token is now added to the prompt on completion endpoint prior running inference.

### 🧪 Checklist

- [x] Unit tests added.
``


src/test/llmnode_test.cpp
```diff
@@ -116,7 +118,11 @@ class LLMFlowHttpTest : public ::testing::Test {
                     break;
                 i++;
                 std::vector<int64_t> tokens = out.generated_ids;
-                SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Generated tokens: {}", tokens);
+                std::cout << "[" << i << "] Generated tokens: ";
```
dtrawins, line 121: shouldn't it be only in TRACE mode?

michalkulakowski (approved):

dtrawins (approved):

PR #2746, opened 2024-10-17
Unit tests for embeddings endpoint

bstrzele:
### 🛠 Summary

[JIRA](https://jira.devtools.intel.com/browse/CVS-154380)

### 🧪 Checklist

- [x] Unit tests added.
``

### DIFF
--- a/src/embeddings/BUILD
+++ b/src/embeddings/BUILD
@@ -40,7 +40,8 @@ mediapipe_proto_library(
 
 cc_library(
     name = "embeddingscalculator",
-    srcs = ["embeddings_calculator.cc", "embeddings_api.cpp", "embeddings_api.hpp"],
+    hdrs = ["embeddings_api.hpp"],
+    srcs = ["embeddings_calculator.cc", "embeddings_api.cpp"],
     deps = [
         "@mediapipe//mediapipe/framework:calculator_framework",
         "@com_github_tencent_rapidjson//:rapidjson",
--- a/src/embeddings/BUILD
+++ b/src/embeddings/BUILD
@@ -40,7 +40,7 @@ mediapipe_proto_library(
 
 cc_library(
     name = "embeddingscalculator",
-    srcs = ["embeddings_calculator.cc"],
+    srcs = ["embeddings_calculator.cc", "embeddings_api.cpp", "embeddings_api.hpp"],
     deps = [
         "@mediapipe//mediapipe/framework:calculator_framework",
         "@com_github_tencent_rapidjson//:rapidjson",
--- a/src/embeddings/embeddings_api.cpp
+++ b/src/embeddings/embeddings_api.cpp
@@ -15,6 +15,7 @@
 //*****************************************************************************
 #include "embeddings_api.hpp"
 
+#include <string>
 #include <variant>
 
 #include "rapidjson/document.h"
--- a/src/embeddings/embeddings_api.cpp
+++ b/src/embeddings/embeddings_api.cpp
@@ -1,0 +1,53 @@
+#include "embeddings_api.hpp"
+
+#include <variant>
+
+#include "rapidjson/document.h"
+
+std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::from_json(rapidjson::Document* parsedJson) {
+    EmbeddingsRequest request;
+    std::vector<std::string> input_strings;
+
+    if (!parsedJson->IsObject())
+        return "Received json is not an object";
+
+    auto it = parsedJson->FindMember("input");
+    if (it != parsedJson->MemberEnd()) {
+        if (it->value.IsString()) {
+            input_strings.push_back(it->value.GetString());
+        } else if (it->value.IsArray()) {
+            for (auto& input : it->value.GetArray()) {
+                // TODO: is array of ints
+                // TODO: is int
+                if (!input.IsString())
+                    return "every element in input array should be string";
+                input_strings.push_back(input.GetString());
+            }
+        } else {
+            return "input should be string or array of strings";
+        }
+    } else {
+        return "input field is required";
+    }
+
+    it = parsedJson->FindMember("encoding_format");
+    request.encoding_format = EncodingFormat::FLOAT;
+    if (it != parsedJson->MemberEnd()) {
+        if (it->value.IsString()) {
+            if (it->value.GetString() == std::string("base64")) {
+                request.encoding_format = EncodingFormat::BASE64;
+            } else if (it->value.GetString() == std::string("float")) {
+                request.encoding_format = EncodingFormat::FLOAT;
+            } else {
+                return "encoding_format should either base64 or float";
+            }
+        } else {
+            return "encoding_format should be string";
+        }
+    }
+
+    // TODO: dimensions (optional)
+    // TODO: user (optional)
+    request.input = input_strings;
+    return request;
+}
--- a/src/embeddings/embeddings_api.cpp
+++ b/src/embeddings/embeddings_api.cpp
@@ -1,3 +1,18 @@
+//*****************************************************************************
+// Copyright 2024 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
 #include "embeddings_api.hpp"
 
 #include <variant>
--- a/src/embeddings/embeddings_api.cpp
+++ b/src/embeddings/embeddings_api.cpp
@@ -83,7 +83,7 @@ absl::Status EmbeddingsHandler::parseRequest() {
     return absl::OkStatus();
 }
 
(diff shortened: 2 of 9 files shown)


src/embeddings/BUILD
```diff
@@ -40,7 +40,7 @@ mediapipe_proto_library(
 
 cc_library(
     name = "embeddingscalculator",
-    srcs = ["embeddings_calculator.cc"],
+    srcs = ["embeddings_calculator.cc", "embeddings_api.cpp", "embeddings_api.hpp"],
```
dtrawins, line 43: shouldn't .hpp be in hdrs  section?

src/embeddings/embeddings_api.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "rapidjson/document.h"
+
+enum class EncodingFormat {
+    FLOAT,
+    BASE64
+};
+
+struct EmbeddingsRequest {
+    std::variant<std::vector<std::string>, std::vector<std::vector<int>>> input;
```
dtrawins, line 30: it could be also std::string or vector<int>

src/embeddings/embeddings_calculator.cc
```diff
@@ -88,62 +89,42 @@ class EmbeddingsCalculator : public CalculatorBase {
...
         SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Tokenizer input name detected: {}", tokenizerInputName);
 
         ::InferenceInput tokenizerInputMap;
-        tokenizerInputMap[tokenizerInputName] = ov::Tensor{
-            ov::element::string,
-            ov::Shape{input_strings.size()},
-            input_strings.data()};
 
         ::InferenceOutput embeddingsOutputMap;
+        size_t expected_batch_size = 1;
```
dtrawins, line 115: shouldn't it be called received_batch_size?

src/embeddings/embeddings_api.hpp
```diff
@@ -0,0 +1,34 @@
...
+    EncodingFormat encoding_format;
```
dkalinowski, line 31: L31/L33 please follow contributing guide https://github.com/openvinotoolkit/model_server/blob/0cef8c9702b49b8ba42bdef1ab63d6c4f141f1c4/docs/contributing.md?plain=1#L31

src/embeddings/embeddings_api.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <variant>
+
+#include "rapidjson/document.h"
+
+std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::from_json(rapidjson::Document* parsedJson) {
+    EmbeddingsRequest request;
+    std::vector<std::string> input_strings;
+
+    if (!parsedJson->IsObject())
+        return "Received json is not an object";
```
dkalinowski, line 28: This is very nice use of std::variant, however I think we should adhere to common patterns in this project which is returning status - either ovms::Status or abseil status in case it is more convenient for calculator
@atobiszei @bstrzele what do you think?

src/embeddings/embeddings_calculator.cc
```diff
@@ -88,62 +89,42 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+            ::InferenceOutput tokenizerOutputMap;
+            if (auto strings = std::get_if<std::vector<std::string>>(&embeddingsRequest.input)) {
+                expected_batch_size = strings->size();
+                tokenizerInputMap[tokenizerInputName] = ov::Tensor{
+                    ov::element::string,
+                    ov::Shape{expected_batch_size},
+                    strings->data()};
+                tokenizerOutputMap = tokenizer_session->infer(tokenizerInputMap);
+            } else {
+                // TODO: input already tokenized
```
dkalinowski, line 126: I know the feature is on the way and will be implemented shortly, but we should at least stop the calculator execution in this case. Otherwise we may end up with segfaults

src/embeddings/embeddings_calculator.cc
```diff
@@ -88,62 +89,42 @@ class EmbeddingsCalculator : public CalculatorBase {
...
-            ov::Shape{input_strings.size()},
-            input_strings.data()};
 
         ::InferenceOutput embeddingsOutputMap;
+        size_t expected_batch_size = 1;
         try {
-            ::InferenceOutput tokenizerOutputMap = tokenizer_session->infer(tokenizerInputMap);
+            ::InferenceOutput tokenizerOutputMap;
+            if (auto strings = std::get_if<std::vector<std::string>>(&embeddingsRequest.input)) {
+                expected_batch_size = strings->size();
```
dkalinowski, line 119: Why do we need this variable? And why setting it only in this if statement? I think it was more readable previously

src/test/mediapipeflow_test.cpp
```diff
@@ -177,6 +177,26 @@ TEST_F(MediapipeEmbeddingsTest, startup) {
     ASSERT_TRUE(mediapipeGraphDefinition->getStatus().isAvailable());
 }
 
+TEST_F(MediapipeEmbeddingsTest, gprcInference) {
```
dkalinowski, line 180:
```suggestion
TEST_F(MediapipeEmbeddingsTest, grpcInference) {
```

src/test/mediapipeflow_test.cpp
```diff
@@ -177,6 +177,26 @@ TEST_F(MediapipeEmbeddingsTest, startup) {
...
+    KFSInferenceServiceImpl& impl = dynamic_cast<const ovms::GRPCServerModule*>(grpcModule)->getKFSGrpcImpl();
+    ::KFSRequest request;
+    ::KFSResponse response;
+    const std::string modelName = "embeddings";
+    request.Clear();
+    response.Clear();
+    inputs_info_t inputsMeta{{"in", {DUMMY_MODEL_SHAPE, precision}}};
+    std::vector<float> requestData1{1., 1., 1., 1., 1., 1., 1., 1., 1., 1.};
+    preparePredictRequest(request, inputsMeta, requestData1);
+    ASSERT_EQ(impl.ModelInfer(nullptr, &request, &response).error_code(), grpc::StatusCode::NOT_FOUND);
```
dkalinowski, line 197: You may want to match the input name so it passes input name validation

src/test/llmnode_test.cpp
```diff
@@ -2809,6 +2810,303 @@ TEST_F(EmbeddingsExtensionTest, simplePositive) {
...
+        {
+            "model": "embeddings",
+            "input": "dummyInput"
+        }
+    )";
+    Status status = handler->dispatchToProcessor(endpointEmbeddings, requestBody, &response, comp, responseComponents, &writer);
+    ASSERT_EQ(status, ovms::StatusCode::MEDIAPIPE_EXECUTION_ERROR) << status.string();
+}
+
+class EmbeddingsInvalidTokenizerConfigTest : public ::testing::Test {
```
dkalinowski, line 3048: You can use inheritence to avoid rewritting entire class for both `EmbeddingsInvalidTokenizerConfigTest` and `EmbeddingsInvalidConfigTest`

src/embeddings/embeddings_api.hpp
bstrzele: in this case a singular string or vector would be represented by either one vector with size 1 simplify declaration of the variant since std::variant<std::vector<std::string>, std::vector<std::vector<int>>, std::string, std::vector<int>> might be unreadable.

src/embeddings/embeddings_api.hpp
bstrzele: According to https://github.com/openvinotoolkit/model_server/blob/0cef8c9702b49b8ba42bdef1ab63d6c4f141f1c4/docs/contributing.md?plain=1#L32 L31 should be fine? And this would be consistent with field name in request json.

src/embeddings/embeddings_calculator.cc
bstrzele: It is meant as a futureproofing for when the expected batch size will be dependent on either the amount of input strings or the amount of int vectors.

src/embeddings/embeddings_api.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    if (auto error = std::get_if<std::string>(&parsed)) {
+        return absl::InvalidArgumentError(*error);
+    }
+    this->request = std::get<EmbeddingsRequest>(parsed);
+    return absl::OkStatus();
+}
+
+std::variant<std::vector<std::string>, std::vector<std::vector<int>>> EmbeddingsHandler::getInput() const {
+    return request.input;
```
dkalinowski, line 87: does this copy underlying data?

src/embeddings/embeddings_api.cpp
```diff
@@ -0,0 +1,91 @@
...
+    return request;
+}
+
+absl::Status EmbeddingsHandler::parseRequest() {
+    auto parsed = EmbeddingsRequest::fromJson(&(this->doc));
+
+    if (auto error = std::get_if<std::string>(&parsed)) {
+        return absl::InvalidArgumentError(*error);
+    }
+    this->request = std::get<EmbeddingsRequest>(parsed);
```
dkalinowski, line 82: I'm pretty sure it does copy the data, can we std::move? or avoid the copies

michalkulakowski (approved):

src/embeddings/embeddings_api.cpp
bstrzele: std::get has not overload that would return a copy https://en.cppreference.com/w/cpp/utility/variant/get. In this case it would return rvalue that would be implicitly moved.

dtrawins (approved):

dkalinowski (approved):

dkalinowski:
This PR is 15 commits behind main, please rebase to make sure you are in line with @rasapala windows integration

PR #2753, opened 2024-10-18
Update URLs in RAG demo

helena-intel:
The existing URLs now resulted in empty documents and a cryptic error. With the URLs from this PR the notebook works as expected.

Also added ipywidgets to requirements.txt (to prevent a message about it) and changed `!pip` to `%pip` because it should be safer with virtualenvs.

FWIW: I also had to build OVMS from source to run this demo, it didn't work with the image pulled from Docker hub


dtrawins (approved):

dtrawins:
Thank you for contribution 👍 
Those changes are included in https://github.com/openvinotoolkit/model_server/pull/2800

PR #2757, opened 2024-10-22
rollback tokenizers and update genai

dtrawins:
### 🛠 Summary

CVS-155625
rollback tokenizer version which reported execution error in embedding models

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ FUZZER_BUILD ?= 0
 #         - adjust binary version path - version variable is not passed to WORKSPACE file!
 OV_SOURCE_BRANCH ?= 03c9ae38292a90ecb5cbfe2c8d5472eed0ec1aa9  # master 2024-10-18
 OV_CONTRIB_BRANCH ?= 4272f47cb3ffbaf5c0fb5db569deb16856c578a1  # master 2024-10-11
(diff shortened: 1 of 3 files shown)


external/cb.patch
```diff
@@ -1,15 +1,18 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
```
mzegla, line 1: I think we can remove this patch entirely if we use latest GenAI master

Makefile
```diff
@@ -76,7 +76,7 @@ FUZZER_BUILD ?= 0
 #         - adjust binary version path - version variable is not passed to WORKSPACE file!
 OV_SOURCE_BRANCH ?= 03c9ae38292a90ecb5cbfe2c8d5472eed0ec1aa9  # master 2024-10-18
 OV_CONTRIB_BRANCH ?= 4272f47cb3ffbaf5c0fb5db569deb16856c578a1  # master 2024-10-11
-OV_TOKENIZERS_BRANCH ?= 57b236f113d78767657676cdefee6695e5e914ff  # master 2024-10-18
+OV_TOKENIZERS_BRANCH ?=  81c067c557d48011e6879a42d4a25147060eaeff # master 2024-10-18
```
mzegla, line 79: Commit changed, but the comment with the date remains the same

mzegla (approved):

PR #2775, opened 2024-10-30
Support for tokens on input in embeddings endpoint

bstrzele:
### 🛠 Summary

[JIRA](https://jira.devtools.intel.com/browse/CVS-154774)

### 🧪 Checklist

- [x] Unit tests added.
``

### DIFF
--- a/src/embeddings/embeddings_api.cpp
+++ b/src/embeddings/embeddings_api.cpp
@@ -31,14 +31,15 @@
 
 using namespace rapidjson;
 
-enum class InputType {
-    NONE,
-    STRING,
-    INT,
-    INT_VEC
-};
+namespace ovms {
 
 std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjson::Document* parsedJson) {
+    enum class InputType {
+        NONE,
+        STRING,
+        INT,
+        INT_VEC
+    };
     EmbeddingsRequest request;
     std::vector<std::string> input_strings;
     std::vector<std::vector<int64_t>> input_tokens;
@@ -58,6 +59,7 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
                         return "input must be homogeneous";
                     input_type = InputType::INT_VEC;
                     std::vector<int64_t> ints;
+                    ints.reserve(input.GetArray().Size());
                     for (auto& val : input.GetArray()) {
                         if (val.IsInt())
                             ints.push_back(val.GetInt());
@@ -129,7 +131,7 @@ absl::Status EmbeddingsHandler::parseRequest() {
 std::variant<std::vector<std::string>, std::vector<std::vector<int64_t>>>& EmbeddingsHandler::getInput() {
     return request.input;
 }
-EncodingFormat EmbeddingsHandler::getEncodingFormat() const {
+EmbeddingsRequest::EncodingFormat EmbeddingsHandler::getEncodingFormat() const {
     return request.encoding_format;
 }
 
@@ -168,7 +170,7 @@ absl::Status EmbeddingsHandler::parseResponse(StringBuffer& buffer, const ov::Te
             std::transform(dataPtr, dataPtrEnd, dataPtr,
                 [denom](auto& element) { return element / denom; });
         }
-        if (getEncodingFormat() == EncodingFormat::BASE64) {
+        if (getEncodingFormat() == EmbeddingsRequest::EncodingFormat::BASE64) {
             std::string_view sv2(reinterpret_cast<char*>(dataPtr), outputShape[2] * sizeof(float));
             std::string escaped;
             absl::Base64Escape(sv2, &escaped);
@@ -198,3 +200,4 @@ absl::Status EmbeddingsHandler::parseResponse(StringBuffer& buffer, const ov::Te
     writer.EndObject();
     return absl::OkStatus();
 }
+}  // namespace ovms
--- a/src/embeddings/embeddings_api.cpp
+++ b/src/embeddings/embeddings_api.cpp
@@ -34,6 +34,7 @@ using namespace rapidjson;
 std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjson::Document* parsedJson) {
     EmbeddingsRequest request;
     std::vector<std::string> input_strings;
+    std::vector<std::vector<int64_t>> input_tokens;
 
     if (!parsedJson->IsObject())
         return "Received json is not an object";
@@ -43,15 +44,39 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
         if (it->value.IsString()) {
             input_strings.push_back(it->value.GetString());
         } else if (it->value.IsArray()) {
+            int input_type = -1;
             for (auto& input : it->value.GetArray()) {
-                // TODO: is array of ints
-                // TODO: is int
-                if (!input.IsString())
-                    return "every element in input array should be string";
-                input_strings.push_back(input.GetString());
+                if (input.IsArray()) {
+                    if (input_type != -1 && input_type != 1)
+                        return "input must be homogeneous";
+                    input_type = 1;
+                    std::vector<int64_t> ints;
+                    for (auto& val : input.GetArray()) {
+                        if (val.IsInt())
+                            ints.push_back(val.GetInt());
+                        else
+                            return "input must be homogeneous";
+                    }
+                    input_tokens.push_back(ints);
+                } else if (input.IsString()) {
+                    if (input_type != -1 && input_type != 2)
+                        return "input must be homogeneous";
+                    input_type = 2;
+                    input_strings.push_back(input.GetString());
+                } else if (input.IsInt()) {
+                    if (input_type != -1 && input_type != 3)
+                        return "input must be homogeneous";
+                    input_type = 3;
+                    if (input_tokens.size() == 0) {
+                        input_tokens.push_back(std::vector<int64_t>());
+                    }
+                    input_tokens[0].push_back(input.GetInt());
+                } else {
+                    return "every element in input array should be either string or int";
+                }
             }
         } else {
-            return "input should be string or array of strings";
+            return "input should be string, array of strings or array of integers";
         }
     } else {
         return "input field is required";
@@ -75,7 +100,12 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
 
     // TODO: dimensions (optional)
     // TODO: user (optional)
-    request.input = input_strings;
+    if (input_strings.size() > 0) {
+        request.input = input_strings;
+    }
+    if (input_tokens.size() > 0) {
+        request.input = input_tokens;
+    }
     return request;
 }
 
@@ -89,7 +119,7 @@ absl::Status EmbeddingsHandler::parseRequest() {
     return absl::OkStatus();
 }
 
-std::variant<std::vector<std::string>, std::vector<std::vector<int>>>& EmbeddingsHandler::getInput() {
+std::variant<std::vector<std::string>, std::vector<std::vector<int64_t>>>& EmbeddingsHandler::getInput() {
     return request.input;
 }
 EncodingFormat EmbeddingsHandler::getEncodingFormat() const {
--- a/src/embeddings/embeddings_api.cpp
+++ b/src/embeddings/embeddings_api.cpp
@@ -31,6 +31,13 @@
 
 using namespace rapidjson;
 
+enum class InputType {
+    NONE,
+    STRING,
+    INT,
+    INT_VEC
+};
+
 std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjson::Document* parsedJson) {
     EmbeddingsRequest request;
     std::vector<std::string> input_strings;
@@ -44,12 +51,12 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
         if (it->value.IsString()) {
             input_strings.push_back(it->value.GetString());
         } else if (it->value.IsArray()) {
-            int input_type = -1;
+            InputType input_type = InputType::NONE;
             for (auto& input : it->value.GetArray()) {
                 if (input.IsArray()) {
-                    if (input_type != -1 && input_type != 1)
+                    if (input_type != InputType::NONE && input_type != InputType::INT_VEC)
                         return "input must be homogeneous";
-                    input_type = 1;
+                    input_type = InputType::INT_VEC;
                     std::vector<int64_t> ints;
                     for (auto& val : input.GetArray()) {
                         if (val.IsInt())
@@ -59,14 +66,14 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
                     }
                     input_tokens.push_back(ints);
                 } else if (input.IsString()) {
-                    if (input_type != -1 && input_type != 2)
+                    if (input_type != InputType::NONE && input_type != InputType::STRING)
                         return "input must be homogeneous";
-                    input_type = 2;
+                    input_type = InputType::STRING;
                     input_strings.push_back(input.GetString());
                 } else if (input.IsInt()) {
-                    if (input_type != -1 && input_type != 3)
+                    if (input_type != InputType::NONE && input_type != InputType::INT)
                         return "input must be homogeneous";
-                    input_type = 3;
+                    input_type = InputType::INT;
                     if (input_tokens.size() == 0) {
                         input_tokens.push_back(std::vector<int64_t>());
                     }
--- a/src/embeddings/embeddings_calculator.cc
+++ b/src/embeddings/embeddings_calculator.cc
@@ -45,6 +45,8 @@ class EmbeddingsCalculator : public CalculatorBase {
     static const std::string INPUT_TAG_NAME;
     static const std::string OUTPUT_TAG_NAME;
     static const std::string EMBEDDINGS_MODEL_INPUT_IDS_NAME;
+    static const std::string EMBEDDINGS_MODEL_ATTENTION_MASK_NAME;
+    static const std::string EMBEDDINGS_MODEL_TOKEN_TYPE_IDS_NAME;
 
     mediapipe::Timestamp timestamp{0};
 
@@ -111,8 +113,11 @@ public:
         SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Tokenizer input name detected: {}", tokenizerInputName);
 
         ::InferenceInput tokenizerInputMap;
-        size_t received_batch_size = 1;
         ::InferenceOutput embeddingsOutputMap;
+        ::InferenceInput embeddingsInputMap;
+        size_t received_batch_size = 1;
+        size_t max_context_length = 512;  // default allowed input length. Otherwise, it will be read from model rt_info>config>max_position_embeddings in the model.xml file
+        ov::AnyMap modelConfig = embeddings_session->getModelConfig();
         try {
             ::InferenceOutput tokenizerOutputMap;
             auto input = handler.getInput();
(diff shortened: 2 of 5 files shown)


src/embeddings/embeddings_api.cpp
```diff
@@ -37,12 +38,26 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
...
+                        input_tokens.push_back(std::vector<int>());
+                    }
+                    input_tokens[0].push_back(input.GetInt());
+                } else {
+                    return "every element in input array should be either string or int";
+                }
+                index++;
             }
         } else {
             return "input should be string or array of strings";
```
dkalinowski, line 63:
```suggestion
            return "input should be string, array of strings or array of integers";
```

src/embeddings/embeddings_api.cpp
```diff
@@ -37,12 +38,26 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
...
+                    input_strings.push_back(input.GetString());
+                } else if (input.IsInt()) {
+                    if (input_tokens.size() == 0) {
+                        input_tokens.push_back(std::vector<int>());
+                    }
+                    input_tokens[0].push_back(input.GetInt());
+                } else {
+                    return "every element in input array should be either string or int";
+                }
+                index++;
```
dkalinowski, line 60: unused variable

src/embeddings/embeddings_api.cpp
```diff
@@ -37,12 +38,26 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
...
-                // TODO: is array of ints
-                // TODO: is int
-                if (!input.IsString())
-                    return "every element in input array should be string";
-                input_strings.push_back(input.GetString());
+                if (input.IsArray()) {
+                    std::vector<int> ints;
+                    for (auto& val : input.GetArray()) {
+                        if (val.IsInt())
+                            ints.push_back(val.GetInt());
```
dkalinowski, line 47: Looks like this deserialization allows:
```
[ [62, 71, True, 5, "abc", 1 ], ... ]
```
is it desired?

src/embeddings/embeddings_api.cpp
```diff
@@ -37,12 +38,26 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
...
+                    }
+                    input_tokens.push_back(ints);
+                } else if (input.IsString()) {
+                    input_strings.push_back(input.GetString());
+                } else if (input.IsInt()) {
```
dkalinowski, line 52: Looks like this deserialization allows:
```
[ [62, 12, 4], 5, 2 ]
```
is it desired?

src/test/llmnode_test.cpp
```diff
@@ -3037,6 +3109,70 @@ TEST(EmbeddingsSerialization, multipleStringInput) {
...
+    ASSERT_EQ(ints->at(0).at(0), 1);
+    ASSERT_EQ(ints->at(0).at(1), 2);
+    ASSERT_EQ(ints->at(0).at(2), 3);
+}
+
+TEST(EmbeddingsSerialization, multipleIntInput) {
+    std::string requestBody = R"(
+        {
+            "model": "embeddings",
+            "input": [[1, 2, 3], [4, 5, 6]]
```
dtrawins, line 3139: add tests with variable tokens length, we should confirm padding and attention mask is created correctly

src/embeddings/embeddings_calculator.cc
```diff
@@ -124,31 +128,60 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                    size_t input_ids_size = tokenizerOutputMap[EMBEDDINGS_MODEL_INPUT_IDS_NAME].get_shape()[1];
+                    if (input_ids_size > max_context_length) {
+                        SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Input size {} exceeds max_context_length {}", input_ids_size, max_context_length);
+                        return absl::InvalidArgumentError(absl::StrCat("Input length ", input_ids_size, " longer than allowed ", max_context_length));
+                    }
+                    RET_CHECK(it != tokenizerOutputMap.end());
+                    SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Embedding model input {} is connected with matching tokenizer output", embeddingsInputName);
+                    embeddingsInputMap[embeddingsInputName] = it->second;
+                }
+            } else if (auto ints = std::get_if<std::vector<std::vector<int>>>(&input)) {
```
dtrawins, line 145: this code doesn't have tests coverage

src/embeddings/embeddings_calculator.cc
```diff
@@ -124,31 +128,60 @@ class EmbeddingsCalculator : public CalculatorBase {
...
-            for (const auto& embeddingsInputName : embeddingsInputNames) {
-                auto it = tokenizerOutputMap.find(embeddingsInputName);
-                RET_CHECK(it != tokenizerOutputMap.end());
-                SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Embedding model input {} is connected with matching tokenizer output", embeddingsInputName);
-                embeddingsInputMap[embeddingsInputName] = it->second;
+                // Check if tokenizer produced at least the number of outputs as there are inputs in embedding model
+                RET_CHECK(tokenizerOutputMap.size() >= embeddingsInputNames.size());
+                for (const auto& embeddingsInputName : embeddingsInputNames) {
+                    auto it = tokenizerOutputMap.find(embeddingsInputName);
+                    RET_CHECK(tokenizerOutputMap[EMBEDDINGS_MODEL_INPUT_IDS_NAME].get_shape().size() == 2);
```
dkalinowski, line 135: This ret check is way too early, at this point you have no idea if EMBEDDINGS_MODEL_INPUT_IDS_NAME exists in tokenizerOutputMap. Therefore using `operator[]` will insert new element to the output map which is not desired

src/embeddings/embeddings_calculator.cc
```diff
@@ -124,31 +128,60 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                for (auto i : *ints) {
+                    max_b = std::max(max_b, i.size());
+                }
+                for (auto arr : *ints) {
+                    for (size_t i = 0; i < max_b; i++) {
+                        if (i < arr.size()) {
+                            input_ids.push_back(arr[i]);
+                            attention_mask.push_back(1);
+                        } else {
+                            input_ids.push_back(0);
```
dtrawins, line 159: why pushing 0? it shoud be  pad_token_id to be checked from modelConfig

src/embeddings/embeddings_calculator.cc
dkalinowski: What is more, I don't think we need to repeat this RET_CHECK N times, where N is equal to embeddingsInputNames.size() (for loop)

src/embeddings/embeddings_calculator.cc
```diff
@@ -124,31 +128,60 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                    if (input_ids_size > max_context_length) {
+                        SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Input size {} exceeds max_context_length {}", input_ids_size, max_context_length);
+                        return absl::InvalidArgumentError(absl::StrCat("Input length ", input_ids_size, " longer than allowed ", max_context_length));
+                    }
+                    RET_CHECK(it != tokenizerOutputMap.end());
+                    SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Embedding model input {} is connected with matching tokenizer output", embeddingsInputName);
+                    embeddingsInputMap[embeddingsInputName] = it->second;
+                }
+            } else if (auto ints = std::get_if<std::vector<std::vector<int>>>(&input)) {
+                size_t max_b = 0;
```
dkalinowski, line 146: Please use more descriptive names, for example `size_t token_count_of_longest_document`

src/embeddings/embeddings_calculator.cc
```diff
@@ -124,31 +128,60 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                std::vector<int> input_ids;
+                std::vector<int> attention_mask;
```
dkalinowski, line 148: `int` size is dependend on architecture, and we need to be sure that it is always 8 bytes long. please use the same precision as model type (int64_t?)

src/embeddings/embeddings_calculator.cc
```diff
@@ -124,31 +128,60 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                    input_ids.data()};
+                embeddingsInputMap[EMBEDDINGS_MODEL_ATTENTION_MASK_NAME] = ov::Tensor{
+                    ov::element::i64,
+                    ov::Shape{received_batch_size, max_b},
+                    attention_mask.data()};
+                if (embeddings_session->getInputNames().size() == 3) {
+                    embeddingsInputMap[EMBEDDINGS_MODEL_TOKEN_TYPE_IDS_NAME] = ov::Tensor{
+                        ov::element::i64,
+                        ov::Shape{received_batch_size, max_b},
+                        attention_mask.data()};
```
dtrawins, line 177: why attention mask? In there would be 3rd input token_type_ids, it would be all ones

src/embeddings/embeddings_calculator.cc
```diff
@@ -124,31 +128,60 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                std::vector<int> input_ids;
+                std::vector<int> attention_mask;
+                received_batch_size = ints->size();
+                for (auto i : *ints) {
+                    max_b = std::max(max_b, i.size());
+                }
+                for (auto arr : *ints) {
+                    for (size_t i = 0; i < max_b; i++) {
+                        if (i < arr.size()) {
+                            input_ids.push_back(arr[i]);
```
dkalinowski, line 156: pushing elements to vector in such a way is slow
my suggestion is to do it similarly to document chunking (ov::Tensor reallocation) https://github.com/openvinotoolkit/model_server/pull/2788

Simply create ov::Tensor with desired shape (you already know all the dimensions - batch size and token_count_of_longest_document
And copy the tokens to preallocated positions, also fill remaining tokens with pad token (which should be read from model config)

@dtrawins @bstrzele

src/test/llmnode_test.cpp
```diff
@@ -2916,6 +2916,78 @@ TEST_F(EmbeddingsHttpTest, simplePositiveBase64) {
     ASSERT_EQ(d["data"][0]["index"], 0);
 }
 
+TEST_F(EmbeddingsHttpTest, simplePositiveInt) {
+    std::string requestBody = R"(
+        {
+            "model": "embeddings",
+            "input": [111, 222, 121]
```
dkalinowski, line 2923: Add tests for the corner cases mentioned in earlier comments

dkalinowski: More corner cases

src/embeddings/embeddings_calculator.cc
dkalinowski: As you can see @dtrawins embedding endpoint will also support only 1 precision for tokens - int64. We could templatize this for both cases, but since we see no model with int32, uint32, uint64 precisions it should be low priority.

@dtrawins

src/embeddings/embeddings_api.cpp
```diff
@@ -43,15 +44,39 @@ std::variant<EmbeddingsRequest, std::string> EmbeddingsRequest::fromJson(rapidjs
...
         } else if (it->value.IsArray()) {
+            int input_type = -1;
             for (auto& input : it->value.GetArray()) {
-                // TODO: is array of ints
-                // TODO: is int
-                if (!input.IsString())
-                    return "every element in input array should be string";
-                input_strings.push_back(input.GetString());
+                if (input.IsArray()) {
+                    if (input_type != -1 && input_type != 1)
```
dkalinowski, line 50: this is very confusing, what do you think about creating enum class?

src/embeddings/embeddings_api.cpp
dkalinowski: @bstrzele

bstrzele: this will not be successfully deserialized https://github.com/openvinotoolkit/model_server/pull/2775/files#diff-2b9c77dffac8a9598e047bebd0ccac92c2e74e198d0664a90f9944d536f6b9beR195

src/embeddings/embeddings_calculator.cc
```diff
@@ -123,42 +128,76 @@ class EmbeddingsCalculator : public CalculatorBase {
...
-                    if (it->second.get_element_type() == ov::element::Type_t::i64) {
-                        for (int i = 0; i < it->second.get_size(); i++) {
-                            attendedTokens += reinterpret_cast<int64_t*>(it->second.data())[i];
-                        }
-                    } else if (it->second.get_element_type() == ov::element::Type_t::i32) {
-                        for (int i = 0; i < it->second.get_size(); i++) {
-                            attendedTokens += reinterpret_cast<int32_t*>(it->second.data())[i];
+                // Check if tokenizer produced at least the number of outputs as there are inputs in embedding model
+                RET_CHECK(tokenizerOutputMap.size() >= embeddingsInputNames.size());
+                RET_CHECK(tokenizerOutputMap[EMBEDDINGS_MODEL_INPUT_IDS_NAME].get_shape().size() == 2);
```
dkalinowski, line 133: Again, not sure where my previous comment go, but this line may insert element with key `EMBEDDINGS_MODEL_INPUT_IDS_NAME` to the tokenizerOutputMap map. This is what `operator[]` is doing when key does not exist

src/embeddings/embeddings_calculator.cc
```diff
@@ -123,42 +128,76 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                                attendedTokens += reinterpret_cast<uint8_t*>(it->second.data())[i];
+                            }
                         }
+                        handler.setPromptTokensUsage(attendedTokens);
                     }
-                    handler.setPromptTokensUsage(attendedTokens);
+                }
+            } else if (auto ints = std::get_if<std::vector<std::vector<int64_t>>>(&input)) {
+                size_t token_count_of_longest_document = 0;
+                int64_t pad_token = modelConfig["pad_token_id"].as<int64_t>();
```
dkalinowski, line 168: What is the type of modelConfig? Is it ov::AnyMap? Is it std::map?

- in a case when it is ov::AnyMap - what does operator[] do? does it throw exception when key does not exist? does it insert with default value? what is the default value in case it inserts it? is it desired default value? do we have unit test for that behavior?

- in case it is std::map - the same questions as above

src/embeddings/embeddings_calculator.cc
```diff
@@ -123,42 +128,76 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                size_t tokens = 0;
+                received_batch_size = ints->size();
+                for (auto i : *ints) {
```
dkalinowski, line 171:
```suggestion
                for (const auto& document_tokens : *tokenized_documents) {
```
Use reference, otherwise the entire vector of integers is coped in every iteration
Use const since you do not need to modify it
Rename it for better readability

src/embeddings/embeddings_calculator.cc
```diff
@@ -123,42 +128,76 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                embeddingsInputMap[EMBEDDINGS_MODEL_INPUT_IDS_NAME] = ov::Tensor{
+                    ov::element::i64,
+                    ov::Shape{received_batch_size, token_count_of_longest_document}};
+                embeddingsInputMap[EMBEDDINGS_MODEL_ATTENTION_MASK_NAME] = ov::Tensor{
+                    ov::element::i64,
+                    ov::Shape{received_batch_size, token_count_of_longest_document}};
+                for (size_t i = 0; i < received_batch_size; i++) {
+                    int64_t* input_ids_start = reinterpret_cast<int64_t*>(embeddingsInputMap[EMBEDDINGS_MODEL_INPUT_IDS_NAME].data()) + i * token_count_of_longest_document;
+                    std::fill(input_ids_start, input_ids_start + token_count_of_longest_document, pad_token);
+                    std::copy(ints->at(i).data(), ints->at(i).data() + ints->at(i).size(), input_ids_start);
```
dkalinowski, line 186: Is `ints` size validated to be equal to `received_batch_size` before hand? If so, you can use `operator[]`. If not, it is good practice to use try catch for throwing functions

src/embeddings/embeddings_calculator.cc
bstrzele: If the output was not created then RET_CHECK one line earlier would fail. If the name of the input will not match then this RET_CHECK would fail anyway, right?

src/embeddings/embeddings_calculator.cc
bstrzele: So would be the case for this one https://github.com/openvinotoolkit/model_server/blob/main/src/embeddings/embeddings_calculator.cc#L164?

src/test/embeddings_handler_test.cpp
```diff
@@ -97,7 +173,77 @@ TEST(EmbeddingsDeserialization, malformedInput) {
...
+        }
+    )";
+    rapidjson::Document d;
+    rapidjson::ParseResult ok = d.Parse(requestBody.c_str());
+    ASSERT_EQ(ok.Code(), 0);
+    EmbeddingsHandler handler(d);
+    ASSERT_EQ(handler.parseRequest(), absl::OkStatus());
+    ASSERT_EQ(handler.getEncodingFormat(), EncodingFormat::FLOAT);
+    auto input = handler.getInput();
+    auto strings = std::get_if<std::vector<std::string>>(&input);
```
dkalinowski, line 241: you could also check that std::get_if<vector<vector<int64>> is false

src/embeddings/embeddings_calculator.cc
dkalinowski: In line 164 you mentioned, we first check if the key exists using `modelConfig.count("max_position_embeddings")`

src/embeddings/embeddings_calculator.cc
dkalinowski: Yes, but it is not the only reason for EMBEDDINGS_MODEL_INPUT_IDS_NAME to be absent. The inference could pass (making`tokenizerOutputMap.size() >= embeddingsInputNames.size()` statement true), but still EMBEDDINGS_MODEL_INPUT_IDS_NAME might not be there

src/embeddings/embeddings_api.cpp
```diff
@@ -31,9 +31,17 @@
 
 using namespace rapidjson;
 
+enum class InputType {
```
dkalinowski, line 34: I just realized that none of those classes are part of ovms namespace. Please do that. When you change it, the InputType will be in ovms namespace which is not good idea either. I think the name is so generic that it could be under EmbeddingsRequest namespace. What's more, I don't think we need to expose it to the API user, it can be hidden (private) within EmbeddingsRequest scope since it is user by this class only anyway.

src/embeddings/embeddings_api.cpp
```diff
@@ -31,9 +31,17 @@
...
+    NONE,
+    STRING,
+    INT,
+    INT_VEC
+};
+
```
dkalinowski, line 40:
```suggestion
enum class InputType {
    NONE,
    STRING,
    TOKENS
};

I don't understand why we do have different type for single batch tokens and multi batch tokens but we do not have one for single string (1 document) and multiple documents (1+ documents). Isn't 1+ strings not supported?

```

bstrzele: In this case single string is represented as vector with size of one and is sharing the same path as when there is multiple strings in the input request

dkalinowski (approved):

src/embeddings/embeddings_calculator.cc
```diff
@@ -123,42 +128,85 @@ class EmbeddingsCalculator : public CalculatorBase {
...
+                        int64_t* input_ids_start = reinterpret_cast<int64_t*>(embeddingsInputMap[EMBEDDINGS_MODEL_INPUT_IDS_NAME].data()) + i * token_count_of_longest_document;
+                        std::fill(input_ids_start, input_ids_start + token_count_of_longest_document, pad_token);
+                        std::copy(tokenized_documents->at(i).data(), tokenized_documents->at(i).data() + tokenized_documents->at(i).size(), input_ids_start);
+
+                        int64_t* attention_mask_start = reinterpret_cast<int64_t*>(embeddingsInputMap[EMBEDDINGS_MODEL_ATTENTION_MASK_NAME].data()) + i * token_count_of_longest_document;
+                        std::fill(attention_mask_start, attention_mask_start + token_count_of_longest_document, 0);
+                        std::fill(attention_mask_start, attention_mask_start + tokenized_documents->at(i).size(), 1);
+                    }
+                } catch (std::out_of_range& e) {
+                    SPDLOG_LOGGER_ERROR(embeddings_calculator_logger, "Caught exception from preparing embeddings inputs(): {}", e.what());
```
dtrawins, line 195:
```suggestion
                    SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Caught exception from preparing embeddings inputs(): {}", e.what());
```
Errors related to individual invalid requests are typically logged in debug mode

dtrawins (approved):

PR #2777, opened 2024-10-30
Enable speculative decoding

mzegla:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``


demos/continuous_batching/speculative_decoding/README.md
```diff
@@ -0,0 +1,181 @@
...
+From the functional perspective both main and draft models must use the same tokenizer, so the tokens from the draft model are correctly matched in the the main model.
+
+From the performance perspective, benefits from speculative decoding are strictly tied to the pair of models used.
+For some models, the performance boost is significant, while for others it's rather negligible. Models sizes and precisions also come into play, so optimal setup shall be found empirically.
+
+In this demo we will use:
+  - [meta-llama/CodeLlama-7b-hf](https://huggingface.co/meta-llama/CodeLlama-7b-hf) as a main model
+  - [AMD-Llama-135m](https://huggingface.co/amd/AMD-Llama-135m) as a draft model
+
+both in FP16 precision.
```
dtrawins, line 27: why FP16? Can it be loaded on dGPU?

demos/continuous_batching/speculative_decoding/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Model preparation
+Here, the original Pytorch LLM models and the tokenizers will be converted to IR format and optionally quantized.
+That ensures faster initialization time, better performance and lower memory consumption.
+LLM engine parameters will be defined inside the `graph.pbtxt` file.
+
+Download export script, install it's dependencies and create directory for the models:
```
dtrawins, line 34:
```suggestion
Download export script, install its dependencies and create directory for the models:
```

demos/continuous_batching/speculative_decoding/README.md
```diff
@@ -0,0 +1,181 @@
...
+        ]
+    }
+}
+```
+
+## Request Generation
+
+Models used in this demo - `meta-llama/CodeLlama-7b-hf` and `AMD-Llama-135m` are not chat models, so we will use `completions` endpoint to interact with the pipeline.
+
+Below you can see an exemplary unary request (you can switch `stream` parameter to enable streamed response). Compared to calls to regular continuous batching model, this request has additional parameter `num_assistant_tokens` which specifies how many tokens should a draft model generate before main model validates them.
```
dtrawins, line 140: What are the default values if both params are omitted?

demos/continuous_batching/speculative_decoding/README.md
```diff
@@ -0,0 +1,181 @@
+# How to serve LLM Models in Speculative Decoding Pipeline{#ovms_demos_continuous_batching_speculative_decoding}
+
+Following [OpenVINO GenAI docs](https://docs.openvino.ai/2024/learn-openvino/llm_inference_guide/genai-guide.html#efficient-text-generation-via-speculative-decoding):
+> Speculative decoding (or assisted-generation) enables faster token generation when an additional smaller draft model is used alongside the main model. This reduces the number of infer requests to the main model, increasing performance.
+> 
+> The draft model predicts the next K tokens one by one in an autoregressive manner. The main model validates these predictions and corrects them if necessary - in case of a discrepancy, the main model prediction is used. Then, the draft model acquires this token and runs prediction of the next K tokens, thus repeating the cycle.
+
+This demo shows how to use speculative decoding in the model serving scenario, by deploying main and draft models in a speculative decoding pipeline in a manner similar to regular deployments with continuous batching. 
+
```
dtrawins, line 9: Add a note that the goal of this algorithm is to reduce the latency while keeping the main model accuracy. It give the biggest gain in low concurrency requests.

demos/continuous_batching/speculative_decoding/README.md
```diff
@@ -0,0 +1,181 @@
...
+  "usage": {
+    "prompt_tokens": 9,
+    "completion_tokens": 100,
+    "total_tokens": 109
+  }
+}
+
+```
+
+Picking the right `num_assistant_tokens` might be tricky. High value brings profit when tokens generated by the draft model mostly match the main model. If they don't, tokens are dropped and both models do additional work. For low values such risk is lower, but the potential performance boost is limited.
```
dtrawins, line 177:
```suggestion
High value for `num_assistant_tokens` brings profit when tokens generated by the draft model mostly match the main model. If they don't, tokens are dropped and both models do additional work. For low values such risk is lower, but the potential performance boost is limited. Usually the value of `5` is a good compromise.
```

demos/continuous_batching/speculative_decoding/README.md
mzegla: There are none. One of those must be provided in the request. 
I didn't specify any default because it's hard to recommend a single value that would work well for different combinations of main and draft models.

dtrawins (approved):

src/llm/llm_calculator.proto
```diff
@@ -45,4 +45,21 @@ message LLMCalculatorOptions {
     optional uint32 max_tokens_limit = 9 [default = 4096];
 
     optional bool enable_prefix_caching = 10 [default = false];
+
+    // speculative decoding - draft model config (ignore below fields if you don't want to use speculative decoding)
+    // when draft_models_path is set, the pipeline will use speculative decoding
+    // other values are by default inherited from the main model when speculative decoding is enabled, but can be overridden
+    optional string draft_models_path = 11;
+
```
dkalinowski, line 53: minor: consider creating internal sub-class for all draft model settings
it would be cleaner to define draft model in inner level:
```
    node_options: {
          models_path: "/ovms/src/test/llm_testing/facebook/opt-125m",
          plugin_config: "{\"INFERENCE_PRECISION_HINT\":\"f32\"}"
          cache_size: 1
          draft_model: {
              model_path: "/ovms/src/test/llm_testing/facebook/opt-125m"
          }
        }
    }
```

you can clearly see that model path is not optional, if we decide to include `draft_model` at all
it would naturally restrict situations when user specifies `draft.device` but does not specify `draft.model_name` which is wrong configuration
it would not require manual handling in llmnoderesources.cpp

src/llm/llmnoderesources.hpp
```diff
@@ -36,6 +36,7 @@
 
 #include "../logging.hpp"
 #include "../stringutils.hpp"
+#include "src/llm/llm_calculator.pb.h"
```
dkalinowski, line 39: we already are in src/llm, why do we need to specify it?
```suggestion
#include "llm_calculator.pb.h"
```

src/llm/llmnoderesources.cpp
```diff
@@ -153,6 +152,21 @@ Status LLMNodeResources::initializeLLMNodeResources(LLMNodeResources& nodeResour
 
     nodeResources.device = nodeOptions.device();
 
+    if (!nodeOptions.draft_models_path().empty()) {
```
dkalinowski, line 155: User is not aware of any issue for pbtxt:

```
    ...
    node_options: {
          models_path: "/ovms/src/test/llm_testing/facebook/opt-125m",
          plugin_config: "{\"INFERENCE_PRECISION_HINT\":\"f32\"}"
          cache_size: 1
          draft_device: "/ovms/src/test/llm_testing/facebook/opt-125m"
        }
    }
    ...
```

src/llm/apis/openai_completions.hpp
```diff
@@ -140,6 +146,11 @@ struct OpenAIChatCompletionsRequest {
 
         if (logprobschat || logprobs > 0)
             config.logprobs = 1;
+        // Speculative decoding specific
+        if (numAssistantTokens.has_value())
+            config.num_assistant_tokens = numAssistantTokens.value();
+        if (assistantConfidenceThreshold.has_value())
```
dkalinowski, line 152: I dont see a place where setting both parameters (which are exclusive) fails the request, do we have test for that?

mzegla: https://github.com/openvinotoolkit/model_server/pull/2777/files#diff-dc1e3d8f1c59d392baedaaa031edd2ef7bbb062fa25cd68fe99b00011f9f770bR492

mzegla: We have only positive tests now

src/llm/llmnoderesources.hpp
mzegla: It seems we have to. Doesn't compile with your suggestion.

src/llm/apis/openai_completions.hpp
mzegla: added negative tests

src/llm/llmnoderesources.cpp
```diff
@@ -165,6 +165,12 @@ Status LLMNodeResources::initializeLLMNodeResources(LLMNodeResources& nodeResour
             ov::genai::scheduler_config(draftSchedulerConfig));
         nodeResources.pluginConfig.insert(draftModelConfig);
         nodeResources.isSpeculativePipeline = true;
+    } else if (nodeOptions.has_draft_max_num_batched_tokens() || nodeOptions.has_draft_cache_size()
+               || nodeOptions.has_draft_dynamic_split_fuse() || nodeOptions.has_draft_max_num_seqs()
+               || nodeOptions.has_draft_block_size() || nodeOptions.has_draft_device()) {
+        // Consider moving draft parameters to separate structure in node options, so it's validated on the proto level
+        SPDLOG_ERROR("Draft model path is not provided, but draft scheduler options are set. Ignoring draft scheduler options.");
```
dkalinowski, line 170: wrong msg

dkalinowski (approved):

src/llm/llmnoderesources.cpp
```diff
@@ -153,6 +152,25 @@ Status LLMNodeResources::initializeLLMNodeResources(LLMNodeResources& nodeResour
...
+            draftModelsPath = fsDraftModelsPath.string();
+        }
+        auto draftSchedulerConfig = prepareDraftModelSchedulerConfig(nodeOptions);
+        auto draftModelConfig = ov::genai::draft_model(draftModelsPath, nodeOptions.draft_device(),
+            ov::genai::scheduler_config(draftSchedulerConfig));
+        nodeResources.pluginConfig.insert(draftModelConfig);
+        nodeResources.isSpeculativePipeline = true;
+    } else if (nodeOptions.has_draft_max_num_batched_tokens() || nodeOptions.has_draft_cache_size() || nodeOptions.has_draft_dynamic_split_fuse() || nodeOptions.has_draft_max_num_seqs() || nodeOptions.has_draft_block_size() || nodeOptions.has_draft_device()) {
+        // Consider moving draft parameters to separate structure in node options, so it's validated on the proto level
+        SPDLOG_ERROR("Draft model path is not provided, but draft scheduler options are set. Ignoring draft scheduler options.");
```
mzegla, line 170:
```suggestion
        SPDLOG_ERROR("Draft model path is not provided, but draft scheduler options are set.");
```

PR #2787, opened 2024-11-04
Model api compatibility

dtrawins:
### 🛠 Summary

CVS-156627
model configuration from HF config and tokenizer_config moved to model_info path in rt_info

### 🧪 Checklist

- [ ] Unit tests added.
- [ ] The documentation updated.
- [ ] Change follows security best practices.
``

PR #2788, opened 2024-11-04
Rerank document chunking

dkalinowski:
### 🛠 Summary

CVS-156597

### 🧪 Checklist

- [x] Unit tests for chunking added, but more to be added when we have e2e tests


src/rerank/rerank_calculator.cc
```diff
@@ -92,6 +96,23 @@ class RerankCalculator : public CalculatorBase {
                 this->sep_token = tokenizer_session->getModelConfig().at("sep_token_id").as<int64_t>();
             }
             this->pad_token = tokenizer_session->getModelConfig().at("pad_token_id").as<int64_t>();
+
+            auto it = rerank_session->getModelConfig().find("config");
+            if (it != rerank_session->getModelConfig().end()) {
+                auto it = rerank_session->getModelConfig().at("config").as<ov::AnyMap>().find("max_position_embeddings");
```
dtrawins, line 102: it will be changed from config to model_info

dtrawins: compare with https://github.com/openvinotoolkit/model_server/pull/2787 - it checks two possible keys

src/rerank/rerank_calculator.cc
```diff
@@ -247,20 +310,18 @@ class RerankCalculator : public CalculatorBase {
...
+        scores.resize(actual_batch_size, 0);
+
+        size_t logits_dim = logits.get_shape()[1];
+
+        for (int i = 0; i < input_ids.get_shape()[0]; ++i) {
+            size_t score_index = chunkMapping[i];
+            if (score_index >= actual_batch_size)
+                throw std::runtime_error("score_index out of bounds");  // should never happen
+            float logit = logits_dim > 1 ? reinterpret_cast<float*>(logits.data())[i * logits_dim + 1] : reinterpret_cast<float*>(logits.data())[i];
+            float score = 1 / (1 + std::exp(-logit));
```
dtrawins, line 322: do we need to calculate it for every chunk or the most likely would be sufficient?

src/rerank/rerank_utils.cpp
```diff
@@ -187,4 +187,104 @@ std::optional<bool> RerankHandler::getReturnDocuments() const { return request.r
 std::optional<std::vector<std::string>> RerankHandler::getRankFields() const { return request.rankFields; }
 std::optional<int> RerankHandler::getMaxChunksPerDoc() const { return request.maxChunksPerDoc; }
 
+absl::Status chunkDocuments(
+    const ov::Tensor& in_input_ids, const ov::Tensor& in_attention_mask,
+    ov::Tensor& out_input_ids, ov::Tensor& out_attention_mask,
+    std::vector<size_t>& chunk_mapping, size_t max_tokens_per_chunk,
+    int64_t pad_token) {
+
```
dtrawins, line 195: comment would be useful with explanations what this method does

dkalinowski: doesnt unit test case explain it? with example input and output
if not, i can also describe it here

src/rerank/rerank_calculator.cc
dkalinowski: what do you mean most likely?
this is exactly what we do - select highest score from all the chunks of given document

dtrawins (approved):

michalkulakowski (approved):

src/rerank/rerank_calculator.cc
dkalinowski: done

src/rerank/rerank_utils.cpp
dkalinowski: added

PR #2810, opened 2024-11-10
decode model name from rest api path

dtrawins:
### 🛠 Summary

CVS-155466
Allow special characters in the model name. Model can can be url encoded in REST API calls. It is in line with triton interface.

### 🧪 Checklist

- [x] Unit tests added.
``


src/http_rest_api_handler.cpp
```diff
@@ -654,6 +655,8 @@ Status HttpRestApiHandler::parseRequestComponents(HttpRequestComponents& request
         return StatusCode::REST_UNSUPPORTED_METHOD;
     }
 
+    SPDLOG_DEBUG("PATH:{}.{}", request_path, urlDecode(request_path));
```
dkalinowski, line 658: Please give more description in the log what is going on. For example:
```
SPDLOG_DEBUG("Parsing http request path. Before parsing: {}; after parsing: {}", request_path, urlDecode(request_path));
```

src/http_rest_api_handler.cpp
```diff
@@ -1123,4 +1126,12 @@ Status HttpRestApiHandler::processConfigStatusRequest(std::string& response, Mod
     return StatusCode::OK;
 }
 
+std::string urlDecode(const std::string& encoded) {
+    int output_length;
+    const auto decoded_value = curl_easy_unescape(nullptr, encoded.c_str(), static_cast<int>(encoded.length()), &output_length);
```
dkalinowski, line 1131: This function may fail: https://curl.se/libcurl/c/curl_easy_unescape.html
Please add check for that and abort in such case

dkalinowski (approved):

PR #2817, opened 2024-11-12
Adjust LLM docs

mzegla:


docs/llm/reference.md
```diff
@@ -161,6 +161,12 @@ convert_tokenizer -o {target folder name} --utf8_replace_mode replace --with-det
 
 Check [tested models](https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/models/real_models).
 
+## Input preprocessing
+
+When loading the pipeline, model server loads not only model files, but also tokenizer configuration present in `tokenizer_config.json`. Make sure your tokenizer configuration is correct - results accuracy might depend on it. Also note that `tokenizer_config.json` with fields filled with incorrect datatypes (for example `"bos_token": {"key": "val"}`) **will not work at all and the pipeline with such tokenizer config will not load**.
```
dtrawins, line 166: I'm not sure if this comment is relevant while this file is generated automatically by optimum-intel package

docs/llm/reference.md
```diff
@@ -172,7 +178,7 @@ Loading chat template proceeds as follows:
 
 **Note**: If both `template.jinja` file and `chat_completion` field from `tokenizer_config.json` are successfully loaded, `template.jinja` takes precedence over `tokenizer_config.json`.
 
-If there are errors in loading or reading files or fields (they exist but are wrong) no template is loaded and servable will not respond to `/chat/completions` calls. 
+If there are errors in loading or reading files or fields (they exist but are wrong) servable will not load at all.
```
dtrawins, line 181:
```suggestion
In case the tokenizer_conifg.json is corrupted, the servable will fail to load.
```

docs/llm/reference.md
mzegla: In general yes, but this part describes serving behavior, so I think it should be kept here as a reference of what to expect.

pgladkows, line 166:
```suggestion
When loading the pipeline, model server loads not only model files, but also tokenizer configuration present in `tokenizer_config.json`. Make sure your tokenizer configuration is correct -  accuracy results might depend on it. Also note that `tokenizer_config.json` with fields filled with incorrect datatypes (for example `"bos_token": {"key": "val"}`) **will not work at all and the pipeline with such tokenizer config will not load**. 
```

pgladkows (approved):

dtrawins (approved):

PR #2885, opened 2024-11-28
Win dependencies - openvino

rasapala:
### 🛠 Summary

JIRA CVS-157903
Adding windows_install_dependencies.bat
It will install openvino and update workspace for the location in CI.

### 🧪 Checklist

- [ ] Unit tests added.
``


windows_install_dependencies.bat
```diff
@@ -0,0 +1,56 @@
...
+
+:: Set proper PATH environment variable: Remove other python paths and add c:\opt with bazel, wget to PATH
+set "PATH=%setPath%"
+
+:: OpenVINO
+set "openvino_dir=w_openvino_toolkit_windows_2024.5.0.17288.7975fa5da0c_x86_64"
+set "openvino_ver=w_openvino_toolkit_windows_2024.5.0.17288.7975fa5da0c_x86_64.zip"
```
mzegla, line 28: Can we update to 2025.0?

windows_install_dependencies.bat
```diff
@@ -0,0 +1,56 @@
...
+set "openvino_http=https://storage.openvinotoolkit.org/repositories/openvino/packages/2024.5/windows/"
+set "openvino_wget=wget -P %BAZEL_SHORT_PATH%\ %openvino_http%%openvino_ver%"
+set "openvino_zip=%BAZEL_SHORT_PATH%\%openvino_ver%"
+set "openvino_workspace=C:\\\\opt\\\\intel\\\\openvino\\\\runtime"
+set "openvino_new_worksapce=C:\\%1\\openvino\\runtime"
```
mzegla, line 33: why do we need this?

windows_install_dependencies.bat
```diff
@@ -0,0 +1,56 @@
...
+
+:: Set proper PATH environment variable: Remove other python paths and add c:\opt with bazel, wget to PATH
+set "PATH=%setPath%"
+
+:: OpenVINO
+set "openvino_dir=w_openvino_toolkit_windows_2024.5.0.17288.7975fa5da0c_x86_64"
+set "openvino_ver=w_openvino_toolkit_windows_2024.5.0.17288.7975fa5da0c_x86_64.zip"
+set "openvino_http=https://storage.openvinotoolkit.org/repositories/openvino/packages/2024.5/windows/"
+set "openvino_wget=wget -P %BAZEL_SHORT_PATH%\ %openvino_http%%openvino_ver%"
```
mzegla, line 30: Why do we keep this command in the variable, but other commands are used directly?

windows_install_dependencies.bat
```diff
@@ -0,0 +1,56 @@
+::
```
mzegla, line 1: Is this supposed to be executed in CI only or also by the user? If it's for the user as well, perhaps we should include it in the docs along with the information that `wget` and `tar` are needed for this to work.

windows_install_dependencies.bat
rasapala: We need to update path to openvino libs in bazel workspace file in the Ci.

windows_install_dependencies.bat
rasapala: I will update the documentation once the script installs all of the dependencies, currently lets stick to manual install instructions.

mzegla (approved):

dkalinowski (approved):

PR #2900, opened 2024-12-06
Add base64 image deserialization for llm calculator

michalkulakowski:
### 🛠 Summary

CVS-148824
CVS-158168

### 🧪 Checklist

- [x] Unit tests added.
``

### DIFF
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -18,11 +18,13 @@
 
 #include <cmath>
 
+#include <opencv2/opencv.hpp>
 #include <rapidjson/stringbuffer.h>
 #include <rapidjson/writer.h>
 
 #include "../../logging.hpp"
 #include "../../profiler.hpp"
+#include "absl/strings/escaping.h"
 
 using namespace rapidjson;
 
@@ -65,7 +67,6 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
             return absl::InvalidArgumentError("echo accepts values true or false");
         request.echo = it->value.GetBool();
     }
-
     // specific part of max_tokens validation due to echo dependency
     if (request.maxTokens == 0) {
         if (!request.echo)
@@ -75,8 +76,30 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
     return absl::OkStatus();
 }
 
-absl::Status OpenAIChatCompletionsHandler::parseChatCompletionsPart() {
-    // messages: [{role: content}, {role: content}, ...]; required
+static ov::element::Type_t getOvTypeFromMatType(int matType) {
+    switch (matType) {
+    case CV_32F:
+        return ov::element::f32;
+    case CV_64F:
+        return ov::element::f64;
+    case CV_16F:
+        return ov::element::f64;
+    case CV_16S:
+        return ov::element::f16;
+    case CV_8U:
+        return ov::element::u8;
+    case CV_8S:
+        return ov::element::i8;
+    case CV_16U:
+        return ov::element::u16;
+    case CV_32S:
+        return ov::element::i32;
+    default:
+        return ov::element::undefined;
+    }
+}
+
+absl::Status OpenAIChatCompletionsHandler::parseMessages() {
     auto it = doc.FindMember("messages");
     if (it == doc.MemberEnd())
         return absl::InvalidArgumentError("Messages missing in request");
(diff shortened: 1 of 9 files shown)


src/llm/apis/openai_completions.cpp
```diff
@@ -75,32 +76,139 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
...
-            chat[member->name.GetString()] = member->value.GetString();
+            if (member->value.IsString()) {
+                continue;
+            } else {
+                if (member->name.GetString() == std::string("content") && member->value.IsArray()) {
+                    jsonChanged = true;
+                    Value contentText;
+                    for (auto& v : member->value.GetArray()) {
+                        if (!v.IsObject()) {
+                            return absl::InvalidArgumentError("Invalid message structure - content array should contains objects");
```
dkalinowski, line 126:
```suggestion
                            return absl::InvalidArgumentError("Invalid message structure - content array should contain objects");
```

src/llm/apis/openai_completions.cpp
```diff
@@ -75,32 +76,139 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
...
+                            if (!entry.HasMember("image_url") || !entry["image_url"].IsObject()) {
+                                return absl::InvalidArgumentError("Invalid message structure - content image_url missing");
+                            }
+                            auto imageUrl = entry["image_url"].GetObject();
+                            if (!imageUrl.HasMember("url") || !imageUrl["url"].IsString()) {
+                                return absl::InvalidArgumentError("Invalid message structure - image_url does not have url field");
+                            }
+                            std::string url = imageUrl["url"].GetString();
+                            std::string pattern = ";base64,";
+                            std::size_t found = url.find(pattern);
```
dkalinowski, line 149:
```suggestion
                            std::size_t patternPosition = url.find(pattern);
```

src/llm/apis/openai_completions.cpp
```diff
@@ -75,32 +76,139 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
...
+                        } else if (type == std::string("image_url")) {
+                            if (!entry.HasMember("image_url") || !entry["image_url"].IsObject()) {
+                                return absl::InvalidArgumentError("Invalid message structure - content image_url missing");
+                            }
+                            auto imageUrl = entry["image_url"].GetObject();
+                            if (!imageUrl.HasMember("url") || !imageUrl["url"].IsString()) {
+                                return absl::InvalidArgumentError("Invalid message structure - image_url does not have url field");
+                            }
+                            std::string url = imageUrl["url"].GetString();
+                            std::string pattern = ";base64,";
```
dkalinowski, line 148: since OpenAI format specifies:
```
"data:image/jpeg;base64,{base64_image}"
```

we should validate for integrity - `data:image/jpeg;base64,` before actual base64 data
another question is, should we also allow for other formats than jpeg?

src/llm/apis/openai_completions.cpp
```diff
@@ -75,32 +76,139 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
...
+                            size_t offset = found + pattern.length();
+                            if (found == std::string::npos) {
+                                offset = 0;
+                            }
+                            url.erase(0, offset);
+                            std::string decoded;
+                            if (!absl::Base64Unescape(url, &decoded)) {
+                                return absl::InvalidArgumentError("Invalid base64 string in request");
+                            }
+                            std::vector<unsigned char> data(decoded.begin(), decoded.end());
```
dkalinowski, line 159: do we really need to copy in order to change from char to unsigned char?

src/llm/apis/openai_completions.cpp
```diff
@@ -75,32 +76,139 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
...
+                                return absl::InvalidArgumentError("Invalid message structure - image_url does not have url field");
+                            }
+                            std::string url = imageUrl["url"].GetString();
+                            std::string pattern = ";base64,";
+                            std::size_t found = url.find(pattern);
+                            size_t offset = found + pattern.length();
+                            if (found == std::string::npos) {
+                                offset = 0;
+                            }
+                            url.erase(0, offset);
```
dkalinowski, line 154: do we need to erase? maybe we could construct the vector by passing begin offseted by patternPosition value?

src/llm/apis/openai_completions.cpp
```diff
@@ -75,32 +76,139 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
...
+                            try {
+                                image = cv::imdecode(dataMat, cv::IMREAD_UNCHANGED);
+                            } catch (const cv::Exception& e) {
+                                return absl::InvalidArgumentError("Error during string to mat conversion");
+                            }
+                            std::vector<size_t> shape;
+                            shape.push_back(image.rows);
+                            shape.push_back(image.cols);
+                            shape.push_back(image.channels());
+                            ov::Tensor tensor(getOvTypeFromMatType(image.depth()), shape);
```
dkalinowski, line 171: shouldnt we check if getOvTypeFromMatType is undefined and reject such image? do we have test for that?

src/llm/apis/openai_completions.cpp
```diff
@@ -75,32 +76,139 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
...
+                        }
+                    }
+                    member->value = contentText;
+                } else {
+                    return absl::InvalidArgumentError("Invalid message structure - content should be string or array");
+                }
+            }
         }
     }
+    if (jsonChanged) {
```
dkalinowski, line 188: I'm not sure if this trick really speeds in the processing since you need to make a copy anyway: `request.processedJson = buffer.GetString();`

src/llm/apis/openai_completions.cpp
```diff
@@ -109,15 +217,27 @@ absl::Status OpenAIChatCompletionsHandler::parseChatCompletionsPart() {
     if (request.logprobschat && request.stream) {
         return absl::InvalidArgumentError("logprobs are not supported in streaming mode.");
     }
-
+    // max_completion_tokens: uint; optional
+    it = doc.FindMember("max_completion_tokens");
```
dkalinowski, line 221: shouldn't it be in parseCommonPart?

does it differ from "max_tokens" processing? if not, we should reorganize it to do not duplicate the code and its value validation

src/llm/apis/openai_completions.hpp
```diff
@@ -172,6 +171,8 @@ class OpenAIChatCompletionsHandler {
     std::optional<std::string> getPrompt() const;
     std::optional<int> getNumReturnSequences() const;
     StreamOptions getStreamOptions() const;
+    std::string& getProcessedJson();
```
dkalinowski, line 174: does it really need to be non-const? if there is any reason for that, why do we need getters if we can edit it anyway?

src/llm/apis/openai_completions.cpp
michalkulakowski: max_completion_tokens refers only to chat completions api, not to the completions api

src/llm/apis/openai_completions.cpp
```diff
@@ -75,32 +76,147 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
...
+                            std::size_t pos = url.find(pattern);
+                            size_t offset = pos + pattern.length();
+                            if (pos == std::string::npos) {
+                                offset = 0;
+                            }
+                            std::string decoded;
+                            if (!absl::Base64Unescape(std::string_view(url.data() + offset, url.size() - offset), &decoded)) {
+                                return absl::InvalidArgumentError("Invalid base64 string in request");
+                            }
+                            cv::Mat rawData(1, decoded.size(), CV_8UC1, (void*)decoded.data());
```
dkalinowski, line 161: what is this magic number `1`?

src/llm/apis/openai_completions.cpp
```diff
@@ -75,32 +76,147 @@ absl::Status OpenAIChatCompletionsHandler::parseCompletionsPart() {
...
+        doc.Accept(writer);
+        request.processedJson = buffer.GetString();
+    }
+    return absl::OkStatus();
+}
+
+const std::string& OpenAIChatCompletionsHandler::getProcessedJson() const {
+    return request.processedJson;
+}
+std::vector<ov::Tensor> OpenAIChatCompletionsHandler::getImages() const {
```
dkalinowski, line 208: can we return const reference to the vector? I know ov::Tensor is not heavy, but constructing the vector is

src/tensor_conversion.cpp
```diff
@@ -78,11 +78,10 @@ static bool isPrecisionEqual(int matPrecision, ovms::Precision tensorPrecision)
 
 static cv::Mat convertStringToMat(const std::string& image) {
     OVMS_PROFILE_FUNCTION();
-    std::vector<unsigned char> data(image.begin(), image.end());
-    cv::Mat dataMat(data, true);
+    cv::Mat rawData(1, image.size(), CV_8UC1, (void*)image.data());
```
dkalinowski, line 81: 1=magic number, please define with name

src/test/http_openai_handler_test.cpp
```diff
@@ -243,3 +245,252 @@ TEST_F(HttpOpenAIHandlerTest, GraphWithANameDoesNotExist) {
...
+            }
+          }
+        ]
+      }
+    ]
+  })";
+    doc.Parse(json.c_str());
+    ASSERT_FALSE(doc.HasParseError());
+    std::shared_ptr<ovms::OpenAIChatCompletionsHandler> apiHandler = std::make_shared<ovms::OpenAIChatCompletionsHandler>(doc, ovms::Endpoint::CHAT_COMPLETIONS, std::chrono::system_clock::now(), *tokenizer);
+    EXPECT_EQ(apiHandler->parseMessages(), absl::OkStatus());
```
dkalinowski, line 279: shouldnt this be assert? if it fails, there is no point in going further with test
the same goes for other checks which may crash further testing (EXPECT does not abort)

src/test/http_openai_handler_test.cpp
```diff
@@ -243,3 +245,252 @@ TEST_F(HttpOpenAIHandlerTest, GraphWithANameDoesNotExist) {
...
+    EXPECT_EQ(image.get_size(), 3);
+    std::vector<uint8_t> expectedBytes = {160, 181, 110};
+    for (size_t i = 0; i < image.get_size(); i++) {
+        EXPECT_EQ(expectedBytes[i], ((uint8_t*)image.data())[i]);
+    }
+    json = apiHandler->getProcessedJson();
+    EXPECT_EQ(json, std::string("{\"model\":\"llama\",\"messages\":[{\"role\":\"user\",\"content\":\"What is in this image?\"}]}"));
+}
+
+TEST_F(HttpOpenAIHandlerParsingTest, ParsingImageJpeg) {
```
dkalinowski, line 293: test name does not say that this is negative scenario - atleast one text message should exist (is it right?)

src/test/http_openai_handler_test.cpp
```diff
@@ -243,3 +245,252 @@ TEST_F(HttpOpenAIHandlerTest, GraphWithANameDoesNotExist) {
...
+        EXPECT_EQ(image.get_size(), 3);
+        for (size_t i = 0; i < image.get_size(); i++) {
+            EXPECT_EQ(expectedBytes[i], ((uint8_t*)image.data())[i]);
+        }
+    }
+    json = apiHandler->getProcessedJson();
+    EXPECT_EQ(json, std::string("{\"model\":\"llama\",\"messages\":[{\"role\":\"user\",\"content\":\"What is in this image?\"},{\"role\":\"user\",\"content\":\"What is in this image?\"},{\"role\":\"system\",\"content\":\"What is Openvino?\"}]}"));
+}
+
+TEST_F(HttpOpenAIHandlerParsingTest, ParsingMessagesWithInvalidContentType) {
```
dkalinowski, line 412: In general, the new test names do not say what is expected behavior, my suggestion is that it could contain it, for example 
ParsingMultipleMessages -> ParsingMultipleTextMessagesWithImagesSucceeds
ParsingMessagesImageStringWithNoPrefix -> ParsingMessagesImageStringWithNoPrefixSucceeds

what do you think?

src/llm/apis/openai_completions.hpp
```diff
@@ -158,7 +157,7 @@ class OpenAIChatCompletionsHandler {
     size_t processedTokens = 0;  // tracks overall number of tokens processed by the pipeline
 
     absl::Status parseCompletionsPart();
-    absl::Status parseChatCompletionsPart();
+    absl::Status parseChatCompletionsPart(uint32_t maxTokensLimit);
```
dtrawins, line 160: is it related to the issues from description?

src/test/http_openai_handler_test.cpp
```diff
@@ -243,3 +245,252 @@ TEST_F(HttpOpenAIHandlerTest, GraphWithANameDoesNotExist) {
...
+            }
+          }
+        ]
+      }
+    ]
+  })";
+    doc.Parse(json.c_str());
+    ASSERT_FALSE(doc.HasParseError());
+    std::shared_ptr<ovms::OpenAIChatCompletionsHandler> apiHandler = std::make_shared<ovms::OpenAIChatCompletionsHandler>(doc, ovms::Endpoint::CHAT_COMPLETIONS, std::chrono::system_clock::now(), *tokenizer);
+    ASSERT_EQ(apiHandler->parseMessages(), absl::OkStatus());
```
dtrawins, line 279: while the payload is invalid, why the assert is expected to pass?

dkalinowski (approved):

src/test/http_openai_handler_test.cpp
```diff
@@ -243,3 +245,252 @@ TEST_F(HttpOpenAIHandlerTest, GraphWithANameDoesNotExist) {
...
+            "image_url": {
+              "url":  "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAAEElEQVR4nGLK27oAEAAA//8DYAHGgEvy5AAAAABJRU5ErkJggg=="
+            }
+          }
+        ]
+      }
+    ]
+  })";
+    doc.Parse(json.c_str());
+    ASSERT_FALSE(doc.HasParseError());
```
dtrawins, line 338: It should be invalid request without base64 type. is it expected to pass?

src/test/llmnode_test.cpp
```diff
@@ -1684,6 +1770,88 @@ TEST_F(LLMHttpParametersValidationTest, maxTokensExceedsUint32Size) {
...
+        handler->dispatchToProcessor(endpointChatCompletions, requestBody, &response, comp, responseComponents, &writer),
+        ovms::StatusCode::MEDIAPIPE_EXECUTION_ERROR);
+}
+
+TEST_F(LLMHttpParametersValidationTest, maxCompletionsTokensExceeds4000WhenIgnoreEosTrue) {
+    std::string requestBody = R"(
+        {
+            "model": "llmDummyKFS",
+            "stream": false,
+            "ignore_eos": true,
```
dtrawins, line 1839: just a note that after we have a proper handling of max context, we could drop this 4000 token limit to ignore_eos

src/llm/apis/openai_completions.hpp
michalkulakowski: yes, it's related to adding support of max_completion_tokens

dtrawins (approved):

PR #2901, opened 2024-12-09
Windows dependencies 3

rasapala:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``


WORKSPACE
```diff
@@ -477,6 +477,3 @@ git_repository(
     remote = "https://github.com/nlohmann/json/",
     tag = "v3.11.3",
 )
-
```
dkalinowski, line 480: it is common practice to leave single blank line at the end of source file

docs/windows_developer_guide.md
```diff
@@ -27,19 +27,38 @@ Install build tools for VS:
 
 https://aka.ms/vs/17/release/vs_BuildTools.exe
 
-Mark c++ Desktop app and v142 CPP platform toolset.
+Mark required options for installation:
+- c++ Desktop development with C++
+- MSVC v143 CPP - VS 2022 C++ platform toolset.
+- C++ CMake tools for Windows platform toolset.
+- MSVC v142 CPP - VS 2022 C++ platform toolset.
```
mzegla, line 34: Do we need both toolsets?

windows_install_dependencies.bat
```diff
@@ -34,8 +34,10 @@ IF "%2"=="1" (
...
+:: Python 39 needs to be first in the windows path
+set "setPath=C:\opt\Python39\;C:\opt\Python39\Scripts\;%PATH%;c:\opt"
```
mzegla, line 40:
```suggestion
set "setPath=C:\opt\Python39\;C:\opt\Python39\Scripts\;%PATH%;C:\opt"
```

windows_install_dependencies.bat
```diff
@@ -145,12 +158,133 @@ set "bazel_path=%opt_install_dir%\bazel.exe"
...
     )
 ) ELSE (
     wget -O %bazel_path% https://github.com/bazelbuild/bazel/releases/download/6.4.0/bazel-6.4.0-windows-x86_64.exe
+    if %errorlevel% neq 0 exit /b %errorlevel%
+)
+
+:::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+::::::::::::::::::::::: Python39
+set "python39_path=%opt_install_dir%\Python39\"
+set "python39_system=C:\Program Files\Python39\"
```
mzegla, line 175: Why do we consider both options? I suppose Python 3.9 is not a default option for Windows now.

docs/windows_developer_guide.md
rasapala: Yes, the 143 is required to install cmake.

WORKSPACE
rasapala: One line is left at the end, the diff shows wrong state.

windows_install_dependencies.bat
rasapala: C:\Program Files\Python39\ this is preinstalled on CI nodes. We use it for the purpose of setting python39 for ovms builds. If the user also has this path or has already installed python in c:/opt the install dependencies will work on laptop. Otherwise it will fail and tell user to install Python.

docs/windows_developer_guide.md
mzegla: Why do we keep information about v142 then?

rasapala: Because we need cmake - required 143, and we need v142 to compile OVMS.

mzegla: Will it be possible to unify in the future? Because it sounds like asking for problems in the future 😕

mzegla (approved):

dkalinowski (approved):

PR #2943, opened 2024-12-23
Update documentation about accelerators

dtrawins:
### 🛠 Summary

CVS-149070
Drop NVIDIA plugin from the maintained list
Drop MULTI plugin - replaced by AUTO with THROUGHPUT_CUMULATIVE hint
Add BATCH virtual device
Refresh documentation about accelerators

### 🧪 Checklist

- [x] Unit tests added.
``


docs/accelerators.md
```diff
@@ -160,52 +139,30 @@ docker run --rm -d --device=/dev/dri --group-add=$(stat -c "%g" /dev/dri/render*
...
 ```bash
-docker run --device /dev/accel -p 9000:9000 --group-add=$(stat -c "%g" /dev/dri/render* | head -n 1) -u $(id -u):$(id -g) \
--v ${PWD}/models/public/resnet-50-tf:/opt/model openvino/model_server:latest --model_path /opt/model --model_name resnet --port 9000 --target_device NPU
-```
-Check more info about the [NPU driver for Linux](https://github.com/intel/linux-npu-driver).
\ No newline at end of file
+docker run -v ${PWD}/models/public/resnet-50-tf:/opt/model -p 9001:9001 openvino/model_server:latest-gpu \
+--model_path /opt/model --model_name resnet --port 9001 \
+--plugin_config '{"AUTO_BATCH_TIMEOUT": 200}' \
+--target_device BATCH:CPU(16)
```
dkalinowski, line 167: You don't specify what is the `(16)` in this example - maybe its worth mentioning? Do I understand it correct that it is max batch size when grouping together?

docs/accelerators.md
```diff
@@ -47,53 +46,19 @@ docker run --rm -it  --device=/dev/dxg --volume /usr/lib/wsl:/usr/lib/wsl -u $(i
...
-      "batch_size": "1",
-      "target_device": "MULTI:GPU,CPU"}
-   }]
-}' >> models/public/resnet-50-tf/config.json
-```
+## Using NPU device Plugin
 
-To start OpenVINO Model Server, with the described config file placed as `./models/config.json`, set the `grpc_workers` parameter to match the `nireq` field in config.json
-and use the run command, like so:
+OpenVINO Model Server can support using [NPU device](https://docs.openvino.ai/2024/openvino-workflow/running-inference/inference-devices-and-modes/npu-device.html)
```
mzegla, line 52:
```suggestion
OpenVINO Model Server supports using [NPU device](https://docs.openvino.ai/2024/openvino-workflow/running-inference/inference-devices-and-modes/npu-device.html)
```
I think we can remove "can" if the plugin is always included in the distribution.

docs/accelerators.md
```diff
@@ -135,6 +100,8 @@ docker run --rm -d --device=/dev/dri --group-add=$(stat -c "%g" /dev/dri/render*
 ```
 
 The `Auto Device` plugin can also use the [PERFORMANCE_HINT](performance_tuning.md) plugin config property that enables you to specify a performance mode for the plugin.
+While LATENCY and THROUGHPUT hint can select one target device with your preferred performance option, the CUMULATIVE_THROUGHPUT option enables running inference on multiple devices for higher throughput. 
+With CUMULATIVE_THROUGHPUT, AUTO loads the network model to all available devices (specified by AUTO) in the candidate list, and then runs inference on them based on the default or specified priority.
```
mzegla, line 104:
```suggestion
With CUMULATIVE_THROUGHPUT hint, AUTO plugin loads the network model to all available devices (specified by the plugin) in the candidate list, and then runs inference on them based on the default or specified priority.
```

src/ov_utils.cpp
```diff
@@ -122,6 +122,11 @@ Status validatePluginConfiguration(const plugin_config_t& pluginConfig, const st
         std::string deviceName;
 
         while (getline(ss, deviceName, deviceDelimiter)) {
+            char bracket = '(';
+            auto bracketPos = deviceName.find(bracket);
+            if (bracketPos != std::string::npos) {
+                deviceName = deviceName.substr(0, bracketPos);
```
mzegla, line 128: Why are we doing this? Don't we exclude batching info here?

dtrawins: to get physical device and retrieve supported properties, we need to omit content in brackets. For the target device in model compilation, it should be included.

mzegla (commented):
If automated tests are supposed to check this I think we should change `console` to `bat` in binary startup commands it's only executed on Windows for now at least.

mzegla (approved):

dkalinowski (approved):

PR #2961, opened 2025-01-10
Upgrade drogon, use setCloseCallback

dkalinowski:
### 🛠 Summary

CVS-159406

### 🧪 Checklist

### DIFF
--- a/src/drogon_http_async_writer_impl.cpp
+++ b/src/drogon_http_async_writer_impl.cpp
@@ -76,6 +76,13 @@ bool DrogonHttpAsyncWriterImpl::IsDisconnected() const {
 }
 
 void DrogonHttpAsyncWriterImpl::RegisterDisconnectionCallback(std::function<void()> callback) {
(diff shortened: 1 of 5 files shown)


src/drogon_http_async_writer_impl.cpp
```diff
@@ -76,6 +76,12 @@ bool DrogonHttpAsyncWriterImpl::IsDisconnected() const {
 }
 
 void DrogonHttpAsyncWriterImpl::RegisterDisconnectionCallback(std::function<void()> callback) {
+    const auto& weakConnPtr = requestPtr->getConnectionPtr();
+    if (auto connPtr = weakConnPtr.lock()) {
```
mzegla, line 80: do we need any handling if this is false?

dkalinowski: this could happen when the client disconnects before we even try to install callback, we can do nothing in this case

dtrawins (approved):

mzegla (approved):

PR #2963, opened 2025-01-13
Disable mmap for models on windows

dkalinowski:
### 🛠 Summary

CVS-159635
CVS-160412

This fixes single model versioning feature on windows (unload/load models when removing/adding model directory)
The trade off is that models are loading slower

### 🧪 Checklist

- [x] Unit tests added.


dkalinowski:
https://github.com/openvinotoolkit/model_server/pull/2976

PR #2985, opened 2025-01-27
Fix python install

rasapala:
### 🛠 Summary

Adding checks on python.exe
New uninstall and install scripts.
Echo off.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/docs/windows_developer_guide.md
+++ b/docs/windows_developer_guide.md
@@ -10,7 +10,7 @@ Following the steps below requires 40GB of free disk space.
 
 ## Install git
 Download from https://github.com/git-for-windows/git/releases/download/v2.47.1.windows.2/Git-2.47.1.2-64-bit.exe
-Or run command in windows cmd.exe terminal:
+Or run command in windows cmd.exe terminal as Administrator:
 ```bat
 winget install --id Git.Git -e --source winget
 ```
@@ -41,7 +41,7 @@ Follow instructions in the link below:
 https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development
 
 ## Run Command Prompt
-Press Windows Start and run the cmd.exe terminal.
+Press Windows Start and run the cmd.exe terminal as Administrator.
 Run commands in this prompt is not stated otherwise.
 
 ## Get the code
--- a/docs/windows_developer_guide.md
+++ b/docs/windows_developer_guide.md
@@ -38,8 +38,8 @@ Press Windows Start and run the cmd.exe terminal as Administrator.
 Run commands in this prompt is not stated otherwise.
 
 ## Pull OpenVINO Model Server source
-Install git from https://git-scm.com/downloads/win
-After that run commands in terminal:
+> `Git` is required to complete this step. If you don't have it on your system, download it from https://git-scm.com/downloads/win and install before you continue.
+Run below commands in terminal to clone model server repository:
 ```bat
 mkdir C:\git
 cd C:\git\
--- a/docs/windows_developer_guide.md
+++ b/docs/windows_developer_guide.md
@@ -8,13 +8,6 @@ This instruction was tested on Windows 11 and Windows 10 OS.
 # Install prerequisites
 Following the steps below requires 40GB of free disk space.
 
-## Install git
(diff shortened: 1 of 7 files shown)


docs/windows_developer_guide.md
```diff
@@ -10,7 +10,7 @@ Following the steps below requires 40GB of free disk space.
 
 ## Install git
 Download from https://github.com/git-for-windows/git/releases/download/v2.47.1.windows.2/Git-2.47.1.2-64-bit.exe
-Or run command in windows cmd.exe terminal:
+Or run command in windows cmd.exe terminal as Administrator:
```
dtrawins, line 13: I think we don't need to maintain instruction for git installation. Maybe it would be sufficient to inform about git installation and point to https://git-scm.com/downloads/win
This section could be called "Pull OpenVINO Model Server source" with info to install git and git clone command (admin permissions not required)

windows_install_build_dependencies.bat
```diff
@@ -270,52 +276,92 @@ IF /I EXIST %python_path% (
...
-        if !errorlevel! neq 0 exit /b !errorlevel!
-        :: setuptools<60.0 required for numpy1.23 on python311 to install
-        %python_path%\python.exe -m pip install "setuptools<60.0" "numpy==1.23" "Jinja2==3.1.5" "MarkupSafe==3.0.2"
+        call :UninstallPython
         if !errorlevel! neq 0 exit /b !errorlevel!
+        call :InstallPython
+        IF /I not EXIST %python_path%\python.exe (
+            echo "Python installation failed. Errorlevel: !errorlevel!"
+            echo "To fix the installation run %opt_install_dir%\%python_full_name%.exe in GUI and press repair."
+            echo "Please review the installation log: python/install.log"
```
dtrawins, line 285: logs should be checked before the attempt to repair manually?

windows_install_build_dependencies.bat
```diff
@@ -270,52 +276,92 @@ IF /I EXIST %python_path% (
...
+            echo "Rerun the windows_install_build_dependencies.bat, once the installation is fixed."
+            exit /b !errorlevel!
+        )
     ) ELSE (
         echo [INFO] ::::::::::::::::::::::: %python_path% already installed
     )
 ) ELSE (
```
dtrawins, line 292: why python is uninstalled when present? It looks like it will be reinstalled with every execution.
maybe just the version should be checked ?

rasapala: The uninstall help with clean installation. Its only done once, as if python.exe is present it will no longer run.

rasapala: The check is if the install_python.exe is present.

docs/windows_developer_guide.md
```diff
@@ -41,10 +34,12 @@ Follow instructions in the link below:
...
 
 ## Run Command Prompt
-Press Windows Start and run the cmd.exe terminal.
+Press Windows Start and run the cmd.exe terminal as Administrator.
 Run commands in this prompt is not stated otherwise.
 
-## Get the code
+## Pull OpenVINO Model Server source
+Install git from https://git-scm.com/downloads/win
+After that run commands in terminal:
```
mzegla, line 42:
```suggestion
> `Git` is required to complete this step. If you don't have it on your system, download it from https://git-scm.com/downloads/win and install before you continue.
Run below commands in terminal to clone model server repository:
```

windows_install_build_dependencies.bat
```diff
@@ -270,52 +276,92 @@ IF /I EXIST %python_path% (
...
+        echo Process "%python_full_name%.exe" finished
+        goto python_install_finished
+    ) else (
+        echo Process "%python_full_name%.exe" in progress ...
+        ping 1.1.1.1 -n 3 > nul
+    )
+)
+:python_install_finished
+exit /b 0
+:::::::::::::::::::::: Uninstall function end
```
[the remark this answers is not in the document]
rasapala: The steps will be executed if we place them anywhere in the begging. Lets keep the Python part together, specifically when we turn the echo off it does not print anything unnecessary to the user.

dtrawins (approved):

mzegla (approved):

PR #3030, opened 2025-02-05
Use GenAI text streamer implementation

mzegla:
Switching text streamer to GenAI implementation. For now we adapt a copy since it's not in GenAI interface now. 

This will likely be useful for streaming enablement in pipelines other than ContinuousBatchingPipeline.

### DIFF
--- a/src/llm/llmnoderesources.hpp
+++ b/src/llm/llmnoderesources.hpp
@@ -24,6 +24,7 @@
 
 #include <openvino/genai/continuous_batching_pipeline.hpp>
 #include <openvino/genai/scheduler_config.hpp>
+#include "openvino/genai/streamer_base.hpp"
 #include <openvino/openvino.hpp>
 
 #pragma warning(push)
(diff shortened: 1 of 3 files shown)


src/llm/llmnoderesources.hpp
```diff
@@ -45,62 +47,83 @@
...
-    std::string end() {
-        if (tokenCache.size() > 0) {
-            std::string text = tokenizer->decode(tokenCache);
-            std::string chunk = std::string{text.data() + printLen, text.size() - printLen};
-            SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Generated tokens: {}", tokenCache);
-            tokenCache.clear();
-            printLen = 0;
-            return chunk;
+        int64_t print_until = m_decoded_lengths[m_decoded_lengths.size() - delay_n_tokens];
+        if (print_until != -1 && print_until > static_cast<int64_t>(m_printed_len)) {
```
mzegla, line 90: GenAI implementation does not have this static cast. We have an error when building without it.

src/llm/llmnoderesources.hpp
```diff
@@ -45,62 +47,83 @@
...
-            // Does not clear the cache.
-            auto lastSpacePos = text.rfind(' ');
-            if (lastSpacePos == std::string::npos || lastSpacePos < printLen) {
-                return std::nullopt;
-            }
-            std::string chunk = std::string{text.data() + printLen, lastSpacePos - printLen + 1};
-            printLen = lastSpacePos + 1;
-            SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Generated tokens: {}", tokenCache);
-            return chunk;
+        constexpr char replacement[] = "\xef\xbf\xbd";  // MSVC with /utf-8 fails to compile <UNK> directly with newline in string literal error.
```
mzegla, line 83: In this comment, GenAI contains special character instead of `<UNK>`. I had to replace it, so our pipeline passes.

michalkulakowski (approved):

PR #3048, opened 2025-02-10
Enable disabled compiler warnings on windows

michalkulakowski:
### 🛠 Summary

CVS-162116

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/http_rest_api_handler.cpp
+++ b/src/http_rest_api_handler.cpp
@@ -144,9 +144,9 @@ Status HttpRestApiHandler::parseModelVersion(std::string& model_version_str, std
     if (!model_version_str.empty()) {
         try {
             model_version = std::stoll(model_version_str.c_str());
-        } catch (std::out_of_range const& ex) {
+        } catch (std::out_of_range const&) {
             return StatusCode::MODEL_VERSION_MISSING;
-        } catch (std::exception& e) {
+        } catch (std::exception&) {
             SPDLOG_DEBUG("Couldn't parse model version {}", model_version_str);
             return StatusCode::REST_COULD_NOT_PARSE_VERSION;
         }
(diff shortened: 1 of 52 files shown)


src/http_rest_api_handler.cpp
```diff
@@ -362,7 +362,7 @@ static std::set<std::string> getRequestedBinaryOutputsNames(::KFSRequest& grpc_r
     std::set<std::string> binaryOutputs;
     bool byDefaultBinaryOutpuRequested = false;
     for (auto& parameter : grpc_request.parameters()) {
-        if (parameter.second.parameter_choice_case(), inference::InferParameter::ParameterChoiceCase::kBoolParam) {
+        if (parameter.second.parameter_choice_case() == inference::InferParameter::ParameterChoiceCase::kBoolParam) {
```
dkalinowski, line 365: does the behavior change? we should probably add unit test for that

src/llm/apis/openai_completions.cpp
```diff
@@ -613,7 +613,7 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const std::vect
         writer.Int(index++);
         // logprobs: object/null; Log probability information for the choice. TODO
         writer.String("logprobs");
-        if (this->request.logprobschat || this->request.logprobs > 0) {
+        if (this->request.logprobschat || this->request.logprobs) {
```
dkalinowski, line 616: is logprobs signed or unsigned int? if its signed we change functionality here (might be <0 which now if will be true)

michalkulakowski: but logprobs is bool, right?

rasapala (approved):

dkalinowski (approved):

PR #3054, opened 2025-02-11
Mediapipe generic metrics demo

bstrzele:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``


demos/common/stream_client/stream_client.py
```diff
@@ -172,7 +173,7 @@ def display(self):
                 self.pq.put((sent_frame_id, received_frame, timestamp))
 
     def get_timestamp(self) -> int:
-        return int(cv2.getTickCount() / cv2.getTickFrequency() * 1e6)
+        return int(time.time() * 1e6)
```
dkalinowski, line 176: why this change? why do we still want to stamp it manually? should we rely on automatic stamping?
otherwise the metric is wrong - we take into account client->server network transfer time

demos/common/stream_client/stream_client.py
```diff
@@ -232,7 +233,7 @@ def start(self, *, ovms_address : str, input_name : str, model_name : str, datat
                     inputs=[grpcclient.InferInput(input_name, np_frame.shape, datatype.string())]
                     inputs[0].set_data_from_numpy(np_frame)
                     if streaming_api:
-                        triton_client.async_stream_infer(model_name=model_name, inputs=inputs, parameters={"OVMS_MP_TIMESTAMP":self.get_timestamp()})
+                        triton_client.async_stream_infer(model_name=model_name, inputs=inputs)
```
dkalinowski, line 236: why do we still have get_timestamp if we dont use it here?

demos/common/stream_client/stream_client.py
bstrzele: fixed that, now benchmarking is based on server side timestamps. Client side timestamps are now only used to order frames while using legacy(non-streaming) method.

dkalinowski (approved):

dtrawins (approved):

PR #3086, opened 2025-02-26
Refactor GenAI pipelines

mzegla:
### 🛠 Summary

- Renaming: LLMNodeResources -> GenAiServable
- Creating common flow for request processing in HttpLLMCalculator reused by all GenAI pipelines
- Implemented common interface for ContinuousBatchingPipeline
- Introduced initializer classes for flexible creation of servables
- New flow based on servable `properties` existing during whole servable lifespan and `execution context` created per request

### 🧪 Checklist

- [ ] Unit tests added. - adjusted existing tests
- [ ] The documentation updated. - no update required at this point
``


src/llm/continuous_batching/servable.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Node resources interface start
+
+ovms::Status ContinuousBatchingServable::initialize() {
+    auto properties = std::static_pointer_cast<ContinuousBatchingServableProperties>(this->properties);
+    if (properties->pipeline == nullptr) {
+        throw std::logic_error("Cannot initiate generation with uninitialized pipeline");
+    }
+    properties->llmExecutorWrapper = std::make_shared<LLMExecutorWrapper>(properties->pipeline);
```
dkalinowski, line 59: we have special initializers now - why not initialize it there?

src/llm/continuous_batching/servable_initializer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    ov::genai::SchedulerConfig config;
+    config.max_num_batched_tokens = nodeOptions.has_draft_max_num_batched_tokens() ? nodeOptions.draft_max_num_batched_tokens() : nodeOptions.max_num_batched_tokens();
+    config.cache_size = nodeOptions.has_draft_cache_size() ? nodeOptions.draft_cache_size() : nodeOptions.cache_size();
+    config.dynamic_split_fuse = nodeOptions.has_draft_dynamic_split_fuse() ? nodeOptions.draft_dynamic_split_fuse() : nodeOptions.dynamic_split_fuse();
+    config.max_num_seqs = nodeOptions.has_draft_max_num_seqs() ? nodeOptions.draft_max_num_seqs() : nodeOptions.max_num_seqs();
+    config.enable_prefix_caching = nodeOptions.enable_prefix_caching();
+    return config;
+}
+
+Status ContinuousBatchingServableInitializer::initializeExperimental(std::shared_ptr<GenAiServable>& servable, const mediapipe::LLMCalculatorOptions& nodeOptions, std::string graphPath) {
```
dkalinowski, line 64: this is actual implementation of initializer - `ContinuousBatchingServableInitializer`, why do we take base class `std::shared_ptr<GenAiServable>& servable` as parameter? we could already use actual derived class to avoid cast

src/llm/continuous_batching/servable_initializer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    auto continousBatchingPipelineConfig = nodeOptions.continuous_batching_pipeline_config();
+    auto mainPipelineConfig = continousBatchingPipelineConfig.main_pipeline_config();
```
dkalinowski, line 66: why do we need to have pairs: servable<->initializer? the initializer only exists because it needs to be used in servable - why cant it be part of servable? is it separate the code for clarity? if we had it inside servable class, we could be using the actual this->properties with known derived type

src/llm/continuous_batching/servable.cpp
```diff
@@ -0,0 +1,131 @@
...
+    return ovms::StatusCode::OK;
+}
+
+std::shared_ptr<GenAiServableExecutionContext> ContinuousBatchingServable::createExecutionContext() {
+    return std::make_shared<ContinuousBatchingServableExecutionContext>();
+}
+
+absl::Status ContinuousBatchingServable::scheduleExecution(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto properties = std::static_pointer_cast<ContinuousBatchingServableProperties>(this->properties);
```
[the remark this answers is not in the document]
mzegla: Good idea. Change applied.

src/llm/continuous_batching/servable.cpp
mzegla: Done

src/llm/continuous_batching/servable_initializer.cpp
mzegla: `initialize` is part of the interface and it's signature must be the same for all pipelines. 
I did it that way to have a unified path and clear convention of initialization for GenAI pipelines.

src/llm/continuous_batching/servable_initializer.cpp
mzegla: Mostly for code separation (less code per file, easier to find stuff) as well as logical separation (initializer is OVMS core thing and has dependency on Mediapipe, while servables are used only in the calculator and potentially could be initialized in a different way (for example if we extract it from the MP framework for some reason).

michalkulakowski (approved):

dkalinowski (approved):

dtrawins (approved):

PR #3102, opened 2025-03-04
Adopt GenAI Public TextStreamer in OVMS

sammysun0711:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

- Current OMVS copied Text streamer implementation copied from GenAI: https://github.com/openvinotoolkit/model_server/blob/822b127993a5c5a5420ceb8617b9a7533aabd6e4/src/llm/servable.hpp#L37-L109
- Recent GenAI make TextStreamer interface public: https://github.com/openvinotoolkit/openvino.genai/pull/1700

This PR aim to adopt GenAI public TextStreamer in OVMS to use updated openvino.genai stream generation implemetation.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/servable.cpp
+++ b/src/llm/servable.cpp
@@ -77,7 +77,7 @@ absl::Status GenAiServable::parseRequest(std::shared_ptr<GenAiServableExecutionC
             return false;
         };
 
-        executionContext->textStreamer = std::make_shared<ov::genai::TextCallbackStreamer>(getProperties()->tokenizer, callback);
+        executionContext->textStreamer = std::make_shared<ov::genai::TextStreamer>(getProperties()->tokenizer, callback);
     }
     return absl::OkStatus();
 }
@@ -134,7 +134,7 @@ absl::Status GenAiServable::preparePartialResponse(std::shared_ptr<GenAiServable
     // so such change should be done in GenAI first
     std::stringstream ss;
     for (const auto& token : generationOutput.generated_ids) {
-        executionContext->textStreamer->put(token);
+        executionContext->textStreamer->write(token);
         ss << executionContext->lastStreamerCallbackOutput;
     }
     std::string lastTextChunk = ss.str();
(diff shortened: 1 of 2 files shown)


src/llm/servable.cpp
```diff
@@ -134,7 +134,7 @@ absl::Status GenAiServable::preparePartialResponse(std::shared_ptr<GenAiServable
     // so such change should be done in GenAI first
     std::stringstream ss;
     for (const auto& token : generationOutput.generated_ids) {
-        executionContext->textStreamer->put(token);
+        executionContext->textStreamer->write(token);
         ss << executionContext->lastStreamerCallbackOutput;
     }
```
mzegla, line 139: According to https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/include/openvino/genai/text_streamer.hpp#L24 we can remove this loop and call `write` directly with `generationOutput.generated_ids`

sammysun0711: Thanks for your feedback. Thanks for your feedback. Currently OVMS use openvino.genai 2025.1.0.0.dev20250221
`StreamingStatus write(const std::vector<int64_t>& tokens) override;` first introduced in genai 0226 nightly build: https://github.com/openvinotoolkit/openvino.genai/pull/1807
Do you think I can upgrade genai package used in OVMS in this PR or better use separate PR.

mzegla: Update is needed also for other PRs, so I already prepared it:
https://github.com/openvinotoolkit/model_server/pull/3104
There's some issue on LLM tests, so can't merge it right away, but once it's resolved we can get back to applying changes here.

mzegla: @sammysun0711, OV and GenAI have been updated to 2025.03.04 nightly, so we can proceed with updates here

sammysun0711:
Open new PR to trigger CI: https://github.com/openvinotoolkit/model_server/pull/3107

PR #3106, opened 2025-03-05
store-ovms-windows-artifacts

pgladkows:
### 🛠 Summary

https://jira.devtools.intel.com/browse/CVS-162510 
Added post action to build_test_OnCommit.groovy to store windows artifacts (ovms.zip) as latest. 
The latest package will be available at the link: http://ov-share-05.sclab.intel.com/cv_bench_cache/OVMS_do_not_remove/ovms-windows-main-latest.zip

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/ci/build_test_OnCommit.groovy
+++ b/ci/build_test_OnCommit.groovy
@@ -185,7 +185,7 @@ pipeline {
             node("${agent_name_windows}") {
                 script {
                     if (windows_success == "True") { // env.BRANCH_NAME == "main" &&
-                        bat(returnStatus:true, script: "copy C:\\Jenkins\\workspace\\ovms_oncommit_main\\dist\\windows\\ovms.zip Z:\\ovms-windows-main-latest.zip")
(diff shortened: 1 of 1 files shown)


rasapala (approved):

ci/build_test_OnCommit.groovy
```diff
@@ -174,8 +176,19 @@ pipeline {
...
-            }           
+            }
           }
         }
     }
+    post {
+        always {
+            script {
+                if (env.BRANCH_NAME == "main" && windows_success == "True") {
+                    build job: "ovms/store_ovms_windows_artifacts"
```
dtrawins, line 187: it might not work because we have 2 workers. hardcoded agent will not work here. Why new job is triggered instead of new local script?

pgladkows: I changed it to copy to ov-share.

ci/build_test_OnCommit.groovy
```diff
@@ -174,8 +176,21 @@ pipeline {
                       }
                   }
               }
-            }           
+            }
           }
         }
     }
+    post {
+        always {
```
dtrawins, line 184: we could have when expression so this step would be initiated only for main branch

pgladkows: done

dkalinowski (approved):

PR #3109, opened 2025-03-06
Inferring GenAI servable type

mzegla:
### 🛠 Summary
 - By default we select pipeline type by checking models directory contents and device choice
 - We also give possibility to enforce pipeline type with limitation that it must still match models directory content (can't use text only pipeline on VLM model and vice-versa)
 - Added supportsSpeculativeDecoding method to servable interface and blocked SD in VLM pipeline
 - Added new tests for initialization

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/demos/common/export_models/export_model.py
+++ b/demos/common/export_models/export_model.py
@@ -38,7 +38,7 @@ parser = argparse.ArgumentParser(description='Export Hugging face models to OVMS
 subparsers = parser.add_subparsers(help='subcommand help', required=True, dest='task')
 parser_text = subparsers.add_parser('text_generation', help='export model for chat and completion endpoints')
 add_common_arguments(parser_text)
-parser_text.add_argument('--pipeline_type', default='CONTINUOUS_BATCHING', help='Type of the pipeline to be used. Can be either CONTINUOUS_BATCHING or VISUAL_LANGUAGE_MODEL.', dest='pipeline_type')
+parser_text.add_argument('--pipeline_type', default=None, help='Type of the pipeline to be used. Can be either TEXT_CB or VLM_CB', dest='pipeline_type')
 parser_text.add_argument('--kv_cache_precision', default=None, choices=["u8"], help='u8 or empty (model default). Reduced kv cache precision to u8 lowers the cache size consumption.', dest='kv_cache_precision')
 parser_text.add_argument('--enable_prefix_caching', action='store_true', help='This algorithm is used to cache the prompt tokens.', dest='enable_prefix_caching')
 parser_text.add_argument('--disable_dynamic_split_fuse', action='store_false', help='The maximum number of tokens that can be batched together.', dest='dynamic_split_fuse')
@@ -144,7 +144,8 @@ node: {
   }
   node_options: {
       [type.googleapis.com / mediapipe.LLMCalculatorOptions]: {
-          pipeline_type: {{pipeline_type}},
+          {%- if pipeline_type %}
+          pipeline_type: {{pipeline_type}},{% endif %}
           models_path: "{{model_path}}",
           plugin_config: '{ {% if kv_cache_precision %}"KV_CACHE_PRECISION": "{{kv_cache_precision}}"{% endif %}}',
           enable_prefix_caching: {% if not enable_prefix_caching %}false{% else %} true{% endif%},
--- a/src/llm/llm_calculator.proto
+++ b/src/llm/llm_calculator.proto
@@ -27,12 +27,22 @@ message LLMCalculatorOptions {
     }
 
     enum PipelineType {
+      // Single modality (text only), text generation based on LLMPipeline
+      TEXT = 0;
+      // Multimodal (text and image), text generation based on LLMPipeline
+      VLM = 1;
+      // Single modality (text only), text generation based on ContinuousBatchingPipeline
+      TEXT_CB = 2;
+      // Multimodal (text and image), text generation based on ContinuousBatchingPipeline
+      VLM_CB = 3;
+      // Default option. If selected, pipeline type will be inferred by the serving
+      UNDEFINED_PIPELINE_TYPE = 4;
+
+      // Compatibility with current configurations, remove below in the next PR
       // Continuous batching pipeline
-      CONTINUOUS_BATCHING = 0;
+      CONTINUOUS_BATCHING = 5;
       // Visual language model pipeline
-      VISUAL_LANGUAGE_MODEL = 1;
-      // LLM pipeline
-      LLM = 2; // To be changed...
+      VISUAL_LANGUAGE_MODEL = 6;
     }
 
     optional string models_path = 1;
@@ -73,7 +83,7 @@ message LLMCalculatorOptions {
 
     optional bool draft_dynamic_split_fuse = 17;
 
-    optional PipelineType pipeline_type = 18 [default = CONTINUOUS_BATCHING];
+    optional PipelineType pipeline_type = 18 [default = UNDEFINED_PIPELINE_TYPE];
     
     // START EXPERIMENTAL OPTIONS
     message PipelineConfig {
--- a/src/llm/servable.cpp
+++ b/src/llm/servable.cpp
@@ -87,8 +87,12 @@ absl::Status GenAiServable::prepareInputs(std::shared_ptr<GenAiServableExecution
         return absl::Status(absl::StatusCode::kInvalidArgument, "API handler is not initialized");
     }
 
-    std::string inputText;
+    // Base servable cannot process images
+    if (executionContext->apiHandler->getImages().size() > 0) {
+        return absl::InternalError("This servable supports only text input, but image_url has been provided");
+    }
 
+    std::string inputText;
     switch (executionContext->endpoint) {
     case Endpoint::CHAT_COMPLETIONS: {
         bool success;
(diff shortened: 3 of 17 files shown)


src/llm/llm_calculator.proto
```diff
@@ -27,12 +27,22 @@ message LLMCalculatorOptions {
...
+      TEXT_CB = 2;
+      // Multimodal (text and image), text generation based on ContinuousBatchingPipeline
+      VLM_CB = 3;
+      // Default option. If selected, pipeline type will be inferred by the serving
+      UNDEFINED_PIPELINE_TYPE = 4;
+
+      // Compatibility with current configurations, remove below in the next PR
       // Continuous batching pipeline
-      CONTINUOUS_BATCHING = 0;
+      CONTINUOUS_BATCHING = 5;
```
dtrawins, line 43: isn't it duplicated?

mzegla: It has to be for now
If I remove old values, current models in CI will not load properly 
We need to merge it with old values for compatibility, regenerate models and remove them in the next PR

src/llm/llm_calculator.proto
```diff
@@ -27,12 +27,22 @@ message LLMCalculatorOptions {
...
+      // Single modality (text only), text generation based on LLMPipeline
+      TEXT = 0;
+      // Multimodal (text and image), text generation based on LLMPipeline
+      VLM = 1;
+      // Single modality (text only), text generation based on ContinuousBatchingPipeline
+      TEXT_CB = 2;
+      // Multimodal (text and image), text generation based on ContinuousBatchingPipeline
+      VLM_CB = 3;
+      // Default option. If selected, pipeline type will be inferred by the serving
+      UNDEFINED_PIPELINE_TYPE = 4;
```
dkalinowski, line 39:
```suggestion
      INFERRED_PIPELINE_TYPE = 4;

```

dkalinowski: or deduced

src/llm/servable_initializer.cpp
```diff
@@ -117,36 +117,107 @@ void GenAiServableInitializer::loadTextProcessor(std::shared_ptr<GenAiServablePr
     }
 }
 
-Status GenAiServableInitializer::parseModelsPath(std::string modelsPath, std::string graphPath) {
+Status parseModelsPath(std::string& outPath, std::string modelsPath, std::string graphPath) {
```
dkalinowski, line 120: if its not class method, but local helper, use static 

the same below in another method

src/llm/servable_initializer.cpp
```diff
@@ -117,36 +117,107 @@ void GenAiServableInitializer::loadTextProcessor(std::shared_ptr<GenAiServablePr
...
+            SPDLOG_LOGGER_ERROR(modelmanager_logger, "LLM node options do not contain any recognized pipeline configuration.");
+            return StatusCode::INTERNAL_ERROR;
+        }
+
+        if (isVLM && (pipelineType == PipelineType::TEXT || pipelineType == PipelineType::TEXT_CB)) {
+            SPDLOG_LOGGER_ERROR(modelmanager_logger, "Models directory content indicates VLM pipeline, but pipeline type is set to non-VLM type.");
+            return StatusCode::INTERNAL_ERROR;
+        }
+
+        if (!isVLM && (pipelineType == PipelineType::VLM || pipelineType == PipelineType::VLM_CB)) {
```
dkalinowski, line 202:
```suggestion
        if (!isVLM && (pipelineType != PipelineType::TEXT && pipelineType != PipelineType::TEXT_CB)) {
```

isnt it safer to manually add it here everytime we need?

src/llm/visual_language_model/servable.cpp
```diff
@@ -41,24 +41,37 @@ std::shared_ptr<GenAiServableProperties> VisualLanguageModelServable::getPropert
...
-    if (chatHistory.size() == 0) {
-        return absl::Status(absl::StatusCode::kInvalidArgument, "Chat history is empty");
+    if (executionContext->endpoint == Endpoint::CHAT_COMPLETIONS) {
+        ov::genai::ChatHistory& chatHistory = vlmExecutionContext->apiHandler->getChatHistory();
+        ImageHistory& imageHistory = vlmExecutionContext->apiHandler->getImageHistory();
+        size_t imageIndex = 0;
+        for (auto& image : imageHistory) {
+            auto& [chatTurnIndex, imageTensor] = image;
+            std::string imageTag = "<ov_genai_image_" + std::to_string(imageIndex++) + ">\n";
+            if (chatHistory[chatTurnIndex].find("content") != chatHistory[chatTurnIndex].end()) {
```
dkalinowski, line 60: undefined behavior when chatHistory does not contain chatTurnIndex

src/llm/visual_language_model/servable.cpp
```diff
@@ -41,24 +41,37 @@ std::shared_ptr<GenAiServableProperties> VisualLanguageModelServable::getPropert
...
     }
-
-    vlmExecutionContext->inputImages = vlmExecutionContext->apiHandler->getImages();
-
-    auto& chatHistory = vlmExecutionContext->apiHandler->getChatHistory();
-    if (chatHistory.size() == 0) {
-        return absl::Status(absl::StatusCode::kInvalidArgument, "Chat history is empty");
+    if (executionContext->endpoint == Endpoint::CHAT_COMPLETIONS) {
+        ov::genai::ChatHistory& chatHistory = vlmExecutionContext->apiHandler->getChatHistory();
+        ImageHistory& imageHistory = vlmExecutionContext->apiHandler->getImageHistory();
```
dkalinowski, line 55: chatHistory is modified here - I understand we need to return non-const ref from getChatHistory()

however, I dont see image history being modified. why couldnt it still be const ref?

src/test/llm/visual_language_model/complete_flow_test.cpp
```diff
@@ -153,7 +163,44 @@ TEST_F(VLMServableFlowTest, unaryBasic) {
     EXPECT_STREQ(parsedResponse["object"].GetString(), "chat.completion");
 }
 
-TEST_F(VLMServableFlowTest, streamBasic) {
+// Only image input is accepted, but expected output can't be predicted
```
dkalinowski, line 166: do we have tests for multiple images and ensuring that order is correct?

demos/common/export_models/export_model.py
```diff
@@ -38,7 +38,7 @@ def add_common_arguments(parser):
 subparsers = parser.add_subparsers(help='subcommand help', required=True, dest='task')
 parser_text = subparsers.add_parser('text_generation', help='export model for chat and completion endpoints')
 add_common_arguments(parser_text)
-parser_text.add_argument('--pipeline_type', default='CONTINUOUS_BATCHING', help='Type of the pipeline to be used. Can be either CONTINUOUS_BATCHING or VISUAL_LANGUAGE_MODEL.', dest='pipeline_type')
+parser_text.add_argument('--pipeline_type', default=None, help='Type of the pipeline to be used. Can be either TEXT_CB or VLM_CB', dest='pipeline_type')
```
dtrawins, line 41:
```suggestion
parser_text.add_argument('--pipeline_type', default=None, help='Type of the pipeline to be used. Can be either TEXT_CB or VLM_CB. When undefined, it will be autodetected', dest='pipeline_type')
```

dtrawins (approved):

src/llm/servable_initializer.cpp
mzegla: Could you elaborate why? Current condition is more readable in my opinion as it's limited to VLM "terminology" (easy to associate isVLM with comparing against VLM pipeline types). Not sure I see the benefit of switching here.

src/llm/visual_language_model/servable.cpp
mzegla: Sure, I can change it back to const

src/llm/visual_language_model/servable.cpp
mzegla: It's not external input. `chatTurnIndex` is prepared along with `chatHistory` in API handler. 
We can check it, but it feels like we don't trust our own implementation.

src/test/llm/visual_language_model/complete_flow_test.cpp
mzegla: see changes made in http_openai_handler_test

src/llm/servable_initializer.cpp
mzegla: This method is also used outside of main servable initializer (namely in derived classes that include this header).
Below method determinePipelineType has dedicated tests specifically checking it's correctness. They are not used only in the scope of basic servable initializer.

src/llm/servable_initializer.cpp
dkalinowski: Readable yes, but in case we add new kind of pipeline, (and we forget to include it here) the if will return true, but it shouldnt right?

src/llm/servable.cpp
```diff
@@ -87,8 +87,12 @@ absl::Status GenAiServable::prepareInputs(std::shared_ptr<GenAiServableExecution
         return absl::Status(absl::StatusCode::kInvalidArgument, "API handler is not initialized");
     }
 
-    std::string inputText;
+    // Base servable cannot process images
+    if (executionContext->apiHandler->getImageHistory().size() > 0) {
+        return absl::InternalError("This servable supports only text input, but image_url has been provided");
```
dkalinowski, line 92: how do we know that image_url was provided? we only see that no images has been sent

dkalinowski (approved):

PR #3120, opened 2025-03-13
Mkulakow/support llm on npu

michalkulakowski:
### 🛠 Summary

CVS-163438
CVS-163439

### 🧪 Checklist

- [ ] Unit tests added.
``


src/llm/legacy/legacy_executor.hpp
```diff
@@ -0,0 +1,136 @@
...
+#pragma GCC diagnostic pop
+
+class LegacyExecutorWrapper {
+    LegacyExecutor legacyExecutor;
+    std::thread legacyExecutorThread;
+    std::atomic<bool> finishExecutorThread = false;
+
+    static void run(LegacyExecutor* legacyExecutor, std::atomic<bool>* receivedEndSignal) {
+        const uint8_t printMetricsEveryNumberOfSteps = 10;
+        uint8_t stepCounter = 0;
```
mzegla, line 100: That's not valid. Let's remove those variables and all of it's usage.

src/llm/legacy/legacy_executor.hpp
```diff
@@ -0,0 +1,136 @@
...
+            try {
+                if (stepCounter == printMetricsEveryNumberOfSteps) {
+                    // legacyExecutor->printMetrics();
+                    stepCounter = 0;
+                }
+                if (legacyExecutor->hasRequests()) {
+                    stepCounter++;
+                    legacyExecutor->step();
+                } else {
+                    SPDLOG_LOGGER_INFO(llm_executor_logger, "All requests: {}; Scheduled requests: {};", 0, 0);
```
mzegla, line 111: Scheduled requests will always be 1 in non CB flow. Let's print "Awaiting requests: <size_of_the_requests_queue>" instead.

src/llm/legacy/servable.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "../text_processor.hpp"
+#include "servable.hpp"
+
+namespace ovms {
+
+static std::string wrapTextInServerSideEventMessage(const std::string& text) {
+    std::stringstream ss;
+    ss << "data: " << text << "\n\n";
+    return ss.str();
+}
```
mzegla, line 45: Don't we already have such methods in basic servable? Looks pretty common to me.

src/llm/legacy/servable.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+absl::Status LegacyServable::readCompleteExecutionResults(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto legacyExecutionContext = std::static_pointer_cast<LegacyServableExecutionContext>(executionContext);
+    if (legacyExecutionContext->payload.client->isDisconnected()) {
+        return absl::CancelledError();
+    }
+    legacyExecutionContext->finished.wait();
+    // std::lock_guard<std::mutex> guard(pipelineMutex);
+    // legacyExecutionContext->results = properties->pipeline->generate(legacyExecutionContext->inputIds, legacyExecutionContext->apiHandler->createGenerationConfig());
```
mzegla, line 72: remove?

src/llm/legacy/servable.cpp
```diff
@@ -0,0 +1,158 @@
...
+    if (!status.ok()) {
+        SPDLOG_LOGGER_ERROR(llm_calculator_logger, "Failed to parse request: {}", status.message());
+        return status;
+    }
+
+    if (executionContext->apiHandler->isStream()) {
+        executionContext->lastStreamerCallbackOutput = "";  // initialize with empty string
+        auto callback = [& lastStreamerCallbackOutput = executionContext->lastStreamerCallbackOutput](std::string text) {
+            SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Streamer callback executed with text: [{}]", text);
+            lastStreamerCallbackOutput += text;
```
mzegla, line 92: Could you check if the same approach works for CB pipelines? 
In both cases we clear `lastStreamerCallbackOutput` after we read from it, so in token-by-token flow of CB pipeline it should not make a difference whever we assign or append. If it works, please add comments so it is clear why we do it that way and why we can do it.

src/llm/legacy/servable.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+absl::Status LegacyServable::preparePartialResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto legacyExecutionContext = std::static_pointer_cast<LegacyServableExecutionContext>(executionContext);
+    if (legacyExecutionContext->payload.client->isDisconnected()) {
+        return absl::CancelledError();
+    }
+    std::stringstream ss;
+    ss << executionContext->lastStreamerCallbackOutput;
+    std::string lastTextChunk = ss.str();
+    executionContext->lastStreamerCallbackOutput = "";
```
mzegla, line 119: executionContext->lastStreamerCallbackOutput is modified in another thread so I suppose we need some access protection

src/llm/legacy/servable.cpp
```diff
@@ -0,0 +1,158 @@
...
+    return absl::OkStatus();
+}
+
+absl::Status LegacyServable::readPartialExecutionResults(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto legacyExecutionContext = std::static_pointer_cast<LegacyServableExecutionContext>(executionContext);
+    if (legacyExecutionContext->payload.client->isDisconnected()) {
+        return absl::CancelledError();
+    }
+    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Relevant properties read");
+    return absl::OkStatus();
```
mzegla, line 152: Do we need all this logic? Looks like we can simply return OkStatus();

src/llm/legacy/servable.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+bool LegacyServable::supportsSpeculativeDecoding() const {
+    return true;
```
mzegla, line 156: Not sure if we should set it to true before testing proves it actually works as expected.

src/llm/legacy/servable.hpp
```diff
@@ -0,0 +1,56 @@
...
+    // Interface methods
+    std::shared_ptr<GenAiServableExecutionContext> createExecutionContext() override;
+    std::shared_ptr<GenAiServableProperties> getProperties() override;
+    absl::Status scheduleExecution(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
+    absl::Status readCompleteExecutionResults(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
+    absl::Status readPartialExecutionResults(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
+    absl::Status prepareCompleteResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
+    absl::Status preparePartialResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
+    bool supportsSpeculativeDecoding() const override;
+    absl::Status parseRequest(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
```
mzegla, line 54: It's a minor, but can we have a method order matching base class? 
Also the implementation in cpp file should match that order

src/llm/legacy/servable_initializer.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+ov::genai::SchedulerConfig LegacyServableInitializer::prepareDraftPipelineSchedulerConfig(const mediapipe::LLMCalculatorOptions& nodeOptions) {
+    ov::genai::SchedulerConfig config;
+    config.max_num_batched_tokens = nodeOptions.has_draft_max_num_batched_tokens() ? nodeOptions.draft_max_num_batched_tokens() : nodeOptions.max_num_batched_tokens();
+    config.cache_size = nodeOptions.has_draft_cache_size() ? nodeOptions.draft_cache_size() : nodeOptions.cache_size();
+    config.dynamic_split_fuse = nodeOptions.has_draft_dynamic_split_fuse() ? nodeOptions.draft_dynamic_split_fuse() : nodeOptions.dynamic_split_fuse();
+    config.max_num_seqs = nodeOptions.has_draft_max_num_seqs() ? nodeOptions.draft_max_num_seqs() : nodeOptions.max_num_seqs();
+    config.enable_prefix_caching = nodeOptions.enable_prefix_caching();
+    return config;
+}
```
mzegla, line 60: This is required if servable supports speculative decoding. Perhaps we could set it to false and comment this out for now until tests are enabled?

src/llm/legacy/servable_initializer.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+Status LegacyServableInitializer::initializeExperimental(std::shared_ptr<GenAiServable>& servable, const mediapipe::LLMCalculatorOptions& nodeOptions, std::string graphPath) {
+    return StatusCode::OK;
+}
```
mzegla, line 64: can probably be removed

src/llm/servable_initializer.cpp
```diff
@@ -234,11 +236,13 @@ Status initializeGenAiServable(std::shared_ptr<GenAiServable>& servable, const :
             ContinuousBatchingServableInitializer cbServableInitializer;
             servable = std::make_shared<VisualLanguageModelServable>();
             status = cbServableInitializer.initialize(servable, nodeOptions, graphPath);
+        } else if (nodeOptions.pipeline_type() == mediapipe::LLMCalculatorOptions::TEXT) {
```
mzegla, line 239: wrong condition

```suggestion
        } else if (pipelineType == PipelineType::TEXT) {
```

src/test/llmnode_test.cpp
```diff
@@ -666,9 +666,8 @@ TEST_F(LLMFlowHttpTest, unaryChatCompletionsJson) {
     std::string requestBody = R"(
         {
             "model": "llmDummyKFS",
-            "stream": false,
+            "stream": true,
```
mzegla, line 669: Why is it changed in unary test?

src/test/llm/graphspeculative.pbtxt
```diff
@@ -31,6 +31,7 @@ node {
           models_path: "/ovms/src/test/llm_testing/facebook/opt-125m"
           cache_size: 1
           draft_models_path: "/ovms/src/test/llm_testing/facebook/opt-125m"
+          pipeline_type: TEXT
```
mzegla, line 34: remove before merge

src/test/llm/graphkfspass.pbtxt
```diff
@@ -30,6 +30,8 @@ node {
         [type.googleapis.com/mediapipe.LLMCalculatorOptions]: {
           models_path: "/ovms/src/test/llm_testing/facebook/opt-125m"
           cache_size: 1
+          pipeline_type: TEXT
```
mzegla, line 33: remove before merge

src/llm/language_model/legacy/servable.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+absl::Status LegacyServable::preparePartialResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto legacyExecutionContext = std::static_pointer_cast<LegacyServableExecutionContext>(executionContext);
+    if (legacyExecutionContext->payload.client->isDisconnected()) {
+        return absl::CancelledError();
+    }
+    std::stringstream ss;
+    ss << executionContext->lastStreamerCallbackOutput;
+    std::string lastTextChunk = ss.str();
+    legacyExecutionContext->mutex.lock();
```
mzegla, line 123: I think it would be safer to move this lock above 
`ss << executionContext->lastStreamerCallbackOutput;`
also, what do you think about closing this logic in a scope and using scoped_lock instead of lock() and unlock() calls?

src/llm/llm_calculator.proto
```diff
@@ -43,6 +43,7 @@ message LLMCalculatorOptions {
       CONTINUOUS_BATCHING = 5;
       // Visual language model pipeline
       VISUAL_LANGUAGE_MODEL = 6;
+      // Legacy pipeline
```
mzegla, line 46: ?

src/test/llmnode_test.cpp
```diff
@@ -668,7 +668,6 @@ TEST_F(LLMFlowHttpTest, unaryChatCompletionsJson) {
             "model": "llmDummyKFS",
             "stream": false,
             "seed" : 1,
-            "best_of" : 16,
```
mzegla, line 671: why is it removed?

mzegla (approved):

mzegla (approved):

src/llm/apis/openai_completions.cpp
```diff
@@ -782,6 +782,93 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const std::vect
     return buffer.GetString();
 }
 
+std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const ov::genai::EncodedResults& results) {  // TODO separate common part with function implemented above
```
dkalinowski, line 785: Is it really a thing, that we have code duplication because non-continuous batching pipelines return different ov::genai objects as result?

If so, isnt there an API that could return us the same type as for continuous batching?
Is it due to the fact that it lacks logprobs? Only because of that?

It will make our code very hard to maintain, since all changes need to be made in multiple places

src/llm/language_model/legacy/legacy_executor.cpp
```diff
@@ -0,0 +1,87 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include "legacy_executor.hpp"
+#include "servable.hpp"
+
+namespace ovms {
+LegacyExecutor::LegacyExecutor(std::shared_ptr<ov::genai::LLMPipeline> pipe) {
```
dkalinowski, line 21: What is LegacyExecutor? Is it single threaded text-generation for non-continuous batching non-visual pipelines? The name states that it is something old we always had and we will be getting rid of it, which is wrong... or i dont understand something

src/llm/language_model/legacy/legacy_executor.cpp
```diff
@@ -0,0 +1,87 @@
...
+    this->pipe = std::move(pipe);
+}
+
+bool LegacyExecutor::hasRequests() {
+    return (requests.size() > 0);
```
dkalinowski, line 26:
```suggestion
    return requests.size() > 0;
```

src/llm/language_model/legacy/legacy_executor.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+bool LegacyExecutor::requestsQueueSize() {
+    return requests.size();
+}
+
+void LegacyExecutor::processRequest() {
+    OVMS_PROFILE_FUNCTION();
+    requests.front()->results = pipe->generate(requests.front()->inputIds, requests.front()->apiHandler->createGenerationConfig(), requests.front()->textStreamer);
```
dkalinowski, line 35: is it really good API in terms of of LegacyExecutor class user? It needs to remember, that calling processRequest() when internal state is that requests is empty will cause segfault. It is not even noted anywhere in the code.

At minimum I'd recomment adding comment here and also in the legacy executor header, so we never call this function on empty requests container

Consider changing the API so that processRequest returns at error when internal state is abormal at the moment of calling it?

@mzegla @atobiszei

src/llm/language_model/legacy/servable.cpp
```diff
@@ -0,0 +1,152 @@
...
+}
+
+bool LegacyServable::supportsSpeculativeDecoding() const {
+    return false;
+}
+
+absl::Status LegacyServable::parseRequest(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto legacyExecutionContext = std::static_pointer_cast<LegacyServableExecutionContext>(executionContext);
+    if (legacyExecutionContext->payload.client->isDisconnected()) {
+        return absl::CancelledError();
```
dkalinowski, line 57: why are we using abseil status when we are in the context of ovms? we have ovms::Status for that
if it because really annoying or just our ovms::Status lacks functionalities to be convertible to abseil in the context of llm calculator, maybe this is finally a time to add the missing bits

src/llm/language_model/legacy/legacy_executor.cpp
mzegla: I suppose it's only called inside `if (legacyExecutor->hasRequests())` block, but I guess I won't hurt us if we add such check here and return internal error in case it's called without requests.

src/llm/language_model/legacy/servable.cpp
```diff
@@ -0,0 +1,152 @@
...
+}
+
+absl::Status LegacyServable::preparePartialResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto legacyExecutionContext = std::static_pointer_cast<LegacyServableExecutionContext>(executionContext);
+    if (legacyExecutionContext->payload.client->isDisconnected()) {
+        return absl::CancelledError();
+    }
+    std::string lastTextChunk;
+    {
+        std::scoped_lock lock(legacyExecutionContext->mutex);
```
dkalinowski, line 122: here you use scoped lock which means that multiple threads can lock it at the same time - is it desired behavior? what do we use lock for? what is the data race we try to avoid?

do we have a test for such cases?

dkalinowski (approved):

PR #3130, opened 2025-03-18
Mkulakow/support vlm on npu

michalkulakowski:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -874,6 +874,89 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const ov::genai
     return buffer.GetString();
 }
 
+std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const ov::genai::VLMDecodedResults& results) {  // TODO separate common part with function implemented above
+    OVMS_PROFILE_FUNCTION();
+    StringBuffer buffer;
+    Writer<StringBuffer> writer(buffer);
+
+    writer.StartObject();  // {
+
+    // choices: array of size N, where N is related to n request parameter
+    writer.String("choices");
+    writer.StartArray();  // [
+    int index = 0;
+    usage.completionTokens = 0;
+    for (int i = 0; i < results.texts.size(); i++) {
+        const std::string& texts = results.texts[i];
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Generated tokens: {}", tokens);
+        writer.StartObject();  // {
+        writer.String("finish_reason");
+        writer.String("stop");
+        // index: integer; Choice index, only n=1 supported anyway
+        writer.String("index");
+        writer.Int(index++);
+        // logprobs: object/null; Log probability information for the choice. TODO
+        // message: object
+        if (endpoint == Endpoint::CHAT_COMPLETIONS) {
+            writer.String("message");
+            writer.StartObject();  // {
+            // content: string; Actual content of the text produced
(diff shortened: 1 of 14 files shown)


src/llm/visual_language_model/legacy/servable.cpp
```diff
@@ -0,0 +1,195 @@
...
+    std::string lastTextChunk;
+    auto generationStatus = legacyExecutionContext->finished.wait_for(std::chrono::nanoseconds::zero());
+    {
+        std::unique_lock lock(legacyExecutionContext->mutex);
+        while (executionContext->lastStreamerCallbackOutput.size() == 0 && generationStatus != std::future_status::ready) {
+            SPDLOG_LOGGER_TRACE(llm_executor_logger, "Waiting for partial data...");
+            legacyExecutionContext->executionInProgress.wait(lock);
+            generationStatus = legacyExecutionContext->finished.wait_for(std::chrono::nanoseconds::zero());
+        }
+        lastTextChunk = executionContext->lastStreamerCallbackOutput;
```
mzegla, line 132: We can properly handle usage parameter here if user requests it. You can add below:
```
        if (!lastTextChunk.empty()) {
            auto tokensTensor = properties->tokenizer.encode(lastTextChunk, ov::genai::add_special_tokens(false)).input_ids;
            auto numTokens = tokensTensor.get_size();
            executionContext->apiHandler->incrementProcessedTokens(numTokens);
```

src/llm/visual_language_model/legacy/servable_initializer.cpp
```diff
@@ -0,0 +1,92 @@
...
+        return StatusCode::LLM_NODE_RESOURCE_STATE_INITIALIZATION_FAILED;
+    }
+
+    status = JsonParser::parsePluginConfig(nodeOptions.plugin_config(), properties->pluginConfig);
+    if (!status.ok()) {
+        SPDLOG_ERROR("Error during llm node plugin_config option parsing to JSON: {}", nodeOptions.plugin_config());
+        return status;
+    }
+
+    properties->tokenizerPluginConfig = {{"PERFORMANCE_HINT", "THROUGHPUT"}};
```
mzegla, line 73: Looks like this is unused

mzegla (commented):
Majority of the code introduced here is a duplication of LM servable logic. 
I think we can merge it to enable testing as soon as possible, but we definitely need to plan refactor and unification of "legacy" servables code

src/llm/apis/openai_completions.cpp
```diff
@@ -874,6 +874,89 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const ov::genai
     return buffer.GetString();
 }
 
+std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const ov::genai::VLMDecodedResults& results) {  // TODO separate common part with function implemented above
```
dkalinowski, line 877: I have already seen such code in last PR 🤔 
each combination of VLM/TEXT x Legacy/CB has its own type we must deserialize from...?

mzegla: I think the best solution would be to make changes in GenAI, so we can handle this without such logic (we could also have some additional data added to the output - like logprobs and finish reason that are missing now). But this will not happen in 2025.1

mzegla (approved):
Please add task to Jira for the next development phase on this servable

dkalinowski (approved):

PR #3133, opened 2025-03-20
Enable LLM tests for stateful pipeline

mzegla:
Changes:
- add stateful servable to LLM tests suite and adjust tests wherever needed
- rename TEXT/TEXT_CB to LM/LM_CB in pipeline type and removed unused types
- fixed usage computation for streaming in LM pipeline

### DIFF
--- a/src/llm/language_model/legacy/legacy_executor.cpp
+++ b/src/llm/language_model/legacy/legacy_executor.cpp
@@ -33,7 +33,9 @@ bool LegacyExecutor::requestsQueueSize() {
 void LegacyExecutor::processRequest() {
     OVMS_PROFILE_FUNCTION();
     SPDLOG_LOGGER_TRACE(llm_executor_logger, "Generation started");
+    pipe->start_chat();
     requests.front()->results = pipe->generate(requests.front()->inputIds, requests.front()->apiHandler->createGenerationConfig(), requests.front()->textStreamer);
+    pipe->finish_chat();
     SPDLOG_LOGGER_TRACE(llm_executor_logger, "Generation ended");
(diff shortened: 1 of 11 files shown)


src/llm/language_model/legacy/legacy_executor.cpp
```diff
@@ -47,7 +49,7 @@ void LegacyExecutor::waitForRequests(std::atomic<bool>* receivedEndSignal) {
 }
 
 void LegacyExecutor::addRequest(std::shared_ptr<LegacyServableExecutionContext> request) {
-    std::lock_guard<std::mutex> guard(queueMutex);
+    std::unique_lock<std::mutex> lock(queueMutex);
```
dkalinowski, line 52: why using unique lock? lock guard is sufficient in this case since we never need to unlock manually

src/test/llm/llmnode_test.cpp
```diff
@@ -425,7 +427,8 @@ TEST_P(LLMFlowHttpTestParameterized, streamCompletionsEchoWithCompletion) {
 TEST_P(LLMFlowHttpTestParameterized, unaryCompletionsJsonEchoOnly) {
     auto params = GetParam();
     // TODO: In the next step we should break this suite into smaller ones, use proper configuration instead of skipping
-    if (params.modelName.find("vlm") != std::string::npos) {
+    if (params.modelName.find("vlm") != std::string::npos || params.modelName.find("legacy") != std::string::npos) {
```
dkalinowski, line 430: when readable skip defines? 😺

michalkulakowski (approved):

dkalinowski (approved):

PR #3247, opened 2025-04-17
Add support for image provided as URL

michalkulakowski:
### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -143,7 +143,7 @@ static size_t appendChunkCallback(void* downloadedChunk, size_t size, size_t nme
         status = setopt;      \
     }
 
-static absl::Status downloadImage(const char* url, std::string& image, const long& sizeLimit) {
+static absl::Status downloadImage(const char* url, std::string& image, const int64& sizeLimit) {
     CURL* curl_handle = curl_easy_init();
     auto status = curl_easy_setopt(curl_handle, CURLOPT_URL, url);
     CURL_SETOPT(curl_easy_setopt(curl_handle, CURLOPT_WRITEFUNCTION, appendChunkCallback))
@@ -234,7 +234,7 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
                                 }
                             } else if (std::regex_match(url.c_str(), std::regex("^(https?:\\/\\/)?([\\da-z\\.-]+)\\.([a-z\\.]{2,6})([\\/\\w \\.-]*)*\\/?$"))) {
                                 curl_global_init(CURL_GLOBAL_ALL);
-                                long sizeLimit = 20000000;  // restrict single image size to 20MB
+                                int64 sizeLimit = 20000000;  // restrict single image size to 20MB
                                 auto status = downloadImage(url.c_str(), decoded, sizeLimit);
                                 if (status != absl::OkStatus()) {
                                     return status;
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -39,6 +39,9 @@
 #include "absl/strings/str_cat.h"
 #pragma warning(pop)
 
+#include <curl/curl.h>
+#include <regex>
+
 using namespace rapidjson;
 
 namespace ovms {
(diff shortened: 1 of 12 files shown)


src/llm/apis/openai_completions.cpp
```diff
@@ -128,6 +131,57 @@ ov::Tensor load_image_stbi(const std::string& imageBytes) {
         SharedImageAllocator{image, desiredChannels, y, x});
 }
 
+static size_t writeMemoryCallback(void* contents, size_t size, size_t nmemb,
+    void* userp) {
+    size_t realsize = size * nmemb;
+    auto& mem = *static_cast<std::string*>(userp);
+    mem.append(static_cast<char*>(contents), realsize);
+    return realsize;
+}
```
mzegla, line 140: Is this function signature imposed on us? I wonder if we could for example rename variables etc.

src/llm/apis/openai_completions.cpp
```diff
@@ -128,6 +131,57 @@ ov::Tensor load_image_stbi(const std::string& imageBytes) {
         SharedImageAllocator{image, desiredChannels, y, x});
 }
 
+static size_t writeMemoryCallback(void* contents, size_t size, size_t nmemb,
+    void* userp) {
+    size_t realsize = size * nmemb;
```
mzegla, line 136: There's a potential overflow here, but since it's handled by third party I'm not sure we do anything about it here (like return 0 if overflow?), maybe curl handles it somehow before passing the values to the callback.

src/llm/apis/openai_completions.cpp
```diff
@@ -128,6 +131,57 @@ ov::Tensor load_image_stbi(const std::string& imageBytes) {
...
+    }
+    if (status == CURLE_OK) {
+        status = curl_easy_setopt(curl_handle, CURLOPT_HTTPPROXYTUNNEL, 1L);
+    }
+    if (status == CURLE_OK) {
+        status = curl_easy_setopt(curl_handle, CURLOPT_HTTPPROXYTUNNEL, 1L);
+    }
+    if (status == CURLE_OK) {
+        status = curl_easy_setopt(curl_handle, CURLOPT_MAXFILESIZE, sizeLimit);
+    }
```
mzegla, line 168: I wonder if we could wrap:
```
    if (status == CURLE_OK) {
        status = curl_easy_setopt(curl_handle, curl_opt, opt_value);
    }
```
in a macro or figure out something else to reduce code size

src/llm/apis/openai_completions.cpp
```diff
@@ -128,6 +131,57 @@ ov::Tensor load_image_stbi(const std::string& imageBytes) {
...
+    if (status == CURLE_OK) {
+        status = curl_easy_perform(curl_handle);
+    }
+
+    if (status != CURLE_OK) {
+        std::stringstream ss;
+        ss << "Downloading image failed: " << curl_easy_strerror(status);
+        SPDLOG_LOGGER_ERROR(llm_calculator_logger, ss.str());
+        return absl::InvalidArgumentError(ss.str());
```
mzegla, line 177: You return the same message that we put in the logs. I don't know the contents of such errors, but wouldn't they reveal to much if we send them back to the client?

src/llm/apis/openai_completions.cpp
```diff
@@ -184,16 +238,32 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
...
+                                return absl::InvalidArgumentError("Url should contain base64 encoded string followed by \"base64,\" prefix or valid URL");
+                            }
+                            try {
+                                ov::Tensor tensor = load_image_stbi(decoded);
+                                request.imageHistory.push_back({i, tensor});
+                            } catch (std::runtime_error& e) {
+                                std::stringstream ss;
+                                ss << "Image parsing failed: " << e.what();
+                                SPDLOG_LOGGER_ERROR(llm_calculator_logger, ss.str());
+                                return absl::InvalidArgumentError(ss.str());
```
mzegla, line 265: Same as above, don't we reveal too much?

mzegla (approved):

src/llm/apis/openai_completions.cpp
```diff
@@ -128,6 +131,45 @@ ov::Tensor load_image_stbi(const std::string& imageBytes) {
...
+    CURL_SETOPT(curl_easy_setopt(curl_handle, CURLOPT_WRITEDATA, &image))
+    CURL_SETOPT(curl_easy_setopt(curl_handle, CURLOPT_SSL_OPTIONS, CURLSSLOPT_NATIVE_CA))
+    CURL_SETOPT(curl_easy_setopt(curl_handle, CURLOPT_TIMEOUT, 20L))
+    CURL_SETOPT(curl_easy_setopt(curl_handle, CURLOPT_FOLLOWLOCATION, 1L))
+    CURL_SETOPT(curl_easy_setopt(curl_handle, CURLOPT_HTTPPROXYTUNNEL, 1L))
+    CURL_SETOPT(curl_easy_setopt(curl_handle, CURLOPT_MAXFILESIZE, sizeLimit))
+
+    if (status == CURLE_OK) {
+        status = curl_easy_perform(curl_handle);
+    }
```
mzegla, line 159: I would consider adding some debug logging. If not on per-opt basis in the macro, then at least something like:
"Image download failed. Could not set all required cURL opts."

src/llm/apis/openai_completions.cpp
```diff
@@ -128,6 +131,45 @@ ov::Tensor load_image_stbi(const std::string& imageBytes) {
...
 }
 
+static size_t appendChunkCallback(void* downloadedChunk, size_t size, size_t nmemb,
+    void* image) {
+    size_t realsize = size * nmemb;
+    auto& mem = *static_cast<std::string*>(image);
+    mem.append(static_cast<char*>(downloadedChunk), realsize);
+    return realsize;
+}
+#define CURL_SETOPT(setopt) \
```
mzegla, line 141: What do you think? Adding `_func_call` could be the indicator what we are passing here
```suggestion
#define CURL_SETOPT(setopt_func_call) \
```

src/llm/apis/openai_completions.cpp
```diff
@@ -128,6 +131,45 @@ ov::Tensor load_image_stbi(const std::string& imageBytes) {
...
+
+    if (status == CURLE_OK) {
+        status = curl_easy_perform(curl_handle);
+    }
+
+    if (status != CURLE_OK) {
+        std::stringstream ss;
+        ss << "Downloading image failed: " << curl_easy_strerror(status);
+        SPDLOG_LOGGER_ERROR(llm_calculator_logger, ss.str());
+        return absl::InvalidArgumentError("Image downloading fialed");
```
mzegla, line 165:
```suggestion
        return absl::InvalidArgumentError("Image download failed");
```

rasapala (approved):

src/llm/apis/openai_completions.cpp
```diff
@@ -184,16 +226,32 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
...
-                            if (!absl::Base64Unescape(std::string_view(url.data() + offset, url.size() - offset), &decoded)) {
-                                return absl::InvalidArgumentError("Invalid base64 string in request");
+                            if (pos != std::string::npos) {
+                                size_t offset = pos + pattern.length();
+                                if (!absl::Base64Unescape(std::string_view(url.data() + offset, url.size() - offset), &decoded)) {
+                                    return absl::InvalidArgumentError("Invalid base64 string in request");
+                                }
+                            } else if (std::regex_match(url.c_str(), std::regex("^(https?:\\/\\/)?([\\da-z\\.-]+)\\.([a-z\\.]{2,6})([\\/\\w \\.-]*)*\\/?$"))) {
+                                curl_global_init(CURL_GLOBAL_ALL);
+                                double sizeLimit = 20000000L;
```
dtrawins, line 237:
```suggestion
                                double sizeLimit = 20000000L;  // restrict single image size to 20MB
```

src/test/http_openai_handler_test.cpp
```diff
@@ -275,6 +275,41 @@ TEST_F(HttpOpenAIHandlerParsingTest, ParsingMessagesSucceeds) {
     EXPECT_EQ(json, std::string("{\"model\":\"llama\",\"messages\":[{\"role\":\"user\",\"content\":\"What is in this image?\"}]}"));
 }
 
+TEST_F(HttpOpenAIHandlerParsingTest, ParsingMessagesSucceedsUrl) {
```
dtrawins, line 278: there is no test for downloading files with multipart and size >20MB. A functional tests should be added. If not in the PR, then later. Confirm at least manually now.
There is also no test for pulling over https using public and private certificate.

windows_create_package.bat
```diff
@@ -80,7 +80,8 @@ copy /Y %cd%\bazel-out\x64_windows-opt\bin\src\git2.dll dist\windows\ovms
 if !errorlevel! neq 0 exit /b !errorlevel!
 copy /Y %dest_dir%\git-lfs.exe dist\windows\ovms
 if !errorlevel! neq 0 exit /b !errorlevel!
-
+copy C:\opt\curl-8.13.0_1-win64-mingw\bin\libcurl-x64.dll dist\windows\ovms
```
dtrawins, line 83: you must add also licenses for curl and it's dependencies.

PR #3253, opened 2025-04-22
Allow IPv6 addresses

stevegrubb:
The REST API works with IPv6 addresses. But the address validation only supports IPv4. gRPC on the other hand is supposed to work with IPv6, but has some kind of problem with the current version of gPRC. It does support IN6ADDR_ANY, though.

The origin of the regex is here:
https://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses

Tested with:
podman network create --subnet 2001:db8::/64 --ipv6 ipv6_network
podman run -it --rm -p 8000:8000 --network ipv6_network --ip 2001:db8::10 --ipc=host -u 0 -v ~/models:/models --entrypoint=/bin/bash model_server:2025.0-ipv6

Then inside the container:
/ovms/bin/ovms --log_level DEBUG --log_path /root/ovms.log --model_path /models/resnet/ --model_name resnet  --port 9000 --rest_bind_address "2001:db8::10" --rest_port 8000

```
netstat -taunp
    Proto Recv-Q Send-Q Local Address       Foreign Address      State       PID/Program name
    tcp6       0      0 :::9000             :::*                 LISTEN      56/ovms
    tcp6       0      0 2001:db8::10:8000   :::*                 LISTEN      56/ovms
```

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -76,16 +76,18 @@ bool Config::check_hostname_or_ip(const std::string& input) {
     }
     bool all_numeric = true;
     for (char c : input) {
-        if (c == '.') {
+        if (c == '.' || c == ':') {
             continue;
         }
-        if (!::isdigit(c)) {
+        if (!::isxdigit(c)) {
             all_numeric = false;
         }
     }
     if (all_numeric) {
-        std::regex valid_ip_regex("^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$");
-        return std::regex_match(input, valid_ip_regex);
+        static const std::regex valid_ipv4_regex("^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$");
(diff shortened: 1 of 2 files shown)


dkalinowski:
@dtrawins @stevegrubb the change looks very good, however unit tests are missing. please take a look at `ovmsconfig_test.cpp`, we have ovms::Config::check_hostname_or_ip unit tests there

stevegrubb:
OK, I added tests for various kinds of address types. Note that Link-local IPv6 with zone index address types are not supported. They add a lot of complication and kubernetes will not normally assign that kind of an address in a cluster. It's more likely with multiple Ethernet cards and specific routing.

Note, I cannot get the unit tests to run via Dockerfile.redhat. If I use the master branch, it fails saying it cannot download a suitable version of OpenVino 2025. If I use the 2025.0 branch, lots of python mismatches in the .sh files. They all need converting to python3. Next, it wants to run export_model.py but that is not copied into the image. I fixed that, now it fails with: openvino.exceptions.OVTypeError: Tokenizer type 'Unigram' is not supported. This is right after the step that does: Exporting tokenizer to  /ovms/src/test/llm_testing/BAAI/bge-reranker-base/tokenizer/1

I did test locally to verify the regex does the right thing with all these different address types.

stevegrubb:
I figured out what was wrong on my end. The unit tests compile and pass. However, export_model.py still needs to be put in the right place. I guess that's a new PR.

dkalinowski (approved):

dkalinowski:
Thanks for contribution!
Merged in #3295  - closing

PR #3259, opened 2025-04-24
Fix running tests for Docker builds

stevegrubb:
If you build with a Dockerfile and pass --build-arg RUN_TESTS=1, it will fail to run the tests because the prepare_llm_models.sh calls a model_export python script at a specific directory location and it's not there. This patch copies model_export.py into the image and places it at the expected location so that models can be downloaded for testing.


dkalinowski:
Thanks for contribution!
Merged in #3298 - closing

PR #3260, opened 2025-04-25
Enable C++ only text generation

mzegla:
### 🛠 Summary

Enabling text generation endpoints in OVMS with Python disabled.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -242,7 +242,7 @@ build:windows --override_repository="boringssl=C:\\opt\\boringSSL-SwiftPM"
 
 # Windows config default flags
 build:windows --define=CLOUD_DISABLE=1
-build:windows --define=PYTHON_DISABLE=0
+build:windows --define=PYTHON_DISABLE=1
 build:windows --define=MEDIAPIPE_DISABLE=0
 # Change this path alongside WORKSPACE:"windows_openvino" path for model_api cmake compilation.
 build:windows --action_env OpenVINO_DIR="c:/opt/openvino/runtime/cmake"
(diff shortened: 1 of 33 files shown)


src/llm/py_jinja_template_processor.hpp
```diff
@@ -0,0 +1,41 @@
+//*****************************************************************************
+// Copyright 2024 Intel Corporation
```
rasapala, line 2: 2025

windows_create_package.bat
```diff
@@ -37,30 +37,34 @@ if !errorlevel! neq 0 exit /b !errorlevel!
 copy C:\%output_user_root%\openvino\runtime\bin\intel64\Release\*.dll dist\windows\ovms
 if !errorlevel! neq 0 exit /b !errorlevel!
 
+set "dest_dir=C:\opt"
+
+:: Remove below if Windows should always come without python
```
rasapala, line 42: Lets add a simple parameter to the script so that we can set one variable inside script and python will be added.
Can we remove install python from build dependencies now or we want to keep it for python nodes ?

.bazelrc
```diff
@@ -242,7 +242,7 @@ build:windows --override_repository="boringssl=C:\\opt\\boringSSL-SwiftPM"
 
 # Windows config default flags
 build:windows --define=CLOUD_DISABLE=1
-build:windows --define=PYTHON_DISABLE=0
+build:windows --define=PYTHON_DISABLE=1
```
rasapala, line 245: This also disables python_nodes. We do not want it on windows ? Also we do not test python compilation in this settings on windows.

rasapala (changes requested):
Lets discuss two build options with and without python on windows.

rasapala (approved):

michalkulakowski (approved):

PR #3261, opened 2025-04-25
Mkulakow/local filesystem image support vlm

michalkulakowski:
### 🛠 Summary

CVS-165618

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -251,7 +251,6 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages(std::optional<std::stri
                                 }
                             } else if (std::regex_match(url.c_str(), std::regex("^(http|https|ftp|sftp|)://(.*)"))) {
                                 SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Loading image using curl");
-                                curl_global_init(CURL_GLOBAL_ALL);
                                 int64_t sizeLimit = 20000000;  // restrict single image size to 20MB
                                 auto status = downloadImage(url.c_str(), decoded, sizeLimit);
                                 if (status != absl::OkStatus()) {
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -272,7 +272,7 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages(std::optional<std::stri
                                 SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Loading image from local filesystem");
                                 const auto firstMissmatch = std::mismatch(url.begin(), url.end(), allowedLocalMediaPath.value().begin(), allowedLocalMediaPath.value().end());
                                 if (firstMissmatch.second != allowedLocalMediaPath.value().end()) {
-                                    return absl::InvalidArgumentError("Given filepath is not subpath of allowed_local_media_path provided in graph.pbtxt");
+                                    return absl::InvalidArgumentError("Given filepath is not subpath of allowed_local_media_path");
                                 }
                                 try {
                                     ov::Tensor tensor = loadImageStbiFromFile(url.c_str());
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -269,11 +269,8 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages(std::optional<std::stri
                                     SPDLOG_LOGGER_ERROR(llm_calculator_logger, ss.str());
                                     return absl::InvalidArgumentError("Image parsing failed");
                                 }
-#ifdef __linux__
-                            } else if (std::regex_match(url.c_str(), std::regex("[^\\0]+"))) {
-#elif _WIN32
-                            } else if (std::regex_match(url.c_str(), std::regex("^[a-zA-Z]:\\(((?![<>:\"\/\\|?*]).)+((?<![ .])\\)?)*$"))) {
-#endif
+
+                            } else {
                                 SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Loading image from local filesystem");
                                 if (allowedLocalMediaPath.has_value()) {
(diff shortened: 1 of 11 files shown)


src/llm/apis/openai_completions.cpp
```diff
@@ -166,7 +161,25 @@ static absl::Status downloadImage(const char* url, std::string& image, const int
     return absl::OkStatus();
 }
 
-absl::Status OpenAIChatCompletionsHandler::parseMessages() {
+ov::Tensor loadImageStbiFromMemory(const std::string& imageBytes) {
+    int x = 0, y = 0, channelsInFile = 0;
```
mzegla, line 165: How is `channelsInFile` used? It's hardcoded here along with `x` and `y`, but other values are passed to `loadImageStbi`

src/llm/apis/openai_completions.cpp
```diff
@@ -166,7 +161,25 @@ static absl::Status downloadImage(const char* url, std::string& image, const int
...
+    constexpr int desiredChannels = 3;
+    unsigned char* image = stbi_load_from_memory(
+        (const unsigned char*)imageBytes.data(), imageBytes.size(),
+        &x, &y, &channelsInFile, desiredChannels);
+    return loadImageStbi(image, x, y, desiredChannels);
+}
+
+ov::Tensor loadImageStbiFromFile(char const* filename) {
+    int x = 0, y = 0, channelsInFile = 0;
```
mzegla, line 174: Same question about `channelsInFile` as above

src/llm/apis/openai_completions.cpp
```diff
@@ -166,7 +161,25 @@ static absl::Status downloadImage(const char* url, std::string& image, const int
...
+    constexpr int desiredChannels = 3;
+    unsigned char* image = stbi_load(
+        filename,
+        &x, &y, &channelsInFile, desiredChannels);
+    return loadImageStbi(image, x, y, desiredChannels);
```
mzegla, line 179: We use raw pointers: `filename` and `image`. Are you sure they are both valid and we don't need to do any checks here?

src/llm/apis/openai_completions.cpp
```diff
@@ -224,30 +237,58 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
...
+                                } catch (std::runtime_error& e) {
+                                    std::stringstream ss;
+                                    ss << "Image parsing failed: " << e.what();
+                                    SPDLOG_LOGGER_ERROR(llm_calculator_logger, ss.str());
+                                    return absl::InvalidArgumentError(ss.str());
+                                }
                             } else if (std::regex_match(url.c_str(), std::regex("^(https?:\\/\\/)?([\\da-z\\.-]+)\\.([a-z\\.]{2,6})([\\/\\w \\.-]*)*\\/?$"))) {
+                                SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Loading image using curl");
                                 curl_global_init(CURL_GLOBAL_ALL);
                                 int64_t sizeLimit = 20000000;  // restrict single image size to 20MB
```
mzegla, line 257: unrelated question - do we have such restriction in general, also for other loading paths or is it just for using curl?

src/llm/apis/openai_completions.cpp
```diff
@@ -224,30 +237,58 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
...
-                            try {
-                                ov::Tensor tensor = load_image_stbi(decoded);
-                                request.imageHistory.push_back({i, tensor});
-                            } catch (std::runtime_error& e) {
-                                std::stringstream ss;
-                                ss << "Image parsing failed: " << e.what();
-                                SPDLOG_LOGGER_ERROR(llm_calculator_logger, ss.str());
-                                return absl::InvalidArgumentError("Image parsing failed");
+                                SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Loading image from local filesystem");
+                                if (allowedLocalMediaPath.has_value()) {
```
mzegla, line 275: If allow local media path does not exist we continue and read url, so we allow everything then, is that intended?

src/llm/llm_calculator.proto
```diff
@@ -78,4 +78,6 @@ message LLMCalculatorOptions {
     optional bool draft_dynamic_split_fuse = 17;
 
     optional PipelineType pipeline_type = 18 [default = AUTO];
+    
+    optional string allowed_local_media_path = 19;
```
mzegla, line 82: Shouldn't it be more deployment-wide setting on the server level? 
It injects frontend option to servable configuration that holds backend settings.

src/test/http_openai_handler_test.cpp
```diff
@@ -547,7 +673,7 @@ TEST_F(HttpOpenAIHandlerParsingTest, ParsingMessagesEmptyImageUrlFails) {
     doc.Parse(json.c_str());
     ASSERT_FALSE(doc.HasParseError());
     std::shared_ptr<ovms::OpenAIChatCompletionsHandler> apiHandler = std::make_shared<ovms::OpenAIChatCompletionsHandler>(doc, ovms::Endpoint::CHAT_COMPLETIONS, std::chrono::system_clock::now(), *tokenizer);
-    EXPECT_EQ(apiHandler->parseMessages(), absl::InvalidArgumentError("Url should contain base64 encoded string followed by \"base64,\" prefix or valid URL"));
+    EXPECT_EQ(apiHandler->parseMessages(), absl::InvalidArgumentError("Image file  parsing failed: can't fopen"));
 }
```
mzegla, line 677: I know it's in another scope but do we have a test for loading image from the network from non existing location?

src/llm/servable.hpp
```diff
@@ -90,6 +90,7 @@ struct GenAiServableProperties {
 #if (PYTHON_DISABLE == 0)
     PyJinjaTemplateProcessor templateProcessor;
 #endif
+    std::optional<std::string> allowedLocalMediaPath;
```
mzegla, line 93: looks like it's not needed anymore

mzegla (approved):

rasapala (approved):

PR #3274, opened 2025-05-05
Install openvino-tokenizer python module in image

stevegrubb:
I was working through a demo where it said to run
"pip install -U optimum[diffusers,openvino]".
When I did that, I noticed that it downloaded:
openvino_tokenizers-2025.1.0.0-py3-none-manylinux2014_x86_64.whl. However, the tokenizer is built and installed in the image by the Dockerfile. After more digging, I found that openvino's python module is not installed in the image. This patch installs the openvino-tokenizer python module in the image and uses a symlink to point to the tokenizer shared object to save space.


mzegla:
@stevegrubb when working with which demo did you encounter this issue?

stevegrubb:
I think it was the stable_diffusion demo. I was logged in to an image built by Dockerfile.redhat. But I guess it doesn't matter which demo, because an end user can write code that uses the tokenizer and it should be preinstalled since we took the time to build it.

stevegrubb:
I updated the tokenizer version 2025.2.

PR #3290, opened 2025-05-14
Validate create graph parameters

michalkulakowski:
### 🛠 Summary

CVS-166727

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -132,7 +132,6 @@ bool Config::validate() {
                 std::cerr << "For now only OpenVINO models are supported";
                 return false;
             }
-            return true;
         }
 
         if (this->serverSettings.hfSettings.task == embeddings) {
@@ -149,11 +148,11 @@ bool Config::validate() {
             }
 
             if (std::find(allowedBoolValues.begin(), allowedBoolValues.end(), settings.truncate) == allowedBoolValues.end()) {
-                std::cerr << "normalize: " << settings.truncate << " is not allowed. Supported values: true, false" << std::endl;
+                std::cerr << "truncate: " << settings.truncate << " is not allowed. Supported values: true, false" << std::endl;
                 return false;
             }
-            return true;
         }
+        return true;
     }
     if (this->serverSettings.serverMode == LIST_MODELS_MODE) {
         if (this->serverSettings.hfSettings.downloadPath.empty()) {
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -86,19 +86,16 @@ bool Config::check_hostname_or_ip(const std::string& input) {
 }
 
 bool Config::validate() {
-    // TODO: CVS-166727 Add validation of all parameters once the CLI model export flags will be implemented
     if (this->serverSettings.serverMode == HF_PULL_MODE) {
         if (this->serverSettings.hfSettings.task == UNKNOWN_GRAPH) {
             std::cerr << "Error: --task parameter not set." << std::endl;
             return false;
         }
-#if (PYTHON_DISABLE == 1)
         if (serverSettings.hfSettings.sourceModel.rfind("OpenVINO/", 0) != 0) {
             std::cerr << "For now OVMS version without python supports pulling OpenVINO models only";
             return false;
         }
-#endif
-        if (this->serverSettings.hfSettings.task == text_generation) {
+        if (this->serverSettings.hfSettings.task == TEXT_GENERATION_GRAPH) {
             if (!std::holds_alternative<TextGenGraphSettingsImpl>(this->serverSettings.hfSettings.graphSettings)) {
                 std::cerr << "Graph options not initialized for text generation.";
                 return false;
@@ -135,7 +132,7 @@ bool Config::validate() {
             }
         }
 
-        if (this->serverSettings.hfSettings.task == embeddings) {
+        if (this->serverSettings.hfSettings.task == EMBEDDINGS_GRAPH) {
             if (!std::holds_alternative<EmbeddingsGraphSettingsImpl>(this->serverSettings.hfSettings.graphSettings)) {
                 std::cerr << "Graph options not initialized for embeddings.";
                 return false;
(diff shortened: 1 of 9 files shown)


src/config.cpp
```diff
@@ -92,6 +92,66 @@ bool Config::validate() {
...
+
+            if (settings.targetDevice != "NPU") {
+                if (settings.pluginConfig.maxPromptLength.has_value()) {
+                    std::cerr << "max_prompt_len is only supported for NPU target device";
+                    return false;
+                }
+            }
+
+            if (serverSettings.hfSettings.sourceModel.rfind("OpenVINO/", 0) != 0) {
+                std::cerr << "For now only OpenVINO models are supported";
```
rasapala, line 132: I think this should be generic check not only for text_generation. @dtrawins

rasapala (changes requested):

src/config.cpp
```diff
@@ -92,6 +92,66 @@ bool Config::validate() {
             std::cerr << "Error: --task parameter not set." << std::endl;
```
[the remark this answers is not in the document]
rasapala: Not removed.

src/config.cpp
```diff
@@ -92,6 +92,67 @@ bool Config::validate() {
...
             return false;
         }
+#if (PYTHON_DISABLE == 1)
```
rasapala, line 95: #if (PYTHON_DISABLE == 1) can be removed IMHO. We base optimum-cli functionality on external tools/python/environment not on internal ovms compilation flags. This check should be generic until I change it in optimum-cli enabling PR. @dtrawins

src/graph_export/graph_export.cpp
```diff
@@ -102,12 +108,31 @@ static Status createTextGenerationGraphTemplate(const std::string& directoryPath
...
+        return StatusCode::MEDIAPIPE_GRAPH_CONFIG_FILE_INVALID;
+    }
     // clang-format on
     std::string fullPath = FileSystem::joinPath({directoryPath, "graph.pbtxt"});
     return FileSystem::createFileOverwrite(fullPath, oss.str());
 }
 
+static Status validateSubconfigSchema(const std::string& subconfig, const std::string& type) {
+    rapidjson::Document subconfigJson;
+    rapidjson::ParseResult parseResult = subconfigJson.Parse(subconfig.c_str());
```
rasapala, line 124: I am for parsing subconfig and mediapipe graph once it is created. @dtrawins want to remove it. Lets decide now.

rasapala (changes requested):
@dtrawins please respond.

rasapala:
Please change std::string "true":"false" flags to bool in cxxopts and settingsImpl.

src/config.cpp
dtrawins: I think this is a valid check. The version without python shouldn't pull models outside of OV organization. Otherwise the pulled model will not work. It might change in the future but for how it looks reasonable.
@rasapala  do you have in mind replacing it with a logic for checking if optimum-cli is present in runtime? It would be also fine when we have this detection.

src/config.cpp
```diff
@@ -92,6 +92,67 @@ bool Config::validate() {
...
+            }
+            auto settings = std::get<EmbeddingsGraphSettingsImpl>(this->serverSettings.hfSettings.graphSettings);
+
+            std::vector allowedBoolValues = {"false", "true"};
+            if (std::find(allowedBoolValues.begin(), allowedBoolValues.end(), settings.normalize) == allowedBoolValues.end()) {
+                std::cerr << "normalize: " << settings.normalize << " is not allowed. Supported values: true, false" << std::endl;
+                return false;
+            }
+
+            if (std::find(allowedBoolValues.begin(), allowedBoolValues.end(), settings.truncate) == allowedBoolValues.end()) {
```
dtrawins, line 151: truncate option will be effective only when convert_tokenizer is used? Maybe this parameter should be dropped for now?

src/graph_export/graph_export.cpp
```diff
@@ -102,12 +108,31 @@ static Status createTextGenerationGraphTemplate(const std::string& directoryPath
         }
     }
     })";
-
+    ::mediapipe::CalculatorGraphConfig config;
+    bool success = ::google::protobuf::TextFormat::ParseFromString(oss.str(), &config);
+    if (!success) {
+        SPDLOG_ERROR("Created graph config couldn't be parsed - some parameters values are invalid.");
```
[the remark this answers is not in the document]
dtrawins: I have mixed feelings if we should include this check in the graph export. It will be checked in the model initialization.
Do we know which parameters should cause such error? How is printed out? The log format might look strange for the CLI interface.

src/graph_export/graph_export.cpp
dtrawins: @rasapala  subconfig will not be used then new embeddings and rerank calculators are merged.

dtrawins (approved):

src/test/ovmsconfig_test.cpp
```diff
@@ -795,6 +825,134 @@ TEST(OvmsGraphConfigTest, positiveDefaultStart) {
...
+        (char*)"ovms",
+        (char*)"--pull",
+        (char*)"--source_model",
+        (char*)modelName.c_str(),
+        (char*)"--model_repository_path",
+        (char*)downloadPath.c_str(),
+    };
+
+    int arg_count = 6;
+    EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(OVMS_EX_USAGE), "or now OVMS version without python supports pulling OpenVINO models only");
```
rasapala, line 953:
```suggestion
    EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(OVMS_EX_USAGE), "Only OpenVINO models supported");
```

rasapala (changes requested):
1 change.

rasapala (approved):

PR #3294, opened 2025-05-15
tools parsing in chat/completions

dtrawins:
### 🛠 Summary

CVS-167117 CVS-167115

### 🧪 Checklist

- [ ] Unit tests added.
``


src/llm/apis/openai_completions.cpp
```diff
@@ -182,10 +182,12 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
         // Add new message to chat history
         request.chatHistory.push_back({});
         for (auto member = obj.MemberBegin(); member != obj.MemberEnd(); member++) {
+            std::cout << "Member name: " << member->name.GetString() << std::endl;   
             if (!member->name.IsString())
                 return absl::InvalidArgumentError("Invalid message structure");
-            if (member->value.IsString()) {
+            if (member->value.IsString() && (member->name.GetString() == std::string("role") || member->name.GetString() == std::string("content"))) { 
                 // Add new field to the last message in history
+                std::cout << "Adding field to last message: " << member->name.GetString() << "value" << member->value.GetString() << std::endl;
```
mzegla, line 190: remove std::couts

src/llm/apis/openai_completions.cpp
```diff
@@ -259,8 +261,6 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
                     if (member->value.IsString()) {
                         request.chatHistory.back().insert({member->name.GetString(), member->value.GetString()});
                     }
-                } else {
-                    return absl::InvalidArgumentError("Invalid message structure - content should be string or array");
```
mzegla, line 263: Why is this removed?

src/llm/apis/openai_completions.cpp
```diff
@@ -278,6 +278,77 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
...
+                } else {
+                    return absl::InvalidArgumentError("tool_choice.function.name is not a valid string");
+                }
+            } else {
+                return absl::InvalidArgumentError("tool_choice.function is not a valid JSON object");
+            }
+        } else
+            return absl::InvalidArgumentError("tool_choice is not a valid JSON object or string");
+    }
+    std::cout << "Tool choice: " << tool_choice << std::endl;
```
mzegla, line 309: remove before merge

src/llm/apis/openai_completions.cpp
```diff
@@ -278,6 +278,77 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
...
+                    tool_choice = nameIt->value.GetString();
+                } else {
+                    return absl::InvalidArgumentError("tool_choice.function.name is not a valid string");
+                }
+            } else {
+                return absl::InvalidArgumentError("tool_choice.function is not a valid JSON object");
+            }
+        } else
+            return absl::InvalidArgumentError("tool_choice is not a valid JSON object or string");
+    }
```
mzegla, line 308: Please align naming:
tool_choice_it ->toolChoiceIt
tool_choice_functionIt -> toolChoiceFunctionIt 
tool_choice -> toolChoice

src/llm/apis/openai_completions.cpp
```diff
@@ -278,6 +278,77 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
...
+    for (size_t i = 0; i < it->value.GetArray().Size(); i++) {
+        auto& obj = it->value.GetArray()[i];
+        if (!obj.IsObject())
+            return absl::InvalidArgumentError("Tool is not a JSON object");
+        auto functionIt = obj.FindMember("function");
+        if (functionIt != obj.MemberEnd() && functionIt->value.IsObject()) {
+            auto nameIt = functionIt->value.GetObject().FindMember("name");
+            if (nameIt != functionIt->value.GetObject().MemberEnd() && nameIt->value.IsString()) {
+                std::string functionName = nameIt->value.GetString();
+                std::cout << "Function name: " << functionName << std::endl;
```
mzegla, line 328: remove before merge

src/llm/apis/openai_completions.cpp
```diff
@@ -278,6 +278,77 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
...
+            return absl::InvalidArgumentError("tool_choice is not a valid JSON object or string");
+    }
+    std::cout << "Tool choice: " << tool_choice << std::endl;
+
+    bool jsonChanged = false;
+    if (tool_choice == "auto")  // for now, with auto choice we don't need to do anything
+        return absl::OkStatus();
+    if (tool_choice == "none") {
+        // remove tools from the request
+        doc.RemoveMember("tools");
```
mzegla, line 316: Is that call safe? We remove `tools` from the docs, but don't we have invalid `it` after that operation?
Shouldn't we exit here and not get into below loop?

src/llm/servable.cpp
```diff
@@ -115,7 +115,8 @@ absl::Status GenAiServable::prepareInputs(std::shared_ptr<GenAiServableExecution
         inputText = executionContext->apiHandler->getPrompt().value();
     }
     }
-
+    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "payload: {}", executionContext->apiHandler->getProcessedJson());
+    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Input text: {}", inputText);
```
mzegla, line 119: Remove or change level to trace?

tests/functional/test_llm_json.py
```diff
@@ -0,0 +1,355 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import numpy as np
+import pytest
+import os
+import logging
```
mzegla, line 20: Can you move all imports in this file to the top level?

tests/functional/test_llm_json.py
```diff
@@ -0,0 +1,355 @@
...
+        response = requests.post(
+            url=f"{base_url}/chat/completions",
+            json=body,
+            headers={"Authorization": f"Bearer unused"}
+        )
+
+        if response.status_code == 200:
+            print("API Response:", response.json())
+        else:
+            print(f"Failed to get response: {response.status_code}, {response.text}")
```
mzegla, line 97: why do we run it also using requests? 
if completion via openai library worked then it doesn't feel like this part adds something

tests/functional/test_llm_json.py
```diff
@@ -0,0 +1,355 @@
...
+            {"role": "user", "content": "What is the weather like in Paris today?"}
+        ]
+        completion = client.chat.completions.create(
+            model=model_name,
+            messages=messages,
+            tools=tools,
+            tool_choice={"type": "function", "function": {"name": "get_weather"}},
+        )
+
+        print("COMPLETION:",completion)
```
mzegla, line 79: Remove or add some assertion?

tests/functional/test_llm_json.py
```diff
@@ -0,0 +1,355 @@
...
+            assert False, f"tool_call is not a valid JSON: {e}"
+        except ValidationError as e:
+            assert False, f"tool_call does not match the schema: {e}"
+
+        messages.append({'role': 'assistant', 'reasoning_content': None, 'content': '', 'tool_calls': [{'id': 'chatcmpl-tool-d39b13c90f9b4d48b08c16455553dbec', 'type': 'function', 'function': {'name': 'get_weather', 'arguments': '{"location": "Paris, France"}'}}]})
+        messages.append({"role": "tool", "tool_call_id": "chatcmpl-tool-d39b13c90f9b4d48b08c16455553dbec", "content": "15 degrees Celsius"})
+
+        print("Messages after tool call:", messages)
+
+        completion2 = client.chat.completions.create(
```
mzegla, line 122: we could overwrite original completion here I guess... 
completion2 looks a bit awkward

tests/functional/test_llm_json.py
```diff
@@ -0,0 +1,355 @@
...
+            assert True, "tool_call is a valid JSON and matches the schema"
+        except json.JSONDecodeError as e:
+            assert False, f"tool_call is not a valid JSON: {e}"
+        except ValidationError as e:
+            assert False, f"tool_call does not match the schema: {e}"
+
+        messages.append({'role': 'assistant', 'reasoning_content': None, 'content': '', 'tool_calls': [{'id': 'chatcmpl-tool-d39b13c90f9b4d48b08c16455553dbec', 'type': 'function', 'function': {'name': 'get_weather', 'arguments': '{"location": "Paris, France"}'}}]})
+        messages.append({"role": "tool", "tool_call_id": "chatcmpl-tool-d39b13c90f9b4d48b08c16455553dbec", "content": "15 degrees Celsius"})
+
+        print("Messages after tool call:", messages)
```
mzegla, line 120: assertion instead of print?

tests/functional/test_llm_json.py
```diff
@@ -0,0 +1,355 @@
...
+
+        completion2 = client.chat.completions.create(
+            model=model_name,
+            messages=messages,
+            tools=tools,
+        )
+        print(completion2.choices[0].message)
```
mzegla, line 127: assertion instead of print?

tests/functional/test_llm_json.py
```diff
@@ -0,0 +1,355 @@
...
+            assert False, f"tool_call does not match the schema: {e}"
+
+        messages.append(completion.choices[0].message)
+        messages.append({"role": "tool", "tool_call_id": completion.choices[0].message.tool_calls[0].id, "content": "15 degrees Celsius"})
+        messages.append({"role": "tool", "tool_call_id": completion.choices[0].message.tool_calls[1].id, "content": "pm10 28µg/m3"})
+
+        print("Messages after tool call:", messages)
+
+        try:
+            completion2 = client.chat.completions.create(
```
mzegla, line 229: assignment is not needed if we expect error

src/llm/apis/openai_completions.cpp
dtrawins: before, there was assumption that only the fields can be in the message, role and content. it is not the case with tool role.

mzegla: This path is executed when content is not an array and it looks to me that it's relevant part of the logic we had so far.
I would make explicit exception for `tool` role instead

src/llm/servable_initializer.cpp
```diff
@@ -99,18 +99,34 @@ void GenAiServableInitializer::loadPyTemplateProcessor(std::shared_ptr<GenAiServ
                 eos_token = data.get("eos_token", "")
                 eos_token = "" if eos_token is None else eos_token  # Null token conversion to empty string.
                 chat_template = data.get("chat_template", default_chat_template)
+                if isinstance(chat_template, list):
+                    for template_entry in chat_template:
+                        if isinstance(template_entry, dict):
+                            if template_entry.get("name") == "default":
+                                chat_template = template_entry.get("template", default_chat_template)
+                            elif template_entry.get("name") == "tool_use":
+                                tool_chat_template = template_entry.get("template", default_chat_template)
```
mzegla, line 108: should default here be the same as for regular chat template?

src/llm/servable_initializer.cpp
```diff
@@ -99,18 +99,34 @@ void GenAiServableInitializer::loadPyTemplateProcessor(std::shared_ptr<GenAiServ
...
 
             if template is None:
                 jinja_env = ImmutableSandboxedEnvironment(trim_blocks=True, lstrip_blocks=True)
                 jinja_env.policies["json.dumps_kwargs"]["ensure_ascii"] = False
                 jinja_env.globals["raise_exception"] = raise_exception
                 template = jinja_env.from_string(chat_template)
+            if tool_chat_template is not None:
+                jinja_env = ImmutableSandboxedEnvironment(trim_blocks=True, lstrip_blocks=True)
+                jinja_env.policies["json.dumps_kwargs"]["ensure_ascii"] = False
+                jinja_env.globals["raise_exception"] = raise_exception
```
mzegla, line 118: we have a lot of duplication here, can we reuse part with setting up jinja env?

src/llm/apis/openai_completions.cpp
```diff
@@ -278,6 +276,74 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
     return absl::OkStatus();
 }
 
+absl::Status OpenAIChatCompletionsHandler::parseTools() {
+    auto it = doc.FindMember("tools");
+    if (it == doc.MemberEnd())
```
dkalinowski, line 281: what happens when someone specifies "tool_choice" but doesnt specify "tools"? As far I see we proceed with OK, meaning that user doesnt see any error, is it intended?

tests/functional/test_llm_json.py
```diff
@@ -0,0 +1,362 @@
...
+# limitations under the License.
+#
+
+import numpy as np
+import pytest
+import os
+import logging
+
+model_name = os.getenv("MODEL_NAME", "meta-llama/Llama-3.1-8B-Instruct")
+base_url = os.getenv("BASE_URL", "http://localhost:8000/v3")
```
dkalinowski, line 23: can we use reservation manager?

tests/functional/test_llm_json.py
```diff
@@ -0,0 +1,362 @@
...
+        except json.JSONDecodeError as e:
+            assert False, f"tool_call is not a valid JSON: {e}"
+        except ValidationError as e:
+            assert False, f"tool_call does not match the schema: {e}"
+
+        messages.append(completion.choices[0].message)
+        messages.append({"role": "tool", "tool_call_id": completion.choices[0].message.tool_calls[0].id, "content": "15 degrees Celsius"})
+        messages.append({"role": "tool", "tool_call_id": completion.choices[0].message.tool_calls[1].id, "content": "pm10 28µg/m3"})
+
+        print("Messages after tool call:", messages)
```
dkalinowski, line 233: should we also assert that tools really get used? are part of the response?

src/llm/apis/openai_completions.cpp
```diff
@@ -278,6 +276,74 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
...
+        auto& obj = it->value.GetArray()[i];
+        if (!obj.IsObject())
+            return absl::InvalidArgumentError("Tool is not a JSON object");
+        auto functionIt = obj.FindMember("function");
+        if (functionIt != obj.MemberEnd() && functionIt->value.IsObject()) {
+            auto nameIt = functionIt->value.GetObject().FindMember("name");
+            if (nameIt != functionIt->value.GetObject().MemberEnd() && nameIt->value.IsString()) {
+                std::string functionName = nameIt->value.GetString();
+                if (tool_choice != functionName) {
+                    it->value.Erase(&obj);
```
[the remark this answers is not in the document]
dtrawins: there is functional test included, a single item is removed and the rest of tools is preserved

tests/functional/test_llm_json.py
dtrawins: not yet, those tests can be executed manually, CI integration will be done later

tests/functional/test_llm_json.py
dtrawins: there is assertion for the response. tool_calls are included in the response when the tool is used.

src/llm/apis/openai_completions.cpp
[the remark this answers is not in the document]
dkalinowski: can you point the functional test that erases tool in a middle? last tool? first tool?
I think we need to split this parser to separate functionality which can be unit tested, theres plenty of use cases which can be tested without really loading the model

from my understanding there is an issue here, example:

tools:
[tool1, tool2, tool3]

tool_choice:
tool1 tool2

we iterate over the tools 3 times. for index 1, we remove the tool (tool2) and increase i++ which is now at index 2, tool3 check is completely ommited

tests/functional/test_llm_json.py
dkalinowski: so for now this functionality has no unit tests or functional tests in CI :(

src/llm/apis/openai_completions.cpp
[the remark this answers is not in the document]
dtrawins: per API specification, the tool choice can't include a list of tools. it can be a string with auto/none value or a single object representing a single tool.

src/llm/apis/openai_completions.cpp
```diff
@@ -259,8 +259,6 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages() {
                     if (member->value.IsString()) {
                         request.chatHistory.back().insert({member->name.GetString(), member->value.GetString()});
                     }
-                } else {
-                    return absl::InvalidArgumentError("Invalid message structure - content should be string or array");
```
dkalinowski, line 263: TODO: removal of this else requires multiple unit tests, removal exposes new paths for tests

dkalinowski (approved):

PR #3297, opened 2025-05-19
Steve's tokenizer changes

dkalinowski:
### 🛠 Summary

Taken from PR https://github.com/openvinotoolkit/model_server/pull/3274


Dockerfile.redhat
```diff
@@ -220,6 +220,17 @@ RUN git clone https://github.com/openvinotoolkit/openvino_tokenizers.git /openvi
 WORKDIR /openvino_tokenizers/build
 RUN cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE="${VERBOSE_LOGS}"  && cmake --build . --parallel $JOBS ; cp /openvino_tokenizers/build/src/lib*.so /opt/intel/openvino/runtime/lib/intel64/
 
+WORKDIR /openvino_tokenizers/
```
dtrawins, line 223: shall we add it also to ubuntu dockerfile?

stevegrubb: Not sure who you're asking...but I originally didn't touch Ubuntu because I didn't want to force changes on another distribution. That said, it makes sense to add so that Dockerfiles are roughly in sync. And if end users need the tokenizer, it's already installed.

dtrawins (approved):

mzegla (approved):

PR #3315, opened 2025-05-27
Include tool calls in OpenAI API serialization

mzegla:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``


dkalinowski:
Please update documentation that generation config defaults differ now. And that these differ per-model.

src/llm/apis/openai_completions.cpp
```diff
@@ -727,7 +728,7 @@ absl::Status OpenAIChatCompletionsHandler::parseRequest(std::optional<uint32_t>
     return status;
 }
 
-std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const std::vector<ov::genai::GenerationOutput>& generationOutputs) {
+std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const std::vector<ov::genai::GenerationOutput>& generationOutputs, const std::string responseParserName) {
```
dkalinowski, line 731: dont copy the string, pass by reference, please check all other places in this PR

src/llm/apis/openai_completions.cpp
```diff
@@ -869,17 +881,36 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const std::vect
...
+
+                writer.String("type");
+                writer.String("function");
+
+                writer.String("function");
+                writer.StartObject();  // {
+                writer.String("name");
+                writer.String(toolCall.name.c_str());
+                writer.String("arguments");
+                writer.String(toolCall.arguments.c_str());  // Assuming toolsResponse is a valid JSON string
```
dkalinowski, line 905: is it validated anywhere?

src/llm/apis/openai_completions.cpp
```diff
@@ -924,7 +955,7 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const std::vect
     return buffer.GetString();
 }
 
-std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const ov::genai::EncodedResults& results) {  // TODO separate common part with function implemented above
+std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const ov::genai::EncodedResults& results, const std::string responseParserName) {  // TODO separate common part with function implemented above
```
dkalinowski, line 958: why do we have 2 functions, that do the same thing? I dont think is clean approach, I see that the arguments differ, but as you can see this is not maintainable

maybe we could convert one arg to another before passing to single function that does the work? or some intermediate format, not sure what would fit here, probably std::string or vector of strings (instead of EncodedResults)

(comment mentioned in group chat)

src/llm/language_model/continuous_batching/servable_initializer.cpp
```diff
@@ -73,6 +73,15 @@ Status ContinuousBatchingServableInitializer::initialize(std::shared_ptr<GenAiSe
     }
     auto properties = std::static_pointer_cast<ContinuousBatchingServableProperties>(servable->getProperties());
     properties->modelsPath = parsedModelsPath;
+    std::filesystem::path modelGenerationConfigPath = std::filesystem::path(parsedModelsPath) / "generation_config.json";
+    if (std::filesystem::exists(modelGenerationConfigPath)) {
+        properties->baseGenerationConfig = ov::genai::GenerationConfig(modelGenerationConfigPath.string());
+    }
+
+    if (nodeOptions.has_response_parser()) {
+        properties->responseParserName = nodeOptions.response_parser();
```
dkalinowski, line 82: what if there is no parser selected, but user selects the tool, and model produces tool output?

src/llm/response_parsers/utils.hpp
```diff
@@ -0,0 +1,38 @@
...
+std::string generateRandomId() {
+    static const char alphanum[] =
+        "0123456789"
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        "abcdefghijklmnopqrstuvwxyz";
+    static constexpr int idLength = 9;
+    static thread_local std::mt19937 rng{std::random_device{}()};
+    static thread_local std::uniform_int_distribution<> dist(0, sizeof(alphanum) - 2);
+
+    std::string id;
```
dkalinowski, line 31: Please verify, if we can rely on this id randomizer from SDL standpoint

dtrawins: what latency is adding this function? I wonder if a sequential number would be quicker and good enough.

src/llm/apis/openai_completions.cpp
mzegla: Yes. Check out last phases of parsing the the parse() method.

dtrawins (approved):

dkalinowski (approved):

PR #3322, opened 2025-05-30
Unit tests for response parsers

mzegla:
### 🛠 Summary

Adding unit tests for parsers introduced in: https://github.com/openvinotoolkit/model_server/pull/3315

### DIFF
--- a/prepare_llm_models.sh
+++ b/prepare_llm_models.sh
@@ -24,10 +24,27 @@ CB_MODEL="facebook/opt-125m"
 EMBEDDING_MODEL="thenlper/gte-small"
 RERANK_MODEL="BAAI/bge-reranker-base"
 VLM_MODEL="OpenGVLab/InternVL2-1B"
-if [ -d "$1/$CB_MODEL" ] && [ -d "$1/$EMBEDDING_MODEL" ] && [ -d "$1/$RERANK_MODEL" ] && [ -d "$1/$VLM_MODEL" ]; then
-  echo "Models directory $1 exists. Skipping downloading models."
+
+# Models for tools testing. Only tokenizers are downloaded.
+QWEN3_MODEL="Qwen/Qwen3-8B"
+LLAMA3_MODEL="meta-llama/Llama-3.1-8B-Instruct"
+HERMES3_MODEL="NousResearch/Hermes-3-Llama-3.1-8B"
+
+MODELS=("$CB_MODEL" "$EMBEDDING_MODEL" "$RERANK_MODEL" "$VLM_MODEL" "$QWEN3_MODEL" "$LLAMA3_MODEL" "$HERMES3_MODEL")
+
+all_exist=true
+for model in "${MODELS[@]}"; do
+  if [ ! -d "$1/$model" ]; then
+    all_exist=false
+    break
+  fi
+done
+
+if $all_exist; then
+  echo "All model directories exist in $1. Skipping downloading models."
   exit 0
 fi
+
 if [ "$(python3 -c 'import sys; print(sys.version_info[1])')" -le "8" ]; then echo "Prepare models with python > 3.8."; exit 1 ; fi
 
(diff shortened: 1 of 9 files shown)


dtrawins (approved):

prepare_llm_models.sh
```diff
@@ -68,3 +85,24 @@ if [ -d "$1/$RERANK_MODEL" ]; then
 else
   python3 demos/common/export_models/export_model.py rerank --source_model "$RERANK_MODEL" --weight-format int8 --model_repository_path $1
 fi
+
+if [ -d "$1/$QWEN3_MODEL" ]; then
+  echo "Models directory $1/$QWEN3_MODEL exists. Skipping downloading models."
+else
+  mkdir -p $1/$QWEN3_MODEL
+  convert_tokenizer $QWEN3_MODEL --with_detokenizer -o $1/$QWEN3_MODEL
```
dkalinowski, line 93: where to take convert_tokenizer from?

src/llm/response_parsers/utils.hpp
```diff
@@ -19,7 +19,7 @@
 #include <string>
 
 namespace ovms {
-std::string generateRandomId() {
+static std::string generateRandomId() {
```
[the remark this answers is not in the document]
dkalinowski: agree! please move it to cpp so we have only one symbol of that

src/llm/response_parsers/qwen3_response_parser.hpp
```diff
@@ -64,9 +71,14 @@ class Qwen3ResponseParser : public BaseResponseParser {
         // Assuming content ends when tool calls start, so we find the first occurrence of <tool_call> after the content start
         auto contentEndIt = std::find(contentStartIt, generatedTokens.end(), toolCallStartTokenId);
 
-        if (contentStartIt != generatedTokens.end() && contentEndIt != generatedTokens.end() && contentStartIt < contentEndIt) {
+        if (contentStartIt == contentEndIt) {
```
dkalinowski: can we already hide implementation inside cpp unit? what was the explanation of why not earlier? this will greatly decrease compilation time

dkalinowski: if we did that, we probably could even hide rapidjson includes from headers. please dont extend already long compile times

dkalinowski: please do that for all new tool parsers

src/test/llm/response_parsers/hermes3_response_parser_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+    // The results should be identical
+    std::vector<std::string> inputs = {inputWithProperClosure, inputWithImproperClosure};
+    for (auto& input : inputs) {
+        auto generatedTensor = tokenizer->encode(input, ov::genai::add_special_tokens(false)).input_ids;
+        std::vector<int64_t> generatedTokens(generatedTensor.data<int64_t>(), generatedTensor.data<int64_t>() + generatedTensor.get_size());
+        ParsedResponse parsedResponse = responseParser->parse(generatedTokens);
+        EXPECT_EQ(parsedResponse.content, "");
+        EXPECT_EQ(parsedResponse.reasoning, "");
+        EXPECT_EQ(parsedResponse.reasoningTokenCount, 0);
+        EXPECT_EQ(parsedResponse.toolCalls.size(), 1);
```
dkalinowski: assert before accessing [0]. we already have enough of segfaults in unit tests!
please do that in all other cases and all other expects that then continue with undefined behavior in negative scenario

prepare_llm_models.sh
mzegla: It comes with openvino-tokenizers which is a dependency of optimum. 
If you have installed packages required by `export_model.py`, you already have `convert_tokenizer`

PR #3323, opened 2025-05-30
Phi4 response parser

mzegla:


dtrawins (approved):

src/llm/response_parsers/phi4_response_parser.hpp
```diff
@@ -0,0 +1,32 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+#pragma once
+
+#include <openvino/genai/tokenizer.hpp>
+#include <string>
```
dkalinowski: why do we need string here?

mzegla: we don't, it's a leftover, good catch

src/llm/response_parsers/phi4_response_parser.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    if (matchCount == 0) {
+        parsedResponse.content = decoded;
+    } else if (matchCount == 1) {
+        std::smatch match = *begin;
+        // Put everything, but functools[...] part into the response content
+        parsedResponse.content = decoded.substr(0, match.position()) +
+                                 decoded.substr(match.position() + match.length());
+
+        std::string toolsStr = match[1].str();
```
dkalinowski, line 55: this is second element of a match, while we validate if the count is == 1, how does it work?

mzegla: Second element is the first captured group of the match. First element is entire matched string, so:
`functools[{"name": [function name], "arguments": [function arguments as JSON]}, ...]`

Second element is the match for (functools\[**(.*?)**\]) - the inside of functools[...], so:
`"name": [function name], "arguments": [function arguments as JSON]}, ...`

Number of complete matches still needs to be one

mzegla:
Merging despite red status. Tests passed - it's the infra issue

PR #3327, opened 2025-06-02
Accept lack of content in OpenAI request

mzegla:
With more complex requests, we must relax our frontend validation to allow requests based on fields other than content get through.


src/llm/apis/openai_completions.cpp
```diff
@@ -310,8 +311,8 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages(std::optional<std::stri
             }
         }
         const auto& lastMessage = request.chatHistory.back();
-        if (lastMessage.find("content") == lastMessage.end() || lastMessage.find("role") == lastMessage.end()) {
-            return absl::InvalidArgumentError("Every message must have both 'content' and 'role' fields");
+        if (lastMessage.find("role") == lastMessage.end()) {
+            return absl::InvalidArgumentError("Every message must have 'role' field");
```
dtrawins, line 315: unit test without content in the message?

dkalinowski (approved):
pls add comments why we do expect "empty prompt error"

PR #3329, opened 2025-06-02
Logistic Regression

darksapien23151:
Logistic Regression model trained on iris dataset


iris_pipeline_project/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM openvino/model_server:latest
+
+USER root
+
+ENV LD_LIBRARY_PATH=/ovms/lib
+ENV PYTHONPATH=/ovms/lib/python
+
+RUN apt update && apt install -y python3-pip git \
+    build-essential python3-dev libatlas-base-dev
```
mzegla: Do we need those packages?
Since we only want to download a few python packages, they don't look necessary.

iris_pipeline_project/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+RUN pip3 install --break-system-packages numpy pandas scikit-learn
```
mzegla: Why using `--break-system-packages`?

iris_pipeline_project/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+
+COPY pipeline/graph.pbtxt /models/iris_pipeline/
+COPY pipeline/ovmsmodel.py /models/iris_pipeline/
+COPY model/model.pkl /models/iris_pipeline/
+COPY model_config.json /model_config.json
```
mzegla: Let's drop those lines. Eventually we would like to have a Dockerfile for multiple models, datasets etc. so users would build the image with all required packages and mount their data while launching the container.

iris_pipeline_project/convert.py
```diff
@@ -0,0 +1,24 @@
+from sklearn.datasets import load_iris
```
mzegla: This entire file should be wrapped in OvmsPythonModel with training done in `execute` method

iris_pipeline_project/model_config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "custom_node_library_config_list": [],
+  "graph_config_list": [
+    {
+      "name": "iris_pipeline",
+      "base_path": "/models/iris_pipeline",
+      "graph_path": "graph.pbtxt"
+    }
+  ]
+}
```
mzegla: This format is not correct. See:
https://github.com/openvinotoolkit/model_server/blob/main/docs/python_support/quickstart.md#step-4-prepare-server-configuration-file
for reference

iris_pipeline_project/ovmsmodel.py
```diff
@@ -0,0 +1,29 @@
+import pandas as pd
```
mzegla: This file is unnecessary. After exporting model to ONNX we can use native KServe API support.

iris_pipeline_project/pipeline/graph.pbtxt
```diff
@@ -0,0 +1,14 @@
...
+  calculator: "InferenceCalculator"
+  input_stream: "inference_input"
+  output_stream: "inference_output"
+  options: {
+    [type.googleapis.com/openvino.CalculatorOptions] {
+      model_name: "iris_pipeline"
+      signature_name: "serving_default"
+    }
+  }
+}
```
mzegla: This will not work. See:
https://github.com/openvinotoolkit/model_server/blob/main/docs/python_support/quickstart.md#step-3-prepare-graph-configuration-file

for reference

mzegla (commented):
Please also provide client sample to interact with deployed Python servable

PR #3330, opened 2025-06-02
Mkulakow/ov rerank calculator

michalkulakowski:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``


rasapala (approved):
LGTM

demos/common/export_models/export_model.py
```diff
@@ -491,6 +508,30 @@ def export_embeddings_model_ov(model_repository_path, source_model, model_name,
     print("Created graph {}".format(os.path.join(model_repository_path, model_name, 'graph.pbtxt')))
     add_servable_to_config(config_file_path, model_name + "_ov", os.path.relpath(os.path.join(model_repository_path, model_name, "ov"), os.path.dirname(config_file_path)))
 
+def export_rerank_model_ov(model_repository_path, source_model, model_name, precision, task_parameters, config_file_path, truncate=True):
```
rasapala: Add cli parameters: max_allowed_chunks, max_position_embeddings description.

src/BUILD
```diff
@@ -3316,3 +3317,25 @@ filegroup(
...
+            "//src:libovmsmodelversioning",
+            "//src:libovms_ovinferrequestsqueue",
+            "@mediapipe//mediapipe/framework:calculator_framework",
+            "//third_party:openvino",
+            "@com_github_tencent_rapidjson//:rapidjson",] + select({
+            "//conditions:default": ["//third_party:genai", "@llm_engine//:llm_engine"],
+            "//:not_genai_bin" : ["@llm_engine//:llm_engine"],
+    }),
+    visibility = ["//visibility:public"],
+    copts = COPTS_ADJUSTED,
```
dkalinowski, line 3343: missing includes. please check C PLUS channel

src/mediapipe_internal/mediapipegraphdefinition.cpp
```diff
@@ -141,6 +143,10 @@ Status MediapipeGraphDefinition::validate(ModelManager& manager) {
         SPDLOG_ERROR("Internal Error: MediaPipe definition is in unexpected state.");
         return StatusCode::INTERNAL_ERROR;
     }
+    if (!this->sidePacketMaps.rerankServableMap.empty()) {
```
dkalinowski: please group all empty checks in single sidePacketMaps.isEmpty() call, this is getting out of hand

src/rerank/rerank_calculator_ov.cc
```diff
@@ -0,0 +1,315 @@
...
+            std::fill(attention_mask_data, attention_mask_data + 1 + query_tokens.size() + 2 + pad_token_index + 1, int64_t(1));
+        }
+
+        return std::make_pair(input_ids, attention_mask);
+    }
+
+    std::vector<float> ComputeScoresUsingRerankModel(ov::Tensor input_ids, ov::Tensor attention_mask, const std::vector<size_t>& chunkMapping, size_t actual_batch_size) const {
+        ModelMetricReporter tmp(nullptr, nullptr, "example_pipeline_name", 1);
+        auto executingStreamIdGuard = std::make_shared<ExecutingStreamIdGuard>(rerank_session->getInferRequestsQueue(), tmp);
+        ov::InferRequest& inferRequest = executingStreamIdGuard->getInferRequest();
```
dkalinowski, line 235: what about the same fixes for embeddings calculator? regarding infer request scope

src/sidepacket_servable.cpp
```diff
@@ -81,24 +60,44 @@ EmbeddingsServable::EmbeddingsServable(const ::mediapipe::CalculatorGraphConfig:
                 if (modelConfig.HasMember("pad_token_id") && modelConfig["pad_token_id"].IsInt64()) {
                     pad_token = modelConfig["pad_token_id"].GetInt64();
                 }
+                if (modelConfig.HasMember("eos_token_id") && modelConfig["eos_token_id"].IsInt64()) {
+                    eos_token = modelConfig["eos_token_id"].GetInt64();
+                }
```
dkalinowski: macro?

src/sidepacket_servable.cpp
```diff
@@ -81,24 +60,44 @@ EmbeddingsServable::EmbeddingsServable(const ::mediapipe::CalculatorGraphConfig:
...
-    embeddings = std::make_shared<EmbeddingsModel>(parsedModelsPath.string(), target_device, embeddingsPoperties);
+
+    ov::Core core;
+    std::shared_ptr<ov::Model> m_model = core.read_model(parsedModelsPath / std::filesystem::path("openvino_model.xml"), {}, properties);
+    compiledModel = core.compile_model(m_model, targetDevice, properties);
+    auto& ovmsConfig = ovms::Config::instance();
+    uint32_t numberOfParallelInferRequests = 1;
+    if (ovmsConfig.nireq() > 0) {
+        // nireq is set globally for all models in ovms startup parameters
+        numberOfParallelInferRequests = ovmsConfig.nireq();
```
dkalinowski, line 92: :( how is nireq config.json thing being tracked?

src/sidepacket_servable.hpp
```diff
@@ -33,48 +33,46 @@
...
     }
-    int64_t& getPadToken() {
+    const int64_t& getPadToken() {
         return pad_token;
     }
-    std::optional<uint32_t>& getMaxModelLength() {
+    const int64_t& getEosToken() {
+        return eos_token;
+    }
+    const int64_t& getBosToken() {
```
dkalinowski: does it really speed up execution that we return const reference of a single integer? i think we can simply return copy of int64_t, it would be equal in terms of functionality, but cleaner

src/test/pythonnode_test.cpp
```diff
@@ -1003,7 +1003,7 @@ class MockedMediapipeGraphExecutorPy : public ovms::MediapipeGraphExecutor {
         const PythonNodeResourcesMap& pythonNodeResourcesMap,
         PythonBackend* pythonBackend,
         MediapipeServableMetricReporter* mediapipeServableMetricReporter) :
-        MediapipeGraphExecutor(name, version, config, inputTypes, outputTypes, inputNames, outputNames, pythonNodeResourcesMap, {}, {}, pythonBackend, mediapipeServableMetricReporter) {}
+        MediapipeGraphExecutor(name, version, config, inputTypes, outputTypes, inputNames, outputNames, pythonNodeResourcesMap, {}, {}, {}, pythonBackend, mediapipeServableMetricReporter) {}
```
dkalinowski, line 1006: when are we switching unit tests to use single constructor, which is used in the actual product, instead of testing different constructor which exists for the purpose of unit tests now?

dkalinowski: @atobiszei was the intention switch at some point, or keep 2 for tests and ovms binary?

atobiszei: Second constuctor (the one used here) calls the first one. But here if we modify those anyway I would use the first (with SidePacketMaps)

atobiszei: image
So we shouldn't have those {}, {}. {}. {} and expand them each time for new servableTypeMap

dkalinowski (approved):

PR #3335, opened 2025-06-03
Cvs 168735 fix config remove

rasapala:
### 🛠 Summary

CVS-168735 and CVS-168734

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -19,6 +19,7 @@
 #include <limits>
 #include <regex>
 #include <thread>
+#include <vector>
 
 #include "logging.hpp"
 #include "ovms_exit_codes.hpp"
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -85,6 +85,33 @@ bool Config::check_hostname_or_ip(const std::string& input) {
     }
 }
 
+bool Config::validateUserSettingsInConfigAddRemoveModel(const ModelsSettingsImpl& modelsSettings) {
+    std::vector<std::string> allowedUserSettings = {"model_name", "model_path"};
+    std::vector<std::string> usedButdisallowedUserSettings;
+    for (const std::string& userSetting : modelsSettings.userSetSingleModelArguments) {
+        bool isAllowed = false;
+        for (const std::string& allowedSetting : allowedUserSettings) {
+            if (userSetting == allowedSetting)
+                isAllowed = true;
+        }
+
+        if (!isAllowed)
+            usedButdisallowedUserSettings.push_back(userSetting);
+    }
+
+    if (!usedButdisallowedUserSettings.empty()) {
+        std::string arguments = "";
+        for (const std::string& userSetting : usedButdisallowedUserSettings) {
+            arguments += userSetting + ", ";
+        }
+        std::cerr << "Starting config.json add/remove model with unsupported model settings: " << arguments << std::endl;
+
+        return false;
+    }
+
+    return true;
+}
+
 bool Config::validate() {
     if (this->serverSettings.serverMode == HF_PULL_MODE) {
         if (!serverSettings.hfSettings.sourceModel.size()) {
@@ -211,6 +238,9 @@ bool Config::validate() {
             std::cerr << "Set model_path or model_repository_path and model_name with add_to_config, remove_from_config" << std::endl;
             return false;
         }
+
+        if (!this->validateUserSettingsInConfigAddRemoveModel(this->modelsSettings))
+            return false;
     }
 
     // check rest_workers value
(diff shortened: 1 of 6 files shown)


src/config.cpp
```diff
@@ -85,6 +86,33 @@ bool Config::check_hostname_or_ip(const std::string& input) {
     }
 }
 
+bool Config::validateUserSettingsInConfigAddRemoveModel(const ModelsSettingsImpl& modelsSettings) {
+    std::vector<std::string> allowedUserSettings = {"model_name", "model_path"};
```
dkalinowski: can be const and static

src/config.cpp
```diff
@@ -85,6 +86,33 @@ bool Config::check_hostname_or_ip(const std::string& input) {
...
+    std::vector<std::string> usedButdisallowedUserSettings;
+    for (const std::string& userSetting : modelsSettings.userSetSingleModelArguments) {
+        bool isAllowed = false;
+        for (const std::string& allowedSetting : allowedUserSettings) {
+            if (userSetting == allowedSetting)
+                isAllowed = true;
```
dkalinowski, line 96: can break earlier

can also use <algorithm> to and use std::find

src/test/config_export_test.cpp
```diff
@@ -68,6 +72,7 @@ const std::string expected2ModelsConfigContents = R"({
                 "base_path": "/model1/Path"
             }
         },
+        {},
```
dkalinowski: why is it expected? shouldnt we also remove entire object?

src/test/config_export_test.cpp
```diff
@@ -116,6 +121,42 @@ TEST_F(ConfigCreationTest, positiveRemoveOneModelToEmptyConfig) {
     ASSERT_EQ(expectedEmptyConfigContents, configContents) << configContents;
 }
 
+TEST_F(ConfigCreationTest, positiveRemoveOneModelToExistingConfig) {
+    std::string configContents = R"(
+{
+  "model_config_list": [
+    {
```
dkalinowski, line 161: more tests with deletion in the middle?
at the end?
remove non existing?

src/config.cpp
rasapala: We want to iterate and find all settings that do not match not only one.

src/test/config_export_test.cpp
rasapala: Ok I will add at the end, first and non existing check.

src/config.cpp
```diff
@@ -85,6 +86,33 @@ bool Config::check_hostname_or_ip(const std::string& input) {
...
+        if (!isAllowed)
+            usedButdisallowedUserSettings.push_back(userSetting);
+    }
+
+    if (!usedButdisallowedUserSettings.empty()) {
+        std::string arguments = "";
+        for (const std::string& userSetting : usedButdisallowedUserSettings) {
+            arguments += userSetting + ", ";
+        }
+        std::cerr << "Starting config.json add/remove model with unsupported model settings: " << arguments << std::endl;
```
dtrawins: what does it mean?

src/test/ovmsconfig_test.cpp
```diff
@@ -673,6 +673,21 @@ TEST_F(OvmsConfigDeathTest, modifyModelConfigEnableButMissingModelPath) {
...
+        "--model_name",
+        "name",
+        "--add_to_config",
+        "/config/path",
+        "--target_device",
+        "GPU",
+        "--model_path",
+        "/model/path"};
+    int arg_count = 9;
+    EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(OVMS_EX_USAGE), "Starting config.json add/remove model with unsupported model settings: target_device,");
```
dtrawins:
```suggestion
    EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(OVMS_EX_USAGE), "Adding or removing models from the configuration file, allows passing only model_name and model_path parameters. There were passed invalid parameter: target_device,");
```
I propose a bit clearer message.

dtrawins (approved):

src/config.cpp
```diff
@@ -85,6 +86,33 @@ bool Config::check_hostname_or_ip(const std::string& input) {
     }
 }
 
+bool Config::validateUserSettingsInConfigAddRemoveModel(const ModelsSettingsImpl& modelsSettings) {
+    const std::vector<std::string> allowedUserSettings = {"model_name", "model_path"};
+    std::vector<std::string> usedButdisallowedUserSettings;
```
dkalinowski:
```suggestion
    std::vector<std::string> usedButDisallowedUserSettings;
```

PR #3368, opened 2025-06-09
Fix global curl cleanup

michalkulakowski:
### 🛠 Summary

curl_global_cleanup() should be called only once in ovms

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/httpservermodule.cpp
+++ b/src/httpservermodule.cpp
@@ -80,6 +80,7 @@ void HTTPServerModule::shutdown() {
     drogonServer->terminate();
     drogonServer.reset();
 #endif
+    curl_global_cleanup();
     SPDLOG_INFO("Shutdown HTTP server");
     state = ModuleState::SHUTDOWN;
 }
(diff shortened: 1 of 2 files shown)


src/httpservermodule.cpp
```diff
@@ -80,6 +80,7 @@ void HTTPServerModule::shutdown() {
     drogonServer->terminate();
     drogonServer.reset();
 #endif
+    curl_global_cleanup();
```
dkalinowski, line 83: Recently there was automatic cleanup introduced: https://github.com/openvinotoolkit/model_server/commit/a02ea5e69e497582fa2ab26740614cec294b2920

Why do we need multiple cleanups?

michalkulakowski: As described in libcurl documentation [https://curl.se/libcurl/c/curl_easy_init.html], we need one global init and one global clean up, and also easy init and easy cleanup every time we are taking CURL handler. Right now we call global clean up per every request which is incorrect and thats why we need this change

atobiszei (approved):

dkalinowski (approved):

PR #3377, opened 2025-06-10
Optimized serving endpoint for classical ML models

darksapien23151:
### 🛠 Summary

GSOC Contribution for the project - Optimized serving endpoint for classical Machine Learning models.
Current status: Finished working on Logistic Regression and KMeans.

Mentors:
Zeglarski, Milosz
Trawinski, Dariusz

### DIFF
--- a/extras/iris_pipeline_project/README.md
+++ b/extras/iris_pipeline_project/README.md
@@ -6,6 +6,16 @@ At the moment, for demonstrating purposes of the established project structure, 
 
 ---
 
+## Instructions for preparing the data
+Run the command to download the Iris dataset, which is taken to be the hello-world dataset of classification datasets for simplicity purposes.
+Split and preprocess(label encoding, normalization and Cross validation) accordingly.
+
+```bash
+curl -o iris.csv https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data
+```
+
+---
+
 ## Step 1: Clone the Repository
 
 ```bash
@@ -38,11 +48,7 @@ docker build --no-cache -t prototype_iris .
 client/
   ├── client_inference.py
   └── client_train.py
-data_folder/
-  ├── iris_train.csv
-  └── iris_test.csv
 pipeline/
-  ├── __pycache__/
   ├── graph.pbtxt
   ├── model.py
   └── ovmsmodel.py
@@ -79,17 +85,10 @@ python client/client_inference.py infer iris_train_nolabel.csv  --target_column 
 ---
 
 For Enabling accelerator support:
-Manually set the ```bool - (use_ipex/use_oneDAL)``` in model.py file under "pipeline" directory to either True/False depending on the necessity.
-
-## Instructions for preparing the data
-Run the command to download the Iris dataset, which is taken to be the hello-world dataset of classification datasets.
-
-```bash
-curl -o iris.csv https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data
-```
+You can pass the ```bool - (use_ipex/use_oneDAL)``` in the "hyperparams.json"/ "kmeans_params.json" file as a key variable pair to either True/False depending on the necessity.
 
 
-Command-Line Usage
+Command-Line Usage:
 
 The training and inference client supports flexible options for both Logistic Regression and KMeans models.
 


extras/iris_pipeline_project/client/client_inference.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+import tritonclient.grpc as grpcclient
+import pandas as pd
+
+
+X_test = pd.read_csv("/home/harshitha/iris_pipeline_project/data/iris_test.csv")
```
mzegla: Do not use absolute path to your own filesystem

extras/iris_pipeline_project/client/client_inference.py
```diff
@@ -0,0 +1,25 @@
...
+X_test = X_test.values.astype(np.float32)
```
mzegla: This fails since `X_test` containts string in the last column. Please map class names to numbers.

extras/iris_pipeline_project/client/client_train.py
```diff
@@ -0,0 +1,26 @@
+import numpy as np
+import pandas as pd
+import tritonclient.grpc as grpcclient
+
+data = pd.read_csv("/home/harshitha/iris_pipeline_project/data/iris_train.csv")
```
mzegla: Do not use absolute path to your own filesystem

extras/iris_pipeline_project/client/client_train.py
```diff
@@ -0,0 +1,26 @@
+import numpy as np
```
mzegla: There is no entry in `model_config.json`, so there's no way to run it.

extras/iris_pipeline_project/data/Iris (copy).csv
```diff
@@ -0,0 +1,151 @@
+Id,SepalLengthCm,SepalWidthCm,PetalLengthCm,PetalWidthCm,Species
```
mzegla: Please remove entire data catalog. Provide instructions or script for data download, split etc.

extras/iris_pipeline_project/model/iris_logreg/1/model.onnx
mzegla, line 1: We don't want to add trained model file here as well. The assumption is that trained model is on the host (or mounted if running in a container).

extras/iris_pipeline_project/pipeline/graph.pbtxt
```diff
@@ -0,0 +1,31 @@
+input_stream: "pipeline_input"
```
mzegla: We don't need to deploy inference pipeline via mediapipe. You already have it deployed for inference in model_config.json using native OV runtime. 

We need mediapipe only to use Python, so the configuration here should describe training endpoint. Refer to:
https://github.com/openvinotoolkit/model_server/blob/main/docs/python_support/quickstart.md#step-3-prepare-graph-configuration-file

extras/iris_pipeline_project/model_config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "model_config_list": [
```
mzegla, line 2: That's a model deployment for inference, we also need to specify training endpoint here (not under model_config_list, but in mediapipe_config_list as done here: [main/docs/python_support/quickstart.md#step-4-prepare-server-configuration-file](https://github.com/openvinotoolkit/model_server/blob/main/docs/python_support/quickstart.md#step-4-prepare-server-configuration-file)

extras/iris_pipeline_project/model_config.json
```diff
@@ -0,0 +1,10 @@
...
+    {
+      "config": {
+        "name": "iris_pipeline",
```
mzegla:
```suggestion
        "name": "iris_logreg",
```

extras/iris_pipeline_project/model_config.json
```diff
@@ -0,0 +1,10 @@
...
+        "base_path": "/models/iris_pipeline"
```
mzegla:
```suggestion
        "base_path": "/models/iris_logreg"
```

extras/iris_pipeline_project/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+USER root
+
+ENV LD_LIBRARY_PATH=/ovms/lib
+ENV PYTHONPATH=/ovms/lib/python
+
+COPY model/iris_logreg /models/iris_logreg
+COPY pipeline/graph.pbtxt /models/iris_pipeline/
+COPY pipeline/ovmsmodel.py /models/iris_pipeline/
+COPY model_config.json /model_config.json
```
mzegla: Let's not copy model and configuration for inference. I would only include training part in the image.

extras/iris_pipeline_project/pipeline/ovmsmodel.py
```diff
@@ -0,0 +1,41 @@
...
+import onnx
+
+class OvmsPythonModel:
+    def initialize(self, kwargs):
+        print("Training handler initialized.")
+
+    def execute(self, inputs):
+        print("Starting training...")
+        input_tensor = inputs[0]
+        input_data = input_tensor.as_numpy()
```
mzegla: I don't think you can do it. `inputs` is a list of `pyovms.Tensor` objects and they don't expose `as_numpy` function. You can try: 
```suggestion
        input_data = np.array(input_tensor)
```

extras/iris_pipeline_project/pipeline/ovmsmodel.py
```diff
@@ -0,0 +1,41 @@
...
+        output_dir = "/model/iris_logreg/1"
+        os.makedirs(output_dir, exist_ok=True)
+
+        onx = to_onnx(model, X[:1], target_opset=12)
+        output_path = os.path.join(output_dir, "logreg_model.onnx")
+        with open(output_path, "wb") as f:
+            f.write(onx.SerializeToString())
+
+        print(f"Model saved to {output_path}")
+        return [Tensor.from_numpy(np.array([True]))]
```
mzegla: `pyovms.Tensor` does not have such method. You can create it with a constructor that also takes tensor name as an argument (it has to be linked with output name specified in `graph.pbxtx`)
```suggestion
        return [Tensor("output", (np.array([True]))]
```

extras/iris_pipeline_project/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+ENV LD_LIBRARY_PATH=/ovms/lib
+ENV PYTHONPATH=/ovms/lib/python
+
+RUN apt-get update && apt-get install -y python3-pip
+RUN pip3 install --break-system-packages pandas numpy scikit-learn joblib skl2onnx onnx onnxruntime
+
+
+WORKDIR /workspace
+ENTRYPOINT ["/ovms/bin/ovms", "--config_path", "/model_config.json", "--grpc_port", "9000"]
```
mzegla:
```suggestion
ENTRYPOINT ["/ovms/bin/ovms"]
```
Let's give user flexibility to specify these parameters in the startup command

extras/iris_pipeline_project/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+This repository demonstrates how to use OpenVINO Model Server (OVMS) with a custom Mediapipe pipeline for the Iris dataset, including both model training and inference through a Python client.
+
+---
+
+## Step 1: Clone the Repository
+
+```bash
+git clone https://github.com/openvinotoolkit/model_server.git
+cd [need to fix]
```
mzegla: Please apply the fix

extras/iris_pipeline_project/.gitignore
```diff
@@ -0,0 +1,4 @@
+data/
+model/
+client/
```
mzegla: Won't we ignore changes in client scripts with this line?

extras/iris_pipeline_project/README.md
```diff
@@ -0,0 +1,115 @@
...
+  ├── client_inference.py
+  └── client_train.py
+model/
+  └── iris_logreg/1/model.onnx
+pipeline/
+  ├── __pycache__/
+  ├── graph.pbtxt
+  └── ovmsmodel.py
+Dockerfile
+model_config.json
```
mzegla: We should not include model at this point (if user goes through that instruction step by step, there will be no model yet).
Also please remove model from the PR.

extras/iris_pipeline_project/README.md
```diff
@@ -0,0 +1,115 @@
...
+```bash
+python client/client_train.py train data/iris_train.csv
+```
+
+### 4.2. Inference
+
+```bash
+python client/client_train.py infer data/iris_test.csv
+# OR
+python client/client_inference.py infer data/iris_test.csv
```
mzegla: Let's drop the `OR` part, make it straightforward. Leave only `client_inference.py` usage here.

extras/iris_pipeline_project/README.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+---
+
+## Step 4: Run Training and Inference
+
+### 4.1. Training
+
+```bash
+python client/client_train.py train data/iris_train.csv
```
mzegla: Can you provide simple script or command line steps to get required datasets (for training and testing)?

extras/iris_pipeline_project/client/client_inference.py
```diff
@@ -0,0 +1,24 @@
+import numpy as np
+import tritonclient.grpc as grpcclient
+import pandas as pd
+import json
+
+df = pd.read_csv("data/iris_test_stripped.csv")
```
mzegla: User might not have this file. Please provide instruction on how to get it first. Also this path could be parametrized, just like in client_train.py

extras/iris_pipeline_project/pipeline/ovmsmodel.py
```diff
@@ -0,0 +1,141 @@
...
+            input_data = input_tensor.data
+            if isinstance(input_data, memoryview):
+                inp_bytes = bytes(input_data)
+            elif isinstance(input_data, (bytes, bytearray)):
+                inp_bytes = input_data
+            elif isinstance(input_data, np.ndarray):
+                if input_data.dtype == object:
+                    inp_bytes = input_data[0]
+                else:
+                    inp_bytes = bytes(input_data)
```
mzegla: `input_tensor.data` will always be a memoryview:
https://github.com/openvinotoolkit/model_server/blob/main/src/python/binding/ovms_py_tensor.cpp#L57
 
I think we can safely get rid of that condition. That object is provided by OVMS, so it must be okay or it would throw error earlier in the model server logic.

extras/iris_pipeline_project/pipeline/ovmsmodel.py
```diff
@@ -0,0 +1,141 @@
...
+            else:
+                raise TypeError(f"Unsupported type for input_data: {type(input_data)}")
+
+            print("input_data preview:", inp_bytes[:40], file=sys.stderr)
+            if inp_bytes[:1] != b'{' and inp_bytes.find(b'{') > 0:
+                first_brace = inp_bytes.find(b'{')
```
mzegla: I can see you run `find` twice. I suppose you can run it once before the check and store the value.
Also I'm not sure if you need first part of the condition: `inp_bytes[:1] != b'{'`. If you search entire `inp_bytes` with `find`, there's no need for a separate check I guess.

extras/iris_pipeline_project/pipeline/ovmsmodel.py
darksapien23151: I wasn’t aware of that earlier, thanks for the clarification!

extras/iris_pipeline_project/.gitignore
darksapien23151: from my knowledge of it, it just wont track the changes to store in version control, rest the code stays the same to be accessed(read/write).
also, should I let the client/ to be there on the repo visible?

extras/iris_pipeline_project/README.md
darksapien23151: sure

extras/iris_pipeline_project/client/client_inference.py
darksapien23151: Understood, will do!

extras/iris_pipeline_project/README.md
darksapien23151: should I link the path of the final directory that this is PR is to be merged into or in the same demo directory?

extras/iris_pipeline_project/README.md
darksapien23151: sure.

extras/iris_pipeline_project/README.md
darksapien23151: Alright!

extras/iris_pipeline_project/README.md
mzegla: Let's assume we are in the working directory so we can put relative path to the catalog
as of now: `cd model_server/extras/iris_pipeline_project`

extras/iris_pipeline_project/.gitignore
mzegla: client directory with the scripts should be visible

extras/iris_pipeline_project/data_preprocess.py
```diff
@@ -0,0 +1,17 @@
+import pandas as pd
+from sklearn.utils import shuffle
+import os
+
+os.makedirs("data_folder", exist_ok=True)
```
mzegla: Let's not implicitly create directories in user filesystem. Instead we could accept script parameter with path to the catalog where we would save the data.

extras/iris_pipeline_project/.gitignore
```diff
@@ -0,0 +1,5 @@
+data/
+model/
+
+
+
```
mzegla: No need for so many newlines

extras/iris_pipeline_project/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+USER root
+
+ENV LD_LIBRARY_PATH=/ovms/lib
+ENV PYTHONPATH=/ovms/lib/python
+
+RUN apt-get update && apt-get install -y python3-pip
+RUN pip install --break-system-packages pandas numpy scikit-learn joblib skl2onnx onnx onnxruntime
+
+
+WORKDIR /workspace
```
mzegla:
```suggestion
```
That's obsolete

extras/iris_pipeline_project/.gitignore
darksapien23151: resolved

extras/iris_pipeline_project/pipeline/ovmsmodel.py
darksapien23151: thank you for pointing out, fixed the code

extras/iris_pipeline_project/data_preprocess.py
darksapien23151: Ahh yeah that makes total sense. Didn’t consider cluttering the user’s fs earlier. Just updated it to take an output path param and tested it. Looks good now!

extras/iris_pipeline_project/model_config.json
darksapien23151: fixed

extras/iris_pipeline_project/README.md
```diff
@@ -0,0 +1,126 @@
...
+### 2.1. Build the Docker Image
+
+```bash
+docker build --no-cache -t iris_logisticreg_ovms .
+```
+
+### 2.2. Run the OVMS Container
+
+```bash
+docker run --rm -it -v "$PWD:/workspace"   -p 9000:9000 -p 8000:8000  iris_logisticreg_ovms --config_path /workspace/model_config.json   --port 9000 --rest_port 8000
```
mzegla:
```suggestion
docker run --rm -it -v "$PWD:/workspace"   -p 9000:9000 iris_logisticreg_ovms --config_path /workspace/model_config.json   --port 9000
```
We can drop rest port usage as it is not used in this demo.

extras/iris_pipeline_project/README.md
```diff
@@ -0,0 +1,126 @@
...
+Run the command to download the Iris dataset, which is taken to be the hello-world dataset of classification datasets.
+
+```bash
+curl -o iris.csv https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data
+```
+Run the following file to prepare the data and split it into data for training and for inferencing.
+
+```bash
+python data_preprocess.py <path to output_dir>
+```
```
mzegla: Please make it a separate step before step 4, so user can follow the guide step-by-step

extras/iris_pipeline_project/data_folder/iris_test.csv
```diff
@@ -0,0 +1,2 @@
+SepalLength,SepalWidth,PetalLength,PetalWidth
```
mzegla: I guess we don't need that in the repo since we have an instruction on preparing the data.

extras/iris_pipeline_project/data_folder/iris_train.csv
```diff
@@ -0,0 +1,508 @@
+SepalLength,SepalWidth,PetalLength,PetalWidth,Species
```
mzegla: I guess we don't need that in the repo since we have an instruction on preparing the data.

extras/iris_pipeline_project/data_preprocess.py
```diff
@@ -0,0 +1,26 @@
...
+import os
+import sys
+
+if len(sys.argv) < 2:
+    print("Usage: python datapreprocess.py <output_directory>")
+    sys.exit(1)
+
+output_dir = sys.argv[1]
+
+#os.makedirs(output_dir, exist_ok=True)
```
mzegla, line 12: let's remove the comment

extras/iris_pipeline_project/README.md
```diff
@@ -0,0 +1,182 @@
...
+python client/client_inference.py infer data_folder/iris_test.csv --target_column Species  --model_class LogisticRegressionTorch
+
+python client/client_inference.py infer iris_train_nolabel.csv  --target_column Species --model_class KMeansSkLearn
+
+```
+
+---
+
+For Enabling accelerator support:
+Manually set the ```bool - (use_ipex/use_oneDAL)``` in model.py file under "pipeline" directory to either True/False depending on the necessity.
```
mzegla: We needed that for benchmarks.In the guide we should not require user to modify the code. 
Either remove it completely and use acceleration by default or pass it in hyperparameters JSON.

extras/iris_pipeline_project/pipeline/ovmsmodel.py
```diff
@@ -0,0 +1,132 @@
...
+            y = payload.get("y", None)
+            params = payload.get("params", {})
+            model_class_name = payload.get("model_class", "LogisticRegressionTorch")
+
+            if model_class_name not in AVAILABLE_MODEL_CLASSES:
+                raise ValueError(f"Unknown model: {model_class_name}")
+
+            model_obj = AVAILABLE_MODEL_CLASSES[model_class_name]()
+
+            if model_class_name == "KMeansSkLearn":
```
mzegla: I don't like this conditional logic. `execute` method should be generic and wrap all possible implementations. Having any model specific code goes against this idea. Please move those implementation specific code blocks to its classes. 
You define the interface and convention. You can make `fit` or `infer` method return not only results, but also metrics that can be used directly to create pyovms.Tensor object like:
```
trained_model, training_metrics = model_obj.fit(X, y, params))
...
return [Tensor("pipeline_output", training_metrics)]
```

PR #3379, opened 2025-06-10
Downgrade opencl headers to 2.* version for RedHat source build

mzegla:
Addresses issue with building OpenVINO from source on UBI 9.5

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -254,17 +254,19 @@ spell: venv-style
 
 $(ACTIVATE):
 	@echo "Updating virtualenv dependencies in: $(VIRTUALENV_DIR)..."
+	@python3 -m pip install virtualenv
 	@test -d $(VIRTUALENV_DIR) || $(VIRTUALENV_EXE) $(VIRTUALENV_DIR)
 	@. $(ACTIVATE); pip3 install --upgrade pip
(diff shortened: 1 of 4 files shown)


Dockerfile.redhat
```diff
@@ -101,7 +101,7 @@ ARG JOBS=40
 ARG VERBOSE_LOGS=OFF
 
 # hadolint ignore=DL3041
-RUN dnf install -y https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/opencl-headers-3.0-6.20201007gitd65bcc5.el9.noarch.rpm && \
+RUN dnf install -y https://rpmfind.net/linux/almalinux/8.10/PowerTools/x86_64/os/Packages/opencl-headers-2.2-1.20180306gite986688.el8.noarch.rpm && \
```
rasapala, line 113: This will disable ocl 3.0 features for the user.

mzegla: What features that user could use in this context you have in mind?

dtrawins:
was it tested in runtime?

ngrozae (approved):

stevegrubb:
I can confirm downgrading opencl allows it to finish a build. Without this, the build fails after 5 minutes.

rasapala (approved):

PR #3431, opened 2025-06-23
Phi4 response parser - drop regex matching

dtrawins:
### 🛠 Summary

CVS-169389
Incorrect response parsing in phi4 model with array in the arguments

### 🧪 Checklist

- [ ] Unit tests added.
``


src/test/llm/response_parsers/phi4_response_parser_test.cpp
```diff
@@ -125,3 +125,20 @@ TEST_F(Phi4ResponseParserTest, ParseToolCallOutputWithMultipleFunctoolsThrows) {
     },
         std::runtime_error);
 }
+
+TEST_F(Phi4ResponseParserTest, ParseToolCallOutputWithArrayArguments) {
```
mzegla, line 131: Please remove commented line and set this test to skip or expected fail so we can merge it and enable in the PR with the fix

dtrawins: @mzegla  can you takeover this PR and apply the fix to make this test work?

PR #3442, opened 2025-06-25
Refactor unary serialization for LLM

mzegla:
This change lays the foundation for tool calls streaming as it makes it possible to maintain state between serialization calls that will be needed to properly handle streaming for tool calls.

It also refactors serialization to split data extraction and JSON building logic.


src/llm/apis/openai_completions.cpp
```diff
@@ -688,222 +689,137 @@ absl::Status OpenAIChatCompletionsHandler::parseRequest(std::optional<uint32_t>
...
+            finishReason = "stop";
             break;
         case ov::genai::GenerationFinishReason::LENGTH:
-            writer.String("length");
+            finishReason = "length";
             break;
         default:
-            writer.Null();
+            finishReason = "unknown";
+            SPDLOG_LOGGER_WARN(llm_calculator_logger, "Unknown finish reason: {}", static_cast<int>(generationOutput.finish_reason));
```
dkalinowski: Warn is visible when INFO (right?), are you sure we want to bloat the logs?

src/llm/apis/openai_json_response.hpp
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    void startArray(const std::string& name = "") {
+        if (!name.empty()) {
+            writer.String(name.c_str());
+        }
+        writer.StartArray();  // [
+    }
+
+    void endArray() {
```
dkalinowski: do we really need that wrapper? maybe inheriting would be enough not to duplicate code? and then add required overrides and custom functions?

mzegla: I added those generic methods to keep it all contained conceptually. I wanted the user to use only `OpenAiJsonResponse` to build JSON response so we call methods of just one object in openai_completions file and at some point maybe keep rapidjson dependency in modules that only interact with JSON.

dtrawins (approved):

dkalinowski (approved):

PR #3447, opened 2025-06-27
Support V3/models endpoint

michalkulakowski:
### 🛠 Summary

CVS-169848

### 🧪 Checklist

- [ ] Unit tests added.
``


dtrawins:
add tests for the following scenarios:
v3/models
v3/models/{model}
v3/v1/models
v3/models/{model}
Compare with vllm if the name should be url encoded

In tests there could be used dummy models - no need to load llms...

src/test/listmodelsendpoint_test.cpp
```diff
@@ -0,0 +1,171 @@
...
+    ASSERT_EQ(d["data"].Size(), 1);
+    ASSERT_EQ(d["data"][0]["object"], "model");
+    ASSERT_EQ(d["data"][0]["id"], "add");
+    ASSERT_TRUE(d["data"][0]["created"].IsInt());
+    ASSERT_EQ(d["data"][0]["owned_by"], "OVMS");
+}
+
+TEST_F(ListModelsEndpointTest, simplePositiveRetrieveGraph) {
+    std::string requestBody = "";
+    std::string endpoint = listModelsEndpoint + "/my_graph";
```
dtrawins: what happens when the model name includes slash? Can you add such test and compare with vLLM?
does this name have to be url encoded or slashes can be be used?

src/http_rest_api_handler.cpp
```diff
@@ -113,6 +114,10 @@ const std::string HttpRestApiHandler::kfs_serverliveRegexExp =
 const std::string HttpRestApiHandler::kfs_servermetadataRegexExp =
     R"(/v2)";
 
+const std::string HttpRestApiHandler::v3_ListModelsRegexExp =
+    R"(/v3/(v1/)?models)";
+const std::string HttpRestApiHandler::v3_RetrieveModelRegexExp =
+    R"(/v3/(v1/)?models/([^:]+))";
```
dtrawins: why testing for semicolon?

dtrawins (approved):

src/test/listmodelsendpoint_test.cpp
```diff
@@ -0,0 +1,187 @@
+//*****************************************************************************
+// Copyright 2024 Intel Corporation
```
rasapala: 2025

rasapala (approved):

PR #3456, opened 2025-06-30
Introduce chat_template_kwargs

mzegla:
### 🛠 Summary

In python configuration, check if request contains `chat_template_kwargs` object. If it does expand it and pass to template engine for rendering.
It makes it possible to disable reasoning, but can be used to pass any other value that template could make use of.


mzegla:
Cherry picked version in: https://github.com/openvinotoolkit/model_server/pull/3459

PR #3458, opened 2025-06-30
Fix and tests for HF_ENDPOINT without https added automatically.

rasapala:
### 🛠 Summary

JIRA CVS-169935

### 🧪 Checklist

- [ ] Unit tests added.
``


src/pull_module/libgit2.cpp
```diff
@@ -173,27 +173,42 @@ Libgt2InitGuard::~Libgt2InitGuard() {
     git_libgit2_shutdown();
 }
 
+const std::string HF_PROTOCOL_SEPARATOR = "://";
```
mzegla: I don't think it's hugging face specific right? Can we drop HF prefix then?

src/pull_module/libgit2.cpp
```diff
@@ -173,27 +173,42 @@ Libgt2InitGuard::~Libgt2InitGuard() {
...
     } else {
         SPDLOG_DEBUG("HF_TOKEN environment variable not set");
+        passRepoUrl += this->hfEndpoint + this->sourceModel;
+        return passRepoUrl;
     }
 
-    passRepoUrl += this->hfEndpoint + this->sourceModel;
+    std::string outputWithPass = "";
+    size_t match = this->hfEndpoint.find(HF_PROTOCOL_SEPARATOR);
+    std::cout << "MATCH: " << match << std::endl;
```
mzegla: drop cout

src/pull_module/libgit2.cpp
```diff
@@ -173,27 +173,42 @@ Libgt2InitGuard::~Libgt2InitGuard() {
...
+    if (match != std::string::npos) {
+        // https://huggingface.co
+        // protocol[match]//address
+        std::string protocol = this->hfEndpoint.substr(0, match);
+        std::string address = this->hfEndpoint.substr(match + 3, this->hfEndpoint.size());
```
mzegla:
```suggestion
        std::string address = this->hfEndpoint.substr(match + HF_PROTOCOL_SEPARATOR.size());
```

src/pull_module/libgit2.cpp
```diff
@@ -173,27 +173,42 @@ Libgt2InitGuard::~Libgt2InitGuard() {
...
 
 std::string HfDownloader::GetRepositoryUrlWithPassword() {
-    std::string passRepoUrl = "https://";
+    std::string passRepoUrl = "";
     if (this->hfToken != "") {
         passRepoUrl += this->hfToken + ":" + this->hfToken + "@";
     } else {
         SPDLOG_DEBUG("HF_TOKEN environment variable not set");
+        passRepoUrl += this->hfEndpoint + this->sourceModel;
+        return passRepoUrl;
```
mzegla:
```suggestion
        return this->hfEndpoint + this->sourceModel;
```

src/pull_module/libgit2.cpp
```diff
@@ -173,27 +173,42 @@ Libgt2InitGuard::~Libgt2InitGuard() {
...
+
 bool HfDownloader::CheckIfProxySet() {
     if (this->httpProxy != "")
         return true;
     return false;
 }
 
 std::string HfDownloader::GetRepositoryUrlWithPassword() {
-    std::string passRepoUrl = "https://";
+    std::string passRepoUrl = "";
```
mzegla: Let's rename it to `repoPass` or something like that. 
I think using this variable in a path where it's a valid URL is not necessary, so we can be more specific about what it contains.

mzegla (approved):

PR #3460, opened 2025-06-30
Partial JSON parser

mzegla:
### 🛠 Summary

In order to stream structured output in JSON format (like tool calls) we need to be able to find deltas between structures as model produces tokens. For example for:

``` 
Model output #1
<tool_call>{"name": "get_weather", "arguments": "\"location\": \"New
Model output #2
<tool_call>{"name": "get_weather", "arguments": "\"location\": \"New York
```

We need to produce response similar to:
```
data: {"id":"abcd1234","object":"chat.completion.chunk","created":1750764695,"model":"Qwen/Qwen3-8B","choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"arguments":" York"}}]},"logprobs":null,"finish_reason":null}]}
```

This PR introduces partial JSON parser that closes incomplete JSONs so they can be compared and we could extract delta in the next phases.

```
Incomplete object
{"name": "get_weather", "arguments": "\"location\": \"New
is changed to:
{"name": "get_weather", "arguments": "\"location\": \"New"}
```

### DIFF
--- a/src/llm/response_parsers/partial_json_parser.hpp
+++ b/src/llm/response_parsers/partial_json_parser.hpp
@@ -1,0 +1,164 @@
+//*****************************************************************************
+// Copyright 2025 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include <string>
+#include <vector>
+#include <iostream>
+
+#pragma warning(push)
+#pragma warning(disable : 6313)
+#include <rapidjson/document.h>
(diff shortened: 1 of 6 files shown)


src/llm/response_parsers/partial_json_parser.hpp
```diff
@@ -0,0 +1,164 @@
...
+            In such cases we need to drop the incomplete part, get back to the last separator position and remove 
+            all the content after it, including the last comma if it exists.
+        */
+        if (lastSeparatorPos != std::string::npos && lastSeparatorPos < closedInput.size()) {
+            closedInput.erase(lastSeparatorPos);
+        }
+    }
+    
+    // Close any unclosed objects/arrays/strings in reverse order
+    for (auto it = openCloseStack.rbegin(); it != openCloseStack.rend(); ++it) {
```
dkalinowski: looks like `::partialParseAndCompleteToJson`

src/llm/response_parsers/partial_json_parser.hpp
```diff
@@ -0,0 +1,164 @@
...
+    AWAITING_VALUE,
+    PROCESSING_VALUE,
+    PROCESSING_STRING,
+    PROCESSING_OBJECT,
+    PROCESSING_ARRAY,
+    AWAITING_ARRAY_ELEMENT,
+    END
+};
+
+rapidjson::Document partialParseToJson(const std::string& input) {
```
dkalinowski: move to cpp

Issue #3417, opened 2025-06-17
Add CORS support


farshidtz:
**Describe the bug**
Because OVMS does not support [CORS](https://developer.mozilla.org/en-US/docs/Glossary/CORS), it is impossible to use it directly from a web browser via a Web client such as [OpenWebUI](https://openwebui.com/). 

This is a missing OVMS feature seen as a bug from the client's perspective. Feel free to remove the bug label.

The web browser denies the requests when the server fails to respond to the [preflight request](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request):
`Access to fetch at 'http://localhost:8081/v3/chat/completions' from origin 'http://localhost:8080' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.`

In the above case, the Web UI at origin `http://localhost:8080` is denied access to OVMS at `http://localhost:8081` by the browser.

Moreover, the responses don't include any of the [CORS headers](https://developer.mozilla.org/en-US/docs/Glossary/CORS#cors_headers).

**To Reproduce**
1. Run OpenWebUI
2. Configure a direct connection. Use `http://localhost:8081/v3` URL and model name (e.g. `DeepSeek-R1-Distill-Qwen-7B-ov-int4` as Model ID
3. Start a chat, select the model, then send a prompt.

![Image](https://github.com/user-attachments/assets/415247b8-4683-4533-a048-cc95677e1e80)

![Image](https://github.com/user-attachments/assets/8a6df13c-3ce5-4fa7-ac54-7bb9017b2df0)

**Expected behavior**
The server should respond to the OPTIONS request and provide appropriate CORS headers in all responses. This could be made configurable via a command line argument.

**Logs**
```
[2025-06-17 15:20:39.052][77509][serving][debug][drogon_http_server.cpp:77] Request URI /v3/chat/completions dispatched to streaming thread pool
[2025-06-17 15:20:39.052][77509][serving][debug][http_server.cpp:192] REST request /v3/chat/completions
[2025-06-17 15:20:39.052][77509][serving][debug][http_server.cpp:200] Processing HTTP request: OPTIONS /v3/chat/completions body: 0 bytes
[...]

PR #3412, opened 2025-06-17
add CORS headers to responses

Huanli-Gong:
### Details:
* add CORS headers to handle cross-origin request.

 
### Tickets:
* Closes [Add CORS support #3417](https://github.com/openvinotoolkit/model_server/issues/3417)

PR #3462, opened 2025-07-01
[GSoC] CORS headers in response 

dkalinowski:
### 🛠 Summary

Following VLLM params: https://github.com/vllm-project/vllm/blob/ed70f3c64f684750edea087e286cbf264e7cc3f3/vllm/entrypoints/openai/cli_args.py#L100-L114

This is related to: https://github.com/openvinotoolkit/model_server/pull/3412
https://github.com/openvinotoolkit/model_server/issues/3417

This allows Open WebUI integration.
CVS-169897

### 🧪 Checklist

- [ ] Unit tests added.

### DIFF
--- a/src/cli_parser.cpp
+++ b/src/cli_parser.cpp
@@ -131,7 +131,23 @@ void CLIParser::parse(int argc, char** argv) {
             ("allowed_local_media_path",
                 "Path to directory that contains multimedia files that can be used as input for LLMs.",
                 cxxopts::value<std::string>(),
-                "ALLOWED_LOCAL_MEDIA_PATH");
+                "ALLOWED_LOCAL_MEDIA_PATH")
+            ("allow_credentials",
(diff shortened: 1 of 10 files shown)


src/cli_parser.cpp
```diff
@@ -394,6 +410,11 @@ void CLIParser::prepareServer(ServerSettingsImpl& serverSettings) {
     if (result->count("trace_path"))
         serverSettings.tracePath = result->operator[]("trace_path").as<std::string>();
 #endif
+
+    serverSettings.allowCredentials = result->operator[]("allow_credentials").as<bool>();
```
rasapala, line 414: Should we validate that only if allowCredentials  is set then other options are allowed ? Or is it not dependent ?

dkalinowski: it is not dependent, we can enable/disable any of those

PR #3471, opened 2025-07-03
[ImageGen] Document NPU usage, Static Shape, Multi Device

dkalinowski:
### 🛠 Summary

CVS-168932
CVS-169867

### 🧪 Checklist


demos/image_generation/README.md
```diff
@@ -109,6 +109,53 @@ ovms --rest_port 8000 ^
...
+> **NOTE:** The NPU device requires the pipeline to be reshaped to static shape, this is why the `--resolution` parameter is used to define the input resolution.
+
+It can be applied using the commands below:
+```bash
+mkdir -p models
+
+docker run -d --rm -p 8000:8000 -v $(pwd)/models:/models/:rw \
+  --user $(id -u):$(id -g) --device /dev/accel --group-add=$(stat -c "%g" /dev/dri/render* | head -n 1) \
+  -e http_proxy=$http_proxy -e https_proxy=$https_proxy -e no_proxy=$no_proxy \
+  openvino/model_server:2025.2-gpu \
```
mzegla: will this work with this tag in public image?

demos/image_generation/README.md
```diff
@@ -233,6 +297,42 @@ ovms --rest_port 8000 ^
 
 ::::
 
+**NPU**  
+
+This feature will be available in 2025.3 and later releases.
```
mzegla: Not sure if that's how we should put it. Looks like we are describing something that is not available, but planned. Maybe we could rephrase that to say that in order to use NPU / mixed devices for image generation user has to build from main?

demos/image_generation/README.md
```diff
@@ -233,6 +297,42 @@ ovms --rest_port 8000 ^
...
+    --model_name OpenVINO/FLUX.1-schnell-int4-ov \
+    --model_path /models/black-forest-labs/FLUX.1-schnell
+```
+
+:::
+
+:::{tab-item} Bare metal (Windows)
+:sync: bare-metal
+
+Depending on how you prepared models in the first step of this demo, they are deployed to either CPU or GPU (it's defined in `config.json`). If you run on GPU make sure to have appropriate drivers installed, so the device is accessible for the model server.
```
mzegla: NPU not mentioned here. Looks like GPU section copy-paste

demos/image_generation/README.md
dkalinowski: it wont. i added note to build from src

demos/image_generation/README.md
dkalinowski: i added note to build from src

demos/image_generation/README.md
dkalinowski: good catch, fixed!

atobiszei (approved):

demos/image_generation/README.md
```diff
@@ -117,17 +117,21 @@ Image generation endpoints consist of 3 steps: text encoding, denoising and vae
...
 It can be applied using the commands below:
 ```bash
 mkdir -p models
 
 docker run -d --rm -p 8000:8000 -v $(pwd)/models:/models/:rw \
-  --user $(id -u):$(id -g) --device /dev/accel --group-add=$(stat -c "%g" /dev/dri/render* | head -n 1) \
+  --user $(id -u):$(id -g) --device /dev/accel --device /dev/dri --group-add=$(stat -c "%g" /dev/dri/render* | head -n 1) \
   -e http_proxy=$http_proxy -e https_proxy=$https_proxy -e no_proxy=$no_proxy \
-  openvino/model_server:2025.2-gpu \
+  openvino/model_server:2025.3-gpu \
```
mzegla: Not sure if we should use this tag anyway. Until it's not released, maybe we could use the default tag that model server image has when built with `make docker_build`. What do you think?

dkalinowski: good idea, done

mzegla (approved):

PR #3506, opened 2025-07-15
Fix fuzz build

rasapala:
### 🛠 Summary

JIRA CVS-168977
Add libgit2.so to fuzz build dependency - libovms_shared.
### 🧪 Checklist

- [ ] Unit tests added.
``


src/BUILD
```diff
@@ -134,6 +134,7 @@ cc_shared_library(
     features = [],
     roots = [
         "//src:ovms_lib",
+        "@llm_engine//:llm_engine", "@libgit2_engine//:libgit2_engine",
```
mzegla: Why is it needed here now and was not needed before?

rasapala: With FUZZER_BUILD=1 we only build this target. The shared bazel library will not build nor add this dependencies so it will not be available in bazel-out directory.

Dockerfile.ubuntu
```diff
@@ -355,7 +355,9 @@ RUN if [ "$FUZZER_BUILD" == "0" ]; then bazel build ${CAPI_FLAGS} --jobs $JOBS /
 COPY MakefileCapi /ovms/
 RUN if [ "$FUZZER_BUILD" == "0" ]; then  CAPI_FLAGS=${CAPI_FLAGS} make -f MakefileCapi cpp && \
                                          CAPI_FLAGS=${CAPI_FLAGS} make -f MakefileCapi c; fi ;
-RUN mkdir -p /ovms_release/lib/ ; find /ovms/bazel-out/k8-*/bin -iname 'libovms_shared.so' -exec cp -v {} /ovms_release/lib/ \;
+RUN mkdir -p /ovms_release/lib/ ; find /ovms/bazel-out/k8-*/bin -iname 'libovms_shared.so' -exec cp -v {} /ovms_release/lib/ \; \
+   find /ovms/bazel-out/k8-*/bin -iname 'libopenvino_genai.so*' -exec cp -v {} /ovms_release/lib/ \; \
```
dkalinowski, line 359: why genai? is it another issue?

mzegla (approved):

Dockerfile.ubuntu
rasapala: libopenvino_genai also missing in the dependencies.

dkalinowski (approved):

PR #3527, opened 2025-07-23
Fix parameter validation

rasapala:
### 🛠 Summary

JIRA CVS-171067

### 🧪 Checklist

- [ ] Unit tests added.
``


dkalinowski (approved):

src/config.cpp
```diff
@@ -114,7 +114,7 @@ bool Config::validateUserSettingsInConfigAddRemoveModel(const ModelsSettingsImpl
 }
 
 bool Config::validate() {
-    if (this->serverSettings.serverMode == HF_PULL_MODE) {
+    if (this->serverSettings.serverMode == HF_PULL_MODE || this->serverSettings.serverMode == HF_PULL_AND_START_MODE) {
```
mzegla, line 117: Do we have some positive test for this mode?

rasapala: Yes, we only run one end to end test with pull and start.

mzegla (approved):

PR #3534, opened 2025-07-25
Introduce tool guided generation

mzegla:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/io_processing/hermes3/generation_config_builder.cpp
+++ b/src/llm/io_processing/hermes3/generation_config_builder.cpp
@@ -1,0 +1,66 @@
+//*****************************************************************************
+// Copyright 2025 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include <string>
+#include <utility>
+#include <openvino/genai/generation_config.hpp>
+
+#include "generation_config_builder.hpp"
+
+namespace ovms {
+
+void Hermes3GenerationConfigBuilder::parseConfigFromRequest(const OpenAIChatCompletionsRequest& request) {
+    // Call the base class method to fill in common configuration
+    BaseGenerationConfigBuilder::parseConfigFromRequest(request);
+
+    // Set tool guided generation config specific to Hermes3 and Qwen3 models
+    ov::genai::StructuralTagsConfig structuralTagsConfig;
+    std::string toolCallTrigger = "<tool_call>";
+    structuralTagsConfig.triggers.push_back(toolCallTrigger);
+
+    for (const auto& [toolName, toolSchema] : request.toolNameSchemaMap) {
+        ov::genai::StructuralTagItem tagItem;
+        tagItem.begin = toolCallTrigger;
+        tagItem.schema = R"({
+            "type": "object",
+            "properties": {
+                "name": {
+                    "type": "string",
+                    "enum": [")" +
+                         toolName + R"("]
+                },
+                "arguments": )" +
+                         toolSchema + R"(
+            },
+            "required": [
+                "name",
+                "arguments"
+            ]
+        })";
+        structuralTagsConfig.structural_tags.push_back(tagItem);
+    }
+
+    // Set the structural tags config in the generation config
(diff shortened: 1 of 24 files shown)


src/llm/language_model/legacy/servable_initializer.cpp
```diff
@@ -114,6 +114,12 @@ Status LegacyServableInitializer::initialize(std::shared_ptr<GenAiServable>& ser
     }
     properties->bestOfLimit = nodeOptions.best_of_limit();
     properties->maxModelLength = parseMaxModelLength(parsedModelsPath);
+
+    if (nodeOptions.has_enable_tool_guided_generation()) {
+        properties->enableToolGuidedGeneration = nodeOptions.enable_tool_guided_generation();
+    } else {
+        properties->enableToolGuidedGeneration = false;
+    }
```
dkalinowski:
```suggestion
    properties->enableToolGuidedGeneration = nodeOptions.enable_tool_guided_generation();
```

nit picking, but i think this is possible since this field has default value
leaving it as is creates 2 sources of truth: default value here and one in .proto

src/llm/io_processing/phi4/generation_config_builder.cpp
```diff
@@ -0,0 +1,84 @@
...
+namespace ovms {
+
+void Phi4GenerationConfigBuilder::parseConfigFromRequest(const OpenAIChatCompletionsRequest& request) {
+    // Call the base class method to fill in common configuration
+    BaseGenerationConfigBuilder::parseConfigFromRequest(request);
+
+    // Set tool guided generation config specific to Phi-4 model as described in template from:
+    // https://github.com/vllm-project/vllm/blob/v0.9.2/examples/tool_chat_template_phi4_mini.jinja
+    ov::genai::StructuralTagsConfig structuralTagsConfig;
+    std::string beginOfToolsString = "functools";
```
dkalinowski: this never changes, it doesnt need to be a variable. if so, at least be it static const local variable, meaning it will be created only once in a program

src/llm/io_processing/phi4/generation_config_builder.cpp
```diff
@@ -0,0 +1,84 @@
...
+            anyOfArray += ",";
+        }
+        first = false;
+        anyOfArray += R"({
+            "type": "object",
+            "properties": {
+                "name": {
+                    "type": "string",
+                    "enum": [")" +
+                      toolName + R"("]
```
dkalinowski, line 51: how do we handle injections? escaping? what if toolName is:
```
hi"], "abc": 512, "def": ["
```
?
it would end up with valid json, but additional field `abc`:
```
"type": "object",
"properties": {
  "name: {
     "type": "string",
     "enum": ["hi"],
     "abc": 512, <----------------
    "def: [], <--------------------
  }
  "arguments": ...
```

src/llm/io_processing/phi4/generation_config_builder.cpp
```diff
@@ -0,0 +1,84 @@
...
+        "items": {
+            "anyOf": )" +
+                     anyOfArray + R"(
+        }
+    })";
+
+    structuralTagsConfig.structural_tags.push_back(tagItem);
+
+    // Set the structural tags config in the generation config
+    if (config.structured_output_config) {
```
dkalinowski: how comes `structured_output_config` can already be populated?

src/llm/io_processing/hermes3/generation_config_builder.cpp
```diff
@@ -0,0 +1,66 @@
...
+            "required": [
+                "name",
+                "arguments"
+            ]
+        })";
+        structuralTagsConfig.structural_tags.push_back(tagItem);
+    }
+
+    // Set the structural tags config in the generation config
+    if (config.structured_output_config) {
```
dkalinowski: this repeats in all config builders

src/llm/io_processing/phi4/generation_config_builder.cpp
mzegla: It's preparation for a change expected to be introduced in another PR (with response format handling) that also uses this field.

src/llm/io_processing/phi4/generation_config_builder.cpp
mzegla: We do not. This data is taken from user anyway. `toolSchema` can be completely broken. 
If user sends bad data, we need to reject it anyway. Injection via `toolName` is nothing specific here.

dkalinowski: how do we reject it?

dkalinowski (approved):

src/llm/io_processing/phi4/generation_config_builder.cpp
mzegla: It's beyond OVMS. XGrammar rejects bad schema on grammar compilation phase which happens in GenAI.

PR #3536, opened 2025-07-28
Fix mp off build.

rasapala:
### 🛠 Summary

Fix mp off build.

### 🧪 Checklist

- [ ] Unit tests added.
``


Dockerfile.ubuntu
```diff
@@ -374,9 +376,7 @@ ARG BASE_OS=ubuntu
 ARG ov_use_binary=0
 ARG FUZZER_BUILD=0
 ARG debug_bazel_flags=--strip=never\ --copt="-g "\ -c\ dbg
-COPY --from=capi-build /ovms_release/lib/libovms_shared.so /ovms_release/lib/
-COPY --from=capi-build /ovms_release/lib/libopenvino_genai.so* /ovms_release/lib/
-COPY --from=capi-build /ovms_release/lib/libgit2.so /ovms_release/lib/
+COPY --from=capi-build /ovms_release/lib/*.so* /ovms_release/lib/
```
mzegla: So we give up on selective copy. Is it safe to assume that at this point `/ovms_release/lib/` contains only what we want in the release package?

rasapala: That is correct.

PR #3546, opened 2025-07-31
fix export for mistral model

dtrawins:
### 🛠 Summary

Fix exporting model Mistral-7B-Instruct-v0.3 which is not detecting correctly the task in optimum lib

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/demos/common/export_models/export_model.py
+++ b/demos/common/export_models/export_model.py
@@ -382,7 +382,7 @@ def export_text_generation_model(model_repository_path, source_model, model_name
                 if task_parameters['extra_quantization_params'] == "":
                     print("Using default quantization parameters for NPU: --sym --ratio 1.0 --group-size -1")
                     task_parameters['extra_quantization_params'] = "--sym --ratio 1.0 --group-size -1"
-            optimum_command = "optimum-cli export openvino --model {} --weight-format {} {} --trust-remote-code {}".format(source_model, precision, task_parameters['extra_quantization_params'], llm_model_path)
+            optimum_command = "optimum-cli export openvino --model {} --weight-format {} {} --task text-generation-with-past --trust-remote-code {}".format(source_model, precision, task_parameters['extra_quantization_params'], llm_model_path)
             if os.system(optimum_command):
                 raise ValueError("Failed to export llm model", source_model)    
     ### Export draft model for speculative decoding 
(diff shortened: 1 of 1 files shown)


demos/common/export_models/export_model.py
```diff
@@ -382,7 +382,7 @@ def export_text_generation_model(model_repository_path, source_model, model_name
                 if task_parameters['extra_quantization_params'] == "":
                     print("Using default quantization parameters for NPU: --sym --ratio 1.0 --group-size -1")
                     task_parameters['extra_quantization_params'] = "--sym --ratio 1.0 --group-size -1"
-            optimum_command = "optimum-cli export openvino --model {} --weight-format {} {} --trust-remote-code {}".format(source_model, precision, task_parameters['extra_quantization_params'], llm_model_path)
+            optimum_command = "optimum-cli export openvino --model {} --weight-format {} {} --task text-generation-with-past --trust-remote-code {}".format(source_model, precision, task_parameters['extra_quantization_params'], llm_model_path)
```
mzegla: is it okay to hardcode it like that for all models?

ngrozae: doesn't work for OpenGVLab/InternVL2-2B and OpenGVLab/InternVL2_5-8B

 ValueError: Asked to export a internvl-chat model for the task text-generation-with-past, but the Optimum OpenVINO exporter only supports the tasks image-text-to-text for internvl-chat. Please use a supported task. Please open an issue at https://github.com/huggingface/optimum/issues if you would like the task text-generation-with-past to be supported in the ONNX export for internvl-chat.

demos/common/export_models/README.md
```diff
@@ -112,16 +112,21 @@ Text generation for NPU target device. Command below sets max allowed prompt siz
 python export_model.py text_generation --source_model meta-llama/Llama-3.2-3B-Instruct --config_file_path models/config_all.json --model_repository_path models --target_device NPU --max_prompt_len 2048 --ov_cache_dir ./models/.ov_cache
 ```
 
+> **Note:** Some models like `mistralai/Mistral-7B-Instruct-v0.3` might fail to export because the task can't be determined automatically. In such situation it can be set in `--extra_quantization_parameters`. For example:
+```console
+python export_model.py text_generation --source_model mistralai/Mistral-7B-Instruct-v0.3 --model_repository_path models --extra_quantization_params "--task text-generation-with-past"
```
mzegla, line 117: Shall we say a few words why this task is selected? (like that this is a default task for text generation in OVMS or something similar)

PR #3548, opened 2025-08-01
Split response parser to tool and reasoning parser

mzegla:
Split response parser to tool and reasoning parser. That enables streaming for Hermes3 model out of the box.


src/llm/io_processing/output_parser.cpp
```diff
@@ -0,0 +1,117 @@
...
+    // If for any processing phase a nullopt should be returned, it should be done in the parser implementation.
+    // Do not return nullopt directly from this method.
+
+    bool reasoningParserExistsAndSupportsStreaming = reasoningParser && !reasoningParser->getParsingStartTag().empty() && !reasoningParser->getParsingEndTag().empty();
+    bool toolParserExistsAndSupportsStreaming = toolParser && !toolParser->getParsingStartTag().empty();
+
+    if (processingPhase == UNKNOWN) {
+        // If we are in the UNKNOWN phase, we need to determine if we should switch to CONTENT, REASONING, or TOOL_CALLS phase.
+        if (reasoningParserExistsAndSupportsStreaming && chunkResponse.find(reasoningParser->getParsingStartTag()) != std::string::npos) {
+            std::cout << "Switching to REASONING phase" << std::endl;
```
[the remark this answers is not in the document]
mzegla: to be removed

src/llm/io_processing/generation_config_builder.hpp
```diff
@@ -32,7 +32,7 @@ class GenerationConfigBuilder {
 
 public:
     GenerationConfigBuilder() = delete;
-    // Using tool parser name to select appropriate builder implementation to avoid introducing additional parameters
+    // Using tool parser name to select appropriate builder implementation to avoid introducing additional parameters. Might be insufficient in the future.
```
dkalinowski, line 35: why?

src/llm/io_processing/hermes3/tool_parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+    if (chunk.empty()) {
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Received empty chunk for Hermes3ToolParser");
+        return std::nullopt;
+    }
+
+    if (chunk.find(toolCallStartTag) != std::string::npos) {
+        lastJson.Clear();
+        jsonBuilder.clear();
+        toolCallIndex++;
+        argumentsDelayWindow[0].clear();
```
dkalinowski, line 115: are you sure both exist?

src/llm/io_processing/hermes3/tool_parser.hpp
```diff
@@ -29,31 +30,33 @@
...
-    explicit Qwen3OutputParser(ov::genai::Tokenizer& tokenizer) :
+    Hermes3ToolParser() = delete;
+    explicit Hermes3ToolParser(ov::genai::Tokenizer& tokenizer) :
         BaseOutputParser(tokenizer) {}
 
-    ParsedOutput parse(const std::vector<int64_t>& generatedTokens) override;
+    void parse(ParsedOutput& parsedOutput, const std::vector<int64_t>& generatedTokens) override;
     std::optional<rapidjson::Document> parseChunk(const std::string& chunk) override;
+    std::string getParsingStartTag() const override {
+        return toolCallStartTag;
```
dkalinowski: why do we need to construct it each time? it is time consuming. does it change at any time? if not, consider api that returns const reference to const pre-initialized string or char array

src/test/llm/output_parsers/hermes3_output_parser_test.cpp
```diff
@@ -107,29 +125,225 @@ TEST_F(Hermes3OutputParserTest, ParseToolCallOutputWithThreeToolCalls) {
...
+        // Parser removes whitespaces, so we expect arguments value to be without spaces
+        EXPECT_EQ(parsedOutput.toolCalls[0].arguments, "{\"arg1\":\"value1\",\"arg2\":42}");
+        EXPECT_EQ(parsedOutput.toolCalls[0].id.empty(), false);  // ID should be generated
+        auto firstToolCallId = parsedOutput.toolCalls[0].id;
+
+        EXPECT_EQ(parsedOutput.toolCalls[2].name, "third_tool");
+        // Parser removes whitespaces, so we expect arguments value to be without spaces
+        EXPECT_EQ(parsedOutput.toolCalls[2].arguments, "{\"key\":\"value\"}");
+        EXPECT_EQ(parsedOutput.toolCalls[2].id.empty(), false);  // ID should be generated
+        auto thirdToolCallId = parsedOutput.toolCalls[2].id;
```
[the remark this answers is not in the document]
dkalinowski: good catch mr copilot

src/test/llm/output_parsers/llama3_output_parser_test.cpp
```diff
@@ -119,10 +152,10 @@ TEST_F(Llama3OutputParserTest, ParseToolCallOutputWithContentAndSingleToolCall)
     generatedTokens.insert(generatedTokens.end(), botTokenId);  // Add bot token ID
     generatedTokens.insert(generatedTokens.end(), generatedToolCallTokens.begin(), generatedToolCallTokens.end());
     // generatedTokens should now contain content followed by bot token ID and then tool call
-    ParsedOutput parsedOutput = outputParser->parse(generatedTokens);
+    ParsedOutput parsedOutput = outputParser->parse(generatedTokens, true);
```
dkalinowski, line 155: do we need separate tests with `false`?

src/llm/io_processing/generation_config_builder.hpp
mzegla: For now, model dependent options in GenerationConfig are only related to tool guided generation. In the future it is possible that there will be other options that would differ depending on the model and it would be impossible to select correct generation config builder based only on tool parser.

src/llm/io_processing/hermes3/tool_parser.cpp
mzegla: The two element array is always initialized with two empty strings as `std::array`.
from `hermes3/tool_parser.hpp`:
`std::array<std::string, 2> argumentsDelayWindow{{"", ""}};`

src/test/llm/output_parsers/llama3_output_parser_test.cpp
mzegla: I added a few for the relevant case i.e. when we don't pass tools and model generates only JSON (when response_format is set for example), we don't want to interpret it as tools.

PR #3552, opened 2025-08-04
Add tests for cli

rasapala:
### 🛠 Summary

JIRA CVS-169285, CVS-169294, CVS-170631

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/pull_module/libgit2.cpp
+++ b/src/pull_module/libgit2.cpp
@@ -56,7 +56,7 @@ int cred_acquire_cb(git_credential** out,
     char *username = NULL, *password = NULL;
     int error = -1;
 
-    fprintf(stdout, "Authentication is required for repository clone.\n");
+    fprintf(stdout, "Authentication is required for repository clone or model is missing.\n");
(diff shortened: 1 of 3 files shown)


src/pull_module/libgit2.cpp
```diff
@@ -56,7 +56,7 @@ int cred_acquire_cb(git_credential** out,
     char *username = NULL, *password = NULL;
     int error = -1;
 
-    fprintf(stdout, "Authentication is required for repository clone.\n");
+    fprintf(stdout, "Authentication is required for repository clone or model is missing.\n");
```
mzegla, line 59: Can we differentiate those cases? I mean separate errors for authentication issue and for missing model would be better in my opinion. We would know more precisely what should be fixed.

rasapala: Unfortunately no. This callback is called once git2 clone failed as it tries to process authentication and try again.

mzegla (approved):

PR #3554, opened 2025-08-05
Tests and version added

rasapala:
### 🛠 Summary

JIRA CVS-168020

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/graph_export/graph_export.cpp
+++ b/src/graph_export/graph_export.cpp
@@ -38,6 +38,7 @@
 #include "../status.hpp"
 #include "../stringutils.hpp"
 #include "../schema.hpp"
+#include "../version.hpp"
 
 #if (MEDIAPIPE_DISABLE == 0)
 #pragma warning(push)
@@ -51,8 +52,16 @@
 #endif
 namespace ovms {
 
-static Status createTextGenerationGraphTemplate(const std::string& directoryPath, const TextGenGraphSettingsImpl& graphSettings) {
+static std::ostringstream GetOvmsVersionStream(std::string projectName, std::string peojectVersion) {
     std::ostringstream oss;
+    std::string project_name(projectName);
+    std::string project_version(peojectVersion);
+    oss << "# " << project_name + " " + project_version << std::endl;
+    return oss;
+}
+
+static Status createTextGenerationGraphTemplate(const std::string& directoryPath, const TextGenGraphSettingsImpl& graphSettings) {
+    std::ostringstream oss = GetOvmsVersionStream(PROJECT_NAME, PROJECT_VERSION);
     // clang-format off
     oss << R"(
     input_stream: "HTTP_REQUEST_PAYLOAD:input"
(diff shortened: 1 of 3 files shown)


src/graph_export/graph_export.cpp
```diff
@@ -51,8 +52,16 @@
 #endif
 namespace ovms {
 
-static Status createTextGenerationGraphTemplate(const std::string& directoryPath, const TextGenGraphSettingsImpl& graphSettings) {
+static std::ostringstream GetOvmsVersionStream(std::string projectName, std::string peojectVersion) {
```
dkalinowski:
```suggestion
static std::ostringstream GetOvmsVersionStream(const std::string& projectName, const std::string& projectVersion) {
```

src/graph_export/graph_export.cpp
```diff
@@ -51,8 +52,16 @@
...
     std::ostringstream oss;
+    std::string project_name(projectName);
```
dkalinowski: why copying string multiple times?

src/graph_export/graph_export.cpp
```diff
@@ -141,7 +150,7 @@ static Status createTextGenerationGraphTemplate(const std::string& directoryPath
 }
 
 static Status createRerankGraphTemplate(const std::string& directoryPath, const RerankGraphSettingsImpl& graphSettings) {
-    std::ostringstream oss;
+    std::ostringstream oss = GetOvmsVersionStream(PROJECT_NAME, PROJECT_VERSION);
```
dkalinowski: I think it would be cleaner to:
```suggestion
    std::ostringstream osss;
    writeOvmsVersionGraphLine(oss, PROJECT_NAME, PROJECT_VERSION);
```

+ GetOvmsVersionStream doesnt follow naming convention (camel case)

atobiszei (approved):

src/graph_export/graph_export.cpp
```diff
@@ -51,8 +52,15 @@
 #endif
 namespace ovms {
 
+static std::string WriteOvmsVersionGraphLine(const std::string& projectName, const std::string& projectVersion) {
+    std::ostringstream oss;
+    oss << "# File created with: " << projectName + " " + projectVersion << std::endl;
+    return oss.str();
+}
```
dkalinowski: you dont need output string stream if all you do is concatenating string (+ camel case convention)
plus why are you mixing + and <<
```suggestion
static std::string writeOvmsVersionGraphLine(const std::string& projectName, const std::string& projectVersion) {
    std::stringstream ss;
    ss << "# File created with: " << projectName << " " << projectVersion << std::endl;
    return ss.str();
}
```

dkalinowski: now as I think of it, why do we need to construct it all the time? cant we use constexpr and save it in memory during compilation?

if thats not the case, cant we construct it once, save it in static const std::string and just oss <<  it everytime?

dkalinowski: `static const std::string OVMS_VERSION_GRAPH_LINE = "# File created with: " + PROJECT_NAME + std::string(" ") + PROJECT_VERSION;

and then each time:
```
oss << OVMS_VERSION_GRAPH_LINE;
```

whenever you need it inside createRerankGraphTemplate for example

dkalinowski (approved):

PR #3579, opened 2025-08-19
Log improvements

przepeck:
### 🛠 Summary

CVS-170725, CVS-171760
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/capi_frontend/server_settings.hpp
+++ b/src/capi_frontend/server_settings.hpp
@@ -176,8 +176,8 @@ struct ServerSettingsImpl {
     std::optional<size_t> grpcMemoryQuota;
     std::string grpcChannelArguments;
     uint32_t filesystemPollWaitMilliseconds = 1000;
-    uint32_t sequenceCleanerPollWaitMinutes = 5;
-    uint32_t resourcesCleanerPollWaitSeconds = 1;
+    uint32_t sequenceCleanerPollWaitMinutes = 0;
+    uint32_t resourcesCleanerPollWaitSeconds = 0;
     std::string cacheDir;
     bool withPython = false;
     bool startedWithCLI = false;
--- a/src/llm/language_model/continuous_batching/servable_initializer.cpp
+++ b/src/llm/language_model/continuous_batching/servable_initializer.cpp
@@ -99,6 +99,25 @@ Status ContinuousBatchingServableInitializer::initialize(std::shared_ptr<GenAiSe
         properties->schedulerConfig.use_cache_eviction = false;
     }
 
+    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "SchedulerConfig: max_num_batched_tokens={}, cache_size={}, dynamic_split_fuse={}, max_num_seqs={}, enable_prefix_caching={}",
+        properties->schedulerConfig.max_num_batched_tokens,
+        properties->schedulerConfig.cache_size,
+        properties->schedulerConfig.dynamic_split_fuse,
+        properties->schedulerConfig.max_num_seqs,
+        properties->schedulerConfig.enable_prefix_caching);
+
+        if(properties->schedulerConfig.use_cache_eviction) {
+            const auto& cacheEvictionConfig = properties->schedulerConfig.cache_eviction_config;
+            SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "CacheEvictionConfig: start_size={}, recent_size={}, max_cache_size={}, aggregation_mode={}, apply_rotation={}",
+                cacheEvictionConfig.get_start_size(),
+                cacheEvictionConfig.get_recent_size(),
+                cacheEvictionConfig.get_max_cache_size(),
+                static_cast<int>(cacheEvictionConfig.aggregation_mode),
+                cacheEvictionConfig.apply_rotation);
+        }else{
+            SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "CacheEvictionConfig is not used");
+        }
+
     properties->device = nodeOptions.device();
 
     if (!nodeOptions.draft_models_path().empty()) {
--- a/src/llm/servable.cpp
+++ b/src/llm/servable.cpp
@@ -37,7 +37,8 @@
 
 namespace ovms {
 absl::Status GenAiServable::loadRequest(std::shared_ptr<GenAiServableExecutionContext>& executionContext, const ovms::HttpPayload& payload) {
-    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Request body: {}", payload.body);
+    std::string requestBody = getBodyWithDecodedContent(payload);
+    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Request body: {}", requestBody);
     SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Request uri: {}", payload.uri);
     // Parsed JSON is not guaranteed to be valid, we may reach this point via multipart content type request with no valid JSON parser
     if (payload.parsedJson->HasParseError()) {
@@ -206,6 +207,26 @@ std::string wrapTextInServerSideEventMessage(const std::string& text) {
     std::stringstream ss;
     ss << "data: " << text << "\n\n";
     return ss.str();
+}
+std::string getBodyWithDecodedContent(const ovms::HttpPayload& payload) {
+    auto json = payload.parsedJson;
+    std::string requestContent;
+    if (json->HasMember("messages") && (*json)["messages"].IsArray() && !(*json)["messages"].Empty()) {
+        const auto& message = (*json)["messages"][0];
+        if (message.HasMember("content") && message["content"].IsString()) {
+            requestContent = message["content"].GetString();
+        }
+    }
+    std::string requestBody = payload.body;
+    std::string contentKey = "\"content\":";
+    auto contentIdx = requestBody.find(contentKey);
+    if (contentIdx != std::string::npos) {
+        requestBody = requestBody.substr(0, contentIdx);
+        requestBody += contentKey;
+        requestBody += requestContent.empty() ? "null" : requestContent;
+        requestBody += payload.body.substr(contentIdx + contentKey.length() + requestContent.length()*2 + 1);
+    }
+    return requestBody;
 }
 #pragma GCC diagnostic pop
 #pragma warning(push)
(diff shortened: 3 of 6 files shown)


src/llm/servable.cpp
```diff
@@ -207,6 +208,26 @@ std::string wrapTextInServerSideEventMessage(const std::string& text) {
     ss << "data: " << text << "\n\n";
     return ss.str();
 }
+std::string getBodyWithDecodedContent(const ovms::HttpPayload& payload) {
```
dtrawins: what is the point of this payload parsing? In a critical performance path, all non essential processing should be removed

src/capi_frontend/server_settings.hpp
```diff
@@ -176,8 +176,8 @@ struct ServerSettingsImpl {
     std::optional<size_t> grpcMemoryQuota;
     std::string grpcChannelArguments;
     uint32_t filesystemPollWaitMilliseconds = 1000;
-    uint32_t sequenceCleanerPollWaitMinutes = 5;
-    uint32_t resourcesCleanerPollWaitSeconds = 1;
+    uint32_t sequenceCleanerPollWaitMinutes = 0;
```
dtrawins: to be checked if this sequencecleaner was doing also something useful like free() to release deallocated memory

src/llm/language_model/continuous_batching/servable_initializer.cpp
```diff
@@ -99,6 +99,25 @@ Status ContinuousBatchingServableInitializer::initialize(std::shared_ptr<GenAiSe
...
+        const auto& cacheEvictionConfig = properties->schedulerConfig.cache_eviction_config;
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "CacheEvictionConfig: start_size={}, recent_size={}, max_cache_size={}, aggregation_mode={}, apply_rotation={}",
+            cacheEvictionConfig.get_start_size(),
+            cacheEvictionConfig.get_recent_size(),
+            cacheEvictionConfig.get_max_cache_size(),
+            static_cast<int>(cacheEvictionConfig.aggregation_mode),
+            cacheEvictionConfig.apply_rotation);
+    } else {
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "CacheEvictionConfig is not used");
+    }
```
mzegla: How about we add a method `log()` to properties class that would handle logging it's properties and then we would call it here like `properties->log()` if we are above info level?
Such method could be first implemented in base properties (in main servable.hpp class) that would print things like device, used parsers (if any) etc. and child implementations (like in servable.hpp in continuous batching) would override it, logging it's specific properties like scheduler config for CB.

src/llm/servable.cpp
```diff
@@ -169,9 +171,9 @@ absl::Status GenAiServable::preparePartialResponse(std::shared_ptr<GenAiServable
     ov::genai::GenerationFinishReason finishReason = generationOutput.finish_reason;
     if (finishReason == ov::genai::GenerationFinishReason::NONE) {  // continue
         if (lastTextChunk.size() > 0) {
-            std::string serializedChunk = executionContext->apiHandler->serializeStreamingChunk(lastTextChunk, finishReason);
+            std::string serializedChunk = " 1 " + executionContext->apiHandler->serializeStreamingChunk(lastTextChunk, finishReason);
             if (!serializedChunk.empty()) {
-                executionContext->response = wrapTextInServerSideEventMessage(serializedChunk);
+                executionContext->response = " 2 " + wrapTextInServerSideEventMessage(serializedChunk);
```
mzegla: Debugging leftovers?

src/modelmanager.cpp
```diff
@@ -189,7 +185,9 @@ Status ModelManager::start(const Config& config) {
         return status;
     }
     startWatcher(isStartedWithConfigFile());
-    startCleaner();
+    if (sequenceCleaupIntervalMinutes > 0 || resourcesCleanupIntervalMillisec > 0)
+        startCleaner();
```
mzegla, line 189: Is it the same cleaner?

src/modelmanager.cpp
```diff
@@ -963,7 +957,7 @@ Status ModelManager::loadMediapipeSubConfigModels(std::vector<ModelConfig>& gate
         const auto mediapipeItr = subconfigJson.FindMember("model_config_list");
 
         if (mediapipeItr == subconfigJson.MemberEnd() || !mediapipeItr->value.IsArray()) {
-            SPDLOG_LOGGER_ERROR(modelmanager_logger, "Configuration file doesn't have models property.");
+            SPDLOG_LOGGER_INFO(modelmanager_logger, "Configuration file doesn't have models property.");
```
mzegla: Why is it changed if we return error?

src/modelmanager.cpp
```diff
@@ -987,7 +981,7 @@ Status ModelManager::loadModelsConfig(rapidjson::Document& configJson, std::vect
     const auto itr = configJson.FindMember("model_config_list");
 
     if (itr == configJson.MemberEnd() || !itr->value.IsArray()) {
-        SPDLOG_LOGGER_ERROR(modelmanager_logger, "Configuration file doesn't have models property.");
+        SPDLOG_LOGGER_INFO(modelmanager_logger, "Configuration file doesn't have models property.");
```
mzegla: Why is it changed if we return error?

src/llm/servable.cpp
```diff
@@ -37,7 +39,7 @@
 
 namespace ovms {
 absl::Status GenAiServable::loadRequest(std::shared_ptr<GenAiServableExecutionContext>& executionContext, const ovms::HttpPayload& payload) {
-    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Request body: {}", payload.body);
+    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Request body: {}", getBodyString(payload));
```
mzegla: Why is it changed? We loaded request body before.

przepeck: we changed it to display utf-8 characters in console

src/modelmanager.cpp
przepeck: this is the false positive error

src/modelmanager.cpp
przepeck: this is the false positive error

src/modelmanager.cpp
przepeck: yes, it has sequenceCleaner and resourcesCleaner only

src/llm/language_model/continuous_batching/servable_initializer.cpp
przepeck: this is great idea, I  will implement that.

src/pull_module/optimum_export.cpp
```diff
@@ -137,7 +137,7 @@ Status OptimumDownloader::checkRequiredToolsArePresent() {
     std::string output = exec_cmd(this->OPTIMUM_CLI_CHECK_COMMAND, retCode);
     if (retCode != 0) {
         SPDLOG_DEBUG("Command output {}", output);
-        SPDLOG_ERROR("optimum-cli executable is not present. Please install python and demos/common/export_models/requirements.txt");
+        SPDLOG_ERROR("To download model {} you need the optimum-cli executable, which is not present. Please install python and demos/common/export_models/requirements.txt", this->hfSettings.sourceModel);
```
dtrawins: Target folder <> not found, trying to pull from <> from HuggingFace but missing optimum-intel. Use the ovms package with optimum-intel.

src/llm/servable.cpp
```diff
@@ -225,6 +227,17 @@ std::string wrapTextInServerSideEventMessage(const std::string& text) {
...
     return ss.str();
 }
+std::string getBodyString(const ovms::HttpPayload& payload) {
+    if (spdlog::default_logger_raw()->level() == spdlog::level::debug) {
+        auto& json = payload.parsedJson;
+        rapidjson::StringBuffer buffer;
+        rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+        json->Accept(writer);
+        return buffer.GetString();
+    }
```
mzegla: I don't like the fact that we create a dependency between return value and logging level. 
If this method is only used above in `loadRequest`, then maybe rename it to `logRequestBody` and call `SPD_LOGGER_DEBUG` here to keep it all in the context of logging.

src/modelmanager.cpp
```diff
@@ -974,7 +968,7 @@ Status ModelManager::loadMediapipeSubConfigModels(std::vector<ModelConfig>& gate
         const auto mediapipeItr = subconfigJson.FindMember("model_config_list");
 
         if (mediapipeItr == subconfigJson.MemberEnd() || !mediapipeItr->value.IsArray()) {
-            SPDLOG_LOGGER_ERROR(modelmanager_logger, "Configuration file doesn't have models property.");
+            SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Configuration file doesn't have models property.");
```
mzegla: "Subconfig" ?

src/llm/servable.cpp
```diff
@@ -37,7 +39,7 @@
 
 namespace ovms {
 absl::Status GenAiServable::loadRequest(std::shared_ptr<GenAiServableExecutionContext>& executionContext, const ovms::HttpPayload& payload) {
-    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Request body: {}", payload.body);
+    logRequestBody(payload.parsedJson);
```
mzegla: How about we expand that to `logRequestDetails(payload)` and inside that function we always log request body and URI.
Then here we could call such method if log level is debug or trace.

src/llm/servable.cpp
```diff
@@ -227,14 +226,16 @@ std::string wrapTextInServerSideEventMessage(const std::string& text) {
     ss << "data: " << text << "\n\n";
     return ss.str();
 }
-void logRequestBody(std::shared_ptr<rapidjson::Document> parsedJson) {
+void logRequestDetails(const ovms::HttpPayload& payload) {
     if (spdlog::default_logger_raw()->level() != spdlog::level::debug)
         return;
```
mzegla: What if we set log level TRACE?

przepeck: it wasn't displayed, I fixed it in the next commit

mzegla (approved):

PR #3577, opened 2025-08-18
fix agentic demo execution and add more models

dtrawins:
### 🛠 Summary

CVS-172089
Swap tool_parser name from qwen3 to hermes3
Rollback transformers to improve accuracy in LLM models
Adjust the demo to include more models
improve the agent script to give option to filter MCP server and set tool_choice

### 🧪 Checklist

- [ ] Unit tests added.
- [ ] The documentation updated.
- [ ] Change follows security best practices.
``

PR #3583, opened 2025-08-20
enable chat_template.jinja to fix transformers upgrade

dtrawins:
### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/demos/common/export_models/requirements.txt
+++ b/demos/common/export_models/requirements.txt
@@ -2,14 +2,14 @@
 --extra-index-url "https://storage.openvinotoolkit.org/simple/wheels/nightly"
 --extra-index-url "https://storage.openvinotoolkit.org/simple/wheels/pre-release"
 --pre
-optimum-intel
+optimum-intel@git+https://github.com/huggingface/optimum-intel.git
 openvino-tokenizers<=2025.3.0.0.dev20250812
 openvino<=2025.3.0.dev20250812
 nncf>=2.11.0 
 sentence_transformers
 sentencepiece==0.2.0
 openai 
-transformers<4.53
+transformers<4.54
 einops
 torchvision
 timm==1.0.15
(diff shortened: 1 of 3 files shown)


demos/common/export_models/requirements.txt
```diff
@@ -2,14 +2,14 @@
...
-optimum-intel
+optimum-intel@git+https://github.com/huggingface/optimum-intel.git
 openvino-tokenizers<=2025.3.0.0.dev20250812
 openvino<=2025.3.0.dev20250812
 nncf>=2.11.0 
 sentence_transformers
 sentencepiece==0.2.0
 openai 
-transformers<4.53
+transformers<4.54
```
ngrozae, line 12: shouldn't it also specify transformers==4.54 ? so earlier version wouldn't be installed

mzegla: or <= if we want to include it...

src/llm/servable_initializer.cpp
```diff
@@ -137,13 +137,16 @@ void GenAiServableInitializer::loadPyTemplateProcessor(std::shared_ptr<GenAiServ
             tool_template = None
 
             # Try to read template from template.jinja file
-            jinja_file = Path(templates_directory + "/template.jinja")
+            jinja_file = Path(templates_directory + "/chat_template.jinja")
```
mzegla, line 140: @ngrozae, @pgladkows note that after this PR is merged we will accept chat_template.jinja, which might simplify some workflows in validation

mzegla (approved):
Do we want to put using template.jinja to legacy/deprecated and remove this path at some point?

demos/common/export_models/requirements.txt
dtrawins: @ngrozae  we don't want to include 4.54. It should be 4.53.x or older.

dtrawins:
> Do we want to put using template.jinja to legacy/deprecated and remove this path at some point?

yes, but we have too many demos and script using it to drop it now

demos/common/export_models/requirements.txt
ngrozae: also there's another PR with transformers<4.52
https://github.com/openvinotoolkit/model_server/pull/3577
which one is correct now?

PR #3589, opened 2025-08-22
llama3 streaming

dkalinowski:
### 🧪 Checklist

- [ ] Unit tests added.

### DIFF
--- a/src/llm/io_processing/llama3/tool_parser.cpp
+++ b/src/llm/io_processing/llama3/tool_parser.cpp
@@ -216,7 +216,7 @@ std::optional<rapidjson::Document> Llama3ToolParser::parseChunk(const std::strin
             newJson = jsonBuilder.add(chunk);
         }
     } catch (const std::exception& e) {
-        SPDLOG_LOGGER_INFO(llm_calculator_logger, "Tool call chunk partial parse failed: {}", e.what());
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Tool call chunk partial parse failed: {}", e.what());
         // Throwing an error since at this point the JSON is broken and next chunks will not make it right.
         throw std::runtime_error("Generated tool call structure is not valid");  // re-throw
     }
@@ -232,7 +232,7 @@ std::optional<rapidjson::Document> Llama3ToolParser::parseChunk(const std::strin
             // We received big chunk with both full function name and parameters, so we get function name from the new JSON
             functionName = newJson["name"].GetString();
         } else {
-            SPDLOG_LOGGER_INFO(llm_calculator_logger, "Tool call name has not been generated and parameters already started");
+            SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Tool call name has not been generated and parameters already started");
             throw std::runtime_error("Tool call name is missing in generated output");
         }
         // Wrap first delta in {"tool_calls":[{"id":<id>,"type":"function","index":<toolCallIndex>,"function":{"name": <functionName>}}]}


src/llm/io_processing/base_output_parser.hpp
```diff
@@ -86,13 +88,17 @@ class BaseOutputParser {
...
+    virtual std::optional<rapidjson::Document> parseChunk(const std::string& chunkResponse, ov::genai::GenerationFinishReason finishReason) = 0;
 
     // Get the tag that marks the beginning of the segment that should be processed by the parser.
     // This method is used in streaming mode to determine if the parser should start processing the content.
     // If empty string is returned, it means that the parser will never start processing the content.
     virtual const std::string& getParsingStartTag() const = 0;
 
+    // During streaming, there might be tags which when appear at the beginning, always mean that this is tool parsing.
+    // This is other than ParsingStartTag, which is used at any point during streaming.
+    virtual const std::unordered_set<std::string>& getBeginningOnlyParsingTags() const = 0;
```
mzegla: This is a base class, so the comment should be more generic (do not mention tool parsing).
I would also change the name of that method to include `ParsingStartTag` since tags returned by this method have the same purpose.

```suggestion
    // Get a vector of additional tags that mark beginning of the segment that should be processed by the parser.
    // These tags are considered only if they the first output produced by the model.
    virtual const std::unordered_set<std::string>& getSpecialParsingStartTags() const = 0;
```

src/llm/io_processing/llama3/tool_parser.cpp
```diff
@@ -110,9 +110,159 @@ void Llama3ToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int64
...
+    if (chunk.empty()) {
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Received empty chunk for Llama3ToolParser");
+        return std::nullopt;
+    }
+
+    // <|python_tag|> appears
+    if (chunk.find(parsingStartTag) != std::string::npos) {
+        this->startNextToolCall();
+        return std::nullopt;  // ignoring the special tag
+    }
```
mzegla, line 143: Doesn't <|python_tag|> appear only once in the entire generation?

src/llm/io_processing/llama3/tool_parser.cpp
```diff
@@ -110,9 +110,159 @@ void Llama3ToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int64
...
+
+        // Handle the case when we are starting to collect parameters/arguments.
+        // Force parameters/arguments string type and fill first element of the delay array.
+        if (argumentsDelayWindow[0].empty()) {
+            // If we are starting to collect parameters/arguments, we add opening quote before the first non-whitespace character
+            size_t firstNonWhitespaceCharacter = modifiedChunk.find_first_not_of(" \t\n\r\f\v");
+            if (firstNonWhitespaceCharacter != std::string::npos) {
+                modifiedChunk.insert(firstNonWhitespaceCharacter, "\"");
+            } else {
+                // If the chunk is all whitespace, just insert at the end
```
mzegla:
```suggestion
                // If the chunk is all whitespace, just insert closing quote at the end
```

src/llm/io_processing/llama3/tool_parser.cpp
```diff
@@ -110,9 +110,159 @@ void Llama3ToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int64
...
+        }
+
+        // If this is end of streaming, we need to manually add closing quote "
+        // We need to place it right before last closing brace
+        if (finishReason == ov::genai::GenerationFinishReason::STOP) {
+            isCurrentToolCallParsingFinished = true;
+            size_t lastClosingBrace = modifiedChunk.find_last_of('}');
+            if (lastClosingBrace != std::string::npos) {
+                modifiedChunk.insert(lastClosingBrace, "\"");
+                argumentsDelayWindow[0] += modifiedChunk;
```
mzegla, line 193: Why +=?

src/llm/io_processing/llama3/tool_parser.cpp
```diff
@@ -110,9 +110,159 @@ void Llama3ToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int64
...
+            }
+            // If this is end of one of the tool calls "in the middle" (; has been found), we need to manually add closing quote "
+            // We need to place it right before last closing brace
+        } else if (modifiedChunk.find(separator) != std::string::npos) {
+            isCurrentToolCallParsingFinished = true;
+            size_t lastClosingBrace = argumentsDelayWindow[0].find_last_of('}');
+            if (lastClosingBrace != std::string::npos) {
+                argumentsDelayWindow[0].insert(lastClosingBrace, "\"");
+            }
```
mzegla, line 202: Might be risky if streamer merges separator with some opening/closing brackets or some other characters. 
Would be good if we could check if such risk is real somehow.

src/llm/io_processing/llama3/tool_parser.cpp
```diff
@@ -110,9 +110,159 @@ void Llama3ToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int64
...
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Tool call chunk partial parse failed: {}", e.what());
+        // Throwing an error since at this point the JSON is broken and next chunks will not make it right.
+        throw std::runtime_error("Generated tool call structure is not valid");  // re-throw
+    }
+
+    rapidjson::Document doc;
+    // Case 1: 'parameters'/'arguments' has just appeared in the current chunk. If so, we return first delta.
+    if (jsonHasArgumentsOrParameters(newJson) && !jsonHasArgumentsOrParameters(lastJson)) {
+        std::string functionName;
+        ensureArgumentsInJson(newJson);
```
mzegla: This function name in that context might be ambiguous imho. Maybe `changeParametersToArguments` or something like that to make it more explicit.

src/llm/io_processing/output_parser.cpp
```diff
@@ -104,7 +105,20 @@ ParsedOutput OutputParser::parse(const std::vector<int64_t>& generatedTokens, co
     return parsedOutput;
 }
 
-std::optional<rapidjson::Document> OutputParser::parseChunk(const std::string& chunkResponse, const bool toolsAvailable) {
+static inline bool isParsingTagPartOfChunk(const std::string& chunk, const std::string& parsingTag) {
+    return chunk.find(parsingTag) != std::string::npos;
+}
```
mzegla, line 110: Do we need to wrap it in a separate function since it's one-liner?

src/llm/io_processing/output_parser.cpp
```diff
@@ -104,7 +105,20 @@ ParsedOutput OutputParser::parse(const std::vector<int64_t>& generatedTokens, co
...
+
+static inline bool isAnyOfBeginningOnlyParsingTagPartOfChunk(const std::string& chunk, const std::unordered_set<std::string>& beginningOnlyParsingTags) {
```
mzegla: `beginningOnly` name does not speak to me to be honest

src/llm/io_processing/llama3/tool_parser.cpp
dkalinowski: yes, why?

src/llm/io_processing/llama3/tool_parser.cpp
dkalinowski: because we might already have something in the window. and if so, we need to add new content to it, which is then added to partialParser

src/llm/io_processing/llama3/tool_parser.cpp
dkalinowski: havent seen such case in 'multiple' dataset, let's work on it during broader testing phase

src/llm/io_processing/output_parser.cpp
dkalinowski: way cleaner to read actual chunk parser logic

src/llm/io_processing/output_parser.cpp
dkalinowski: to be neither, any ideas?

src/llm/io_processing/llama3/tool_parser.cpp
mzegla: I see, we have toolCallIndex on -1 at this point. Nevermind.

src/llm/io_processing/llama3/tool_parser.cpp
mzegla: I see. We need the same behavior in other parsers as well so we don't drop last chunks, but it's out of scope of this PR.

src/llm/io_processing/output_parser.cpp
mzegla: `bool chunkContainsSpecialStartTag(const std::string& chunk, const std::unordered_set<std::string>& specialStartTags)` ?
I keep thinking about the naming for those additional start tags. "special", "secondary" or something else. I think we should give them a name so we can differentiate them from the regular start tag and then base method and variables naming on that.

src/llm/io_processing/base_output_parser.hpp
dkalinowski: done

src/llm/io_processing/llama3/tool_parser.cpp
dkalinowski: done

src/llm/io_processing/llama3/tool_parser.cpp
dkalinowski: done

src/llm/io_processing/output_parser.cpp
dkalinowski: changed

PR #3618, opened 2025-09-02
Add file model checks

rasapala:
### 🛠 Summary

JIRA CVS-172909

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/prepare_llm_models.sh
+++ b/prepare_llm_models.sh
@@ -22,7 +22,7 @@ fi
 
 CB_MODEL="facebook/opt-125m"
 TOKENIZER_FILE="openvino_tokenizer.bin"
-MODEL_FILE="1/model.bin"
+LEGACY_MODEL_FILE="1/model.bin"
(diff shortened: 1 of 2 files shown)


dkalinowski (approved):

prepare_llm_models.sh
```diff
@@ -21,6 +21,8 @@ if [ -z "$1" ]; then
 fi
 
 CB_MODEL="facebook/opt-125m"
+TOKENIZER_FILE="openvino_tokenizer.bin"
+MODEL_FILE="1/model.bin"
```
mzegla: Are you sure it's not 	`openvino_model.xml/bin`?

rasapala: Yes, this is old type.

mzegla: Then let's rename to LEGACY_MODEL_FILE or something like that

mzegla (approved):

PR #3624, opened 2025-09-02
Fix openai packages versions for agentic demo

mzegla:


dkalinowski (approved):

dtrawins (approved):

demos/continuous_batching/agentic_ai/requirements.txt
```diff
@@ -1,3 +1,4 @@
-openai-agents==0.0.19
+openai-agents==0.0.17
```
dtrawins, line 1: version 0.0.19 also works fine

mzegla: I feel like 0.0.17 is safer for now. Let's make general update as a separate task.

PR #3633, opened 2025-09-04
Update Continue demo with agentic model

dkalinowski:
### 🛠 Summary

CVS-173017


demos/code_local_assistant/README.md
```diff
@@ -24,6 +24,46 @@ Export `codellama/CodeLlama-7b-Instruct-hf`:
 python export_model.py text_generation --source_model codellama/CodeLlama-7b-Instruct-hf --weight-format int4 --config_file_path models/config_all.json --model_repository_path models --target_device GPU --overwrite_models
 ```
 
+> **Note:** Use `--target_device NPU` for Intel NPU or omit this parameter to run on Intel CPU
+
+## Prepare Agentic Model 
+We need specialized model that is able to produce tool calls. For this task we will use Qwen3-8B quantized to int4. We will use automatic pulling of HF models, so export script is not required.
+
+Pull and add `OpenVINO/Qwen3-8B-int4-ov` (Linux):
```
mzegla, line 32: Did you check Qwen3-4B? Maybe it would be enough in terms of quality and less demanding in terms of resources.

demos/code_local_assistant/README.md
```diff
@@ -172,8 +255,28 @@ context:
 ![final](final.png)
 
 
-## Troubleshooting
+## Agentic behavior in action
```
mzegla:
```suggestion
## AI Agents in action
```
What do you think? "Agentic behavior in action" does not sound very natural to me.

demos/code_local_assistant/README.md
```diff
@@ -172,8 +255,28 @@ context:
...
+Continue.dev plugin is shipped with multiple built-in tools. For full list please [visit Continue documentation](https://docs.continue.dev/features/agent/how-it-works#what-tools-are-available-in-plan-mode-read-only).
+
+To use them, select Agent Mode:
+
+![select agent](./select_agent.png)
+
+Select Agentic model from model list:
```
mzegla:
```suggestion
Select model that support tool calling from model list:
```
It's more explicit I think

PR #3637, opened 2025-09-06
build ov from src

dtrawins:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``


windows_create_package.bat
```diff
@@ -13,7 +13,7 @@
 :: See the License for the specific language governing permissions and
 :: limitations under the License.
 ::
-echo off
+@echo on
```
rasapala, line 16: We want to have all logs in CI ?

windows_install_build_dependencies.bat
```diff
@@ -184,14 +192,68 @@ IF /I EXIST %BAZEL_SHORT_PATH%\openvino (
...
+    set "OV_SOURCE_ORG=e-ddykim"
+)
+if "%TOKENIZER_SOURCE_ORG%"=="" (
+    set "TOKENIZER_SOURCE_ORG=openvinotoolkit"
+)
+if "%TOKENIZER_SOURCE_BRANCH%"=="" (
+    set "TOKENIZER_SOURCE_BRANCH=master"
+)
+
+IF /I NOT EXIST openvino (
```
rasapala, line 214: Lets keep the build outside of model_server git repository.
instead of openvino change path to c:\!output_user_root!\openvino_src and use it in consecutive commands.

windows_install_build_dependencies.bat
```diff
@@ -184,14 +192,68 @@ IF /I EXIST %BAZEL_SHORT_PATH%\openvino (
...
+cmake.exe --build . --config Release --verbose -j
+if !errorlevel! neq 0 exit /b !errorlevel!
+cmake.exe --install . --config Release --prefix C:\\%output_user_root%\\openvino
+cd ..\..
+call C:\\%output_user_root%\\openvino\\setupvars.bat
+set "OVMS_DIR=C:\\%output_user_root%\\openvino"
+IF /I NOT EXIST openvino_tokenizers (
+    git clone https://github.com/%TOKENIZER_SOURCE_ORG%/openvino_tokenizers.git
+)
+cd openvino_tokenizers
```
rasapala, line 236: Lets keep the build outside of model_server git repository.
instead of openvino_tokenizerschange path to c:\!output_user_root!\openvino_tokenizers_src and use it in consecutive commands.

rasapala (commented):
Change build paths. Is it good idea to take the parameter from env variables. Maybe better to pass it as a 3rd argument to the scripts ?

windows_create_package.bat
dtrawins: it shows the commands executed. Otherwise there is no information about the fail reason.

PR #3710, opened 2025-10-17
Demo export_models --help output fix, wrong CLI param fix

dkalinowski:
### 🛠 Summary

CVS-172617

--tools_model_type has been changed to --tool_parser
and also qwen3 is not a tool parser type anymore, it is hermes3


porlows1 (approved):
Approved

PR #3714, opened 2025-10-20
Adjust ChatHistory usage and update OV deps

mzegla:


src/llm/apis/openai_completions.cpp
```diff
@@ -287,16 +287,16 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages(std::optional<std::stri
...
                     if (member->value.IsString()) {
-                        request.chatHistory.back().insert({member->name.GetString(), member->value.GetString()});
+                        request.chatHistory.last().push_back({member->name.GetString(), member->value.GetString()});
                     }
                 }
             }
         }
-        auto& lastMessage = request.chatHistory.back();
-        if (lastMessage.find("role") == lastMessage.end()) {
+        auto lastMessage = request.chatHistory.last();
```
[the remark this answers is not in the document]
mzegla: Not possible due to last() method signature. Not required as `JsonContainer` is mostly a wrapper and holds a shared pointer to the `Impl` object.

src/llm/apis/openai_completions.cpp
```diff
@@ -184,7 +184,7 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages(std::optional<std::stri
             if (member->value.IsString() && (member->name.GetString() == std::string("role") || member->name.GetString() == std::string("content"))) {
                 // Add new field to the last message in history
                 // tools handing to be done later
-                request.chatHistory.back().insert({member->name.GetString(), member->value.GetString()});
+                request.chatHistory.last()[member->name.GetString()] = member->value.GetString();
```
dkalinowski, line 187: are we sure that there is at least one chat in the history?

mzegla: in line 180 there is a push with empty object and since we never remove messages from chatHistory it always has at least one element

PR #3720, opened 2025-10-22
Removing BDBA reports after archiving it 

przepeck:
### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/ci/windows_bdba.bat
+++ b/ci/windows_bdba.bat
@@ -4,6 +4,8 @@ set "OVMS_PATH=..\%2"
 set "CONFIG_PATH=..\%3"
 cd repo_ci_infra
 
+del ovms_windows*
+
 python -m venv venv
 
 call venv\Scripts\activate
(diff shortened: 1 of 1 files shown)


ci/windows_bdba.bat
```diff
@@ -33,4 +33,5 @@ python binary_scans\ovms_bdba.py --key %BDBA_KEY% --config_dir=%CONFIG_PATH% --t
 if errorlevel 1 exit /b %errorlevel%
 
 tar -a -c -f ..\ovms_windows_bdba_reports.zip ovms_windows*
+del ovms_windows*
```
dtrawins: it would be safer to delete the folder at the beginning of stage

dtrawins (approved):

ci/windows_bdba.bat
```diff
@@ -4,6 +4,8 @@ set "OVMS_PATH=..\%2"
 set "CONFIG_PATH=..\%3"
 cd repo_ci_infra
 
+del ovms_windows*
```
dtrawins, line 7: what will happen if this folder is missing?
Shouldn't it be like :
IF EXIST ovms_windows* del ovms_windows
?

przepeck: the message will be printed 
Could Not Find c:\Jenkins\workspace\ovmsc\windows\Windows_SDL\repo_ci_infra\ovms_windows*
it won't affect step's result

PR #3730, opened 2025-10-24
Deprecated EmbeddingsCalculator removal 

dkalinowski:


dtrawins (commented):
remove also from prepare_llm_models

PR #3737, opened 2025-10-28
Fix tokenizers

rasapala:
### 🛠 Summary
Fix: 
optimum cli error: Tokenizer won't be converted.
/usr/bin/python3: Error while finding module specification for 'openvino_tokenizers.cli' (ModuleNotFoundError: No module named 'openvino_tokenizers')

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -192,7 +192,7 @@ http_archive(
 git_repository(
     name = "mediapipe",
     remote = "https://github.com/openvinotoolkit/mediapipe",
-    commit = "742e351997b339ad50fc59bca6d9a73a596bd5ed", # main as of 28/08/2025
+    commit = "b2dccf07dc3cd8ae23040d20a8ca68d35d6c75db", # main as of 06/11/2025
 )
 
 # DEV mediapipe 1 source - adjust local repository path for build
--- a/demos/common/export_models/README.md
+++ b/demos/common/export_models/README.md
@@ -16,18 +16,20 @@ python export_model.py --help
 ```
 Expected Output:
 ```console
-usage: export_model.py [-h] {text_generation,embeddings_ov,rerank,rerank_ov,image_generation} ...
+usage: export_model.py [-h] {text_generation,embeddings_ov,rerank,rerank_ov,image_generation,text2speech,speech2text} ...
 
 Export Hugging face models to OVMS models repository including all configuration for deployments
 
 positional arguments:
-  {text_generation,embeddings_ov,rerank,rerank_ov,image_generation}
+  {text_generation,embeddings_ov,rerank,rerank_ov,image_generation,text2speech,speech2text}
                         subcommand help
     text_generation     export model for chat and completion endpoints
     embeddings_ov       export model for embeddings endpoint with directory structure aligned with OpenVINO tools
     rerank              [deprecated] export model for rerank endpoint with models split into separate, versioned directories
     rerank_ov           export model for rerank endpoint with directory structure aligned with OpenVINO tools
     image_generation    export model for image generation endpoint
+    text2speech         export model for text2speech endpoint
+    speech2text         export model for speech2text endpoint
 
 options:
   -h, --help            show this help message and exit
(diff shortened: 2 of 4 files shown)


Dockerfile.redhat
```diff
@@ -219,17 +219,20 @@ ARG ov_tokenizers_org=openvinotoolkit
 ARG SDL_OPS="-fpic -O2 -U_FORTIFY_SOURCE -fstack-protector -fno-omit-frame-pointer -D_FORTIFY_SOURCE=1 -fno-strict-overflow -Wall -Wno-unknown-pragmas -Wno-error=sign-compare -fno-delete-null-pointer-checks -fwrapv -fstack-clash-protection -Wformat -Wformat-security -Werror=format-security -s -D_GLIBCXX_USE_CXX11_ABI=1 -Wno-error=deprecated-declarations -Wuninitialized"
 # hadolint ignore=DL3003
 RUN git clone https://github.com/$ov_tokenizers_org/openvino_tokenizers.git /openvino_tokenizers && cd /openvino_tokenizers && git checkout $ov_tokenizers_branch && git submodule update --init --recursive
+RUN if ! [[ $debug_bazel_flags == *"_py_off"* ]]; then \
+    mkdir -p /opt/intel/openvino/python/openvino_tokenizers/lib ; \
```
mzegla, line 223: why are we moving this logic here?

there is usage of `libopenvino_tokenizers.so` here, but tokenizer building happens in the next section, is that not a problem?

Dockerfile.ubuntu
```diff
@@ -198,18 +198,19 @@ ARG ov_tokenizers_branch=master
...
     mkdir -p /opt/intel/openvino/python/openvino_tokenizers/lib ; \
     cp -r python/* /opt/intel/openvino/python/ ; \
-    cp build/python/* /opt/intel/openvino/python/openvino_tokenizers/ ; \
     mkdir -p /opt/intel/openvino/python/openvino_tokenizers-2025.4.dist-info ; \
     echo $'Metadata-Version: 1.0\nName: openvino-tokenizers\nVersion: 2025.4\nRequires-Python: >=3.9\nRequires-Dist: openvino~=2025.4.0' > /opt/intel/openvino/python/openvino_tokenizers-2025.4.dist-info/METADATA ; \
     ln -s /ovms/lib/libopenvino_tokenizers.so /opt/intel/openvino/python/openvino_tokenizers/lib/libopenvino_tokenizers.so ; \
     fi
-
+# hadolint ignore=DL3003
+RUN if [ "$ov_use_binary" == "0" ]; then true ; else exit 0 ; fi ; \
```
mzegla, line 209: same as for redhat, why do we need to move this logic?

Dockerfile.redhat
rasapala: We only create directories and links. We need to copy the python folder for both cases - ov_us_binary=1 and =0, thus we move the logic outside.

Dockerfile.ubuntu
rasapala: We only create directories and links. We need to copy the python folder for both cases - ov_us_binary=1 and =0, thus we move the logic outside. Gen_ai bin package does not provide python tokenizers setup for python.

Dockerfile.redhat
```diff
@@ -406,7 +409,7 @@ RUN if [ -f /usr/bin/dnf ] ; then export DNF_TOOL=dnf ; echo -e "max_parallel_do
     if ! [[ $debug_bazel_flags == *"py_off"* ]]; then \
         $DNF_TOOL install -y python3-libs --setopt=install_weak_deps=0 --nodocs; \
     fi ; \
-    $DNF_TOOL install -y shadow-utils; \
+    $DNF_TOOL install -y shadow-utils which; \
```
dtrawins: why `which` is installed here?

rasapala: Required for convert_tokenizers check.

dtrawins: could we use `command -v convert_tokenizer` instead?

PR #3749, opened 2025-10-30
Allow more than one http rest & grpc listeners 

dkalinowski:
### 🛠 Summary

CVS-170537

- added support for comma separated bind address list via CLI/C-API
- changed localhost to 127.0.0.1 in docs where requests pip package is used because request localhost can introduce initial 2s delay on windows systems due to ipv6 connection try before actual ipv4 connection is estabilished
- updated security considerations

### DIFF
--- a/docs/performance_tuning.md
+++ b/docs/performance_tuning.md
@@ -146,34 +146,19 @@ $ cpupower frequency-set --min 3.1GHz
 
 ## Network Configuration for Optimal Performance
 
-When clients connect to the server using hostname resolution (particularly "localhost"), the system may attempt IPv6 resolution first before falling back to IPv4. If IPv6 is disabled, misconfigured, or unavailable, this can cause connection timeouts and delays before the IPv4 fallback occurs, which is especially noticeable when minimizing time to first token in generative AI applications.
+By default, OVMS endpoints are bound to all ipv4 addresses. On same systems, which route localhost name to ipv6 address, it might cause extra time on the client side to switch to ipv4. It can effectively results with extra 1-2s latency.
+It can be overcome by switching the API URL to `http://127.0.0.1` on the client side.
 
 To optimize network connection performance:
 
-**For local secured environments (restricted to localhost only):**
-- If dual-stack networking is configured properly, binding to IPv6 localhost is sufficient: `--grpc_bind_address ::1 --rest_bind_address ::1` (both IPv4 and IPv6 will work)
-- For systems without proper dual-stack support, specify both addresses to avoid resolution delays when clients use "localhost": `--grpc_bind_address 127.0.0.1,::1 --rest_bind_address 127.0.0.1,::1`
-- If IPv6 is disabled or not available in the environment, bind only to IPv4: `--grpc_bind_address 127.0.0.1 --rest_bind_address 127.0.0.1`
-
-**For public deployments:**
-- If dual-stack networking is configured properly, binding to IPv6 is sufficient: `--grpc_bind_address :: --rest_bind_address ::` (both IPv4 and IPv6 will work)
-- For systems without proper dual-stack support, specify both addresses: `--grpc_bind_address 0.0.0.0,:: --rest_bind_address 0.0.0.0,::`
-- Or just configure clients to connect directly to specific IP addresses (127.0.0.1 or ::1) rather than using "localhost" hostname
-
-Example for local secured access using OpenVINO Model Server binary:
-
-Linux/macOS:
-```bash
-./ovms --model_path /path/to/model --model_name resnet --port 9001 \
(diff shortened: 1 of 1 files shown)


docs/parameters.md
```diff
@@ -36,8 +36,8 @@ Configuration options for the server are defined only via command-line options a
 |---|---|---|
 | `port` | `integer` | Number of the port used by gRPC sever. |
 | `rest_port` | `integer` | Number of the port used by HTTP server (if not provided or set to 0, HTTP server will not be launched). |
-| `grpc_bind_address` | `string` | Network interface address or a hostname, to which gRPC server will bind to. Default: all interfaces: 0.0.0.0 |
-| `rest_bind_address` | `string` | Network interface address or a hostname, to which REST server will bind to. Default: all interfaces: 0.0.0.0 |
+| `grpc_bind_address` | `string` | Network interface address or a hostname, to which gRPC server will bind to. Default: all interfaces (both ipv4 and ipv6): 0.0.0.0 |
```
dkalinowski: todo: document in security considerations how to limit to 127.0.0.1,::1

src/test/ovmsconfig_test.cpp
```diff
@@ -1997,6 +1998,30 @@ TEST_F(OvmsParamsTest, hostname_ip_regex) {
     EXPECT_EQ(ovms::Config::check_hostname_or_ip("::ffff:192.0.2.128"), true);
     //  IPv4-translated IPv6 addresses
     EXPECT_EQ(ovms::Config::check_hostname_or_ip("::ffff:0:192.0.2.128"), true);
+
+    // Multiple selections
+    EXPECT_EQ(ovms::Config::check_hostname_or_ip("0.0.0.0"), true);
```
dkalinowski, line 2309: 127.0.0.1,::1 as actual e2e unit test?

docs/parameters.md
dkalinowski: change localhost to 127.0.0.1 in demos

src/test/ovmsconfig_test.cpp
dkalinowski: not all build environments support ipv6, therefore this is bad idea to make such e2e test

docs/parameters.md
dkalinowski: changed, but only in places where `requests` package is used: https://github.com/openvinotoolkit/model_server/pull/3749/commits/4ef60deda480d615ddc24ecd2a41495493ad735d

docs/security_considerations.md
```diff
@@ -13,6 +13,11 @@ docker run --rm -d --user $(id -u):$(id -g) --read-only --tmpfs /tmp -p 9000:900
 ---
 OpenVINO Model Server currently does not provide access restrictions and traffic encryption on gRPC and REST API endpoints. The endpoints can be secured using network settings like docker network settings or network firewall on the host. The recommended configuration is to place OpenVINO Model Server behind any reverse proxy component or load balancer, which provides traffic encryption and user authorization.
 
+When deploying in environments where only local access is required, administrators can configure the server to bind exclusively to localhost addresses. This can be achieved by setting the bind address to `127.0.0.1` for IPv4 or `::1` for IPv6, which restricts incoming connections to the local machine only. This configuration prevents external network access to the server endpoints, providing an additional layer of security for local development or testing environments.
```
dkalinowski, line 16: + performance guideline doc about potential 2s issue on windows

docs/performance_tuning.md
```diff
@@ -144,6 +144,38 @@ To save power, the OS can decrease the CPU frequency and increase a volatility o
 $ cpupower frequency-set --min 3.1GHz
 ```
 
+## Network Configuration for Optimal Performance
+
+When clients connect to the server using hostname resolution (particularly "localhost"), the system may attempt IPv6 resolution first before falling back to IPv4. If IPv6 is disabled, misconfigured, or unavailable, this can cause connection timeouts and delays before the IPv4 fallback occurs, which is especially noticeable when minimizing time to first token in generative AI applications.
```
dtrawins: By default, OVMS endpoints are bound to all ipv4 addresses. On same systems, which route `localhost` name to ipv6 address, it might cause extra time on the client side to switch to ipv4. It can effectively results with extra 1-2s latency.
It can be overcome by switching the API URL to http://127.0.0.1  instead.

Alternatively ipv6 can be enabled in the model server using --grpc_bind_address and --rest_bind_address. 
For example:
`--grpc_bind_address 127.0.0.1,::1 --rest_bind_address 127.0.0.1,::1`
or 
`--grpc_bind_address 0.0.0.0,:: --rest_bind_address 0.0.0.0,::`

dtrawins (approved):

docs/performance_tuning.md
dkalinowski: changed

PR #3751, opened 2025-10-30
Mistral tool parser: ignore missing beginning of tools token

dkalinowski:
### 🛠 Summary

This change forgives model when `[TOOL_CALL]` is not generated (but expected), however (valid) generations starting with `[{` will be treated as tools, which is incorrect

bfcl simple: 0.89->0.895
bfcl multiple: multiple 0.745->0.895

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,9 +75,9 @@ FUZZER_BUILD ?= 0
 #         - uncomment source build section, comment binary section
 #         - adjust binary version path - version variable is not passed to WORKSPACE file!
 
-OV_SOURCE_BRANCH ?= d19dc49a3adc1c8d76f97961452e79809d7ab615 # master 2025/10/22
-OV_TOKENIZERS_BRANCH ?= 760f9140ab1fd330d62ec171673ca0705abe6aa0 # master 2025/09/23
-OV_GENAI_BRANCH ?= 5099a65f86ded4e281519b51c2cd15f863163f71 # master 2025/10/22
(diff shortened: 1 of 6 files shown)


src/llm/io_processing/mistral/tool_parser.cpp
```diff
@@ -38,8 +38,12 @@ void MistralToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int6
     // In regular parsing, the parser will consume entire model output only if the first generated token is the beginning of tools token.
     // In immediate parsing, the parser will consume entire model output regardless of the first token.
     if (generatedTokens[0] != this->botTokenId && !immediateParsingEnabled) {
-        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Begin of tools token has not been found in the model output. Exiting parser.");
-        return;
+        if (parsedOutput.content.size() >= 2 && parsedOutput.content[0] == '[' && parsedOutput.content[1] == '{') {
```
mzegla, line 41: Can't we rely on `generatedTokens`? We already base on checking it in condition above, so why not keep it continuous?

dkalinowski: I don't think its good idea, [{ is regular text (non special token)
Do you want me to handle all cases where:
- [ and { are separate tokens? x, y
- [{ is glued together into 1 token? x

mzegla: No, in that case I think it's fine.

PR #3778, opened 2025-11-07
Fix non python, source build in Dockerfiles

mzegla:


src/test/http_rest_api_handler_test.cpp
```diff
@@ -1315,7 +1315,7 @@ TEST_F(ConfigStatus, isAuthorized) {
     ovms::Server& ovmsServer = ovms::Server::instance();
     std::string contents;
     auto fs = std::make_shared<ovms::LocalFileSystem>();
-    fs->readTextFile(getGenericFullPathForSrcTest("/ovms/src/test/mediapipe/config_mediapipe_add_adapter_full.json"), &contents);
+    fs->readTextFile(getGenericFullPathForSrcTest("/ovms/src/test/configs/config_cpu_dummy.json"), &contents);
```
dkalinowski, line 1318: why is that?

mzegla: So this test can work with build with mediapipe disabled

PR #3785, opened 2025-11-13
Support device properties in plugin config

michalkulakowski:
### 🛠 Summary

CVS-171019

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/json_parser.cpp
+++ b/src/json_parser.cpp
@@ -31,7 +31,7 @@ namespace ovms {
 
 using plugin_config_t = std::map<std::string, ov::Any>;
 
-bool validType(const rapidjson::Value::ConstMemberIterator& node){
+bool validType(const rapidjson::Value::ConstMemberIterator& node) {
     return (node->value.IsString() || node->value.IsBool() || node->value.IsInt64() || node->value.IsDouble());
 }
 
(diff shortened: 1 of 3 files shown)


src/json_parser.cpp
```diff
@@ -46,7 +46,30 @@ Status JsonParser::parsePluginConfig(const rapidjson::Value& node, plugin_config
     }
 
     for (auto it = node.MemberBegin(); it != node.MemberEnd(); ++it) {
-        if (it->value.IsString()) {
+        if (it->value.IsObject() && it->name.GetString() == std::string("DEVICE_PROPERTIES")) {
```
dkalinowski, line 53: Please add unit test if possible

src/json_parser.cpp
```diff
@@ -46,7 +46,30 @@ Status JsonParser::parsePluginConfig(const rapidjson::Value& node, plugin_config
...
+                    for (auto propertyIt = deviceProperties.MemberBegin(); propertyIt != deviceProperties.MemberEnd(); ++propertyIt) {
+                        if(propertyIt->value.IsString()){
+                            properties[propertyIt->name.GetString()] = propertyIt->value.GetString();
+                        }
+                        if(propertyIt->value.IsInt64()){
+                            properties[propertyIt->name.GetString()] = propertyIt->value.GetInt64();
+                        }
+                        if(propertyIt->value.IsDouble()){
+                            properties[propertyIt->name.GetString()] = propertyIt->value.GetDouble();
+                        }
```
mzegla: Maybe we can make it `if..else if...else...` block with `else` returning error if property is of unsupported type (array, object, bool)? btw. shouldn't we also accept some of those types - boolean for example sounds like a reasonable data type for a property.

src/json_parser.cpp
mzegla, line 53: General comment to below code. I see occurrences of code like:
```
deviceProperties[propertiesIt->name.GetString()] = ov::AnyMap{};
auto& properties = deviceProperties[propertiesIt->name.GetString()].as<ov::AnyMap>();
```
I wonder if it wouldn't be cleaner to operate on ov maps and then put them inside original objects like:
```
auto propertiesMap = ov::AnyMap{};
// Fill propertiesMap with values...
deviceProperties[propertiesIt->name.GetString()] = propertiesMap;
```

mzegla (approved):

dkalinowski (approved):

PR #3810, opened 2025-11-21
Add note about system message handling in phi4 model

mzegla:


demos/continuous_batching/agentic_ai/README.md
```diff
@@ -83,6 +83,22 @@ curl -L -o models/mistralai/Mistral-7B-Instruct-v0.3/chat_template.jinja https:/
 python export_model.py text_generation --source_model microsoft/Phi-4-mini-instruct --weight-format int8 --config_file_path models/config.json --model_repository_path models --tool_parser phi4 --cache_size 2
 curl -L -o models/microsoft/Phi-4-mini-instruct/chat_template.jinja https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/releases/2025/3/extras/chat_template_examples/chat_template_phi4_mini.jinja
 ```
+**Note**: Recommended Phi4-mini-instruct template for tool calling does not include system message passed in messages with role "system". If you want to set specific system message, you can do it via chat template argument `system_message` set in the request body. For example:
```
dtrawins, line 86: why don't we change the chat template instead?

PR #3845, opened 2025-12-08
[2025.4.1] Ignore usage statistics (0 tokens) for legacy text generation pipelines

dkalinowski:
### 🛠 Summary

Previously there was a fail during last streaming response
Now it streams correctly, but with usage=0 since it is not supported
It enables use of Continue.dev plugin with legacy pipelines

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,7 +275,8 @@ $(ACTIVATE):
 
 $(ACTIVATE_STYLE):
 	@echo "Updating virtualenv dependencies in: $(VIRTUALENV_STYLE_DIR)..."
-	@test -d $(VIRTUALENV_STYLE_DIR) || python3 -m venv $(VIRTUALENV_STYLE_DIR)
+	@python3 -m pip install virtualenv
+	@test -d $(VIRTUALENV_STYLE_DIR) || $(VIRTUALENV_EXE) $(VIRTUALENV_STYLE_DIR)
 	@. $(ACTIVATE_STYLE); pip3 install --upgrade pip
 	@. $(ACTIVATE_STYLE); pip3 install -vUqq "setuptools<80"
 	@. $(ACTIVATE_STYLE); pip3 install -qq -r ci/style_requirements.txt


src/llm/language_model/legacy/servable.cpp
```diff
@@ -206,11 +201,11 @@ absl::Status LegacyServable::preparePartialResponse(std::shared_ptr<GenAiServabl
             executionContext->response = wrapTextInServerSideEventMessage(serializedChunk);
         }
 
-        // Disabling usage in streaming mode in legacy servable due to the issue with token counting.
+        // Fake usage in streaming mode in legacy servable due to the issue with token counting.
```
mzegla: TODO or FIXME for this? 
We definitely don't want to leave such comment here for a longer period.

src/test/llm/llmnode_test.cpp
```diff
@@ -2049,26 +2049,31 @@ TEST_P(LLMFlowHttpTestParameterized, streamChatCompletionsUsage) {
...
 
+        ASSERT_GT(responses.size(), 0);
         ASSERT_TRUE(responses.back().find("\"completion_tokens\":5") != std::string::npos);
         ASSERT_TRUE(responses.back().find("\"prompt_tokens\"") != std::string::npos);
         ASSERT_TRUE(responses.back().find("\"total_tokens\"") != std::string::npos);
         if (params.checkFinishReason) {
             ASSERT_TRUE(responses.back().find("\"finish_reason\":\"length\"") != std::string::npos);
         }
-        // For non-continuous batching servables usage is not supported
+        // For non-continuous batching servables usage is faked, always returns 0
```
mzegla: I don't know if I would use word `fake` - maybe `fixed` would sound better. And I would keep `not supported`, because that is the actual state of that feature.

src/llm/language_model/legacy/servable.cpp
dtrawins:
```suggestion
        // Usage in streaming mode in legacy servable is not supported yet hence we report 0 as a missing metric.
```

src/test/llm/llmnode_test.cpp
dkalinowski: fixed

src/llm/language_model/legacy/servable.cpp
dkalinowski: fixed

PR #3849, opened 2025-12-08
Tokenize endpoint for LLMs and Visual models

przepeck:
### 🛠 Summary

[CVS-177520](https://jira.devtools.intel.com/browse/CVS-177520)
Adding tokenize endpoint same as for embeddings -https://github.com/openvinotoolkit/model_server/blob/main/demos/embeddings/README.md#usage-of-tokenize-endpoint-release-20254-or-weekly

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/servable.cpp
+++ b/src/llm/servable.cpp
@@ -173,7 +173,12 @@ absl::Status GenAiServable::prepareInputs(std::shared_ptr<GenAiServableExecution
 #else
         ov::genai::ChatHistory& chatHistory = executionContext->apiHandler->getChatHistory();
         constexpr bool add_generation_prompt = true;  // confirm it should be hardcoded
-        inputText = getProperties()->tokenizer.apply_chat_template(chatHistory, add_generation_prompt);
+        try {
+            inputText = getProperties()->tokenizer.apply_chat_template(chatHistory, add_generation_prompt);
+        } catch (const std::exception& e) {
+            SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Failed to apply chat template: {}", e.what());
+            return absl::Status(absl::StatusCode::kInvalidArgument, "Failed to apply chat template. The model either does not have chat template or has an invalid one.");
+        }
 #endif
         if (inputText.size() == 0) {
             return absl::Status(absl::StatusCode::kInvalidArgument, "Final prompt after applying chat template is empty");
--- a/ci/lib_search.py
+++ b/ci/lib_search.py
@@ -153,6 +153,7 @@ def check_dir(start_dir):
         "windows_sign.bat",
         "config.yaml",
         "kserve-openvino.yaml",
+        "dummy_facebook_template.jinja",
         ]
 
     exclude_directories = ['/dist/', 'release_files/thirdparty-licenses', 'extras/chat_template_examples']
--- a/demos/audio/README.md
+++ b/demos/audio/README.md
@@ -195,7 +195,7 @@ Transcript file that was previously generated with audio/speech endpoint.
 
 
 ```bash
-curl http://localhost:8000/v3/audio/transcriptions -H "Content-Type: multipart/form-data" -F file="@speech.wav" -F model="OpenVINO/whisper-large-v3-fp16-ov"
+curl http://localhost:8000/v3/audio/transcriptions -H "Content-Type: multipart/form-data" -F file="@speech.wav" -F model="openai/whisper-large-v3-turbo"
 ```
 ```json
 {"text": " The quick brown fox jumped over the lazy dog."}
@@ -217,7 +217,7 @@ client = OpenAI(base_url=url, api_key="not_used")
 
 audio_file = open(filename, "rb")
 transcript = client.audio.transcriptions.create(
-  model="OpenVINO/whisper-large-v3-fp16-ov",
+  model="openai/whisper-large-v3-turbo",
   file=audio_file
 )
 
--- a/demos/code_local_assistant/README.md
+++ b/demos/code_local_assistant/README.md
@@ -105,7 +105,7 @@ Pull and add the model on Windows:
 ::::{tab-set}
 :::{tab-item} Qwen/Qwen3-Coder-30B-A3B-Instruct
 :sync: Qwen/Qwen3-Coder-30B-A3B-Instruct
-```bash
+```bat
 python export_model.py text_generation --source_model Qwen/Qwen3-Coder-30B-A3B-Instruct --weight-format int8 --config_file_path models/config_all.json --model_repository_path models --target_device GPU --tool_parser qwen3coder --cache_size 2 --overwrite_models
 curl -L -o models/Qwen/Qwen3-Coder-30B-A3B-Instruct/chat_template.jinja https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/main/extras/chat_template_examples/chat_template_qwen3coder_instruct.jinja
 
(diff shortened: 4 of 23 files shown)


src/tokenize/tokenize_parser.cpp
```diff
@@ -45,7 +45,7 @@ absl::Status TokenizeParser::parseTokenizeResponse(rapidjson::StringBuffer& buff
         int64_t* attentionMaskPtr = reinterpret_cast<int64_t*>(attentionMaskTensor.data()) + batchIterator * size;
         for (size_t i = 0; i < size; ++i) {
             if (attentionMaskPtr[i] == 0 && !pad_to_max_length) {
-                break;
+                continue;
```
mzegla, line 48: Why do we change that? Do we expect that once we see first 0 in the attention mask any further elements can be different (non-zero)?

src/test/llm/llmnode_test.cpp
```diff
@@ -4530,4 +4536,381 @@ TEST_F(IsolatedServableTests, PromtSizeBetweenDefaultAndNonDefaultMaxPromptLenNP
     ASSERT_EQ(status, absl::OkStatus());
 }
 
-// TODO: Add missing tests for reading max prompt len property from configuration
+
+class LLMHttpTokenizeTest : public ::testing::Test {
```
mzegla: Could you try extracting those tests to new file? like `src/test/llm/tokenize_endpoint_test.cpp`? 
This file is too large already and I would prefer not to expand it if possible...

src/llm/http_llm_calculator.cc
```diff
@@ -90,7 +91,36 @@ class HttpLLMCalculator : public CalculatorBase {
...
+                    StringBuffer responseBuffer;
+                    auto responseStatus = ovms::TokenizeParser::parseTokenizeResponse(responseBuffer, tokens, tokenizeRequest.parameters);
+
+                    if (!responseStatus.ok()) {
+                        return responseStatus;
+                    }
+                    cc->Outputs().Tag(OUTPUT_TAG_NAME).Add(new std::string(responseBuffer.GetString()), iterationBeginTimestamp);
+                    return absl::OkStatus();
+                }
+                auto status = servable->loadRequest(executionContext, payload);
```
mzegla: I would prefer if this logic would be more attached to the general flow like:

```
auto status = servable->loadRequest(executionContext, cc->Inputs().Tag(INPUT_TAG_NAME).Get<ovms::HttpPayload>());
if (status != absl::OkStatus())
    return status;
    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "LLMCalculator  [Node: {}] Request loaded successfully", cc->NodeName());

if (executionContext.endpoint == ovms::Endpoint::TOKENIZE) {
    return servable->processTokenizeRequest(executionContext.payload.parsedJson);
```

src/tokenize/tokenize_parser.cpp
przepeck: I encountered the issue that with batched data, attention mask was starting with 0 and then the output was correct

src/llm/servable.cpp
```diff
@@ -63,13 +64,46 @@ absl::Status GenAiServable::loadRequest(std::shared_ptr<GenAiServableExecutionCo
...
 
+absl::Status GenAiServable::processTokenizeRequest(std::shared_ptr<GenAiServableExecutionContext>& executionContext)
+{
+    ovms::TokenizeRequest tokenizeRequest;
+    auto status = ovms::TokenizeParser::parseTokenizeRequest(*executionContext->payload.parsedJson, tokenizeRequest);
+    if (status != absl::OkStatus()) {
+        return status;
+    }
+    auto tokenizer = getProperties()->tokenizer;
+    auto input = tokenizeRequest.input;
```
mzegla: is it needed? looks like `input` is only used in one place so maybe we could only use `tokenizeRequest` and void making copy?

src/llm/http_llm_calculator.cc
```diff
@@ -95,6 +95,18 @@ class HttpLLMCalculator : public CalculatorBase {
                     return status;
                 SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "LLMCalculator  [Node: {}] Request loaded successfully", cc->NodeName());
 
+                if (executionContext->endpoint == Endpoint::TOKENIZE) {
```
mzegla, line 99: Could you add a comment above this like saying that for tokenize endpoint we do not execute full servable path and return early as we don't need additional parsing or streaming handling etc. 
I think for the future reference it would make it clear that the whole execution ends within that if block.

src/llm/servable.hpp
```diff
@@ -190,6 +190,13 @@ class GenAiServable {
     Base implementation uses textStreamer to create text chunk, attempts to serialize it, and sets sendLoopbackSignal according to generation status.
     */
     virtual absl::Status preparePartialResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext);
+
+    // ----------- Tokenize scenario ------------
+    /*
+    processTokenizeRequest method should implement tokenization of the input text provided in executionContext payload.
+    Implementation MUST fill executionContext response field.
+    */
+    virtual absl::Status processTokenizeRequest(std::shared_ptr<GenAiServableExecutionContext>& executionContext);
```
mzegla: I don't think we need this in the interface section. Tokenizer is currently a necessary element of every pipeline under GenAiServable, so child servables should not be expected to implement that method as of now. 
I think this could be moved up, under determineDecodingMethod with comment modified to avoid words like "should" or "must", so:

processTokenizeRequest method **implements** tokenization ...
Implementation **fills** executionContext response field.

For the second sentence I would also shortly say what is in executionContext.response field after the method is executed. Like is it some specific format (wrapped in json) or raw numbers etc.

src/test/llm/tokenize_endpoint_test.cpp
```diff
@@ -0,0 +1,426 @@
...
+    ASSERT_TRUE(tokens.IsArray());
+    ASSERT_EQ(tokens.Size(), 2);
+    for (rapidjson::SizeType i = 0; i < tokens.Size(); ++i) {
+        const auto& tokenArray = tokens[i];
+        ASSERT_TRUE(tokenArray.IsArray());
+        ASSERT_EQ(tokenArray[0].GetInt(), params.paddingTokenId);
+        ASSERT_EQ(tokenArray.Size(), 25);
+    }
+}
+
```
mzegla, line 417: Looks like there is no test for `add_special_tokens=true`

przepeck: This is problematic for vison model used here since special tokens are reserved here for images 🤔 I can implement this method only for lm models.

docs/llm/quickstart.md
```diff
@@ -161,6 +161,40 @@ Expected output:
 Paris, the charming City of Light, is renowned for its rich history, iconic landmarks, architectural splendor, and artistic
 ```
 
+## Tokenization (release 2025.4.1 or weekly)
```
dkalinowski: Is this PR going to get merged for releases/2025/4 branch as well?

docs/llm/quickstart.md
```diff
@@ -161,6 +161,40 @@ Expected output:
...
+
+The `tokenize` endpoint provides a simple API for tokenizing input text using the same tokenizer as the deployed LLM. This allows you to see how your text will be split into tokens before feature extraction or inference. The endpoint accepts a string or list of strings and returns the corresponding token IDs.
```
dkalinowski: I think we are missing here which kinds of graphs are currently supported. What do you think?

przepeck: Yes, I will add this

docs/llm/quickstart.md
przepeck: I asked and it won't. I will change it to 2026.0

src/llm/http_llm_calculator.cc
```diff
@@ -95,6 +95,19 @@ class HttpLLMCalculator : public CalculatorBase {
                     return status;
                 SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "LLMCalculator  [Node: {}] Request loaded successfully", cc->NodeName());
 
+                // Tokenize endpoint doesn't require full full servable path and it ends workflow after tokenization, it does not need additional processing
```
mzegla:
```suggestion
                // Tokenize endpoint doesn't require full servable path and it ends workflow after tokenization, it does not need additional processing
```

mzegla (approved):

PR #3851, opened 2025-12-09
devstral tool parser for tool calling

dtrawins:
### 🛠 Summary

CVS-177455
Added tool parser for devstral model.

### 🧪 Checklist

- [ ] Unit tests added.
``


src/llm/io_processing/devstral/tool_parser.hpp
```diff
@@ -0,0 +1,80 @@
...
+    int toolCallIndex = -1;
+    std::string streamContent = "";  // content accumulated from stream chunks
+    std::string toolName = "";
+    std::optional<rapidjson::Document> sendFullDelta(ToolCall& toolCall);
+
+public:
+    DevstralToolParser() = delete;
+    DevstralToolParser(ov::genai::Tokenizer& tokenizer, const ToolsSchemas_t& toolSchemas) :
+        BaseOutputParser(tokenizer),
+        argsTokenId(tokenizer.encode("[ARGS]", {{"add_special_tokens", false}}).input_ids.data<int64_t>()[0]),
```
dkalinowski: how are we ensured that [ARGS] / [TOOL_CALLS] are single tokens, treated as special, not as string, for example `[AR` and `GS]`?

src/llm/io_processing/devstral/tool_parser.hpp
```diff
@@ -0,0 +1,80 @@
...
+        botTokenId(tokenizer.encode("[TOOL_CALLS]", {{"add_special_tokens", false}}).input_ids.data<int64_t>()[0]),
```
dkalinowski: validate if input_ids token count is == 1?

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "First tool token index is greater than first args token index.");
+        return;
+    }
+    std::vector<int64_t> tool_name_tokens(generatedTokens.begin() + (firstToolTokenIndex + 1), generatedTokens.begin() + (firstArgsTokenIndex));
+    std::vector<int64_t> arguments_tokens(generatedTokens.begin() + (firstArgsTokenIndex + 1), generatedTokens.end());
+
+    ToolCall toolCall;
+    std::string tool_name = tokenizer.decode(tool_name_tokens, ov::AnyMap{ov::genai::skip_special_tokens(true)});
+    if (this->toolSchemas.find(tool_name) == this->toolSchemas.end()) {
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Tool name '{}' not valid.", tool_name);
```
dkalinowski: this is behavior we havent implemented in other parsers, is it really worth return early? if we return function name that is not part of the toolschemas spec, we might be able to debug it in bfcl

src/llm/io_processing/devstral/tool_parser.hpp
dtrawins: Those are a special tokens.

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+    this->streamContent += chunk;
+    if (this->internalState == AWAITING_START_TAG) {
+        size_t pos = chunk.find("[TOOL_CALLS]");
+        if (pos != std::string::npos) {
+            this->internalState = AWAITING_ARGS_TAG;
+            this->toolCallIndex++;
+            if (pos == 0) {
+                this->streamContent.clear();
+            } else {
+                this->streamContent = this->streamContent.substr(pos + 13);  // "[TOOLS_CALLS]" length is 13
```
dkalinowski: we should avoid magic numbers, this way if we change `this->streamingParsingToolCallsStartTag` to another value, this part will be incorrect

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+        if (end_pos != std::string::npos) {
+            arguments = this->streamContent.substr(0, end_pos);
+        } else {
+            arguments = this->streamContent;
+        }
+        if (!arguments.empty()) {
+            ToolCall toolCall;
+            toolCall.arguments = arguments;
+            toolCall.name = this->toolName;
+            return sendFullDelta(toolCall);
```
dkalinowski: shouldntg we stream partial function argument chunks? if i understand correctly you send full delta at the end of generation

src/llm/io_processing/output_parser.hpp
```diff
@@ -87,7 +87,7 @@ class OutputParser {
     std::optional<rapidjson::Document> parseChunk(const std::string& chunkResponse, const bool toolsAvailable, ov::genai::GenerationFinishReason finishReason);
 
     bool requiresStreamingWithSpecialTokens() const {
-        return (reasoningParser && reasoningParser->requiresStreamingWithSpecialTokens()) &&
+        return (reasoningParser && reasoningParser->requiresStreamingWithSpecialTokens()) ||
                (toolParser && toolParser->requiresStreamingWithSpecialTokens());
```
dkalinowski: @mzegla when i implemented it i remember your comment why it should really be `&&` instead of `||`, do you remember what was the reason?

src/test/llm/output_parsers/devstral_output_parser_test.cpp
```diff
@@ -0,0 +1,303 @@
...
+    EXPECT_EQ(parsedOutput.content, "");
+    EXPECT_EQ(parsedOutput.reasoning, "");
+    ASSERT_EQ(parsedOutput.toolCalls.size(), 1);
+    EXPECT_EQ(parsedOutput.toolCalls[0].name, "example_tool");
+    EXPECT_EQ(parsedOutput.toolCalls[0].arguments, "{\"arg1\":\"value1\",\"arg2\":42}");
+    EXPECT_EQ(parsedOutput.toolCalls[0].id.empty(), false);
+}
+
+TEST_F(DevstralOutputParserTest, ParseToolCallOutputWithSingleToolCall_MissingEndTag) {
+    std::string testInput = "Reasoninig before tool call [TOOL_CALLS]example_tool[ARGS]{\"arg1\":\"value1\",\"arg2\":42}";
```
dkalinowski: can you add test for scenarios with whitespace between the tags? i saw other models often put spaces or new lines before/after the function name

src/test/llm/output_parsers/devstral_output_parser_test.cpp
```diff
@@ -0,0 +1,303 @@
...
+        {"example", ov::genai::GenerationFinishReason::NONE, R"({"delta":{"content":"example"}})"},
+        {"[TOOL_CALLS]", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {"get", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {"_", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {"weather", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {"[ARGS]", ov::genai::GenerationFinishReason::NONE, R"({"delta":{"tool_calls":[{"id":"XXXXXXXXX","type":"function","index":0,"function":{"name":"get_weather"}}]}})"},
+        {"{\"", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {"city\":", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {" \"Paris", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        // Last chunk is added in the for loop below
```
dkalinowski, line 212: missing test for enclosing </s> scenario

src/llm/io_processing/devstral/tool_parser.hpp
dkalinowski: that doesnt answer my question

src/llm/io_processing/devstral/tool_parser.cpp
dkalinowski: you can look up how Adrian handles that in qwen coder 
```
this->lastProcessedPosition = pos + Qwen3CoderToolParser::PARAMETER_END_TAG.length();
```

src/llm/io_processing/devstral/tool_parser.hpp
mzegla: agreed, we could also do that for `argsTokenId` and fail if in any of those cases we have more than one token from encoding

src/llm/io_processing/devstral/tool_parser.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include "src/llm/apis/tool_schema_wrapper.hpp"
+
+namespace ovms {
+class DevstralToolParser : public BaseOutputParser {
+    const int64_t argsTokenId;  // [ARGS]
+    const int64_t botTokenId;   // [TOOL_CALLS]
+
+    // in streaming mode we can rely on tags in string format as tokens are not available
+    const std::string streamingParsingArgsStartTag = "[ARGS]";
+    const std::string streamingParsingToolCallsStartTag = "[TOOL_CALLS]";
```
mzegla: Those tags/tokens are not specific to streaming, so I think we can drop `streamingParsing` prefix
Those are variables are about the same thing - please unify naming:
either botToken or toolCallsStart:
`toolCallsStartTokenId, toolCallsStartTag` or
`botTokenId, botTag`
and either argsTokenId or ArgsStartTag:
`argsStartTokenId, argsStartTag` or 
`argsTokenId, argsTag`

src/llm/io_processing/output_parser.hpp
mzegla: This has been implemented this way to make sure we don't allow two parsers with different special tokens approach as they will receive the same model output, so they must both either require it or not. 
I guess for this case, where we don't have reasoning parser but want to require special tokens for tool parser we should modify this function like:
```
if (!reasoningParser) {
	return toolParser && toolParser->requiresStreamingWithSpecialTokens();
} else if (!toolParser) {
	return reasoningParser && reasoningParser->requiresStreamingWithSpecialTokens();
} else {
	return (reasoningParser && reasoningParser->requiresStreamingWithSpecialTokens()) &&
    	(toolParser && toolParser->requiresStreamingWithSpecialTokens());
}
```

src/test/llm/output_parsers/devstral_output_parser_test.cpp
```diff
@@ -0,0 +1,303 @@
...
+    std::string input = "[TOOL_CALLS]wrong_name[ARGS]{ \"filepath\": \"/var/log/db.log\"}</s>";
+    std::string testInput = input;
+    auto generatedTensor = devstralTokenizer->encode(testInput, ov::genai::add_special_tokens(false)).input_ids;
+    std::vector<int64_t> generatedTokens(generatedTensor.data<int64_t>(), generatedTensor.data<int64_t>() + generatedTensor.get_size());
+    ParsedOutput parsedOutput = outputParserWithRegularToolParsing->parse(generatedTokens, true);
+    EXPECT_EQ(parsedOutput.content, "wrong_name{ \"filepath\": \"/var/log/db.log\"}");
+    EXPECT_EQ(parsedOutput.reasoning, "");
+    ASSERT_EQ(parsedOutput.toolCalls.size(), 0);
+}
+
```
mzegla: could you also add test for coding-like output? see:
https://github.com/openvinotoolkit/model_server/blob/main/src/test/llm/output_parsers/qwen3_output_parser_test.cpp#L311

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "../../../logging.hpp"
+#include "tool_parser.hpp"
+#include "../utils.hpp"
+#include "src/stringutils.hpp"
+
+namespace ovms {
+
+void DevstralToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int64_t>& generatedTokens) {
+    std::vector<std::string> tools;
+    // Parser will consume entire model output only if the first generated token is the beginning of tools token.
```
mzegla: Does not look like this comment is true for this parser

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+    size_t firstToolTokenIndex;
+    auto it = std::find(generatedTokens.begin(), generatedTokens.end(), this->botTokenId);
+    if (it != generatedTokens.end()) {
+        firstToolTokenIndex = std::distance(generatedTokens.begin(), it);
+    } else {
+        return;
+    }
+
+    size_t firstArgsTokenIndex;
+    auto it_args = std::find(generatedTokens.begin() + firstToolTokenIndex, generatedTokens.end(), this->argsTokenId);
```
mzegla: Please keep the naming convention
```suggestion
    auto itArgs = std::find(generatedTokens.begin() + firstToolTokenIndex, generatedTokens.end(), this->argsTokenId);
```

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+        firstArgsTokenIndex = std::distance(generatedTokens.begin(), it_args);
+    } else {
+        return;
+    }
+    if (firstToolTokenIndex > firstArgsTokenIndex) {
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "First tool token index is greater than first args token index.");
+        return;
+    }
+    std::vector<int64_t> tool_name_tokens(generatedTokens.begin() + (firstToolTokenIndex + 1), generatedTokens.begin() + (firstArgsTokenIndex));
+    std::vector<int64_t> arguments_tokens(generatedTokens.begin() + (firstArgsTokenIndex + 1), generatedTokens.end());
```
mzegla:
```suggestion
    std::vector<int64_t> toolNameTokens(generatedTokens.begin() + (firstToolTokenIndex + 1), generatedTokens.begin() + (firstArgsTokenIndex));
    std::vector<int64_t> argumentsTokens(generatedTokens.begin() + (firstArgsTokenIndex + 1), generatedTokens.end());
```

src/llm/io_processing/devstral/tool_parser.cpp
mzegla: This is not in line with current behavior in other parsers. I wouldn't do that check if it's only for this parser. Either drop it or create a task for alignment of other parsers.

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    ToolCall toolCall;
+    std::string tool_name = tokenizer.decode(tool_name_tokens, ov::AnyMap{ov::genai::skip_special_tokens(true)});
```
mzegla:
```suggestion
    std::string toolName = tokenizer.decode(tool_name_tokens, ov::AnyMap{ov::genai::skip_special_tokens(true)});
```

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+    std::string arguments = tokenizer.decode(arguments_tokens, ov::AnyMap{ov::genai::skip_special_tokens(true)});
+
+    toolCall.name = tool_name;
+    toolCall.arguments = arguments;
+    toolCall.id = generateRandomId();  // Generate a random ID for the tool call
+    parsedOutput.toolCalls.push_back(toolCall);
+
+    // get subset of generatedTokens starting from begin() to firstArgsTokenIndex
+    std::vector<int64_t> content_tokens;
```
mzegla:
```suggestion
    std::vector<int64_t> contentTokens;
```

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+        AWAITING_START_TAG,
+        AWAITING_ARGS_TAG,
+        PROCESSING_ARGS
+
+    We store the history of chunks in streamContent string. After state changes are detected, we clear the streamContent to only keep unprocessed part.
+    */
+
+    this->streamContent += chunk;
+    if (this->internalState == AWAITING_START_TAG) {
+        size_t pos = chunk.find("[TOOL_CALLS]");
```
mzegla: I think we should look up on `streamContent`

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+                this->streamContent.clear();
+            } else {
+                this->streamContent = this->streamContent.substr(pos + 13);  // "[TOOLS_CALLS]" length is 13
+            }
+        } else {
+            return std::nullopt;
+        }
+    }
+    if (this->internalState == AWAITING_ARGS_TAG) {
+        // check if [ARGS] tag is present in the chunk and update state accordingly
```
mzegla:
```suggestion
        // check if [ARGS] tag is present in the streamContent and update state accordingly
```

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+        size_t pos = this->streamContent.find("[ARGS]");
+        if (pos != std::string::npos) {
+            this->internalState = PROCESSING_ARGS;
+            this->toolName = this->streamContent.substr(0, pos);
+            if (this->toolSchemas.find(this->toolName) == this->toolSchemas.end()) {
```
mzegla: As for the unary part - this check is unique to this parser and I don't think it's a good idea to have different behavior for different parsers. Either remove or create a task for alignment of other parsers.

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+                SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Tool name '{}' not valid.", this->toolName);
+                return std::nullopt;
+            }
+            this->streamContent = this->streamContent.substr(pos + 6);  // "[ARGS]" length is 6
```
mzegla: Magic number

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,163 @@
...
+            return wrapFirstDelta(this->toolName, this->toolCallIndex);
+        } else {
+            return std::nullopt;
+        }
+    }
+    if (finishReason != ov::genai::GenerationFinishReason::NONE) {
+        size_t end_pos = this->streamContent.find("</s>");
```
mzegla: What is this token? If it has some significant value for the parsing it should be a member of the parser class like args and tool calls token. Also:
```suggestion
        size_t endPos = this->streamContent.find("</s>");
```

src/llm/io_processing/devstral/tool_parser.cpp
mzegla: We already accepted such approach for qwen3 coder, so I suppose we can have it in other parsers as well unless there are specific requirements for "real" streaming.

src/llm/io_processing/devstral/tool_parser.cpp
dtrawins: first chunk starts with [TOOL_CALLS] and it is never split as it is one token

src/llm/io_processing/devstral/tool_parser.cpp
dtrawins: technically we check streamContent but it will be the case only if [ARGS] is added in the chunk. Otherwise it would be different state

src/llm/io_processing/devstral/tool_parser.hpp
dtrawins: devstral parser is setting requiresStreamingWithSpecialTokens() as true

windows_prepare_llm_models.bat
```diff
@@ -202,4 +203,17 @@ if not exist "%~1\%GPTOSS_MODEL%\%TOKENIZER_FILE%" (
   exit /b 1
 )
 
+if exist "%~1\%DEVSTRAL_MODEL%\%TOKENIZER_FILE%" (
```
atobiszei: I think this whole section in bat file could be replace with sth like:
```
:: ===========================
:: Tokenizer-only model calls
:: ===========================

call :EnsureTokenizer "%~1" "%QWEN3_MODEL%"   "%TOKENIZER_FILE%" "Qwen3"
call :EnsureTokenizer "%~1" "%LLAMA3_MODEL%"  "%TOKENIZER_FILE%" "Llama3.1"
call :EnsureTokenizer "%~1" "%HERMES3_MODEL%" "%TOKENIZER_FILE%" "Hermes3"
call :EnsureTokenizer "%~1" "%PHI4_MODEL%"    "%TOKENIZER_FILE%" "Phi-4"
call :EnsureTokenizer "%~1" "%MISTRAL_MODEL%" "%TOKENIZER_FILE%" "Mistral"
call :EnsureTokenizer "%~1" "%GPTOSS_MODEL%"  "%TOKENIZER_FILE%" "GPT-OSS"
call :EnsureTokenizer "%~1" "%DEVSTRAL_MODEL%" "%TOKENIZER_FILE%" "Devstral"

endlocal
exit /b 0

:: =========================================================
:: Function: EnsureTokenizer
:: %1 = base directory
:: %2 = model name
:: %3 = tokenizer file
:: %4 = display name
:: =========================================================
:EnsureTokenizer
if exist "%~1\%~2\%~3" (
  echo Models file %~1\%~2\%~3 exists. Skipping downloading models.
) else (
  echo Downloading tokenizer and detokenizer for %~4 model to %~1\%~2 directory.
  mkdir "%~1\%~2" 2>nul
  convert_tokenizer "%~2" --with_detokenizer -o "%~1\%~2"
  if errorlevel 1 exit /b %errorlevel%
)

if not exist "%~1\%~2\%~3" (
  echo Models file %~1\%~2\%~3 does not exist.
  exit /b 1
)

exit /b 0
```

src/test/llm/output_parsers/devstral_output_parser_test.cpp
atobiszei: Add test for handling special characters - ie sth like python script example:
https://github.com/openvinotoolkit/model_server/blob/main/src/test/llm/output_parsers/qwen3coder_output_parser_test.cpp#L748

src/test/llm/output_parsers/devstral_output_parser_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+// limitations under the License.
+//*****************************************************************************
+#include <gtest/gtest.h>
+#include <openvino/genai/tokenizer.hpp>
+#include <string>
+#include <vector>
+
+#include "../../../llm/io_processing/base_output_parser.hpp"
+#include "../../../llm/io_processing/output_parser.hpp"
+#include "../../platform_utils.hpp"
```
atobiszei:
```suggestion
#include "src/llm/io_processing/base_output_parser.hpp"
#include "src/llm/io_processing/output_parser.hpp"
#include "src/platform_utils.hpp"
```

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+//*****************************************************************************
+
+#include <openvino/genai/tokenizer.hpp>
+#include <string>
+#include <vector>
+#include <regex>
+
+#include "src/port/rapidjson_document.hpp"
+
+#include "../../../logging.hpp"
```
atobiszei:
```suggestion
#include "src/logging.hpp"
```

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include "tool_parser.hpp"
+#include "../utils.hpp"
```
atobiszei:
```suggestion
#include "src/utils.hpp"
```

src/test/llm/output_parsers/devstral_output_parser_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+        {"Reasoning", ov::genai::GenerationFinishReason::NONE, R"({"delta":{"content":"Reasoning"}})"},
+        {"example", ov::genai::GenerationFinishReason::NONE, R"({"delta":{"content":"example"}})"},
+        {"[TOOL_CALLS]", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {"get", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {"_", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {"weather", ov::genai::GenerationFinishReason::NONE, std::nullopt},
+        {"[ARGS]", ov::genai::GenerationFinishReason::NONE, R"({"delta":{"tool_calls":[{"id":"XXXXXXXXX","type":"function","index":0,"function":{"name":"get_weather"}}]}})"},
+        {"{\"", ov::genai::GenerationFinishReason::NONE, R"({"delta":{"tool_calls":[{"index":0,"function":{"arguments":"{\""}}]}})"},
+        {"city\":", ov::genai::GenerationFinishReason::NONE, R"({"delta":{"tool_calls":[{"index":0,"function":{"arguments":"city\":"}}]}})"},
+        {" \"Paris", ov::genai::GenerationFinishReason::NONE, R"({"delta":{"tool_calls":[{"index":0,"function":{"arguments":" \"Paris"}}]}})"},
```
atobiszei, line 209: Aren't we missing \ before " here?
```suggestion
        {" \"Paris", ov::genai::GenerationFinishReason::NONE, R"({"delta":{"tool_calls":[{"index":0,"function":{"arguments":" \"Paris\"}}]}})"},
```

src/test/llm/output_parsers/devstral_output_parser_test.cpp
atobiszei: Add test for empty args

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+    toolCallObj.AddMember("index", toolCallIndex, wrappedDelta.GetAllocator());
+    rapidjson::Value functionObj(rapidjson::kObjectType);
+    rapidjson::Value nameValue(toolCall.name.c_str(), wrappedDelta.GetAllocator());
+    functionObj.AddMember("name", nameValue, wrappedDelta.GetAllocator());
+    // now we need to add string toolCall.arguments to argumentsWrapper under "arguments" key
+
+    toolCallsString.SetString(toolCall.arguments.c_str(), wrappedDelta.GetAllocator());
+    functionObj.AddMember("arguments", toolCallsString, wrappedDelta.GetAllocator());
+    toolCallObj.AddMember("function", functionObj, wrappedDelta.GetAllocator());
+    toolCalls.PushBack(toolCallObj, wrappedDelta.GetAllocator());
```
mzegla, line 114: Why do you mix function name and arguments in one delta? Other parsers follow the pattern where the first delta contains function name and other metadata and arguments are sent in the following deltas.

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+    We have three processing states:
+        AWAITING_START_TAG,
+        AWAITING_ARGS_TAG,
+        PROCESSING_ARGS
+
+    We store the history of chunks in streamContent string. After state changes are detected, we clear the streamContent to only keep unprocessed part.
+    */
+
+    this->streamContent += chunk;
+    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Chunk content: '{}', StreamContent: '{}', State: {}", chunk, this->streamContent, std::to_string(this->internalState));
```
mzegla: That's quite detailed information. Should it be on DEBUG not on TRACE?

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+            chunk.substr(chunk.size() - this->streamingEndTag.size()) == this->streamingEndTag) {
+            // remove </s> from streamContent
+            this->streamContent = this->streamContent.substr(0, this->streamContent.size() - this->streamingEndTag.size());
+            SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Found end tag in chunk while awaiting start tag. Returning content chunk.");
+            return parseContentChunk();
+        }
+        size_t pos = chunk.find(this->streamingParsingToolCallsStartTag);
+        if (pos != std::string::npos) {
+            this->internalState = AWAITING_ARGS_TAG;
+            std::cout << "Found [TOOL_CALLS] tag in chunk."
```
mzegla: Remove or switch to use logger

src/llm/io_processing/devstral/tool_parser.cpp
mzegla: Still, in the line below we check `streamContent` not a `chunk`

src/llm/io_processing/devstral/tool_parser.hpp
mzegla: ^^

mzegla (commented):
Please mention devstral in docs/llm/reference.md:
https://github.com/openvinotoolkit/model_server/blob/main/docs/llm/reference.md#output-parsing-settings

src/llm/io_processing/devstral/tool_parser.hpp
mzegla: So those variables are used only in streaming mode now, that's why such naming was selected, but I'm not sure about it since it's essentially just a to-string mapping of special tokens that are not limited to streaming.

@dkalinowski , @atobiszei - what's your opinion on this?

src/llm/io_processing/devstral/tool_parser.cpp
mzegla: Looks like this particular case is not covered in other parsers - likely due to more verbose format where we don't see whole output provided as one chunk. In such case I'm more open to accepting that change even though it breaks "no arguments in first delta" rule that we followed so far.

src/llm/io_processing/devstral/tool_parser.hpp
```diff
@@ -0,0 +1,97 @@
...
+    std::string streamContent = "";  // content accumulated from stream chunks
+    std::string toolName = "";
+    std::optional<rapidjson::Document> sendFullDelta(ToolCall& toolCall);
+
+public:
+    DevstralToolParser() = delete;
+    DevstralToolParser(ov::genai::Tokenizer& tokenizer, const ToolsSchemas_t& toolSchemas) :
+        BaseOutputParser(tokenizer),
+        argsTokenId([&tokenizer, this]() {
+            // can not use streamingParsingArgsStartTag because object is not initialized yet
```
dkalinowski: it is not initialized because it is class field, which adds memory overhead, why cant this be static const field? it would be initialized early and it could be used here

dkalinowski: To do that, simply declare fields as

```
static const std::string streamingEndTag;";
```
in header file

and then implement in cpp file:
```
const std::string DevstralParser::streamingEndTag = "</s>";
```

src/llm/io_processing/devstral/tool_parser.hpp
```diff
@@ -0,0 +1,97 @@
...
+                throw std::runtime_error("[ARGS] must be a single token in the tokenizer vocabulary.");
+            }
+            return encoded.data<int64_t>()[0];
+        }()),
+        botTokenId([&tokenizer, this]() {
+            // can not use streamingParsingToolCallsStartTag because object is not initialized yet
+            auto encoded = tokenizer.encode("[TOOL_CALLS]", {{"add_special_tokens", false}}).input_ids;
+            if (encoded.get_shape()[0] != 1) {
+                throw std::runtime_error("[TOOL_CALLS] must be a single token in the tokenizer vocabulary.");
+            }
```
dkalinowski:
```suggestion
            if (encoded.get_shape().size() != 1) {
                throw std::runtime_error("[TOOL_CALLS] token shape must have 1 dimension");
            }
            if (encoded.get_shape()[0] != 1) {
                throw std::runtime_error("[TOOL_CALLS] must be a single token in the tokenizer vocabulary.");
            }
```

src/llm/io_processing/devstral/tool_parser.hpp
dkalinowski: Sure, checking other parsers I see the name convention to be:

```
    static const std::string parsingStartTag;
    static const std::string parsingStartTag2;
    static const std::string parsingEndTag;
```

```
    const std::string parsingStartTag = "<|python_tag|>";
    const std::string parsingEndTag = "";
```

so following our convention it should be
```
    static const std::string parsingArgsStartTag; // = "[ARGS]"; <---- and implementation moved to cpp
    staitc const std::string parsingToolCallsStartTag; // = "[TOOL_CALLS]"; <--- and implementation moved to cpp
```

src/llm/io_processing/devstral/tool_parser.hpp
```diff
@@ -0,0 +1,97 @@
...
+        return toolCallStartTags;
+    }
+    const std::vector<std::string>& getSpecialParsingStartTags() const override {
+        static const std::vector<std::string> specialParsingStartTags{};
+        return specialParsingStartTags;
+    }
+    // Tools calls are expected to be the last part of the content, so we do not specify an end tag.
+    const std::string& getParsingEndTag() const override {
+        static const std::string toolCallEndTag = "</s>";
+        return toolCallEndTag;
```
dkalinowski:
```suggestion
        return streamingEndTag;
```

code duplication

src/llm/io_processing/devstral/tool_parser.cpp
dtrawins:
```suggestion
    SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Chunk content: '{}', StreamContent: '{}', State: {}", chunk, this->streamContent, std::to_string(this->internalState));
```

src/llm/io_processing/devstral/tool_parser.cpp
dtrawins:
```suggestion
```

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+                      << " Current state: " << this->internalState << std::endl;
```
dtrawins:
```suggestion
```

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+            } else {
+                this->streamContent = this->streamContent.substr(pos + this->streamingParsingToolCallsStartTag.length());  // "[TOOLS_CALLS]" length is 13
+                return parseContentChunk();
+            }
+        } else {
+            return parseContentChunk();
+        }
+    }
+    if (this->internalState == AWAITING_ARGS_TAG) {
+        // check if [ARGS] tag is present in the chunk and update state accordingly
```
dtrawins:
```suggestion
```

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -0,0 +1,227 @@
...
+#include "src/port/rapidjson_document.hpp"
+#include "src/logging.hpp"
+#include "src/llm/io_processing/utils.hpp"
+#include "src/stringutils.hpp"
+#include "tool_parser.hpp"
+
+namespace ovms {
+
+void DevstralToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int64_t>& generatedTokens) {
+    std::vector<std::string> tools;
```
mzegla: unused

src/llm/io_processing/devstral/tool_parser.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+namespace ovms {
+class DevstralToolParser : public BaseOutputParser {
+    static const int64_t argsTokenId;  // [ARGS]
+    static const int64_t botTokenId;   // [TOOL_CALLS]
+
+    // in streaming mode we can rely on tags in string format as tokens are not available
+    static const std::string ParsingArgsStartTag;
+    static const std::string ParsingToolCallsStartTag;
+    static const std::string ParsingEndTag;
```
dkalinowski:
```suggestion
    static const std::string parsingArgsStartTag;
    static const std::string parsingToolCallsStartTag;
    static const std::string parsingEndTag;
```

src/llm/io_processing/devstral/tool_parser.hpp
```diff
@@ -0,0 +1,81 @@
...
+        return toolCallStartTags;
+    }
+    const std::vector<std::string>& getSpecialParsingStartTags() const override {
+        static const std::vector<std::string> specialParsingStartTags{};
+        return specialParsingStartTags;
+    }
+    // Tools calls are expected to be the last part of the content, so we do not specify an end tag.
+    const std::string& getParsingEndTag() const override {
+        static const std::string toolCallEndTag = "</s>";
+        return toolCallEndTag;
```
dkalinowski:
```suggestion
        static const std::string toolCallEndTag = "</s>";
        return this->parsingEndTag;
```
The defines are here for a reason 😄

src/test/llm/output_parsers/devstral_output_parser_test.cpp
dkalinowski: @dtrawins

src/test/llm/output_parsers/devstral_output_parser_test.cpp
dkalinowski: @dtrawins

src/test/llm/output_parsers/devstral_output_parser_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+    EXPECT_EQ(parsedOutput.toolCalls[0].arguments, "{\"arg1\":\"value1\",\"arg2\":42}");
+    EXPECT_EQ(parsedOutput.toolCalls[0].id.empty(), false);
+}
+
+TEST_F(DevstralOutputParserTest, ParseToolCallOutputWithSingleToolCall_MissingEndTag) {
+    std::string testInput = "Reasoninig before tool call [TOOL_CALLS]example_tool[ARGS]{\"arg1\":\"value1\",\"arg2\":42}";
+    auto generatedTensor = devstralTokenizer->encode(testInput, ov::genai::add_special_tokens(false)).input_ids;
+    std::vector<int64_t> generatedTokens(generatedTensor.data<int64_t>(), generatedTensor.data<int64_t>() + generatedTensor.get_size());
+    ParsedOutput parsedOutput = outputParserWithRegularToolParsing->parse(generatedTokens, true);
+    EXPECT_EQ(parsedOutput.content, "Reasoninig before tool call ");
```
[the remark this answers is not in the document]
dkalinowski: @dtrawins

src/test/llm/output_parsers/devstral_output_parser_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+    auto generatedTensor = devstralTokenizer->encode(input, ov::genai::add_special_tokens(false)).input_ids;
+    std::vector<int64_t> generatedTokens(generatedTensor.data<int64_t>(), generatedTensor.data<int64_t>() + generatedTensor.get_size());
+    ParsedOutput parsedOutput = outputParserWithRegularToolParsing->parse(generatedTokens, true);
+    EXPECT_EQ(parsedOutput.content, "This is a regular model response without tool calls.");
+    ASSERT_EQ(parsedOutput.toolCalls.size(), 0);
+    EXPECT_EQ(parsedOutput.reasoning, "");
+}
+
+TEST_F(DevstralOutputParserTest, ParseToolCallOutputWithContentAndSingleToolCall) {
+    std::string testInput = "Reasoninig before tool call [TOOL_CALLS]example_tool[ARGS]{\"arg1\":\"value1\",\"arg2\":42}</s>";
```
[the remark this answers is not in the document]
dkalinowski: @dtrawins

src/test/llm/output_parsers/devstral_output_parser_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+    auto generatedTensor = devstralTokenizer->encode(testInput, ov::genai::add_special_tokens(false)).input_ids;
+    std::vector<int64_t> generatedTokens(generatedTensor.data<int64_t>(), generatedTensor.data<int64_t>() + generatedTensor.get_size());
+    ParsedOutput parsedOutput = outputParserWithRegularToolParsing->parse(generatedTokens, true);
+    EXPECT_EQ(parsedOutput.content, "Reasoninig before tool call ");
```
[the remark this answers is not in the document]
dkalinowski: @dtrawins

src/llm/io_processing/devstral/tool_parser.hpp
mzegla: I accepted, `toolCallEndTag` in this part could be removed right?

dkalinowski (approved):

mzegla (approved):

PR #3852, opened 2025-12-10
update ov to 10th dec build

dtrawins:
### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/demos/common/export_models/requirements.txt
+++ b/demos/common/export_models/requirements.txt
@@ -6,11 +6,10 @@ optimum-intel@git+https://github.com/huggingface/optimum-intel.git
 accelerate==1.11.0
 diffusers==0.35.2  # for image generation
 einops==0.8.1
-nncf
+nncf@git+https://github.com/openvinotoolkit/nncf.git
 numpy==2.2.6
-openai
-openvino-tokenizers==2025.4.0.0rc3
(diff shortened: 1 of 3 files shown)


demos/common/export_models/requirements.txt
```diff
@@ -6,11 +6,10 @@ optimum-intel@git+https://github.com/huggingface/optimum-intel.git
 accelerate==1.11.0
 diffusers==0.35.2  # for image generation
 einops==0.8.1
-nncf
+nncf@git+https://github.com/openvinotoolkit/nncf.git
```
dkalinowski, line 9: why do we need latest commit from github?

dtrawins: nncf is adding now improvements for gpt other moe models quantization

dkalinowski: doesnt it mean we should re-generate gpt models?

dkalinowski (approved):

demos/common/export_models/requirements.txt
dtrawins: there will be added option to use quantization parameters --awq --gptq --scale-estimation --dataset
such quantization last much longer but might slightly improve accuracy

atobiszei (approved):

PR #3855, opened 2025-12-10
Add speaker embeddings support

michalkulakowski:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``


src/audio/text_to_speech/t2s_servable.hpp
```diff
@@ -38,25 +38,52 @@
...
 struct TtsServable {
     std::filesystem::path parsedModelsPath;
     std::shared_ptr<ov::genai::Text2SpeechPipeline> ttsPipeline;
     std::mutex ttsPipelineMutex;
+    std::unordered_map<std::string, ov::Tensor> voices;
 
-    TtsServable(const mediapipe::T2sCalculatorOptions& nodeOptions, const std::string& graphPath) {
-        auto fsModelsPath = std::filesystem::path(nodeOptions.models_path());
+    TtsServable(const std::string& modelDir, const std::string& targetDevice, const google::protobuf::RepeatedPtrField<mediapipe::T2sCalculatorOptions_SpeakerEmbeddings>& graphVoices, const std::string& pluginConfig, const std::string& graphPath) {
+        auto fsModelsPath = std::filesystem::path(modelDir);
```
dkalinowski: can you move this constructor to cpp? this way you could move read_speaker_embedding to cpp as well.

src/audio/text_to_speech/t2s_servable.hpp
```diff
@@ -38,25 +38,52 @@
...
+    std::ifstream input(file_path, std::ios::binary);
+    OPENVINO_ASSERT(input, "Failed to open file: " + file_path.string());
+
+    // Get file size
+    input.seekg(0, std::ios::end);
+    size_t buffer_size = static_cast<size_t>(input.tellg());
+    input.seekg(0, std::ios::beg);
+
+    // Check size is multiple of float
+    OPENVINO_ASSERT(buffer_size % sizeof(float) == 0, "File size is not a multiple of float size.");
```
dkalinowski: i think it might be the first time we use openvino_assert
why are we introducing yet another way of reporting errors?
for now we have:
- ovms::Status
- absl::Status
- std::variant<Obj, std::string(error)>
- and now this... :(

src/audio/text_to_speech/t2s_servable.hpp
```diff
@@ -38,25 +38,52 @@
...
-            SPDLOG_ERROR("Error during llm node plugin_config option parsing to JSON: {}", nodeOptions.plugin_config());
+            SPDLOG_ERROR("Error during llm node plugin_config option parsing to JSON: {}", pluginConfig);
             throw std::runtime_error("Error during plugin_config option parsing");
         }
-        ttsPipeline = std::make_shared<ov::genai::Text2SpeechPipeline>(parsedModelsPath.string(), nodeOptions.target_device(), config);
+        ttsPipeline = std::make_shared<ov::genai::Text2SpeechPipeline>(parsedModelsPath.string(), pluginConfig, config);
+        for (auto voice : graphVoices) {
+            if (!std::filesystem::exists(voice.path()))
+                throw std::runtime_error{"Requested voice speaker embeddings file does not exist."};
+            voices[voice.name()] = read_speaker_embedding(voice.path());
```
dkalinowski: do we catch exceptions thrown inside read_speaker_embedding?

demos/audio/README.md
```diff
@@ -47,6 +47,41 @@ python export_model.py text2speech --source_model microsoft/speecht5_tts --weigh
 
 The default configuration should work in most cases but the parameters can be tuned via `export_model.py` script arguments. Run the script with `--help` argument to check available parameters and see the [T2s calculator documentation](../../docs/speech_generation/reference.md) to learn more about configuration options and limitations.
 
+### Speaker embeddings
+
+Instead of generating speech with default model voice you can create speaker embeddings with [this script](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/speech_generation/create_speaker_embedding.py)
+```bash
+curl --output create_speaker_embedding.py "https://raw.githubusercontent.com/openvinotoolkit/openvino.genai/refs/heads/master/samples/python/speech_generation/create_speaker_embedding.py"
```
dkalinowski, line 54: yet another link for manual change, it wouldnt even hit sed command which i use because its genai and master
do we want to keep this reference to master and risk of change and getting outdated demos? or do we want to hardcode it? or maybe do we want to keep it in our repo?

prepare_llm_models.sh
```diff
@@ -78,6 +79,16 @@ if [ ! -f "$1/$FACEBOOK/chat_template.jinja" ]; then
     cp src/test/llm/dummy_facebook_template.jinja "$1/$FACEBOOK/chat_template.jinja"
 fi
 
+if [ -f "$1/$TTS_MODEL/$TOKENIZER_FILE" ]; then
+  echo "Model file $1/$TTS_MODEL/$TOKENIZER_FILE exists. Skipping downloading models."
+else
+  python3 demos/common/export_models/export_model.py text2speech --source_model "$TTS_MODEL" --weight-format int4 --model_repository_path $1 --vocoder microsoft/speecht5_hifigan
```
dkalinowski, line 85: what about windows script?

src/test/audio/text2speech_test.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+TEST_F(Text2SpeechHttpTest, simplePositive) {
+    std::string requestBody = R"(
+        {
+            "model": ")" + modelName +
+                              R"(",
+            "input": "The quick brown fox jumped over the lazy dog."
+        }
+    )";
+    ASSERT_EQ(
```
dkalinowski, line 64: tests check only "OK" and no errors, but what about actual output?

dkalinowski (approved):

src/audio/text_to_speech/t2s_calculator.cc
```diff
@@ -103,8 +103,23 @@ class T2sCalculator : public CalculatorBase {
             if (streamIt != payload.parsedJson->MemberEnd()) {
                 return absl::InvalidArgumentError("streaming is not supported");
             }
+            std::optional<std::string> voice;
+            auto voiceIt = payload.parsedJson->FindMember("voice");
+            if (voiceIt != payload.parsedJson->MemberEnd() && voiceIt->value.IsString()) {
+                voice = voiceIt->value.GetString();
+            }
+            if (voice.has_value()) {
+                if (pipe->voices.find(voice.value()) == pipe->voices.end())
```
mzegla: Are `voices` static at this point? Don't we need to lock it for the thread as well for the access?

src/audio/text_to_speech/t2s_servable.cpp
```diff
@@ -0,0 +1,85 @@
...
+    input.seekg(0, std::ios::end);
+    size_t buffer_size = static_cast<size_t>(input.tellg());
+    input.seekg(0, std::ios::beg);
+
+    // Check size is multiple of float
+    if (buffer_size % sizeof(float) != 0) {
+        throw std::runtime_error("File size is not a multiple of float size.");
+    }
+    size_t num_floats = buffer_size / sizeof(float);
+    if (num_floats != 512) {
```
mzegla, line 52: Why 512? 
Can we have more context for this number for future reference?
Is it like embedding dim size or something?

src/audio/text_to_speech/t2s_calculator.cc
```diff
@@ -103,8 +103,23 @@ class T2sCalculator : public CalculatorBase {
             if (streamIt != payload.parsedJson->MemberEnd()) {
                 return absl::InvalidArgumentError("streaming is not supported");
             }
+            std::optional<std::string> voice;
```
mzegla:
```suggestion
            std::optional<std::string> voiceName;
```
?
To distinguish from `pipe->voices` elements

src/audio/text_to_speech/t2s_calculator.cc
```diff
@@ -103,8 +103,23 @@ class T2sCalculator : public CalculatorBase {
...
+            auto voiceIt = payload.parsedJson->FindMember("voice");
+            if (voiceIt != payload.parsedJson->MemberEnd() && voiceIt->value.IsString()) {
+                voice = voiceIt->value.GetString();
+            }
+            if (voice.has_value()) {
```
mzegla: Do we need separate condition here? Can we bring content from this block to the condition above?
`if (voiceIt != payload.parsedJson->MemberEnd() && voiceIt->value.IsString())` I suppose it's the same.

mzegla (approved):

src/audio/text_to_speech/t2s_calculator.cc
dkalinowski: @michalkulakowski whats the answer?

michalkulakowski: answer is yes, they are static, loaded once during pipeline initialization

PR #3858, opened 2025-12-11
[2025.4.1] Gpt-oss multi turn template + add to agentic demo

dkalinowski:
### 🛠 Summary

0.35 BFCL multiturn unary

Changes to original chat template:
- introducing reasoning=`none` - this automaticly adds empty reasoning channel so it forces the model to skip the reasoning part. in reasoning effort placeholder it puts "low", since "none" was not present during gpt-oss training phase
- removed exception when chat history contains both: content and reasoning_content. For some reason, during multi-turn bfcl benchmark i saw model generating both channels, which caused exception during history rendering. instead of exception, the `content` is rendered and `reasoning_content` omitted (no idea why i just assumed content might be more insightful)
- replaced `thinking` with `reasoning_content` - for some reason openai used thinking field to render reasoning from history. replaced with `reasoning_content` which is present in requests received during bfcl benchmark.
-  fixed issue with parsing empty `tool_calls: []` array when rendering chat history. in some cases gpt-oss generates reasoning/content but no tool calls, bfcl sends empty array and chat template accessed index 0 assuming there always is some tool call. new chat templates ignores empty arrays now
- removed `|tojson` from tool argument rendering. this introduced string escaping which causes escaping in output (chain reaction). openai harmony format assumes no escaping. this was related to both: function call output (result from mcp servers) and function call arguments (input to mcp servers) in chat history

### DIFF
--- a/demos/continuous_batching/agentic_ai/README.md
+++ b/demos/continuous_batching/agentic_ai/README.md
@@ -99,7 +99,7 @@ curl -L -o models/Qwen/Qwen3-Coder-30B-A3B-Instruct/chat_template.jinja https://
 :sync: gpt-oss-20b
 ```console
 python export_model.py text_generation --source_model openai/gpt-oss-20b --weight-format int4 --config_file_path models/config.json --model_repository_path models --tool_parser gptoss --reasoning_parser gptoss --pipeline_type LM
-curl -L -o models/openai/gpt-oss-20b/chat_template.jinja https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/gpt-multi-turn/extras/chat_template_examples/chat_template_gpt_oss_multiturn.jinja
+curl -L -o models/openai/gpt-oss-20b/chat_template.jinja https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/releases/2025/4/extras/chat_template_examples/chat_template_gpt_oss_multiturn.jinja
 ```
 > **Note:**: Use `--pipeline_type LM` for export and `--target device GPU` for deployment. Expect continuous batching and CPU support in weekly or 2026.0+ releases.
 :::


extras/chat_template_examples/chat_template_gpt_oss_multiturn.jinja
```diff
@@ -0,0 +1,369 @@
+{#-
+  Modifications to original chat template:
+    * Allowing reasoning_effort=none; this automaticly adds empty reasoning channel at the end. It should force the model to follow with regular response/tool call immediately.
+      IMPORTANT: When none used, reasoning_effort is rendered as low in the reasoning definition slot (as it is the lowest possible reasoning during model training).
```
mzegla: Why not just use `low`? Accepting `none` would create false expectation that there will be no reasoning happening.

dkalinowski: Model has never seen `none` during training. I think it might greatly impact generated responses. It was verified with "low" to be useful in Continue.dev autocompletion feature. @mzegla

mzegla: But user passes `none` right?

dkalinowski: yes, this was my idea to use reasoning_effort none. maybe you have better one - maybe we should reuse enable_thinking as in other models? @mzegla  but then we would need to handle reasoning_effort+enable_thinking usage at the same time

mzegla: I just didn't want to make user think that reasoning_effort=none disables thinking completely. Can we just pass `low` and not make implicit substitution?

dkalinowski: we agreed to keep "none" to "disable completely", but still render "low", as lowest possible reasoning during training phase. having disable switch is crucial, because "low" still introduces some reasoning. it is unacceptable for autocompletions continue.dev feature

PR #3896, opened 2026-01-08
Cvs 178976 rh python312

rasapala:
### 🛠 Summary

Changing redhat core python from 3.9 to 3.12
3.9 still required for tensorflow compilation.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/python/pythoninterpretermodule.cpp
+++ b/src/python/pythoninterpretermodule.cpp
@@ -39,10 +39,10 @@ Status PythonInterpreterModule::start(const ovms::Config&) {
     if (!Py_IsInitialized()) {
         SPDLOG_INFO("Initializing python interpreter", PYTHON_INTERPRETER_MODULE_NAME);
         py::initialize_interpreter();
-        hasModuleInitializedTheInterpreter = true;
+        ownsInterpreter = true;
     } else {
         SPDLOG_INFO("Python interpreter already initialized", PYTHON_INTERPRETER_MODULE_NAME);
-        hasModuleInitializedTheInterpreter = false;
+        ownsInterpreter = false;
     }
 
     py::gil_scoped_acquire acquire;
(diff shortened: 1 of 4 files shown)


run_unit_tests.sh
```diff
@@ -75,43 +75,12 @@ if [ "$RUN_TESTS" == "1" ] ; then
...
-          rm -rf ${FAIL_LOG}
-        fi
-    else
-        # For RH UBI and Ubuntu20
-        if ! bazel test --jobs=$JOBS ${debug_bazel_flags} ${SHARED_OPTIONS} --test_summary=detailed --test_output=streamed --test_filter="*" //src:ovms_test > ${TEST_LOG} 2>&1 ; then
-            failed=1
-        fi
-        cat ${TEST_LOG} | tail -500
+
+    # For RH UBI and Ubuntu20
```
dtrawins:
```suggestion
    # For RH UBI and Ubuntu
```

src/test/mediapipe/python/scripts/symmetric_identity_fp32_only.py
```diff
@@ -1,37 +0,0 @@
-#*****************************************************************************
```
dtrawins, line 1: why this is removed?

rasapala: Not used anywhere.

src/python/pythoninterpretermodule.hpp
```diff
@@ -32,6 +32,7 @@ class PythonInterpreterModule : public Module {
     std::unique_ptr<PythonBackend> pythonBackend;
     mutable std::unique_ptr<py::gil_scoped_release> GILScopedRelease;
     std::thread::id threadId;
+    bool hasModuleInitializedTheInterpreter;
```
mzegla: A bit shorter option
```suggestion
    bool ownsInterpreter;
```

src/python/pythoninterpretermodule.cpp
```diff
@@ -63,10 +72,14 @@ void PythonInterpreterModule::shutdown() {
...
     SPDLOG_INFO("{} shutdown", PYTHON_INTERPRETER_MODULE_NAME);
-    py::finalize_interpreter();
+    if (hasModuleInitializedTheInterpreter)
+        py::finalize_interpreter();
 }
 
 void PythonInterpreterModule::releaseGILFromThisThread() const {
+    // No need to release GIL it must be handled with the initialize interpreter class. Currently in gtest PythonEnvironment.
+    if (!hasModuleInitializedTheInterpreter)
+        return;
```
mzegla: IMHO the comment does not bring too much context and unnecessarily gives information about testing - it's not bound. I also don't think this logic should be here. Can you move such check to the place this method is used - server.cpp and go with something like:
```
        auto pythonModule = dynamic_cast<const PythonInterpreterModule*>(it->second.get());
		if(pythonModule->ownsInterpreter()) {
			// Natively GIL is held by the thread that initialized interpreter, so we only need to release it, if we own the interpreter. 
			// If it was initialized externally, then the external thread shall release the GIL before launching that module.
        	pythonModule->releaseGILFromThisThread();
		}
```

PR #3902, opened 2026-01-12
Exec_cmd fixes

atobiszei:
Tighten parsing CLI arguments for use case with optimum-cli.

Ticket:CVS-177708

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -254,6 +254,7 @@ build:win_mp_on_py_off --define=MEDIAPIPE_DISABLE=0
 build:win_mp_on_py_off --define=PYTHON_DISABLE=1
 
 # Tests settings ############################################################################################################################
+test --test_timeout=1200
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -254,6 +254,7 @@ build:win_mp_on_py_off --define=MEDIAPIPE_DISABLE=0
 build:win_mp_on_py_off --define=PYTHON_DISABLE=1
 
 # Tests settings ############################################################################################################################
+test --test_timeout=1200
```
rasapala, line 257: will this overwrite one from command line ? used in run_unit_tests 1800s ?

atobiszei: The other way arround. I changed this since now when run inside build container it doesn't end within default timeout ~5m.

jszczepa (approved):
Looks ok

PR #3905, opened 2026-01-13
Bfcl v4

dtrawins:
### 🛠 Summary

CVS-179106

### 🧪 Checklist

- [ ] Unit tests added.
``


tests/accuracy/test_small_models.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash -x
+#
+# Copyright (c) 2024 Intel Corporation
```
atobiszei: 2026

tests/accuracy/test_single_model.sh
```diff
@@ -0,0 +1,22 @@
+export MODEL=$1
```
atobiszei, line 18: copyright header

tests/accuracy/test_single_model.sh
```diff
@@ -0,0 +1,22 @@
...
+export PRECISION=$2
+
+
+docker stop ovms 2>/dev/null
+docker run -d --name ovms --user $(id -u):$(id -g) --rm -p 8000:8000 -v $(pwd)/models:/models openvino/model_server:latest \
```
atobiszei, line 24: Most probably we would need to pass image label/tag so this could be the default, but otherwise user passed arguments should be used here?

mzegla (approved):

dkalinowski (approved):

PR #3929, opened 2026-01-28
Prefix caching for NPU in nested plugin config structure

mzegla:


michalkulakowski (approved):

demos/common/export_models/export_model.py
```diff
@@ -459,6 +458,17 @@ def export_text_generation_model(model_repository_path, source_model, model_name
     if "HETERO" in task_parameters['target_device']:
         plugin_config['MODEL_DISTRIBUTION_POLICY'] = 'PIPELINE_PARALLEL'
 
+    if task_parameters['target_device'] == 'NPU':
+        max_prompt_len = task_parameters['max_prompt_len']
```
dkalinowski, line 463: previously we accessed `task_parameters['max_prompt_len']` only if we previously ` if task_parameters['max_prompt_len'] is not None:`

now we dont have such check, wont there be KeyError?

mzegla: We have this value added as `None` by default. See the `parser_text.add_argument(...)` section.

dkalinowski (approved):

PR #3933, opened 2026-01-28
Embeddings npu3, embeddings_calculator_ov

rasapala:
### 🛠 Summary

JIRA CVS-179110

### 🧪 Checklist

- [ ] Unit tests added.
``


src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
+                size_t input_ids_size = tokens.input_ids.get_shape()[1];
+                size_t attention_mask_size = tokens.attention_mask.get_shape()[1];
+                size_t typeIds_size = 0;
+                if (embeddings_session->getNumberOfModelInputs() == 3) {
+                    typeIds_size = tokens.attention_mask.get_shape()[1];
+                }
+                for (uint64_t i = 0; i < received_batch_size; i++) {
+                    // TODO: Check optimal inferRequest init
+                    ov::InferRequest& inferRequest = executingStreamIdGuard->getInferRequest();
+                    ov::Tensor oneBatchInputIdsTensor = ov::Tensor(tokens.input_ids, {i, 0}, {i + 1, input_ids_size});
```
dkalinowski: What is this constructor? it is not clear to me what is {1, 0} and {i+1, input...}
I guess one of those is shape, but what about the other?
To make it cleaner I suggest creating variables with a name to indicate what it is.

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
-            inferRequest.set_tensor(EMBEDDINGS_MODEL_INPUT_IDS_NAME, tokens.input_ids);
-            inferRequest.set_tensor(EMBEDDINGS_MODEL_ATTENTION_MASK_NAME, tokens.attention_mask);
-            if (embeddings_session->getNumberOfModelInputs() == 3) {
-                inferRequest.set_tensor(EMBEDDINGS_MODEL_TOKEN_TYPE_IDS_NAME, typeIds);
+            // NPU embeddings dynamic model case for batch size grater than 1
+            if (embeddings_session->getTargetDevice() == "NPU" && received_batch_size > 1) {
+                SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Embeddings batch NPU request split for BS {}", received_batch_size);
+                size_t input_ids_size = tokens.input_ids.get_shape()[1];
+                size_t attention_mask_size = tokens.attention_mask.get_shape()[1];
+                size_t typeIds_size = 0;
```
dkalinowski:
```suggestion
                size_t type_ids_size = 0;
```
either snake case or camel case - not both at the same time :D
and also I think we agreed on camel case 5 years ago

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
                 SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Embeddings input is of not supported type");
                 return absl::InvalidArgumentError("Input should be string, array of strings or array of integers");
             }
+
+            std::vector<ov::Tensor> embeddingsTensors;
```
dkalinowski, line 257:
```suggestion
            std::vector<ov::Tensor> embeddingsInputIdsTensors;
```
am I right here? guessing

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
+                    ov::Tensor oneBatchAttentionMaskTensor = ov::Tensor(tokens.attention_mask, {i, 0}, {i + 1, attention_mask_size});
+
+                    inferRequest.set_tensor(EMBEDDINGS_MODEL_INPUT_IDS_NAME, oneBatchInputIdsTensor);
+                    inferRequest.set_tensor(EMBEDDINGS_MODEL_ATTENTION_MASK_NAME, oneBatchAttentionMaskTensor);
+
+                    if (embeddings_session->getNumberOfModelInputs() == 3) {
+                        ov::Tensor oneBatchTypeIdsTensor = ov::Tensor(typeIds, {i, 0}, {i + 1, typeIds_size});
+                        inferRequest.set_tensor(EMBEDDINGS_MODEL_TOKEN_TYPE_IDS_NAME, oneBatchTypeIdsTensor);
+                    }
+                    // TODO: Optimize and test execution here
```
dkalinowski: ?

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
+                        int targetOutputIndex = embeddings_session->getTargetOutputIndex();
+                        RET_CHECK(targetOutputIndex >= 0) << "No output with 3 dimensions found";  // this should never happen as pipeline is unavailable if pooling operation could not be added
+                        outputTensorName = inferRequest.get_compiled_model().outputs()[targetOutputIndex].get_any_name();
+                        SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Multiple embedding model outputs found, 3-dim output with name {} will be used", outputTensorName);
+                    } else {  // BGE
+                        RET_CHECK(inferRequest.get_compiled_model().outputs().size() == 1);
+                        outputTensorName = inferRequest.get_compiled_model().outputs().begin()->get_any_name();
+                        SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Single embedding model output found with name {}", outputTensorName);
+                    }
+                    embeddingsTensors.push_back(inferRequest.get_tensor(outputTensorName.c_str()));
```
dkalinowski: this tensor you push to embeddingsTensors is still tied to inferRequest and data underneath will be overriden if you perform another inference on it. and you do, in next for loop iteration. please correct me if I'm wrong here

Did you run any stress tests, high load tests + accuracy tests with different concurrency/queue size? This would showcase the issue which I point out

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
-                RET_CHECK(inferRequest.get_compiled_model().outputs().size() == 1);
-                outputTensorName = inferRequest.get_compiled_model().outputs().begin()->get_any_name();
-                SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Single embedding model output found with name {}", outputTensorName);
+
+            // NPU embeddings dynamic model case
+            if (embeddings_session->isNpuPostprocessingRequired()) {
+                SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "NPU embeddings dynamic model additional inference");
+                ModelMetricReporter unused2(nullptr, nullptr, "unused2", 1);
+                auto executingStreamIdGuard2 = std::make_unique<ExecutingStreamIdGuard>(embeddings_session->getPostProcInferRequestsQueue(), unused2);
+                ov::InferRequest& inferRequest2 = executingStreamIdGuard2->getInferRequest();
```
dkalinowski:
```suggestion
                auto executingStreamIdGuardForPostprocessingModel = std::make_unique<ExecutingStreamIdGuard>(embeddings_session->getPostProcInferRequestsQueue(), unused2);
                ov::InferRequest& inferRequestForPostprocessingMode = executingStreamIdGuard2->getInferRequest();
```

I dont think naming variables with numbers is good idea, makes the code less readable

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
+                ov::Shape input_shape;
+                if (received_batch_size > 1) {
+                    input_shape = embeddingsTensors[0].get_shape();
```
dkalinowski: what is this line? it is not used

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
+                    inferRequest2.set_tensors("attention_mask", embeddingsAttentionMasks);
+                    inferRequest2.set_tensors("embedding_hidden_state", embeddingsTensors);
+                } else {
+                    input_shape = embeddingsTensor.get_shape();
+
+                    RET_CHECK(input_shape.size() > 1) << "Embeddings result shape is too small";
```
dkalinowski:
```suggestion
                    RET_CHECK(input_shape.size() > 1) << "Embeddings result shape must have more than 1 dimension but contains " << input_shape.size();
```
there is no such thing as "small" shape :D

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
+                ModelMetricReporter unused2(nullptr, nullptr, "unused2", 1);
+                auto executingStreamIdGuard2 = std::make_unique<ExecutingStreamIdGuard>(embeddings_session->getPostProcInferRequestsQueue(), unused2);
+                ov::InferRequest& inferRequest2 = executingStreamIdGuard2->getInferRequest();
+                ov::Shape input_shape;
+                if (received_batch_size > 1) {
+                    input_shape = embeddingsTensors[0].get_shape();
+                    inferRequest2.set_tensors("attention_mask", embeddingsAttentionMasks);
+                    inferRequest2.set_tensors("embedding_hidden_state", embeddingsTensors);
+                } else {
+                    input_shape = embeddingsTensor.get_shape();
```
dkalinowski:
```suggestion
                    embeddingsResultShape = embeddingsTensor.get_shape();
```
deduced from ret check below

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -242,27 +242,118 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
+                    // which is greater than the original tokens.attention_mask size of 3800. We need to zero-fill
+                    // the remaining elements in the attention_mask_tensor to ensure correct masking behavior.
+                    if (sequence_length > original_mask_size) {
+                        std::fill_n(attention_mask_tensor.data<int64_t>() + original_mask_size,
+                            sequence_length - original_mask_size,
+                            0);
+                    }
+
+                    // Run post-processing inference
+                    inferRequest2.set_tensor("attention_mask", attention_mask_tensor);
```
dkalinowski: Previously there was no need for second inference pass (2 separate models), can you explain how comes? Did you test entire list of models we supported till now if the accuracy remains the same? We should compare it with the results I gathered in October when I've been adding new pre-post processing layers (poolings).

src/embeddings/embeddings_servable.cpp
```diff
@@ -31,7 +35,177 @@ using namespace ov;
 
 namespace ovms {
 
-// From OpenVINO GenAI repository
+// Start code from OpenVINO GenAI repository
+struct KVAxesPosition {
```
dkalinowski, line 40: can we use existing structures from ov::genai?

src/embeddings/embeddings_servable.cpp
```diff
@@ -31,7 +35,177 @@ using namespace ov;
...
+        }
+        if (value < 0) {
+            OPENVINO_THROW(key + " cannot be negative!");
+        }
+        return std::make_optional(static_cast<uint32_t>(value));
+    }
+    return std::nullopt;
+}
+
+KVAxesPosition get_kv_axes_pos(std::shared_ptr<const ov::Model> model) {
```
dkalinowski, line 80: is this entire copy paste code planned to be somehow merged with genai? are we prioritizing this task in 2026.1? it doesnt look good to keep it longerm
@dtrawins @rasapala 


if we do, i'd suggest separating this code to genai utils module, because more calculators may need it in future

src/embeddings/embeddings_servable.cpp
```diff
@@ -90,50 +315,100 @@ static std::shared_ptr<op::Op> get_last_token_pooling_op(std::shared_ptr<Model>
...
+        config.normalize = this->normalizeEmbeddings;
+        // Compile additional CPU model for NPU dynamic model case
+        auto post_model = create_post_model(model, config);
+        const std::string postModelDevice = "CPU";
+        postProcCompiledModel = core.compile_model(post_model, postModelDevice, properties);
+
+        auto& ovmsConfig = ovms::Config::instance();
+        uint32_t numberOfParallelInferRequests = 1;
+        if (ovmsConfig.nireq() > 0) {
+            // nireq is set globally for all models in ovms startup parameters
```
dkalinowski: no, depending on startup type, ovms allows specifying different nireq per model, in config.json

src/embeddings/embeddings_servable.cpp
```diff
@@ -90,50 +315,100 @@ static std::shared_ptr<op::Op> get_last_token_pooling_op(std::shared_ptr<Model>
...
+            numberOfParallelInferRequests = ovmsConfig.nireq();
```
dkalinowski, line 403: This line isnt even effective, you always overwrite it...

src/embeddings/embeddings_servable.cpp
```diff
@@ -90,50 +315,100 @@ static std::shared_ptr<op::Op> get_last_token_pooling_op(std::shared_ptr<Model>
...
-
-    if (this->normalizeEmbeddings) {
-        SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Applying L2 normalization to embeddings output");
-        processor.output(this->targetOutputIndex).postprocess().custom([](const ov::Output<ov::Node>& node) {
-            auto axis_const = std::make_shared<op::v0::Constant>(ov::element::i32, ov::Shape{1}, std::vector{1});
-            return std::make_shared<op::v0::NormalizeL2>(node, axis_const, 1e-12, op::EpsMode::MAX);
+
+        config.normalize = this->normalizeEmbeddings;
+        // Compile additional CPU model for NPU dynamic model case
+        auto post_model = create_post_model(model, config);
```
dkalinowski, line 394: What is the second model exactly? What does it consist of? Is it only the post-processing layers which we added back in october?

If so, next question is, why cant those be executed on NPU? Those seem basic operations.
If the post-processing model consists of something different, please explain what

src/embeddings/embeddings_servable.cpp
```diff
@@ -90,50 +315,100 @@ static std::shared_ptr<op::Op> get_last_token_pooling_op(std::shared_ptr<Model>
...
+            numberOfParallelInferRequests = 1u;
+        }
+        SPDLOG_DEBUG("Setting post processing inference queue for {} with {} parallel requests", postModelDevice, numberOfParallelInferRequests);
+        postProcInferRequestsQueue = std::make_unique<OVInferRequestsQueue>(postProcCompiledModel, numberOfParallelInferRequests);
+        npuPostprocessingRequired = true;
+
+        // Set additional properties for NPU model
+        auto kv_pos = get_kv_axes_pos(model);
+        KVDesc kv_desc;
+        get_npu_text_embedding_config(properties, kv_pos, kv_desc, config);
```
dkalinowski, line 422: why are you reading the `config` here? It is not used anywhere below

src/sidepacket_servable.cpp
```diff
@@ -57,7 +58,8 @@ SidepacketServable::SidepacketServable(const std::string& modelDir, const std::s
     return;
 }
 
-void SidepacketServable::initialize(const std::string& modelDir, const std::string& targetDevice, const std::string& pluginConfig, const std::string& graphPath) {
+void SidepacketServable::initialize(const std::string& modelDir, const std::string& inTargetDevice, const std::string& pluginConfig, const std::string& graphPath) {
```
dkalinowski: What do you mean with **in**TargetDevice?

src/sidepacket_servable.hpp
```diff
@@ -77,8 +78,12 @@ struct SidepacketServable {
         return compiledModel.inputs().size();
     }
 
+    const std::string getTargetDevice() {
```
dkalinowski: Is something restricting us to return const reference? Do we need to copy? (overhead)

Dockerfile.ubuntu
```diff
@@ -340,6 +342,28 @@ COPY client /client
...
+            mkdir /tmp/npu_deps && cd /tmp/npu_deps ; \
+            apt-get update && apt-get install -y --no-install-recommends libtbb12 libicu-dev && \
+            if [[ "$BASE_IMAGE" == *"22.04"* ]] ; then \
+                curl -L -O https://github.com/intel/linux-npu-driver/releases/download/v1.26.0/linux-npu-driver-v1.26.0.20251125-19665715237-ubuntu2204.tar.gz && \
+                tar -xf linux-npu-driver-v1.26.0.20251125-19665715237-ubuntu2204.tar.gz && \
+                curl -L -O https://github.com/oneapi-src/level-zero/releases/download/v1.24.2/level-zero_1.24.2+u22.04_amd64.deb     ; \
+            else \
+                curl -L -O https://github.com/intel/linux-npu-driver/releases/download/v1.28.0/linux-npu-driver-v1.28.0.20251218-20347000698-ubuntu2404.tar.gz && \
+                tar -xf linux-npu-driver-v1.28.0.20251218-20347000698-ubuntu2404.tar.gz && \
+                curl -L -O https://github.com/oneapi-src/level-zero/releases/download/v1.24.2/level-zero_1.24.2+u24.04_amd64.deb ; \
```
dkalinowski: This is just pure copy-paste from the release image part. It is unnecessary repetition that will make it harder for us to maintain our dockerfile.

And btw, why do you need GPU and NPU drivers in builder image?

src/embeddings/embeddings_calculator_ov.cc
rasapala: This is only for NPU dynamic model case. There is no reference data this is the NPU enabling for embeddings. The CPU and GPU path remains the same.

src/embeddings/embeddings_servable.cpp
rasapala: They are hidden deep in helper headers not in the public API.

src/embeddings/embeddings_servable.cpp
rasapala: They cant be added because dynamic model is not supported on NPU.

Dockerfile.ubuntu
rasapala: For development and testing.

src/embeddings/embeddings_calculator_ov.cc
rasapala: Good catch

src/embeddings/embeddings_calculator_ov.cc
rasapala: Not this is the vector of embeddings from the execution on NPU that will be used for post processing.

src/embeddings/embeddings_servable.cpp
rasapala: Used in the function to get proper model properties: text_embed_config

src/sidepacket_servable.cpp
rasapala: inputTargetDevice - got from user, not to confuse with the class member.

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -45,6 +46,48 @@ class EmbeddingsServable;
 
 namespace mediapipe {
 
+void printTensor(const ov::Tensor& tensor) {
```
michalkulakowski: this should be probbably removed?

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -161,20 +204,20 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
                 if (!tokenizationStatus.ok()) {
                     return tokenizationStatus;
                 }
 
-                size_t input_ids_size = tokens.input_ids.get_shape()[1];
-                if (input_ids_size > max_context_length) {
-                    SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Input size {} exceeds max_context_length {}", input_ids_size, max_context_length);
-                    return absl::InvalidArgumentError("Input length " + std::to_string(input_ids_size) + " longer than allowed " + std::to_string(max_context_length));
+                size_t inputIdsSize = tokens.input_ids.get_shape()[1];
+                if (inputIdsSize > maxContextLength) {
```
michalkulakowski: maybe you could create method for this check and use it also in  line 256?

michalkulakowski (approved):

demos/embeddings/compare_results.py
dkalinowski, line 96: why are you changing this script this way? switching from asserts to prints changes the exit code, meaning that scripts using this will silently show pass. it means automated demo tests but also possibly anyone using this script

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -198,71 +207,173 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
+                    int targetOutputIndex = embeddings_session->getTargetOutputIndex();
+                    RET_CHECK(targetOutputIndex >= 0) << "No output with 3 dimensions found";  // this should never happen as pipeline is unavailable if pooling operation could not be added
+                    outputTensorName = inferRequest.get_compiled_model().outputs()[targetOutputIndex].get_any_name();
+                    SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Multiple embedding model outputs found, 3-dim output with name {} will be used", outputTensorName);
+                } else {  // BGE
+                    RET_CHECK(inferRequest.get_compiled_model().outputs().size() == 1);
+                    outputTensorName = inferRequest.get_compiled_model().outputs().begin()->get_any_name();
+                    SPDLOG_LOGGER_DEBUG(embeddings_calculator_logger, "Single embedding model output found with name {}", outputTensorName);
+                }
+                embeddingsTensor = inferRequest.get_tensor(outputTensorName.c_str());
```
dkalinowski, line 332: Don't we have the same issue with concurrency here? inferRequest is reverted to the queue here and the data will be overriden. Please check all other places you modify

src/embeddings/embeddings_calculator_ov.cc
```diff
@@ -198,71 +207,173 @@ class EmbeddingsCalculatorOV : public CalculatorBase {
...
+
+                    // Run post-processing inference
+                    inferRequestForPostprocessingMode.set_tensor("attention_mask", attentionMaskTensor);
+                    inferRequestForPostprocessingMode.set_tensor("embedding_hidden_state", embeddingsTensor);
+                }
+
+                inferRequestForPostprocessingMode.start_async();
+                inferRequestForPostprocessingMode.wait();
+
+                embeddingsTensor = inferRequestForPostprocessingMode.get_tensor(outputTensorName.c_str());
```
dkalinowski, line 376: The same issue here?

src/test/test_utils.cpp
```diff
@@ -53,6 +53,48 @@ void preparePredictRequest(ovms::InferenceRequest& request, inputs_info_t reques
...
 
+void printTensor(const ov::Tensor& tensor) {
+    const auto& elementType = tensor.get_element_type();
+    // Get pointer to data
+    const void* dataPtr = tensor.data();
+    // Handle different data types (example for float)
+    if (elementType == ov::element::f32) {
+        const float* data = static_cast<const float*>(dataPtr);
+        std::cout << "Tensor data (f32): ";
+        for (size_t i = 0; i < 20; ++i) {
```
dkalinowski: ensure there are at least 20 elements? otherwise we will have crashes in tests

dkalinowski: I dont think its used somewhere?

src/sidepacket_servable.cpp
```diff
@@ -128,22 +130,26 @@ void SidepacketServable::initialize(const std::string& modelDir, const std::stri
...
     }
 
-    ov::Core core;
-    std::shared_ptr<ov::Model> m_model = core.read_model(parsedModelsPath / std::filesystem::path("openvino_model.xml"), {}, properties);
-    m_model = this->applyPrePostProcessing(m_model);
-    compiledModel = core.compile_model(m_model, targetDevice, properties);
     auto& ovmsConfig = ovms::Config::instance();
     uint32_t numberOfParallelInferRequests = 1;
     if (ovmsConfig.nireq() > 0) {
-        // nireq is set globally for all models in ovms startup parameters
```
dkalinowski, line 138: I dont think removing this comment is helpful

src/test/test_utils.cpp
rasapala: Currenlty not, I have added it because if is usefull for debug.

src/embeddings/embeddings_calculator_ov.cc
dkalinowski: ?

src/embeddings/embeddings_calculator_ov.cc
dkalinowski: ?

src/test/test_utils.cpp
```diff
@@ -53,6 +53,48 @@ void preparePredictRequest(ovms::InferenceRequest& request, inputs_info_t reques
...
 
+void printTensor(const ov::Tensor& tensor) {
+    const auto& elementType = tensor.get_element_type();
+    // Get pointer to data
+    const void* dataPtr = tensor.data();
+    // Handle different data types (example for float)
+    if (elementType == ov::element::f32) {
+        const float* data = static_cast<const float*>(dataPtr);
+        std::cout << "Tensor data (f32): ";
+        for (size_t i = 0; i < tensor.get_size(); ++i) {
```
dkalinowski: I dont think its good idea to flood the console, keep it 20 but dont introduce segfaults

dkalinowski (approved):

dtrawins (approved):

PR #3934, opened 2026-01-29
CVS-179718 More drogon logs

dkalinowski:
### 🛠 Summary

CVS-179718

### DIFF
--- a/src/drogon_http_server.cpp
+++ b/src/drogon_http_server.cpp
@@ -44,11 +44,12 @@ DrogonHttpServer::DrogonHttpServer(size_t numWorkersForUnary, size_t numWorkersF
 
     const char* envVarValue = std::getenv("DROGON_LOG_LEVEL");
     if (envVarValue != nullptr) {
-        int logLevel = std::atoi(envVarValue);
-        if (logLevel < 0 || logLevel >= trantor::Logger::kNumberOfLogLevels) {
-            SPDLOG_WARN("Invalid DROGON_LOG_LEVEL value {}, using default log level INFO", logLevel);
+        auto logLevelOpt = stoi32(envVarValue);
+        if (!logLevelOpt.has_value() || logLevelOpt.value() < 0 || logLevelOpt.value() >= trantor::Logger::kNumberOfLogLevels) {
+            SPDLOG_WARN("Invalid DROGON_LOG_LEVEL value, using default log level INFO", envVarValue);
             trantor::Logger::setLogLevel(trantor::Logger::kInfo);
         } else {
+            int logLevel = logLevelOpt.value();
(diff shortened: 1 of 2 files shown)


third_party/drogon/ovms_drogon_trantor.patch
```diff
@@ -270,10 +270,22 @@ index 2aff5a0..145f9eb 100755
...
+index 6a75707..79d042d 100644
 --- a/trantor/trantor/net/inner/TcpConnectionImpl.cc
 +++ b/trantor/trantor/net/inner/TcpConnectionImpl.cc
-@@ -391,7 +391,9 @@ void TcpConnectionImpl::sendInLoop(const char *buffer, size_t length)
+@@ -252,7 +252,10 @@ void TcpConnectionImpl::writeCallback()
+     }
+     else
+     {
+-        LOG_SYSERR << "no writing but write callback called";
++        LOG_WARN << "write callback but not writing, fd=" << socketPtr_->fd()
```
dkalinowski, line 281: ensure we still have it logged

mzegla (approved):

third_party/drogon/ovms_drogon_trantor.patch
dkalinowski:
```
    enum LogLevel
    {
        kTrace = 0,
        kDebug,
        kInfo,
        kWarn,
        kError,
        kFatal,
        kNumberOfLogLevels
    };
```

warn is above info, just like fatal, we use info in ovms.
the log will remain visible

atobiszei: It would be useful to had variable like OVMS_DROGON_LOG_LEVEL to be able to set what is printed additionally?
Beside that LGTM.

dkalinowski: done

src/drogon_http_server.cpp
```diff
@@ -41,7 +41,24 @@ DrogonHttpServer::DrogonHttpServer(size_t numWorkersForUnary, size_t numWorkersF
     SPDLOG_DEBUG("Starting http thread pool for streaming ({} threads)", numWorkersForStreaming);
     pool->StartWorkers();  // this tp is for streaming workload which cannot use drogon's internal listener threads
     SPDLOG_DEBUG("Thread pool started");
-    trantor::Logger::setLogLevel(trantor::Logger::kInfo);
+
+    const char* envVarValue = std::getenv("DROGON_LOG_LEVEL");
+    if (envVarValue != nullptr) {
+        int logLevel = std::atoi(envVarValue);
```
atobiszei: Use ovms string utils?

atobiszei (approved):

PR #3935, opened 2026-01-29
Remove npu compiler, plugin and gpu plugin from cpu docker image

dkalinowski:
### 🛠 Summary

CVS-180114
-150mb

### DIFF
--- a/Dockerfile.ubuntu
+++ b/Dockerfile.ubuntu
@@ -445,9 +445,9 @@ COPY --from=build /usr/local/lib/python3.*/dist-packages/markupsafe /ovms/python
 
 # For CPU images remove unused libraries
 RUN if [ "$GPU" = "0" ]; then \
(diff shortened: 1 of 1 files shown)


Dockerfile.ubuntu
```diff
@@ -443,6 +443,13 @@ COPY --from=build /usr/local/lib/python3.*/dist-packages/jinja2 /ovms/python_dep
 COPY --from=build /usr/local/lib/python3.*/dist-packages/jinja2-3.1.6.dist-info /ovms/python_deps/jinja2-3.1.6.dist-info
 COPY --from=build /usr/local/lib/python3.*/dist-packages/markupsafe /ovms/python_deps/markupsafe
 
+# For CPU images remove unused libraries
+RUN if [ "$GPU" = "0" ]; then \
```
dkalinowski, line 447: it still increases final image size, we should do it other way

PR #3949, opened 2026-02-04
Add whisper timestamps support + fixes

michalkulakowski:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/audio/speech_to_text/s2t_calculator.cc
+++ b/src/audio/speech_to_text/s2t_calculator.cc
@@ -107,124 +107,129 @@ public:
         if (endpoint == Endpoint::UNSUPPORTED) {
             return absl::InvalidArgumentError(absl::StrCat("Unsupported URI: ", payload.uri));
         }
-
-        if (payload.multipartParser->hasParseError())
-            return absl::InvalidArgumentError("Failed to parse multipart data");
-
-        std::string_view stream = payload.multipartParser->getFieldByName("stream");
-        if (!stream.empty()) {
-            return absl::InvalidArgumentError("streaming is not supported");
-        }
-        std::string_view file = payload.multipartParser->getFileContentByFieldName("file");
-        if (file.empty()) {
-            return absl::InvalidArgumentError(absl::StrCat("File parsing fails"));
-        }
-
-        std::vector<float> rawSpeech;
         try {
-            if (isWavBuffer(std::string(file))) {
-                SPDLOG_DEBUG("Received file format: wav");
-                rawSpeech = readWav(file);
-            } else {
-                rawSpeech = readMp3(file);
-                SPDLOG_DEBUG("Received file format: mp3");
+            if (payload.multipartParser->hasParseError())
+                return absl::InvalidArgumentError("Failed to parse multipart data");
+
+            std::string_view stream = payload.multipartParser->getFieldByName("stream");
+            if (!stream.empty()) {
+                return absl::InvalidArgumentError("streaming is not supported");
             }
-        } catch (std::exception&) {
-            return absl::InvalidArgumentError("Received input file is not valid wav nor mp3 audio file");
-        }
-        rapidjson::StringBuffer buffer;
-        rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
-        writer.StartObject();
-        writer.String("text");
-        if (endpoint == Endpoint::TRANSCRIPTIONS) {
-            ov::genai::WhisperGenerationConfig config = pipe->sttPipeline->get_generation_config();
-            std::string_view language = payload.multipartParser->getFieldByName("language");
-            if (!language.empty()) {
-                if (language.size() > ISO_LANG_CODE_MAX) {
-                    return absl::InvalidArgumentError("Invalid language code.");
-                }
-                config.language = "<|" + std::string(language) + "|>";
+            std::string_view file = payload.multipartParser->getFileContentByFieldName("file");
+            if (file.empty()) {
+                return absl::InvalidArgumentError(absl::StrCat("File parsing fails"));
             }
-            std::string_view timestampsType = payload.multipartParser->getFieldByName("timestamp_granularities[]");
-            config.word_timestamps = false;
-            if (!timestampsType.empty()) {
-                auto type = std::string(timestampsType);
-                if (type == "segment") {
-                    config.return_timestamps = true;
-                } else if (type == "word") {
-                    if (!pipe->enableWordTimestamps)
-                        return absl::InvalidArgumentError("Word timestamps not supported for this model");
-                    config.word_timestamps = true;
+
+            std::vector<float> rawSpeech;
+            try {
+                if (isWavBuffer(std::string(file))) {
+                    SPDLOG_DEBUG("Received file format: wav");
+                    rawSpeech = readWav(file);
                 } else {
-                    return absl::InvalidArgumentError("Invalid timestamp_granularities type. Allowed types: \"segment\", \"word\"");
+                    rawSpeech = readMp3(file);
+                    SPDLOG_DEBUG("Received file format: mp3");
+                }
+            } catch (std::exception&) {
+                return absl::InvalidArgumentError("Received input file is not valid wav nor mp3 audio file");
+            }
+            rapidjson::StringBuffer buffer;
+            rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+            writer.StartObject();
+            writer.String("text");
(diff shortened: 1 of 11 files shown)


src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -136,22 +136,75 @@ class S2tCalculator : public CalculatorBase {
...
+                config.language = "<|" + std::string(language) + "|>";
+            }
+            std::string_view timestampsType = payload.multipartParser->getFieldByName("timestamp_granularities");
+            if (!timestampsType.empty()) {
+                if (std::string(timestampsType) == "segment")
+                    config.return_timestamps = true;
+                else if (std::string(timestampsType) == "word")
+                    config.word_timestamps = true;
+                else
+                    return absl::InvalidArgumentError("Invalid timestamp_granularities type.");
```
dkalinowski: Should we list allowed types in the log?
Do we have task for documentation of this feature?

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -136,22 +136,75 @@ class S2tCalculator : public CalculatorBase {
...
+            }
+            std::string_view temperature = payload.multipartParser->getFieldByName("temperature");
+            if (!temperature.empty()) {
+                double temp;
+                try {
+                    temp = stod(std::string(temperature));
```
dkalinowski: Implement equivalent in ours string_utils?
https://github.com/openvinotoolkit/model_server/pull/3934#discussion_r2763528908

@atobiszei @michalkulakowski

dkalinowski: `std::optional<float> stof(const std::string& str);`
we need our stod

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -136,22 +136,75 @@ class S2tCalculator : public CalculatorBase {
...
+                } catch (...) {
+                    return absl::InvalidArgumentError("Invalid temperature type.");
```
dkalinowski: whats the limit? what is the range supported? how is it limited in official apis?

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -136,22 +136,75 @@ class S2tCalculator : public CalculatorBase {
...
+                }
+                config.temperature = temp;
             }
+            std::unique_lock lock(pipe->sttPipelineMutex);
+            auto result = pipe->sttPipeline->generate(rawSpeech, config);
```
dkalinowski: What's `auto`? Is it std::string or string reference? If thats just std::string (copy), you dont need to lock the sttPipelineMutex for the entire serialization time, but immediately after generate call.
Is concurrency planned? Is there epic for that?

mzegla: Yeah. Based on previous code, looks like we have unnecessary copy here. Locking indeed seems overextended.

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -136,22 +136,75 @@ class S2tCalculator : public CalculatorBase {
...
+            std::string generatedText = result;
```
mzegla: `result` likely have member like `text` right?
in such case we could go with
`std::string& generatedText = result.text;`
or even operate on such member directly since result lifetime is not limited and it's still available during writer call

src/audio/speech_to_text/s2t_servable.hpp
```diff
@@ -55,6 +55,9 @@ struct SttServable {
             SPDLOG_ERROR("Error during llm node plugin_config option parsing to JSON: {}", nodeOptions.plugin_config());
             throw std::runtime_error("Error during plugin_config option parsing");
         }
+        if (nodeOptions.target_device() == "NPU")
+            config["STATIC_PIPELINE"] = true;
+        config["word_timestamps"] = true;
```
mzegla: make configurable from graph.pbtxt?

src/multi_part_parser.hpp
```diff
@@ -16,6 +16,7 @@
 
 #include <set>
 #include <string>
+#include <vector>
```
mzegla: remove?

src/audio/speech_to_text/s2t_servable.hpp
```diff
@@ -55,6 +56,11 @@ struct SttServable {
             SPDLOG_ERROR("Error during llm node plugin_config option parsing to JSON: {}", nodeOptions.plugin_config());
             throw std::runtime_error("Error during plugin_config option parsing");
         }
+        if (nodeOptions.target_device() == "NPU")
+            config["STATIC_PIPELINE"] = true;
+        enableWordTimestamps = nodeOptions.enable_word_timestamps();
+        if (enableWordTimestamps)
+            config["word_timestamps"] = true;
```
mzegla: Aren't we at risk of not having the key in config if word timestamps are not enabled?
Since it's a parameter that has a default in node options then maybe we can simply go:
`config["word_timestamps"] = nodeOptions.enable_word_timestamps();` ?

mzegla (approved):

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -136,21 +137,80 @@ class S2tCalculator : public CalculatorBase {
...
                     return absl::InvalidArgumentError("Invalid language code.");
                 }
-                std::string genaiLanguage = "<|" + std::string(language) + "|>";
-                std::unique_lock lock(pipe->sttPipelineMutex);
-                std::string generatedText = pipe->sttPipeline->generate(rawSpeech, ov::genai::language(genaiLanguage.c_str()));
-                lock.unlock();
-                writer.String(generatedText.c_str());
+                config.language = "<|" + std::string(language) + "|>";
+            }
+            std::string_view timestampsType = payload.multipartParser->getFieldByName("timestamp_granularities[]");
```
dkalinowski: please add note that only first one is supported and add cvs

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -136,21 +137,81 @@ class S2tCalculator : public CalculatorBase {
...
-                std::unique_lock lock(pipe->sttPipelineMutex);
-                std::string generatedText = pipe->sttPipeline->generate(rawSpeech);
-                lock.unlock();
-                writer.String(generatedText.c_str());
+                config.temperature = 1.0;  // default value
+            }
+            std::unique_lock lock(pipe->sttPipelineMutex);
+            const auto& result = pipe->sttPipeline->generate(rawSpeech, config);
+            std::vector<std::string> texts = result;
+            const std::string& generatedText = result;
```
dkalinowski:
```suggestion
            const WhisperDecodedResults result = pipe->sttPipeline->generate(rawSpeech, config);
            const std::string generatedText = result;  // combination of subwords into full sentence
```
I suggest to keep the class names so that we are sure which functions are called

dkalinowski (approved):

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -136,21 +137,80 @@ class S2tCalculator : public CalculatorBase {
...
-                std::unique_lock lock(pipe->sttPipelineMutex);
-                std::string generatedText = pipe->sttPipeline->generate(rawSpeech);
-                lock.unlock();
-                writer.String(generatedText.c_str());
+                config.temperature = 1.0;  // default value
+            }
+            std::unique_lock lock(pipe->sttPipelineMutex);
+            const auto result = pipe->sttPipeline->generate(rawSpeech, config);
+            const std::string generatedText = result;
+            lock.unlock();
```
dkalinowski:
```suggestion
            const WhisperDecodedResults result = pipe->sttPipeline->generate(rawSpeech, config);
            lock.unlock();
            const std::string generatedText = result;  // word chunks concatenation to single string
```

src/audio/speech_to_text/s2t_calculator.cc
dkalinowski:
```suggestion
             // Currently supports only 1 granularity at once, CVS-XXX 
            std::string_view timestampsType = payload.multipartParser->getFieldByName("timestamp_granularities[]");
```

src/audio/speech_to_text/s2t_calculator.cc
dkalinowski: @michalkulakowski

PR #3950, opened 2026-02-04
Fix hang on legacy LLM/VLM pipelines

atobiszei:
Culprit is the missed notification for std::condition_variable from LegacyExecutor, not received in blocking wait in legacy servable.
Now we will timeout that wait and another check will happen that relies on std::future status so we will unblock waiting in servable in case notification was missed.

JIRA:CVS-179700

### DIFF
--- a/src/llm/language_model/legacy/servable.cpp
+++ b/src/llm/language_model/legacy/servable.cpp
@@ -171,9 +171,9 @@ absl::Status LegacyServable::preparePartialResponse(std::shared_ptr<GenAiServabl
         std::unique_lock lock(legacyExecutionContext->mutex);
         while (executionContext->lastStreamerCallbackOutput.size() == 0 && generationStatus != std::future_status::ready) {
             SPDLOG_LOGGER_TRACE(llm_executor_logger, "Waiting for partial data...");
-            auto cvStatus = legacyExecutionContext->executionInProgress.wait_for(lock, std::chrono::milliseconds(10));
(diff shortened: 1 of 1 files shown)


src/llm/language_model/legacy/servable.cpp
```diff
@@ -171,8 +171,11 @@ absl::Status LegacyServable::preparePartialResponse(std::shared_ptr<GenAiServabl
         std::unique_lock lock(legacyExecutionContext->mutex);
         while (executionContext->lastStreamerCallbackOutput.size() == 0 && generationStatus != std::future_status::ready) {
             SPDLOG_LOGGER_TRACE(llm_executor_logger, "Waiting for partial data...");
-            legacyExecutionContext->executionInProgress.wait(lock);
+            auto cvStatus = legacyExecutionContext->executionInProgress.wait_for(lock, std::chrono::milliseconds(10));
```
mzegla: Pretty generic name. Would something like `executionProgressStatus` make sense?

mzegla (commented):
I wonder if 10ms is not too small. Since it's more of a safety measure for sporadic situations then maybe even 1 second would be good?

atobiszei:
> I wonder if 10ms is not too small. Since it's more of a safety measure for sporadic situations then maybe even 1 second would be good?

Its still happening ~1/10 runs of full test fixture. I don't want to bring it higher as in those cases we would have response delayed by this amount and tracking this down in performance results would be a nightmare. Downside is that we do additional future status check & 2 boolean comparisons each time, but this should be negligible.

PR #3952, opened 2026-02-04
Fix dangling TCP connection issue

mzegla:


src/drogon_http_async_writer_impl.cpp
```diff
@@ -72,7 +72,10 @@ void DrogonHttpAsyncWriterImpl::PartialReplyBegin(std::function<void()> actualWo
 }
 void DrogonHttpAsyncWriterImpl::PartialReplyEnd() {
     this->stream->close();
+    //this->requestPtr.reset();
+    //this->responsePtr.reset();
```
mzegla:
```suggestion
```
Running checks to confirm it's not needed...

src/drogon_http_async_writer_impl.cpp
```diff
@@ -85,8 +88,18 @@ bool DrogonHttpAsyncWriterImpl::IsDisconnected() const {
 void DrogonHttpAsyncWriterImpl::RegisterDisconnectionCallback(std::function<void()> onDisconnectedCallback) {
     const auto& weakConnPtr = requestPtr->getConnectionPtr();
     if (auto connPtr = weakConnPtr.lock()) {
-        connPtr->setCloseCallback([onDisconnectedCallback = std::move(onDisconnectedCallback)](const trantor::TcpConnectionPtr& conn) {
-            onDisconnectedCallback();
+        // Get the original callback set by TcpServer (does connSet_.erase())
+        auto originalCallback = connPtr->getCloseCallback();
+        connPtr->setCloseCallback([originalCallback, onDisconnectedCallback = std::move(onDisconnectedCallback)](const trantor::TcpConnectionPtr& conn) {
+            // Run original callback first to ensure proper cleanup
+            if (originalCallback) {
```
dkalinowski, line 93: this could be moved to trantor/drogon logic - handling such internals of drogon should be made by drogon, not ovms. then you could remove "getCloseCallback" from api which you added in patch. what do you think?

src/http_rest_api_handler.cpp
```diff
@@ -771,12 +772,12 @@ Status HttpRestApiHandler::processV3(const std::string_view uri, const HttpReque
             auto& executor = executorWrapper.getObjectHolder()->get();
 
             if (request == nullptr || executor == nullptr) {  // should not happen
-                SPDLOG_ERROR("Not all resources for streaming inference have been properly initialized");
                 throw std::runtime_error("Not all resources for streaming inference have been properly initialized");
             }
-
+
```
dkalinowski: linter?

src/drogon_http_async_writer_impl.cpp
mzegla: Ideally the `setCloseCallback` should be useable, by default on trantor level, I agree. 
From our perspective making the change you want would require more changes in trantor, potentially even in very internal parts of the logic and expanding the patch. I'm not sure if making such change shortly before release and in the patch is a good idea.

src/http_rest_api_handler.cpp
mzegla: probably some leftover

dkalinowski (approved):

dtrawins (approved):

PR #3967, opened 2026-02-10
Updates to the Red Hat Dockerfile for Konflux

stevegrubb:
The konflux build system has trouble in 2 areas.

1) It cuts of network access at certain times. The solution was to build a group of the software in one shot.

2) Konflux causes bazel to do things differently which results in not being able to find libovms_shared.so or capi_benchmark. So, it was changed to try to put it in a known spot for searching.

These changes do not affect the build done by Docker or podman.

### DIFF
--- a/Dockerfile.redhat
+++ b/Dockerfile.redhat
@@ -347,6 +347,7 @@ ARG LTO_ENABLE=OFF
 WORKDIR /ovms
 # Konflux has a hard time locating the built binaries. Need to put it
 # in a known location.
+# hadolint ignore=DL4006 
 RUN mkdir -p /tmp/bazel-output && \
     bazel --output_base=/tmp/bazel-output \
         build --jobs=$JOBS ${CAPI_FLAGS} //src:ovms_shared && \
@@ -363,6 +364,7 @@ RUN bazel build --jobs $JOBS ${CAPI_FLAGS} //src:capi_cpp_example
 # C-API benchmark app
 # Konflux has a hard time locating the built binaries. Need to put it
 # in a known location.
+# hadolint ignore=DL4006 
 RUN mkdir -p /tmp/bazel-output && \
     bazel --output_base=/tmp/bazel-output \
         build --jobs=$JOBS ${CAPI_FLAGS} //src:capi_benchmark && \


Dockerfile.redhat
```diff
@@ -310,17 +306,18 @@ RUN bash -c "sed -i -e 's|REPLACE_PROJECT_VERSION|${PROJECT_VERSION}|g' /ovms/sr
...
 ARG OPTIMIZE_BUILDING_TESTS=0
 RUN rm -f /usr/lib64/cmake/OpenSSL/OpenSSLConfig.cmake
 
 # Builds unit tests together with ovms server in one step
 # It speeds up CI when tests are executed outside of the image building
+# Konflux build system cuts off network access at times. Build all in 1 shot.
 # hadolint ignore=SC2046
-RUN bazel build --jobs=$JOBS ${debug_bazel_flags} ${minitrace_flags} //src:ovms $(if [ "$OPTIMIZE_BUILDING_TESTS" == "1" ] ; then echo -n //src:ovms_test; fi)
+RUN bazel build --jobs=$JOBS ${debug_bazel_flags} //:ovms_dependencies @com_google_googletest//:gtest && \
+  bazel build --jobs=$JOBS ${debug_bazel_flags} //src:release_custom_nodes && \
```
dtrawins, line 318: we are using separate docker commands to take better advantage of docker cache. Usually the changes are from the target ovms and ovms_test. Why is it more reliable in one command?

Dockerfile.redhat
```diff
@@ -348,14 +345,32 @@ ARG CAPI_FLAGS="--strip=always  --config=mp_off_py_off --//:distro=redhat"
 ARG JOBS=40
 ARG LTO_ENABLE=OFF
 WORKDIR /ovms
-RUN bazel build --jobs $JOBS ${CAPI_FLAGS} //src:ovms_shared
+# Konflux has a hard time locating the built binaries. Need to put it
+# in a known location.
+RUN mkdir -p /tmp/bazel-output && \
```
dtrawins, line 351:
```suggestion
# hadolint ignore=DL4006 
RUN mkdir -p /tmp/bazel-output && \
```

Dockerfile.redhat
```diff
@@ -348,14 +345,32 @@ ARG CAPI_FLAGS="--strip=always  --config=mp_off_py_off --//:distro=redhat"
...
 
 # C api app with bazel
 # hadolint ignore=DL3059
 RUN bazel build --jobs $JOBS ${CAPI_FLAGS} //src:capi_cpp_example
 
 # C-API benchmark app
-RUN bazel build --jobs=$JOBS ${CAPI_FLAGS} //src:capi_benchmark && ./bazel-bin/src/capi_benchmark --niter 2 --nstreams 1 --servable_name "dummy"
+# Konflux has a hard time locating the built binaries. Need to put it
+# in a known location.
+RUN mkdir -p /tmp/bazel-output && \
```
dtrawins, line 368:
```suggestion
# hadolint ignore=DL4006 
RUN mkdir -p /tmp/bazel-output && \
```

Dockerfile.redhat
stevegrubb: There are times when bazel can't resolve packages due to konflux. (Konflux is designed to do hermetic builds. Of course, this is impossible due to bazel. But it is strict in it's interpretation of standards.) The only fix we could think of is move everything to a window when it has network access. The problem doesn't show up in podman or docker. But the solution doesn't affect them either. They still build.

stevegrubb: And maybe konflux blows up the cache between RUN commands? Konflux takes a very strict interpretation of what the OCI standard says. I wasn't the one that worked on this. Our DevOps team was tasked with the konflux migration. I am passing along what I heard discussed as they tried to solve it. When all 3 are in the same RUN command, it works.

PR #4039, opened 2026-03-04
Mkulakow/responses api

michalkulakowski:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``


demos/continuous_batching/vlm/README.md
```diff
@@ -119,6 +119,45 @@ curl http://localhost:8000/v3/chat/completions  -H "Content-Type: application/js
 ```
 :::
 
+:::{dropdown} **Unary call with curl using responses endpoint**
+**Note**: using urls in request requires `--allowed_media_domains` parameter described [here](../../../docs/parameters.md)
```
dkalinowski:
```suggestion
:::{dropdown} **Unary call with cURL using Responses API**
**Note**: Using cURL in request requires `--allowed_media_domains` parameter described [here](../../../docs/parameters.md).
```

demos/continuous_batching/vlm/README.md
```diff
@@ -119,6 +119,45 @@ curl http://localhost:8000/v3/chat/completions  -H "Content-Type: application/js
...
+
+```bash
+curl http://localhost:8000/v3/responses  -H "Content-Type: application/json" -d "{ \"model\": \"OpenGVLab/InternVL2-2B\", \"input\":[{\"role\": \"user\", \"content\": [{\"type\": \"input_text\", \"text\": \"Describe what is on the picture.\"},{\"type\": \"input_image\", \"image_url\": \"http://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/releases/2025/3/demos/common/static/images/zebra.jpeg\"}]}], \"max_output_tokens\": 100}"
```
dkalinowski: Can you break down the command into multiple lines to showcase what is actually being sent?

demos/continuous_batching/README.md
```diff
@@ -148,9 +148,76 @@ curl -s http://localhost:8000/v3/chat/completions -H "Content-Type: application/
 :::
 
 
+### Unary calls to responses endpoint using cURL
```
dkalinowski:
```suggestion
### Unary calls via Responses API using cURL
```

demos/continuous_batching/README.md
```diff
@@ -148,9 +148,76 @@ curl -s http://localhost:8000/v3/chat/completions -H "Content-Type: application/
...
+          "annotations": []
+        }
+      ]
+    }
+  ],
+  "usage": {
+    "input_tokens": 27,
+    "input_tokens_details": { "cached_tokens": 0 },
+    "output_tokens": 30,
+    "output_tokens_details": { "reasoning_tokens": 0 },
```
dkalinowski: Is reasoning_tokens supported when the model actually reason?
I also cannot see any example showcasing how message looks whenever there is reasoning

demos/continuous_batching/vlm/README.md
```diff
@@ -223,6 +262,79 @@ The picture features a zebra standing in a grassy area. The zebra is characteriz
 
 :::
 
+:::{dropdown} **Streaming request with OpenAI client using responses endpoint**
```
dkalinowski:
```suggestion
:::{dropdown} **Streaming request with OpenAI client via Responses API**
```

demos/continuous_batching/vlm/README.md
```diff
@@ -223,6 +262,79 @@ The picture features a zebra standing in a grassy area. The zebra is characteriz
...
+Output:
+```
+The picture features a zebra standing in a grassy area. The zebra is characterized by its distinctive black and white striped pattern, which covers its entire body, including its legs, neck, and head. Zebras have small, rounded ears and a long, flowing tail. The background appears to be a natural grassy habitat, typical of a savanna or plain.
+```
+
+:::
+
+## Benchmarking text generation with high concurrency
+
+OpenVINO Model Server employs efficient parallelization for text generation. It can be used to generate text also in high concurrency in the environment shared by multiple clients.
```
dkalinowski: Why is it being added in responses api PR? Is it related?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,231 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+
+    if (includeUsage) {
+        writer.String("usage");
+        writer.StartObject();
+        writer.String("input_tokens");
+        writer.Uint64(static_cast<uint64_t>(usage.promptTokens));
+        writer.String("input_tokens_details");
+        writer.StartObject();
+        writer.String("cached_tokens");
+        writer.Uint64(0);
```
dkalinowski: please add note here that it is not supported

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,231 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+        writer.EndObject();
+        writer.String("output_tokens");
+        writer.Uint64(static_cast<uint64_t>(usage.completionTokens));
+        writer.String("output_tokens_details");
+        writer.StartObject();
+        writer.String("reasoning_tokens");
+        writer.Uint64(0);
```
dkalinowski: can we support it in this release?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,6 +95,106 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    writer.String("output_tokens");
+    writer.Int(usage.completionTokens);
+    writer.String("output_tokens_details");
+    writer.StartObject();
+    writer.String("reasoning_tokens");
+    writer.Int(0);
+    writer.EndObject();
+    writer.String("total_tokens");
+    writer.Int(usage.calculateTotalTokens());
+    writer.EndObject();
```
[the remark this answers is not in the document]
dkalinowski: @michalkulakowski can you protect against such corner cases?

src/llm/apis/openai_completions.cpp
```diff
@@ -247,6 +471,193 @@ absl::Status OpenAIChatCompletionsHandler::ensureArgumentsInToolCalls(Value& mes
...
+    if (inputIt == doc.MemberEnd()) {
+        return absl::InvalidArgumentError("input missing in request");
+    }
+
+    auto& allocator = doc.GetAllocator();
+    rapidjson::Value messages(rapidjson::kArrayType);
+
+    if (inputIt->value.IsString()) {
+        request.prompt = inputIt->value.GetString();
+        if (!request.prompt.has_value() || request.prompt.value().empty()) {
```
dkalinowski: request.prompt must have value if you set it one line earlier

src/llm/apis/openai_completions.cpp
```diff
@@ -247,6 +471,193 @@ absl::Status OpenAIChatCompletionsHandler::ensureArgumentsInToolCalls(Value& mes
...
+                        } catch (std::runtime_error& e) {
+                            std::stringstream ss;
+                            ss << "Image file " << imageUrl.c_str() << " parsing failed: " << e.what();
+                            SPDLOG_LOGGER_DEBUG(llm_calculator_logger, ss.str());
+                            return absl::InvalidArgumentError(ss.str());
+                        }
+                    }
+                    request.imageHistory.push_back({i, tensor});
+                } else {
+                    return absl::InvalidArgumentError("Unsupported content type");
```
dkalinowski: can you include whats the content type? this way we can easily see in logs what was expected supported content type by user

src/llm/apis/openai_completions.cpp
```diff
@@ -247,6 +471,193 @@ absl::Status OpenAIChatCompletionsHandler::ensureArgumentsInToolCalls(Value& mes
...
+        }
+    } else {
+        return absl::InvalidArgumentError("input is not a string or array");
+    }
+
+    auto existingMessages = doc.FindMember("messages");
+    if (existingMessages != doc.MemberEnd()) {
+        existingMessages->value = messages;
+    } else {
+        doc.AddMember("messages", messages, allocator);
```
dkalinowski: Why is the document being edited? If I understand correctly, we pare `doc` to our `request` format. Why do we need to edit the `doc`? Shouldnt it be left untouched and just work on `request` in next steps? Why do we parse if we later not use it?

src/llm/apis/openai_completions.cpp
```diff
@@ -444,38 +865,71 @@ absl::Status OpenAIChatCompletionsHandler::parseTools() {
             auto& obj = it->value.GetArray()[i];
             if (!obj.IsObject())
                 return absl::InvalidArgumentError("Tool is not a JSON object");
+            rapidjson::Value* functionObj = nullptr;
```
dkalinowski: functionObj is never used outside of scope, you dont need to work on pointers, contrary to parametersValue where it is required

src/llm/apis/openai_completions.cpp
```diff
@@ -444,38 +865,71 @@ absl::Status OpenAIChatCompletionsHandler::parseTools() {
...
+                // now we want to insert to a mapping of
+                // tool name -> tool schema representations struct
+                // Dump parameters object to string since this is the schema format expected by GenAI
+                // Keep the rapidjson::Value object as well to avoid re-parsing in outputParsers
+                rapidjson::StringBuffer buffer;
+                rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+                parametersValue->Accept(writer);
+                std::string parametersStr = buffer.GetString();
+                ToolSchemaWrapper schemaReprs{parametersValue, std::move(parametersStr)};
+                request.toolNameSchemaMap[functionNameCStr] = std::move(schemaReprs);
```
dkalinowski:
```suggestion
                request.toolNameSchemaMap[functionName] = std::move(schemaReprs);
```

src/llm/apis/openai_completions.cpp
```diff
@@ -656,6 +756,89 @@ absl::Status OpenAIChatCompletionsHandler::parseChatCompletionsPart(std::optiona
...
+    // logprobs: bool; optional - defaults to false
+    it = doc.FindMember("logprobs");
+    if (it != doc.MemberEnd() && !it->value.IsNull()) {
+        if (!it->value.IsBool())
+            return absl::InvalidArgumentError("logprobs accepts values true or false");
+        request.logprobschat = it->value.GetBool();
+    }
+    if (request.logprobschat && request.stream) {
+        return absl::InvalidArgumentError("logprobs are not supported in streaming mode.");
+    }
```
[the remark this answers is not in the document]
dkalinowski: its probably old comment which already been fixed, right? @michalkulakowski

src/llm/apis/openai_completions.cpp
```diff
@@ -863,12 +1393,15 @@ absl::Status OpenAIChatCompletionsHandler::parseCommonPart(std::optional<uint32_
...
     // Supported by OpenAI API and vLLM, supported in CB lib by mapping to num_return_sequences param
+    // Not supported for RESPONSES streaming - output_index is hardcoded to 0
     it = doc.FindMember("n");
     if (it != doc.MemberEnd() && !it->value.IsNull()) {
         if (!it->value.IsUint())
             return absl::InvalidArgumentError("n is not an unsigned integer");
         if (it->value.GetUint() == 0)
             return absl::InvalidArgumentError("n value should be greater than 0");
+        if (endpoint == Endpoint::RESPONSES && request.stream && it->value.GetUint() > 1)
+            return absl::InvalidArgumentError("n greater than 1 is not supported for responses streaming");
```
dkalinowski:
```suggestion
            return absl::InvalidArgumentError("n greater than 1 is not supported for Responses API streaming");
```

src/llm/apis/openai_completions.cpp
```diff
@@ -1112,6 +1658,15 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::Enco
     OVMS_PROFILE_FUNCTION();
     usage.promptTokens = results.perf_metrics.get_num_input_tokens();
     usage.completionTokens = results.perf_metrics.get_num_generated_tokens();
+    if (endpoint == Endpoint::RESPONSES) {
+        std::vector<ParsedOutput> parsedOutputs;
+        for (const auto& tokens : results.tokens) {
+            updateUsage(usage, tokens, request.echo);
```
dkalinowski: We just recently switched to using PerfMetrics @mzegla 
Why cant we use it here?

src/llm/apis/openai_completions.cpp
```diff
@@ -1248,6 +1824,166 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::VLMD
...
+        const auto createdAt = std::chrono::duration_cast<std::chrono::microseconds>(created.time_since_epoch()).count();
+        const std::string responseId = "resp-" + std::to_string(createdAt);
+        const std::string outputItemId = "msg-0";
+
+        std::vector<std::string> events;
+        if (!responsesStreamingInitialized) {
+            events.emplace_back(serializeResponsesEvent([this, &responseId, createdAt](Writer<StringBuffer>& writer) {
+                writer.StartObject();
+                writer.String("type");
+                writer.String("response.created");
```
dkalinowski: it might not be good place for this event to be streamed, please verify vs other servings maybe if OpenAI is not specific enough

dkalinowski: and all other events

demos/continuous_batching/vlm/README.md
```diff
@@ -119,6 +119,65 @@ curl http://localhost:8000/v3/chat/completions  -H "Content-Type: application/js
...
+          "type": "output_text",
+          "text": "The picture features a zebra standing in a grassy plain. Zebras are known for their distinctive black and white striped patterns, which help them blend in for camouflage purposes.",
+          "annotations": []
+        }
+      ]
+    }
+  ],
+  "usage": {
+    "input_tokens": 19,
+    "input_tokens_details": { "cached_tokens": 0 },
```
mzegla, line 172: what is that cached_tokens field?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
     throw std::invalid_argument("Unsupported JSON value type");
 }
 
+std::string serializeResponsesEvent(const std::function<void(Writer<StringBuffer>&)>& eventSerializer) {
```
mzegla: how does it work? evenSerializer carries the data itself?
it's not clear to me what the output string would be

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+void serializeNotSupportedZeroField(Writer<StringBuffer>& writer, const char* fieldName) {
+    writer.String(fieldName);
+    writer.Uint64(0);
+}
+
+void serializeNotSupportedEmptyArrayField(Writer<StringBuffer>& writer, const char* fieldName) {
+    writer.String(fieldName);
+    writer.StartArray();
+    writer.EndArray();
+}
```
mzegla: why not supported?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+        writer.String("name");
+        writer.String(toolName.c_str());
+        writer.String("parameters");
+        writer.RawValue(toolSchemaWrapper.stringRepr.c_str(), toolSchemaWrapper.stringRepr.size(), rapidjson::kObjectType);
+        writer.EndObject();
+    }
+    writer.EndArray();
+}
+
+void OpenAIChatCompletionsHandler::serializeResponsesResponseObject(Writer<StringBuffer>& writer, const std::string& responseId, int64_t createdAt,
```
mzegla: serializeResponsesResponseObject -> too much of a wrap to me

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    const char* status, const std::string& fullOutputText, bool includeUsage,
+    const char* incompleteReason, const char* errorMessage, const char* errorCode) const {
```
mzegla: Lots of parameters. Can we use std::strings and get const char* internally, so it's more straightforward to use?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    writer.Bool(false);
+    serializeNotSupportedNullField(writer, "previous_response_id");
+    serializeNotSupportedNullField(writer, "reasoning");
+    writer.String("store");
+    writer.Bool(true);
+    writer.String("temperature");
+    if (request.temperature.has_value()) {
+        writer.Double(static_cast<double>(request.temperature.value()));
+    } else {
+        writer.Double(1.0);
```
mzegla: is it still our default? temperature=1?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    writer.String(status);
+
+    writer.String("parallel_tool_calls");
+    writer.Bool(false);
+    serializeNotSupportedNullField(writer, "previous_response_id");
+    serializeNotSupportedNullField(writer, "reasoning");
+    writer.String("store");
+    writer.Bool(true);
+    writer.String("temperature");
+    if (request.temperature.has_value()) {
```
mzegla: Do we put sampling parameter ins the response?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    writer.StartObject();
+    writer.String("format");
+    writer.StartObject();
+    writer.String("type");
+    writer.String("text");
+    writer.EndObject();
+    writer.EndObject();
+    serializeResponsesToolChoice(writer);
+    serializeResponsesTools(writer);
+    writer.String("top_p");
```
mzegla: Same as temperature. Shall we include it?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+        writer.String("message");
+        writer.String("role");
+        writer.String("assistant");
+        writer.String("status");
+        if (std::string(status) == "completed") {
+            writer.String("completed");
+        } else if (std::string(status) == "incomplete") {
+            writer.String("incomplete");
+        } else {
+            writer.String("in_progress");
```
mzegla: can this happen? in unary mode we either have completed or not

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    writer.String(outputItemId.c_str());
+    writer.String("type");
+    writer.String("message");
+    writer.String("role");
+    writer.String("assistant");
+    writer.String("status");
+    writer.String(status);
+    writer.String("content");
+    writer.StartArray();
+    if (withContent) {
```
mzegla: do we need the bool? could we rely on empty string for example? 
of empty content is still a valid thing to send?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+        serializeResponsesPart(writer, text);
+    }
+    writer.EndArray();
+    writer.EndObject();
+}
+
+void OpenAIChatCompletionsHandler::serializeResponsesPart(Writer<StringBuffer>& writer, const std::string& text) {
```
mzegla: "responses part" - are those the only things specific to responses endpoint?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    serializeNotSupportedNullField(writer, "error");
+    writer.String("model");
+    writer.String(request.model.c_str());
+    writer.String("status");
+    writer.String(responseStatus);
+
+    writer.String("parallel_tool_calls");
+    writer.Bool(false);
+    serializeNotSupportedNullField(writer, "previous_response_id");
+    serializeNotSupportedNullField(writer, "reasoning");
```
mzegla: reasoning is an unsupported field?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    writer.StartObject();
+    writer.String("format");
+    writer.StartObject();
+    writer.String("type");
+    writer.String("text");
+    writer.EndObject();
+    writer.EndObject();
+    serializeResponsesToolChoice(writer);
+    serializeResponsesTools(writer);
+    writer.String("top_p");
```
mzegla: Hmm, that function looks familiar. Doesn't it duplicate a lot of logic from one of the previous functions?

src/llm/apis/openai_completions.cpp
```diff
@@ -95,8 +96,343 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    writer.Uint64(static_cast<uint64_t>(usage.promptTokens));
+    writer.String("input_tokens_details");
+    writer.StartObject();
+    serializeNotSupportedZeroField(writer, "cached_tokens");
+    writer.EndObject();
+    writer.String("output_tokens");
+    writer.Uint64(static_cast<uint64_t>(usage.completionTokens));
+    writer.String("output_tokens_details");
+    writer.StartObject();
+    serializeNotSupportedZeroField(writer, "reasoning_tokens");
```
mzegla: Is it  specified in the API that it should be 0 for not supported?
If not, then this might be misleading as user might think there really were no reasoning tokens and it might not always be the case

src/llm/apis/openai_completions.cpp
```diff
@@ -247,6 +583,171 @@ absl::Status OpenAIChatCompletionsHandler::ensureArgumentsInToolCalls(Value& mes
...
+                        auto urlIt = imageUrlObj.FindMember("url");
+                        if (urlIt == imageUrlObj.MemberEnd() || !urlIt->value.IsString()) {
+                            return absl::InvalidArgumentError("input_image.image_url.url is missing or invalid");
+                        }
+                        imageUrl = urlIt->value.GetString();
+                    } else {
+                        return absl::InvalidArgumentError("input_image.image_url must be a string or object");
+                    }
+
+                    std::string pattern = "base64,";
```
mzegla: why ',' at the end?

src/llm/apis/openai_completions.cpp
```diff
@@ -247,6 +583,171 @@ absl::Status OpenAIChatCompletionsHandler::ensureArgumentsInToolCalls(Value& mes
...
+                            tensor = loadImageStbiFromMemory(decoded);
+                        } catch (std::runtime_error& e) {
+                            std::stringstream ss;
+                            ss << "Image parsing failed: " << e.what();
+                            SPDLOG_LOGGER_DEBUG(llm_calculator_logger, ss.str());
+                            return absl::InvalidArgumentError(ss.str());
+                        }
+                    } else if (std::regex_match(imageUrl.c_str(), std::regex("^(http|https|ftp|sftp|)://(.*)"))) {
+                        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Loading image using curl");
+                        int64_t sizeLimit = 20000000;  // restrict single image size to 20MB
```
mzegla: Don't we have such limit somewhere else as well? Maybe we could have some global static const, so we define it in one place?

src/llm/apis/openai_completions.cpp
```diff
@@ -247,6 +583,171 @@ absl::Status OpenAIChatCompletionsHandler::ensureArgumentsInToolCalls(Value& mes
...
+                            return absl::InvalidArgumentError(ss.str());
+                        }
+                    }
+                    request.imageHistory.push_back({i, tensor});
+                } else {
+                    return absl::InvalidArgumentError("Unsupported content type. Supported types are input_text and input_image.");
+                }
+            }
+
+            request.chatHistory.last()["content"] = contentText;
```
mzegla: Not sure if I follow nesting correctly, but if there is different ordering (for example input text is first and input image is second) wouldn't we overwrite "content" with uninitialized contentText?
We have contentText as local variable in line 638 and only write to it in 655. 
So we could write to it input text on i=0 and overwrite it with empty string in i=1 right?

src/llm/apis/openai_completions.cpp
```diff
@@ -415,16 +916,26 @@ absl::Status OpenAIChatCompletionsHandler::parseTools() {
             if (tool_choice != "none" && tool_choice != "auto" && tool_choice != "required")
                 return absl::InvalidArgumentError("tool_choice should be either 'none' or 'auto' or 'required'");
         } else if (tool_choice_it->value.IsObject()) {
-            auto tool_choice_functionIt = tool_choice_it->value.GetObject().FindMember("function");
-            if (tool_choice_functionIt != tool_choice_it->value.GetObject().MemberEnd() && tool_choice_functionIt->value.IsObject()) {
+            auto toolChoiceObj = tool_choice_it->value.GetObject();
+            auto tool_choice_functionIt = toolChoiceObj.FindMember("function");
+            if (tool_choice_functionIt != toolChoiceObj.MemberEnd() && tool_choice_functionIt->value.IsObject()) {
```
mzegla: Can we also fix naming to follow one convention?

src/llm/apis/openai_completions.cpp
mzegla, line 1: General comment - I don't like the idea of expanding this file for new endpoint.
This catalog was meant to contain support for different APIs in different files.
Can we extract responses stuff to: openai_responses.hpp/cpp and for common stuff we could have openai_common.cpp/hpp?

src/llm/apis/openai_completions.cpp
```diff
@@ -444,38 +955,71 @@ absl::Status OpenAIChatCompletionsHandler::parseTools() {
             auto& obj = it->value.GetArray()[i];
             if (!obj.IsObject())
                 return absl::InvalidArgumentError("Tool is not a JSON object");
+            rapidjson::Value* functionObj = nullptr;
+            rapidjson::Value* parametersValue = nullptr;
+            const char* functionNameCStr = nullptr;
```
mzegla: Do we really need all those raw pointers?

src/llm/apis/openai_completions.cpp
```diff
@@ -444,38 +955,71 @@ absl::Status OpenAIChatCompletionsHandler::parseTools() {
...
+                }
+                functionNameCStr = nameIt->value.GetString();
+
+                auto parametersIt = obj.FindMember("parameters");
+                if (parametersIt != obj.MemberEnd()) {
+                    parametersValue = &parametersIt->value;
+                }
+            }
+
+            std::string functionName = functionNameCStr;
```
mzegla: we could write to functionName directly right? 
that functionNameCStr seems like unnecessary intermediate step

src/llm/apis/openai_completions.cpp
```diff
@@ -1172,6 +1829,27 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::VLMD
     OVMS_PROFILE_FUNCTION();
     usage.promptTokens = results.perf_metrics.get_num_input_tokens();
     usage.completionTokens = results.perf_metrics.get_num_generated_tokens();
+    if (endpoint == Endpoint::RESPONSES) {
+        std::vector<ParsedOutput> parsedOutputs;
+        usage.completionTokens = 0;
+        for (const std::string& text : results.texts) {
+            auto result = tokenizer.encode(text);
```
mzegla: Do we really need to do the tokenization here?

src/llm/apis/openai_completions.cpp
```diff
@@ -1172,6 +1829,27 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::VLMD
     OVMS_PROFILE_FUNCTION();
     usage.promptTokens = results.perf_metrics.get_num_input_tokens();
     usage.completionTokens = results.perf_metrics.get_num_generated_tokens();
+    if (endpoint == Endpoint::RESPONSES) {
+        std::vector<ParsedOutput> parsedOutputs;
+        usage.completionTokens = 0;
```
mzegla: Why override with 0 here?

src/llm/apis/openai_completions.cpp
```diff
@@ -1246,8 +1924,190 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::VLMD
...
+                writer.String("output_index");
+                writer.Uint64(0);
+                writer.String("content_index");
+                writer.Uint64(0);
+                writer.String("item_id");
+                writer.String(outputItemId.c_str());
+                writer.String("part");
+                serializeResponsesPart(writer, responsesStreamingOutputText);
+                writer.EndObject();
+            }));
```
mzegla: Could such insertions be wrapped? Feels like a lot of replicated code is there to handle JSON building.

src/llm/apis/openai_completions.hpp
```diff
@@ -121,12 +140,16 @@ class OpenAIChatCompletionsHandler {
...
+    const rapidjson::Value* getRawChatTemplateKwargs() const;
 
     std::string serializeUnaryResponse(const std::vector<ov::genai::GenerationOutput>& generationOutputs);
     std::string serializeUnaryResponse(ov::genai::EncodedResults& results);
     std::string serializeUnaryResponse(ov::genai::VLMDecodedResults& results);
     std::string serializeStreamingChunk(const std::string& chunkResponse, ov::genai::GenerationFinishReason finishReason);
     std::string serializeStreamingUsageChunk();
     std::string serializeStreamingHandshakeChunk();
+    std::string serializeResponsesStreamingInitEvents();
+    std::string serializeResponsesFailedEvent(const std::string& errorMessage, const char* errorCode = "server_error");
```
mzegla: error code as string? maybe we could have some enum here

src/llm/http_llm_calculator.cc
```diff
@@ -160,8 +176,22 @@ class HttpLLMCalculator : public CalculatorBase {
...
+            }
             return absl::InvalidArgumentError(e.what());
         } catch (...) {
+            if (executionContext->apiHandler && executionContext->apiHandler->isStream() && executionContext->endpoint == Endpoint::RESPONSES) {
+                std::string failedEvent = executionContext->apiHandler->serializeResponsesFailedEvent("Response generation failed");
+                executionContext->response = wrapTextInServerSideEventMessage(failedEvent);
+                executionContext->response += wrapTextInServerSideEventMessage("[DONE]");
+                cc->Outputs().Tag(OUTPUT_TAG_NAME).Add(new std::string{std::move(executionContext->response)}, iterationBeginTimestamp);
+                return absl::OkStatus();
+            }
```
mzegla: Are you sure that it will be handled correctly with OK status? I mean for example, will it exit properly?

src/llm/http_llm_calculator.cc
```diff
@@ -125,6 +125,22 @@ class HttpLLMCalculator : public CalculatorBase {
...
+                // immediately after scheduling, before blocking on readPartialExecutionResults.
+                // This reduces perceived latency - the client sees the response is created right away.
+                if (executionContext->apiHandler->isStream() && executionContext->endpoint == Endpoint::RESPONSES) {
+                    std::string initEvents = executionContext->apiHandler->serializeResponsesStreamingInitEvents();
+                    if (!initEvents.empty()) {
+                        executionContext->response = wrapTextInServerSideEventMessage(initEvents);
+                        cc->Outputs().Tag(OUTPUT_TAG_NAME).Add(new std::string{std::move(executionContext->response)}, iterationBeginTimestamp);
+                        executionContext->response = "";
+                    }
+                    cc->Outputs().Tag(LOOPBACK_TAG_NAME).Add(new bool{true}, iterationBeginTimestamp);
```
mzegla: Is that whole thing okay? You schedule and don't read/check output before return. 
Aren't we at risk of skipping chunk? Especially if it's generated fast?

src/llm/py_jinja_template_processor.cpp
```diff
@@ -89,4 +91,123 @@ bool PyJinjaTemplateProcessor::applyChatTemplate(PyJinjaTemplateProcessor& templ
     return false;
 }
 
+static py::object rapidJsonValueToPyObject(const rapidjson::Value& value) {
```
mzegla: Is that whole thing necessary? Why don't we serialize to string -> pass to python -> read to json?

src/llm/servable.cpp
```diff
@@ -204,6 +206,50 @@ absl::Status GenAiServable::prepareInputs(std::shared_ptr<GenAiServableExecution
...
+            }
+        } else {
+            auto prompt = executionContext->apiHandler->getPrompt();
+            if (!prompt.has_value()) {
+                return absl::Status(absl::StatusCode::kInvalidArgument, "input is missing");
+            }
+            inputText = prompt.value();
+        }
+        break;
+    }
```
mzegla, line 257: Not sure where (maybe it's already done in the common part), but we want to print inputText on DEBUG or TRACE log level

mzegla (commented):
It's a big change, let's review it iteratively. Pushing comments only to the logic - did not review tests just yet.

src/llm/servable.cpp
michalkulakowski: Its already done in the common part

demos/continuous_batching/README.md
michalkulakowski: Its supported, its just not documented right now

demos/continuous_batching/accuracy/gorilla.patch
```diff
@@ -29,6 +38,30 @@ index bb625d2..7204adb 100644
...
++        url="http://localhost:8000/v3",
++        org="ovms",
++        license="apache-2.0",
++        model_handler=OVMSResponsesHandler,
++        input_price=None,
++        output_price=None,
++        is_fc_model=True,
++        underscore_to_dot=True,
++    ),
++    "ovms-model-stream-responses": ModelConfig(
```
mzegla: maybe `ovms-model-responses-stream`?
more straightforward switching

src/llm/servable.cpp
```diff
@@ -204,6 +206,45 @@ absl::Status GenAiServable::prepareInputs(std::shared_ptr<GenAiServableExecution
...
+    case Endpoint::RESPONSES: {
+        if (executionContext->apiHandler->getChatHistory().size() > 0) {
+#if (PYTHON_DISABLE == 0)
+            bool success = PyJinjaTemplateProcessor::applyChatTemplate(getProperties()->templateProcessor, getProperties()->modelsPath, executionContext->apiHandler->getProcessedJson(), inputText);
+            if (!success) {
+                return absl::Status(absl::StatusCode::kInvalidArgument, inputText);
+            }
+#else
+            ov::genai::ChatHistory& chatHistory = executionContext->apiHandler->getChatHistory();
+            constexpr bool add_generation_prompt = true;
```
mzegla:
```suggestion
            constexpr bool addGenerationPrompt = true;
```
?

src/llm/apis/openai_completions.cpp
```diff
@@ -97,6 +99,372 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+        writer.String(status.c_str());
+        writer.String("call_id");
+        writer.String(toolCall.id.c_str());
+        writer.String("name");
+        writer.String(toolCall.name.c_str());
+        writer.String("arguments");
+        writer.String(toolCall.arguments.c_str());
+        writer.EndObject();
+    }
+    if (!fullOutputText.empty() || responsesState.toolCalls.empty()) {
```
mzegla: If we have assumption that if there were tool calls then there is no other content we should document it - for example as a comment here to be explicit.

src/llm/apis/openai_completions.cpp
mzegla: @michalkulakowski

src/llm/apis/openai_completions.cpp
mzegla: @michalkulakowski

src/llm/apis/openai_completions.cpp
```diff
@@ -444,38 +983,67 @@ absl::Status OpenAIChatCompletionsHandler::parseTools() {
...
+
+                auto parametersIt = obj.FindMember("parameters");
+                if (parametersIt != obj.MemberEnd()) {
+                    parametersValue = &parametersIt->value;
+                }
+            }
+
+            // If tool_choice is set to "auto", we keep all tools
+            // If tool_choice is set to a specific function name, we keep only that tool
+            if (tool_choice != "auto" && tool_choice != "required" && tool_choice != functionName) {
```
mzegla: `toolChoice`?

src/llm/apis/openai_completions.cpp
```diff
@@ -987,6 +1708,20 @@ ParsedOutput OpenAIChatCompletionsHandler::parseOutputIfNeeded(const std::vector
 
 std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(const std::vector<ov::genai::GenerationOutput>& generationOutputs) {
     OVMS_PROFILE_FUNCTION();
+    if (endpoint == Endpoint::RESPONSES) {
+        std::vector<ParsedOutput> parsedOutputs;
+        usage.completionTokens = 0;
+        ov::genai::GenerationFinishReason responsesFinishReason = ov::genai::GenerationFinishReason::STOP;
+        for (const ov::genai::GenerationOutput& generationOutput : generationOutputs) {
+            updateUsage(usage, generationOutput.generated_ids, request.echo);
```
mzegla: do we support `echo` in responses?
if not just paste false or `bool echo{false};` and then pass it, so there is no doubt

src/llm/apis/openai_completions.cpp
```diff
@@ -1172,6 +1911,34 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::VLMD
...
+            if (outputParser != nullptr) {
+                // Same workaround as in chat completions, line part
+                auto result = tokenizer.encode(text);
+                auto& input_ids = result.input_ids;
+                if (input_ids.get_shape().size() != 2)
+                    throw std::runtime_error("input_ids should have 2 dimensions");
+                if (input_ids.get_shape()[0] != 1)
+                    throw std::runtime_error("input_ids should have 1 batch size");
+                if (input_ids.get_element_type() != ov::element::i64)
+                    throw std::runtime_error("input_ids should have i64 element type");
```
mzegla: Does user see those errors or are they on our DEBUG logging?
User does not see input_ids, so it's strictly debugging information

src/llm/apis/openai_completions.cpp
```diff
@@ -1172,6 +1911,34 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::VLMD
...
+
+                int64_t* inputIdsData = reinterpret_cast<int64_t*>(input_ids.data());
+                std::vector<int64_t> generatedTokens(inputIdsData, inputIdsData + input_ids.get_shape()[1]);
+                parsedOutputs.push_back(parseOutputIfNeeded(generatedTokens));
```
mzegla: parseOutputIfNeeded strictly requires tokens? that's why we need to do this entire operation here?

src/llm/apis/openai_completions.cpp
```diff
@@ -1234,20 +2001,522 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::VLMD
...
+                    if (responsesState.reasoningInitialized && !responsesState.reasoningCompleted) {
+                        events.emplace_back(serializeReasoningSummaryTextDoneEvent(reasoningItemId));
+                        events.emplace_back(serializeReasoningSummaryPartDoneEvent(reasoningItemId));
+                        events.emplace_back(serializeReasoningOutputItemDoneEvent(reasoningItemId));
+                        responsesState.reasoningCompleted = true;
+                    }
+                    const auto& toolCallsArr = deltaObj["tool_calls"];
+                    for (rapidjson::SizeType i = 0; i < toolCallsArr.Size(); ++i) {
+                        const auto& tc = toolCallsArr[i];
+                        int tcIndex = tc.HasMember("index") ? tc["index"].GetInt() : 0;
```
mzegla: Is it possible that for two iterations in the loop we get to have `tcIndex = 0` here and overwrite?

src/llm/apis/openai_completions.cpp
```diff
@@ -97,6 +99,372 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    } else {
+        writer.Null();
+    }
+    writer.String("model");
+    writer.String(request.model.c_str());
+    writer.String("status");
+    writer.String(status.c_str());
+
+    writer.String("parallel_tool_calls");
+    writer.Bool(false);
```
dkalinowski: We have sup port for paralellel tool calls, why is it hardcoded to false here?

src/llm/apis/openai_completions.cpp
```diff
@@ -97,6 +99,372 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+    // TODO: previous_response_id not supported
+    writer.String("store");
+    writer.Bool(true);
+    // TODO: temperature/top_p are only included when explicitly provided in the request
```
dkalinowski: Is it 1:1 what standard says? Or standard says that we should always inform user which default was used?

src/llm/apis/openai_completions.cpp
```diff
@@ -97,6 +99,372 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+        writer.Uint64(static_cast<uint64_t>(request.maxTokens.value()));
+    }
+
+    writer.String("output");
+    writer.StartArray();
+    // Include reasoning output item if reasoning was produced during streaming
+    if (!responsesState.reasoningText.empty()) {
+        writer.StartObject();
+        writer.String("id");
+        writer.String("rs-0");
```
dkalinowski: what is rs-0? is it duplicated in multi-turn conversation? or is it for cases where model mixes reasoning and content multiple times within 1 unary response?

src/llm/apis/openai_completions.cpp
```diff
@@ -97,6 +99,372 @@ ov::genai::JsonContainer rapidJsonValueToJsonContainer(const rapidjson::Value& v
...
+        writer.EndObject();
+    }
+    writer.EndArray();
+
+    if (includeUsage) {
+        writer.String("usage");
+        writer.StartObject();
+        writer.String("input_tokens");
+        writer.Uint64(static_cast<uint64_t>(usage.promptTokens));
+        // TODO: input_tokens_details.cached_tokens not supported
```
dkalinowski: There is table for completions and chat completions API what we support and what we dont. Is there a chance that we could have the same for Responses API? This way, in future we will be able to talk about it clearny and see what is missing and estimate next tasks correctly. I'm just afraid people will assume something is done therefore where in fact its not

src/llm/apis/openai_completions.cpp
```diff
@@ -247,6 +615,169 @@ absl::Status OpenAIChatCompletionsHandler::ensureArgumentsInToolCalls(Value& mes
...
+                    return absl::InvalidArgumentError("Unsupported content type. Supported types are input_text and input_image.");
+                }
+            }
+
+            request.chatHistory.last()["content"] = contentText;
+        }
+    } else {
+        return absl::InvalidArgumentError("input is not a string or array");
+    }
+
```
dkalinowski: It's missing my bugfix: https://github.com/openvinotoolkit/model_server/pull/4079/changes

src/llm/apis/openai_completions.cpp
```diff
@@ -656,6 +1224,145 @@ absl::Status OpenAIChatCompletionsHandler::parseChatCompletionsPart(std::optiona
...
+            }
+            const std::string effort = effortIt->value.GetString();
+            if (effort != "low" && effort != "medium" && effort != "high") {
+                return absl::InvalidArgumentError("reasoning.effort must be one of: low, medium, high");
+            }
+            // Inject enable_thinking: true into chat_template_kwargs if not already explicitly set
+            auto kwargsIt = doc.FindMember("chat_template_kwargs");
+            if (kwargsIt == doc.MemberEnd()) {
+                rapidjson::Value kwargs(rapidjson::kObjectType);
+                kwargs.AddMember("enable_thinking", true, doc.GetAllocator());
```
dkalinowski: Barely any model really uses this to enable thinking. Qwen3-VL for example doesnt. Is it an issue? @michalkulakowski  @dtrawins

src/llm/apis/openai_completions.cpp
```diff
@@ -1332,8 +2599,33 @@ std::string OpenAIChatCompletionsHandler::serializeStreamingChunk(const std::str
...
+            events.emplace_back(std::move(initEvents));
+        }
+    }
+
+    events.emplace_back(serializeResponseFailedEventBody(responseId, createdAt, errorMessage, errorCode));
+
+    std::stringstream ss;
+    ss << events.front();
+    for (size_t i = 1; i < events.size(); ++i) {
+        ss << "\n\ndata: " << events[i];
```
dkalinowski: There's already util for that, why cant we reuse it here? wrapTextInServerSideEventMessage

src/llm/apis/openai_completions.cpp
michalkulakowski: This is similar logic we have for chat/completions part - we edit request json with removing images from every entry of input array. If given message contains both input_text and input_image, we change its value to just input text. If there is no input text, content value is set to "".

src/llm/apis/openai_completions.cpp
michalkulakowski: exactly

src/llm/apis/openai_completions.cpp
michalkulakowski: I assume these exception should never happen, but we have same check in chat completions part few lines below.

src/llm/apis/openai_completions.cpp
michalkulakowski: I assume that its not possible, since BaseOutputParser always include id in wrapDelta methods

src/llm/http_llm_calculator.cc
michalkulakowski: Yes, because here we emits responses api events that don't contain data generate by pipeline

src/llm/http_llm_calculator.cc
michalkulakowski: Yes, in responses api streaming, opposites to chat/completions, error handling is realized through special event rather than status code.

demos/continuous_batching/vlm/README.md
```diff
@@ -119,6 +119,65 @@ curl http://localhost:8000/v3/chat/completions  -H "Content-Type: application/js
...
+      {
+        "role": "user",
+        "content": [
+          {
+            "type": "input_text",
+            "text": "Describe what is on the picture."
+          },
+          {
+            "type": "input_image",
+            "image_url": "http://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/releases/2025/3/demos/common/static/images/zebra.jpeg"
```
mzegla: Can we use later tag? `releases/2025/3` for newly added example does not feel right

demos/continuous_batching/vlm/README.md
```diff
@@ -119,6 +119,65 @@ curl http://localhost:8000/v3/chat/completions  -H "Content-Type: application/js
...
+          "annotations": []
+        }
+      ]
+    }
+  ],
+  "usage": {
+    "input_tokens": 19,
+    "input_tokens_details": { "cached_tokens": 0 },
+    "output_tokens": 83,
+    "output_tokens_details": { "reasoning_tokens": 0 },
```
mzegla: What was the decision about returning 0 tokens? Maybe it could be `null` if we don't support it?

src/llm/apis/openai_api_handler.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+// Abstract base class for OpenAI API handlers.
+// Holds common state (request, doc, tokenizer, usage, output parser) and implements
+// shared parsing logic. Endpoint-specific parsing and serialization are pure virtual.
+class OpenAIApiHandler {
+protected:
+    Document& doc;
+    Endpoint endpoint;
+    CompletionUsageStatistics usage;
+    OpenAIChatCompletionsRequest request;
```
mzegla: Should this be in common handler? Seems like chat completions specific

src/llm/apis/openai_api_handler.hpp
```diff
@@ -0,0 +1,187 @@
...
+    std::chrono::time_point<std::chrono::system_clock> created;
+    ov::genai::Tokenizer tokenizer;
+    size_t processedTokens = 0;              // tracks overall number of tokens processed by the pipeline
+    bool toolCallsDetectedInStream = false;  // tracks whether tool calls were detected in any streaming chunk
```
mzegla: Is it possible on `completions` endpoint? Also not usable via `tokenize`.

src/llm/apis/openai_api_handler.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+    // Output parser is used to parse chat completions response to extract specific fields like tool calls and reasoning.
+    std::unique_ptr<OutputParser> outputParser = nullptr;
+
+    // Shared parsing helpers
+    absl::Status parseCommonPart(std::optional<uint32_t> maxTokensLimit, uint32_t bestOfLimit, std::optional<uint32_t> maxModelLength);
+    absl::Status ensureArgumentsInToolCalls(Value& messageObj, bool& jsonChanged);
+    ParsedOutput parseOutputIfNeeded(const std::vector<int64_t>& generatedIds);
```
mzegla, line 106: If this handler is supposed to cover `tokenize` as well. Then we leak functionality here. Maybe we should handle `tokenize` separately? Or introduce another abstraction layer?

src/llm/apis/openai_api_handler.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+public:
+    OpenAIApiHandler(Document& doc, Endpoint endpoint, std::chrono::time_point<std::chrono::system_clock> creationTime,
+        ov::genai::Tokenizer tokenizer, const std::string& toolParserName = "", const std::string& reasoningParserName = "") :
+        doc(doc),
+        endpoint(endpoint),
+        created(creationTime),
+        tokenizer(tokenizer) {
+        // TODO we should delay creating output parser until we have request with toolNameSchemaMap parsed
+        // now we pass it now but it has to be populated first before first use
```
mzegla:
```suggestion
        // we pass it now, but it has to be populated first before first use
```

src/llm/apis/openai_api_handler.hpp
```diff
@@ -0,0 +1,187 @@
...
+    virtual std::string serializeUnaryResponse(ov::genai::EncodedResults& results) = 0;
+    virtual std::string serializeUnaryResponse(ov::genai::VLMDecodedResults& results) = 0;
+    virtual std::string serializeStreamingChunk(const std::string& chunkResponse, ov::genai::GenerationFinishReason finishReason) = 0;
+    virtual std::string serializeStreamingUsageChunk() = 0;
+    virtual std::string serializeStreamingHandshakeChunk() = 0;
+
+    // Streaming lifecycle events - default no-ops for non-responses handlers
+    virtual std::string serializeStreamingCreatedEvent();
+    virtual std::string serializeStreamingInProgressEvent();
+    virtual std::string serializeFailedEvent(const std::string& errorMessage, ResponsesErrorCode errorCode = ResponsesErrorCode::SERVER_ERROR);
```
mzegla, line 175: It's `responses` specific right?

demos/continuous_batching/vlm/README.md
```diff
@@ -119,6 +119,65 @@ curl http://localhost:8000/v3/chat/completions  -H "Content-Type: application/js
...
 :::
 
+:::{dropdown} **Unary call with cURL using Responses API**
+**Note**: Using urls in request requires `--allowed_media_domains` parameter described [here](../../../docs/parameters.md)
+
+```bash
+curl http://localhost:8000/v3/responses \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "OpenGVLab/InternVL2-2B",
```
dkalinowski: it will conflict with https://github.com/openvinotoolkit/model_server/pull/4073

demos/continuous_batching/vlm/README.md
```diff
@@ -223,6 +282,48 @@ The picture features a zebra standing in a grassy area. The zebra is characteriz
...
+```console
+pip3 install openai
+```
+```python
+from openai import OpenAI
+import base64
+base_url='http://localhost:8080/v3'
+model_name = "OpenGVLab/InternVL2-2B"
+
+client = OpenAI(api_key='unused', base_url=base_url)
```
dkalinowski, line 295: it will also conflict with https://github.com/openvinotoolkit/model_server/pull/4073

docs/model_server_rest_api_responses.md
```diff
@@ -0,0 +1,282 @@
...
+          "type": "output_text",
+          "text": "OpenVINO is an open-source toolkit ...",
+          "annotations": []
+        }
+      ]
+    }
+  ],
+  "usage": {
+    "input_tokens": 5,
+    "output_tokens": 42,
```
dkalinowski, line 58: what about input_tokens_details and output_tokens_details? in other documents it is visible

demos/continuous_batching/accuracy/gorilla.patch
```diff
@@ -60,6 +79,79 @@ index 357584f..e45e12c 100644
...
+-        if base_url := os.getenv("OPENAI_BASE_URL"):
++        if base_url := os.getenv("OPENAI_BASE_URL","http://localhost:8000/v3"):
+             kwargs["base_url"] = base_url
+ 
+         if headers_env := os.getenv("OPENAI_DEFAULT_HEADERS"):
+@@ -99,25 +99,12 @@ class OpenAIResponsesHandler(BaseHandler):
+         kwargs = {
+             "input": message,
+             "model": self.model_name,
+-            "store": False,
```
dkalinowski: remove unnecessary patch changes (comment from chat)

docs/model_server_rest_api_responses.md
```diff
@@ -0,0 +1,282 @@
...
+|-----|----------|----------|---------|-----|
+| model | ✅ | ✅ | string (required) | Name of the model to use. From administrator point of view it is the name assigned to a MediaPipe graph configured to schedule generation using desired model. |
+| input | ✅ | ✅ | string or array (required) | The input to generate a response for. Accepts a plain string or an array of message items with `input_text` / `input_image` types. |
+| stream | ✅ | ✅ | bool (optional, default: `false`) | If set to true, partial message deltas will be sent to the client as [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format) as they become available, with the stream terminated by a `data: [DONE]` message. See [Streaming events](#streaming-events) section for details. |
+| stream_options | ✅ | ❌ | object (optional) | Options for streaming response. Only set this when you set stream: true. |
+| stream_options.include_usage | ✅ | ❌ | bool (optional) | If set, usage statistics are tracked. Note: for responses endpoint usage is always included in the `response.completed` event. |
```
dkalinowski: does it mean we include it in both places? why? do we really need stream_options.include_usage if usage is already supported in another output field?

docs/model_server_rest_api_responses.md
```diff
@@ -0,0 +1,282 @@
...
+| `response.content_part.added` | After `response.output_item.added` | A new content part (output_text) has been initialized. Contains `output_index`, `content_index`, `item_id` and the part object. |
+| `response.output_text.delta` | For each text chunk during generation | Contains the text `delta`, `output_index`, `content_index`, and `item_id`. May be emitted many times. |
+| `response.output_text.done` | When text generation is finalized | Contains the full accumulated `text`. |
+| `response.content_part.done` | After `response.output_text.done` | The content part is complete. Contains the final part object with full text. |
+| `response.output_item.done` | After `response.content_part.done` | The output item is complete. Contains the final item object with `status: "completed"`. |
+| `response.completed` | Last event before `[DONE]` | Contains the full response object with `status: "completed"` and `usage` statistics. |
+| `response.incomplete` | Last event before `[DONE]` (when truncated) | Emitted instead of `response.completed` when generation was stopped due to `max_output_tokens` limit. Contains the response object with `status: "incomplete"` and `incomplete_details`. |
```
dkalinowski, line 226: is finish reason part of "completed" and "incomplete" event? if so, is tool call reason supported?

src/llm/apis/openai_api_handler.cpp
```diff
@@ -0,0 +1,783 @@
...
+
+}  // namespace
+
+// Default no-op implementations for streaming lifecycle events
+std::string OpenAIApiHandler::serializeStreamingCreatedEvent() {
+    return "";
+}
+
+std::string OpenAIApiHandler::serializeStreamingInProgressEvent() {
+    return "";
```
dkalinowski, line 102: is it really needed? you can make this method pure virtual and force other classes to implement it

src/llm/apis/openai_api_handler.hpp
dkalinowski: I think the naming is the only issue here, right @michalkulakowski 
From my understanding this is Request object which holds everything required for inference. All apis need to deserialize into it. Just the name is misleading?

src/llm/apis/openai_api_handler.hpp
dkalinowski: yes, therefore shouldnt it be in openai_api_responses?

src/llm/apis/openai_api_handler.cpp
dkalinowski: or as I think of it now, it shouldnt really be part of base class, it is responses api specific

src/llm/apis/openai_completions.cpp
```diff
@@ -284,7 +205,11 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages(std::optional<std::stri
             }
             // Handle tool_calls array for function calling
             if (memberName == "tool_calls" && member->value.IsArray()) {
-                request.chatHistory.last()[memberName] = rapidJsonValueToJsonContainer(member->value);
+                // Use JSON string conversion for tool_calls array
+                rapidjson::StringBuffer tcBuffer;
+                rapidjson::Writer<rapidjson::StringBuffer> tcWriter(tcBuffer);
+                member->value.Accept(tcWriter);
+                request.chatHistory.last()[memberName] = ov::genai::JsonContainer::from_json_string(tcBuffer.GetString());
```
dkalinowski: from_json_string can fail/throw, can you handle it?

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1331 @@
...
+#include "../../logging.hpp"
+#include "../../profiler.hpp"
+#include "../../filesystem.hpp"
+#pragma warning(push)
+#pragma warning(disable : 6001 4324 6385 6386)
+#include "absl/strings/escaping.h"
+#include "absl/strings/str_cat.h"
+#pragma warning(pop)
+
+#include <regex>
```
dkalinowski: move up to other base includes

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1331 @@
...
+// limitations under the License.
+//*****************************************************************************
+
+#include "openai_responses.hpp"
+
+#include <cmath>
+#include <limits>
+#include <memory>
+#include <stdexcept>
+#include "src/port/rapidjson_stringbuffer.hpp"
```
dkalinowski: why is it here, in between base includes?

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1331 @@
...
+    }
+
+    if (events.empty()) {
+        return "";
+    }
+
+    std::stringstream ss;
+    ss << events.front();
+    for (size_t i = 1; i < events.size(); ++i) {
+        ss << "\n\ndata: " << events[i];
```
dkalinowski: there is unanswered comment regarding this line, which got lost after code was moved
https://github.com/openvinotoolkit/model_server/pull/4039#pullrequestreview-3992666145

src/llm/apis/openai_api_handler.hpp
dkalinowski: ok now I see why its made like this -> in calculator empty return of base class means we should not send the event at all. what do you think about it? @mzegla

src/llm/http_llm_calculator.cc
```diff
@@ -160,8 +177,26 @@ class HttpLLMCalculator : public CalculatorBase {
                     cc->Outputs().Tag(LOOPBACK_TAG_NAME).Add(new bool{true}, iterationBeginTimestamp);
             }
         } catch (ov::AssertFailure& e) {
+            if (executionContext->apiHandler && executionContext->apiHandler->isStream()) {
+                std::string failedEvent = executionContext->apiHandler->serializeFailedEvent(e.what());
+                if (!failedEvent.empty()) {
+                    executionContext->response = wrapTextInServerSideEventMessage(failedEvent);
+                    executionContext->response += wrapTextInServerSideEventMessage("[DONE]");
+                    cc->Outputs().Tag(OUTPUT_TAG_NAME).Add(new std::string{std::move(executionContext->response)}, iterationBeginTimestamp);
+                    return absl::OkStatus();
```
dkalinowski:
```suggestion
                    return preparePartialResponse(cc, failedEvent);
```

wrap into function and reuse in all new places you add?

src/llm/apis/openai_completions.cpp
dkalinowski: i still dont get why is it hardcoded. i saw it is changed from false to true in recent commits, but why? what is this output field doing? shouldnt it be dynamic?

src/llm/apis/openai_api_handler.hpp
michalkulakowski: thats right

src/llm/apis/openai_completions.cpp
michalkulakowski: we cant, because this part handles emergency situation when we would need to serialize few events at once

docs/model_server_rest_api_responses.md
michalkulakowski: You're right, that's redundant

demos/continuous_batching/vlm/README.md
dkalinowski: it now conflicts with main
this demo doesnt use InternVL2-2B any more @michalkulakowski 
please rebase/merge main to this branch and verify the documentation still holds

src/llm/apis/openai_completions.cpp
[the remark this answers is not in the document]
dkalinowski: impossible to reach 2kk tokens probably

src/llm/apis/openai_completions.cpp
dkalinowski: it is removed anyway

src/llm/apis/openai_completions.cpp
dkalinowski: it will not get added in this release, not sure when (from call)

src/llm/apis/openai_completions.cpp
dkalinowski: new responses api got error handling, old code still remains with "Unsupported content type", lets keep it like this

src/llm/apis/openai_completions.cpp
dkalinowski: @michalkulakowski on call we deduced it most likely should be removed

src/llm/apis/openai_completions.cpp
dkalinowski: from call: standard says we should always inform in response what has been requested/used during generation. this feature is skipped for now (please note in limitation docs?)

src/llm/apis/openai_completions.cpp
dkalinowski: from call: Responses API allows returning multiple reasonings, then the id is used to distinguish between those. for now, we only support one reasoning per response, therefore rs-0 is hardcoded

please verify @michalkulakowski

src/llm/apis/openai_completions.cpp
dkalinowski: probably not an issue

docs/model_server_rest_api_responses.md
dkalinowski: from call: it got removed, so it is correct

docs/model_server_rest_api_responses.md
dkalinowski: @michalkulakowski

src/llm/apis/openai_completions.cpp
michalkulakowski: confirmed

demos/continuous_batching/vlm/README.md
```diff
@@ -119,6 +119,65 @@ curl http://localhost:8000/v3/chat/completions  -H "Content-Type: application/js
...
 :::
 
+:::{dropdown} **Unary call with cURL using Responses API**
+**Note**: Using urls in request requires `--allowed_media_domains` parameter described [here](../../../docs/parameters.md)
+
+```bash
+curl http://localhost:8000/v3/responses \
+  -H "Content-Type: application/json" \
+  -d '{
+    "model": "Junrui2021/Qwen3-VL-8B-Instruct-int4",
```
mzegla, line 129: Can we use the model from openvino organization?

demos/continuous_batching/vlm/README.md
mzegla: @michalkulakowski ^^

src/llm/apis/openai_completions.hpp
```diff
@@ -15,119 +15,34 @@
...
-class OpenAIChatCompletionsHandler {
-    Document& doc;
-    Endpoint endpoint;
-    CompletionUsageStatistics usage;
-    OpenAIChatCompletionsRequest request;
-    std::chrono::time_point<std::chrono::system_clock> created;
-    ov::genai::Tokenizer tokenizer;
-    size_t processedTokens = 0;              // tracks overall number of tokens processed by the pipeline
+// Handler for OpenAI Completions (/v3/completions) and Chat Completions (/v3/chat/completions) APIs.
+class OpenAIChatCompletionsHandler : public OpenAIApiHandler {
```
mzegla, line 26: Should we have same handler for `/chat/completions` and `/completions`?

src/llm/servable.cpp
mzegla: @michalkulakowski

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1201 @@
...
+        }
+    } else {
+        return absl::InvalidArgumentError("input is not a string or array");
+    }
+
+    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Parsed responses input directly to chat history without mutating request JSON");
+    return absl::OkStatus();
+}
+
+absl::Status OpenAIResponsesHandler::parsePart(std::optional<uint32_t> maxTokensLimit, std::optional<std::string> allowedLocalMediaPath, std::optional<std::vector<std::string>> allowedMediaDomains) {
```
mzegla: `parsePart` seems a bit ambiguous - can we have a comment in the header about what this method actually consumes or expand the name to be more specific (part of what are we parsing?) like `parseInputChunk` or something like that

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1201 @@
...
+    // Reject stream_options — usage is always included in response.completed event
+    if (doc.FindMember("stream_options") != doc.MemberEnd()) {
+        return absl::InvalidArgumentError("stream_options is not supported in Responses API. Usage statistics are always included in the response.");
```
mzegla: Info about usage does not seem relevant at this point (theoretically stream options could have more options than just usage). I would put that information in the docs or somewhere in the headers.

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1201 @@
...
+    writer.String(request.model.c_str());
+    writer.String("status");
+    writer.String(status.c_str());
+
+    writer.String("parallel_tool_calls");
+    writer.Bool(true);
+    // TODO: previous_response_id not supported
+    writer.String("store");
+    writer.Bool(true);
+    // TODO: temperature are only included when explicitly provided in the request
```
mzegla: what's the todo here?

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1201 @@
...
+    writer.StartObject();
+    writer.String("format");
+    writer.StartObject();
+    writer.String("type");
+    writer.String("text");
+    writer.EndObject();
+    writer.EndObject();
+    serializeToolChoice(writer);
+    serializeTools(writer);
+    // TODO: top_p are only included when explicitly provided in the request
```
mzegla: what's the todo here?

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1201 @@
...
+    const std::optional<std::string>& incompleteReason, const std::optional<std::string>& errorMessage, ResponsesErrorCode errorCode) const {
+    serializeResponseEnvelope(writer, responseId, createdAt, status, incompleteReason, errorMessage, errorCode);
+
+    writer.String("output");
+    writer.StartArray();
+    // Include reasoning output item if reasoning was produced during streaming
+    if (!responsesState.reasoningText.empty()) {
+        writer.StartObject();
+        writer.String("id");
+        writer.String("rs-0");
```
mzegla: should id be hardcoded?

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1201 @@
...
+        writer.String("name");
+        writer.String(toolCall.name.c_str());
+        writer.String("arguments");
+        writer.String(toolCall.arguments.c_str());
+        writer.EndObject();
+    }
+    {
+        writer.StartObject();
+        writer.String("id");
+        writer.String("msg-0");
```
mzegla: should id be hardcoded?

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1201 @@
...
+}
+
+void OpenAIResponsesHandler::writeReasoningLocation(Writer<StringBuffer>& writer, const std::string& itemId) {
+    writer.String("output_index");
+    writer.Uint64(0);
+    writer.String("summary_index");
+    writer.Uint64(0);
+    writer.String("item_id");
+    writer.String(itemId.c_str());
+}
```
mzegla, line 705: those indexes should always be 0?

src/llm/apis/openai_responses.cpp
```diff
@@ -0,0 +1,1201 @@
...
+    writer.EndObject();
+    return buffer.GetString();
+}
+
+std::string OpenAIResponsesHandler::serializeStreamingChunk(const std::string& chunkResponse, ov::genai::GenerationFinishReason finishReason) {
+    OVMS_PROFILE_FUNCTION();
+    const auto createdAt = std::chrono::duration_cast<std::chrono::seconds>(created.time_since_epoch()).count();
+    const std::string responseId = "resp-" + std::to_string(createdAt);
+    const std::string outputItemId = "msg-0";
+    const std::string reasoningItemId = "rs-0";
```
mzegla: hardcoded?

dkalinowski (approved):

src/llm/apis/openai_responses.cpp
michalkulakowski: Yes, reasoning is always first item

mzegla (approved):

PR #4054, opened 2026-03-11
Introduce GenAI parser adapter

mzegla:
Added an adapter class to bridge OVMS output parsing machinery and GenAI parsers interface + Llama3 parser implementation via this adapter.

### DIFF
--- a/src/llm/io_processing/genai_parser_adapter.cpp
+++ b/src/llm/io_processing/genai_parser_adapter.cpp
@@ -1,0 +1,197 @@
+//*****************************************************************************
+// Copyright 2026 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include "genai_parser_adapter.hpp"
+
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace ovms {
+
+GenAIParserAdapter::GenAIParserAdapter(
+    ov::genai::Tokenizer& tokenizer,
+    std::shared_ptr<BaseGenAIParser> parser,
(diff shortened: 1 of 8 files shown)


src/test/llm/output_parsers/llama3_output_parser_test.cpp
```diff
@@ -37,7 +37,15 @@ static std::unique_ptr<ov::genai::Tokenizer> llama3Tokenizer;
 // Id of the <|python_tag|> which is a special token used to indicate the start of a tool calls
 constexpr int64_t botTokenId = 128010;
 
-class Llama3OutputParserTest : public ::testing::Test {
+// Parametrized fixture — runs every test with both "llama3" (native) and "llama3_genai" (adapter) parsers.
+//
+// Known behavioral differences between "llama3" and "llama3_genai":
```
dkalinowski, line 42: So we drop support for a case where model generates text and tools separated by <|python_tag|>?
```
abcd abcd <|python_tag|> tools...
```
Isnt it big issue?

src/llm/io_processing/genai_parser_adapter.cpp
```diff
@@ -0,0 +1,197 @@
...
+GenAIParserAdapter::GenAIParserAdapter(
+    ov::genai::Tokenizer& tokenizer,
+    std::shared_ptr<BaseGenAIParser> parser,
+    std::shared_ptr<BaseGenAIIncrementalParser> incrementalParser) :
+    BaseOutputParser(tokenizer),
+    parser(std::move(parser)),
+    incrementalParser(std::move(incrementalParser)) {}
+
+// static
+void GenAIParserAdapter::extractToolCalls(const ov::genai::JsonContainer& message, ToolCalls_t& toolCalls) {
```
dkalinowski, line 34:
```suggestion
ToolCalls_t GenAIParserAdapter::extractToolCalls(const ov::genai::JsonContainer& message) {
```
Why not just work with return value?

src/llm/io_processing/genai_parser_adapter.cpp
```diff
@@ -0,0 +1,197 @@
...
+        ToolCall toolCall;
+        if (tc.contains("id") && tc["id"].is_string()) {
+            toolCall.id = tc["id"].get_string();
+        }
+        if (tc.contains("function") && tc["function"].is_object()) {
+            const ov::genai::JsonContainer fn = tc["function"];
+            if (fn.contains("name") && fn["name"].is_string()) {
+                toolCall.name = fn["name"].get_string();
+            }
+            if (fn.contains("arguments") && fn["arguments"].is_string()) {
```
dkalinowski, line 51: plenty of checks do not return an error, but silently continue. is it good idea?

PR #4074, opened 2026-03-20
Fix for tool_calls finish reason when last streamed message is empty

dkalinowski:
### 🛠 Summary

CVS-182445

### DIFF
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -1274,9 +1274,14 @@ std::string OpenAIChatCompletionsHandler::serializeStreamingChunk(const std::str
         if (outputParser != nullptr) {
             std::optional<Document> delta = outputParser->parseChunk(chunkResponse, areToolsAvailable(), finishReason);
             if (!delta.has_value()) {
-                return "";
+                // If the generation is still ongoing, there is nothing to emit yet
+                if (finishReason == ov::genai::GenerationFinishReason::NONE) {
+                    return "";
+                }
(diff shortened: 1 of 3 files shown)


src/llm/apis/openai_completions.cpp
```diff
@@ -1292,7 +1295,7 @@ std::string OpenAIChatCompletionsHandler::serializeStreamingChunk(const std::str
         choice.AddMember("text", Value(chunkResponse.c_str(), allocator), allocator);
     }
 
-    auto serializedFinishReason = mapFinishReason(finishReason, hasToolCalls);
+    auto serializedFinishReason = mapFinishReason(finishReason, hasToolCalls || toolCallsDetectedInStream);
```
mzegla, line 1303: That looks like overhead. Won't they always be exactly the same?

src/llm/apis/openai_completions.cpp
```diff
@@ -1280,6 +1280,9 @@ std::string OpenAIChatCompletionsHandler::serializeStreamingChunk(const std::str
                 // Deep copy the "delta" member value into the choice object
                 choice.AddMember("delta", Value((*delta)["delta"], allocator), allocator);
                 hasToolCalls = hasToolCallsInStreamingDelta(*delta);
+                if (hasToolCalls) {
+                    toolCallsDetectedInStream = true;
```
mzegla, line 1289: Is this needed? Couldn't we rely on `hasToolCalls`?

mzegla (approved):

dtrawins (approved):

PR #4080, opened 2026-03-24
update OV runtime to master with 2026.2

dtrawins:
### 🧪 Checklist

- [ ] Unit tests added.
``


demos/common/export_models/requirements.txt
```diff
@@ -7,10 +7,10 @@ accelerate
 datasets
 diffusers  # for image generation
 einops
-nncf@git+https://github.com/openvinotoolkit/nncf.git
+nncf
```
dkalinowski, line 10: dont we need latest nncf possible on main branch? @dtrawins

dkalinowski (approved):

demos/common/export_models/requirements.txt
dtrawins: it will speedup out CI flow. Right now we mostly focus on models already converted to IR. Using latest public nncf should be sufficient for CI scenarios

mzegla (approved):

PR #4085, opened 2026-03-25
LFM2 parser

przepeck:
### 🛠 Summary

[CVS-182500](https://jira.devtools.intel.com/browse/CVS-182500)
Adding toolp parser for LFM2 model

### 🧪 Checklist

- [ ] Unit tests added.
``


src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+const std::string Lfm2ToolParser::TOOL_RESPONSE_START_TAG = "<|tool_response_start|>";
+const std::string Lfm2ToolParser::TOOL_RESPONSE_END_TAG = "<|tool_response_end|>";
+
+const std::string Lfm2ToolParser::TOOL_LIST_START_INDICATOR = "[";
+const std::string Lfm2ToolParser::TOOL_LIST_END_INDICATOR = "]";
+const std::string Lfm2ToolParser::TOOL_ARGS_START_INDICATOR = "(";
+const std::string Lfm2ToolParser::TOOL_ARGS_END_INDICATOR = ")";
+const std::string Lfm2ToolParser::TOOL_SEPARATOR_STR = ", ";
+
+void Lfm2ToolParser::writeArgumentOfAnyType(const rapidjson::Value& arg, rapidjson::Writer<rapidjson::StringBuffer>& writer) {
```
mzegla: If that's a standalone function could it be static? 
Also it sounds quite generic. Can we extract it to utils? Not sure if we already have such file in io_processing...

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+        writer.EndArray();
+    } else if (arg.IsObject()) {
+        writer.StartObject();
+        for (auto it = arg.MemberBegin(); it != arg.MemberEnd(); ++it) {
+            writer.Key(it->name.GetString());
+            writeArgumentOfAnyType(it->value, writer);
+        }
+        writer.EndObject();
+    } else {
+        writer.String("");
```
mzegla: Should we write anything if logging error?

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+        return arg;
+    }
+
+    std::string normalized = arg;
+    
+    const char first = normalized.front();
+    const char last = normalized.back();
+    if ((first == '{' && last == '}') || (first == '[' && last == ']')) {
+        std::replace(normalized.begin(), normalized.end(), '\'', '"');
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Argument contains curly braces or square brackets, replaced single quotes with double quotes for JSON parsing. Modified string: {}", normalized);
```
mzegla: Is it always safe? What if we get argument with list of strings:
```
["hello", "it's me"] 
```
?

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    if (normalized[0] == '"' && normalized.back() == '"' && normalized.find("\\") != std::string::npos) {
```
mzegla: Don't we already have some methods that implement escaping? Please check and if they are, we could extract them and put to some utils file.
Note that other special chars might require escaping too. Like `/n`, `/t` etc.

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+        }
+        normalized = escaped;
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Argument is a quoted string containing backslashes. Modified string: {}", normalized);
+
+    }
+
+    std::string lowerArg = normalized;
+    std::transform(lowerArg.begin(), lowerArg.end(), lowerArg.begin(), ::tolower);
+    
+    if (lowerArg == "true" || lowerArg == "false") {
```
mzegla: We don't want to capture situation where false/true has for example trailing whitespace? Or it's not expected in that format.

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+Lfm2ToolParser::Argument Lfm2ToolParser::parseSingleArgument(const std::string& argumentStr){
+    Lfm2ToolParser::Argument argument;
+
+    size_t equalPos = argumentStr.find('=');
+    if (equalPos != std::string::npos) {
+        argument.name = argumentStr.substr(0, equalPos);
+        argument.value = argumentStr.substr(equalPos + 1);
+
+        if (argument.value[0] == '\'' && argument.value.back() == '\'') {
```
mzegla: Is it always the case that model gives argument in string format as something like:
```arg='hello there'```
Can't it generate string as:
```arg=hello there```
?
Is that confirmed argument format for string?

File: src/llm/io_processing/lfm2/lfm2_tool_parser.cpp [hunk 7]
```diff
@@ -0,0 +1,444 @@
...
+        if (argument.value[0] == '\'' && argument.value.back() == '\'') {
+            argument.value[0] = '"';
+            argument.value[argument.value.size() - 1] = '"';
+            SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Argument value is enclosed in single quotes, replaced with double quotes. Modified value: {}", argument.value);
+        } 
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Parsed argument - name: {}, value: {}", argument.name, argument.value);
+    } else {
+        argument.name = argumentStr;
+        argument.value = "";
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Argument string: {} does not contain '=', setting name as entire string and value as empty", argumentStr);
```
mzegla, line 167: at this point we already have complete argument? we won't read next part of the output that could have value right?

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+    return argument;
+}
+
+std::vector<Lfm2ToolParser::Argument> Lfm2ToolParser::parseArguments(const std::string& argumentsStr) {
+    std::vector<std::string> args;
+    std::vector<Lfm2ToolParser::Argument> parsedArgs;
+
+    size_t argPos = 0;
+    while (argPos < argumentsStr.length()) {
+        size_t commaPos = findInStringRespectingSpecialChars(argumentsStr, TOOL_SEPARATOR_STR, argPos);
```
mzegla, line 178: comma is not very specific character - what if argument string has a comma like
arg='hello,there'
won't it break this logic?

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    for (const std::string& arg : args) {
+        parsedArgs.push_back(parseSingleArgument(arg));
+    }
+    return parsedArgs;
+}
+
+bool Lfm2ToolParser::parseInContentState() {
+    size_t pos = this->streamingContent.find(TOOL_CALL_START_TAG, this->streamingPosition);
```
mzegla: toolCallStartTagPos?

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+        parsedArgs.push_back(parseSingleArgument(arg));
+    }
+    return parsedArgs;
+}
+
+bool Lfm2ToolParser::parseInContentState() {
+    size_t pos = this->streamingContent.find(TOOL_CALL_START_TAG, this->streamingPosition);
+    size_t toolCallEndTagPos = this->streamingContent.find(TOOL_CALL_END_TAG, this->streamingPosition);
+    if (toolCallEndTagPos != std::string::npos && pos == std::string::npos) {
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Detected end of tool call at position: {}", toolCallEndTagPos);
```
mzegla, line 201: We are in parseInContentState and detect end of tool call explicitly? Why?

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+    size_t toolCallEndTagPos = this->streamingContent.find(TOOL_CALL_END_TAG, this->streamingPosition);
+    if (toolCallEndTagPos != std::string::npos && pos == std::string::npos) {
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Detected end of tool call at position: {}", toolCallEndTagPos);
+        this->streamingPosition = toolCallEndTagPos + TOOL_CALL_END_TAG.length();
+        return false;
+    }
+    if (pos != std::string::npos) {
+        this->streamingPosition = pos + TOOL_CALL_START_TAG.length() + TOOL_LIST_START_INDICATOR.length();
```
mzegla: We only checked for TOOL_CALL_START_TAG existence. Is moving position with length of TOOL_LIST_START_INDICATOR an assumption? Do we have guarantee that it will happen? 
Don't we risk invalidating position here?

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+    SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Parsed tool name: {}", toolName);
+    this->streamingPosition = pos + TOOL_ARGS_START_INDICATOR.length();
+    this->currentState = State::ToolCallParameters;
+    this->toolCallIndex++;
+    return true;
+}
+
+
+bool Lfm2ToolParser::parseToolCallParametersState() {
+    size_t pos = this->streamingContent.find(TOOL_ARGS_END_INDICATOR, this->streamingPosition);
```
mzegla: Again. ')' is not very specific - what if argument is `arg='y = x(x+1)'`
Won't we have problem there?

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+    argsWriter.EndObject();
+    this->toolCall.arguments = sb.GetString();
+    this->currentState = State::ToolCallEnded;
+    this->streamingPosition = pos + TOOL_ARGS_END_INDICATOR.length();
+
+    return true;
+}
+
+bool Lfm2ToolParser::parseInToolCallEndedState() {
```
mzegla, line 268: Once more, relying on characters like `, ( ) [ ]` seems like very fragile solution, but maybe it's okay in this method were we are after or in between tools and not processing arguments...

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+    rapidjson::Document doc(rapidjson::kObjectType);
+    rapidjson::Document argsDoc;
+    argsDoc.Parse(argsStr.c_str());
+    if (!argsDoc.HasParseError()) {
+        doc.AddMember("arguments", argsDoc, doc.GetAllocator());
+    } else {
+        doc.AddMember("arguments", rapidjson::Value(argsStr.c_str(), doc.GetAllocator()), doc.GetAllocator());
+    }
+    return BaseOutputParser::wrapDelta(doc, toolCallIndex);
+}
```
mzegla: Why can't we use base methods as they are?

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
```diff
@@ -0,0 +1,441 @@
...
+        if (this->currentState == State::Content) {
+            auto content = this->streamingContent.substr(this->streamingPosition);
+            this->streamingPosition += content.size();
+            return wrapDeltaContent(content);
+        }
+        if (this->currentState == State::AfterToolCall) {
+            this->currentState = State::Content;
+        }
+    }
+
```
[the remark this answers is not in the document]
mzegla: Consider this. We need to remain functional even if max_tokens is exceeded

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+    return false;
+}
+
+void Lfm2ToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int64_t>& generatedTokens) {
+    std::vector<std::string> tools;
+    std::vector<std::pair<size_t, size_t>> toolCallPositions;
+    size_t pos = 0;
+    int main_guard = 0;
+
+    while (pos != std::string::npos && main_guard < MAX_TOOL_CALLS) {
```
mzegla: Is this guard really needed?

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+        if (start == std::string::npos) {
+            break;
+        }
+        toolCallPosition.first = start;
+        start += TOOL_CALL_START_TAG.length();
+        size_t end = parsedOutput.content.find(TOOL_CALL_END_TAG, start);
+        if(end == std::string::npos) {
+            end = parsedOutput.content.rfind(TOOL_LIST_END_INDICATOR, end);
+            if (end == std::string::npos) {
+                SPDLOG_LOGGER_ERROR(llm_calculator_logger, "Malformed tool call in content, no tool end tag or tool list end tag found for tool call starting at position {}", start);
```
mzegla: I would put it to trace or debug at most

src/test/llm/output_parsers/lfm2_output_parser_test.cpp
```diff
@@ -0,0 +1,691 @@
...
+    EXPECT_EQ(parsedOutput.content, "This is a content part and next will be a tool call.\n\n");
+    EXPECT_EQ(parsedOutput.reasoning, "");
+
+    ASSERT_EQ(parsedOutput.toolCalls.size(), 1);
+    EXPECT_EQ(parsedOutput.toolCalls[0].name, "example_tool");
+    EXPECT_EQ(parsedOutput.toolCalls[0].arguments, "{\"arg1\":\"value1\",\"arg2\":42}");
+    EXPECT_EQ(parsedOutput.toolCalls[0].id.empty(), false);
+}
+
+TEST_F(LFM2OutputParserTest, HolisticStreaming) {
```
mzegla, line 413: For better coverage of escaping, could you also add test where we have long argument in string format? 
I mean like:
```
impl = "import package\nimport package2\n\ndef func(a, b):\n\td={"python": "dict"}\n\tl = ["list \"with escaped text\"", 123, []]
```

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+}
+
+std::optional<rapidjson::Document> Lfm2ToolParser::parseChunk(const std::string& chunk, ov::genai::GenerationFinishReason finishReason) {
+    if (chunk.empty()) {
+        return std::nullopt;
+    }
+
+    this->streamingContent += chunk;
+
+    if (parseNewContent()) {
```
mzegla, line 340: What happens if you receive chunk with both name and first argument? like:
`dummy(arg1=`
Do we support scenario with bigger chunks that would fit in multiple states?

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: in other case the key will be left like this  {key: bad_value, another_key: value} -> {"key", "another_key":"value"}

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: it's right, I will think how can I change it

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
```diff
@@ -0,0 +1,441 @@
...
+bool Lfm2ToolParser::parseSingleToolCall(const std::string& toolStr, ToolCall& toolCall) {
+    size_t argsPos = toolStr.find(TOOL_ARGS_START_INDICATOR);
+    if (argsPos != std::string::npos) {
+        std::string toolName = toolStr.substr(0, argsPos);                
+        SPDLOG_LOGGER_INFO(llm_calculator_logger, "Parsed tool name: {}", toolName);
+
+        int argsStrLen = toolStr.length() - argsPos - TOOL_ARGS_START_INDICATOR.length() - TOOL_ARGS_END_INDICATOR.length();
+        std::string argsStr = toolStr.substr(argsPos + TOOL_ARGS_START_INDICATOR.length(), argsStrLen);
+        SPDLOG_LOGGER_INFO(llm_calculator_logger, "Parsed args string: {}", argsStr);
+        std::vector<Lfm2ToolParser::Argument> arguments = parseArguments(argsStr);
```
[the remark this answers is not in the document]
przepeck: It won't happen, if tool lacks `)` this function wont be executed.

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
przepeck: TOOL_SEPARATOR_STR is ", " with space, what's more function `findInStringRespectingSpecialChars` is taking into account quotes and brackets, it shouldn't break the logic.
It has been checked in tests like `ParseToolCallWithStringArgumentsContainingBracesAndBrackets`.

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
przepeck: There is a logic in tests which I was following, that sometime it's possible that tool doesn't generate tool call end token. So I am stopping at `]` - which is the indictator of end of tool list then we are changing state to content but still this tag should be skipped

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: This parser will be used with models from 2.6B to 24B, if it would use 2.6B for multiturn it gets confused and generates tool calls and doesn't follow it's usual template. To avoid infinity loops I would live it

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
przepeck: I will add the test for that, I was working on chunks generated by the model

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
przepeck: yes, this function won't process part of argument, it will get complete one

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: we do, but wrapDelta needs to get rapidjson::value  with member "arguments" to be consistant with other parsers

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
przepeck: I have added test for that `StreamingWithBiggerChunks`, it's passing.

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: I've added to the test `StreamingWithContentBetweenToolCalls` this case. It's passing now.

src/test/llm/output_parsers/lfm2_output_parser_test.cpp
przepeck: Implemented in `ParseToolCallWithStringArgumentsContainingSpecialCharacters` and it's passing

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
przepeck: I am not sure, what we could rely on if not on the toolcall template. Do you have any ideas?

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: changed it to skip ` + TOOL_LIST_START_INDICATOR` and add it in `parseInToolCallState` not interrupting the flow

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: I've changed it and tested with tests `ParseToolCallWithStringArgumentsArrayWithStringsContainingQuotes` and `ParseToolCallWithStringArgumentsObjectWithStringsContainingQuotes`

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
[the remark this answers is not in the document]
przepeck: implemented

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
```diff
@@ -0,0 +1,441 @@
...
+        this->streamingPosition = toolCallEndTagPos + TOOL_CALL_END_TAG.length();
+        return false;
+    }
+    if (pos != std::string::npos) {
+        this->streamingPosition = pos + TOOL_CALL_START_TAG.length() + TOOL_LIST_START_INDICATOR.length();
+        this->currentState = State::ToolCallStarted;
+        SPDLOG_LOGGER_INFO(llm_calculator_logger, "Detected start of tool call at position: {}", pos);
+        return false;
+    }
+
```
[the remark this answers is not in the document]
przepeck: handled

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,474 @@
...
+
+    for (size_t i = 1; i < argumentStr.size() - 1; ++i) {
+        if (argumentStr[i] != '\'') {
+            continue;
+        }
+
+        bool isLastElement = (i == argumentStr.size() - 2);
+        bool isFollowedByComma = !isLastElement && argumentStr[i + 1] == ',';
+
+        if (i == 1 || quoteDepth == 0) {
```
mzegla: Is `i` check needed? We never decrement `i`, so it can only happen once in the very first iteration right?
In first iteration `quoteDepth` has to be 0 at this point.

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,474 @@
...
+std::string Lfm2ToolParser::parseArrayParameter(std::string argumentStr) {
+    int quoteDepth = 0;
+
+    for (size_t i = 1; i < argumentStr.size() - 1; ++i) {
+        if (argumentStr[i] != '\'') {
+            continue;
+        }
+
+        bool isLastElement = (i == argumentStr.size() - 2);
+        bool isFollowedByComma = !isLastElement && argumentStr[i + 1] == ',';
```
mzegla, line 48: What if array has strings? Don't we need tracking if we are inside string or not?
I mean, will the case `"argument": ["Hello", "family, friends, strangers"]` be handled properly?

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,474 @@
...
+        bool isFollowedByColon = !isLastElement && argumentStr[i + 1] == ':';
```
mzegla, line 72: Same question as for array. Do we have string tracking so we are aware when ':' is part of the internal string and when it's on the object level?

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,474 @@
...
+
+        if (i == 1 || quoteDepth == 0) {
```
mzegla: Same case as for array. Do we need `i==1` check?

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,444 @@
...
+        parsedArgs.push_back(parseSingleArgument(arg));
+    }
+    return parsedArgs;
+}
+
+bool Lfm2ToolParser::parseInContentState() {
+    size_t pos = this->streamingContent.find(TOOL_CALL_START_TAG, this->streamingPosition);
+    size_t toolCallEndTagPos = this->streamingContent.find(TOOL_CALL_END_TAG, this->streamingPosition);
+    if (toolCallEndTagPos != std::string::npos && pos == std::string::npos) {
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Detected end of tool call at position: {}", toolCallEndTagPos);
```
mzegla: So in that point we know that tools were generated? Won't it catch `end of tool` even if model did not output any tools but for example just and array of numbers?

przepeck: I think it won't. If it will generate array of numbers as a parameter it should be in state `ToolCallParameters` where it waits for `)`, than  in `ToolCallEnded` it's possible to catch `]` as a end of tool list and stop tool call generation.

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: Removed

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: removed

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: I've added quoteDepth check

src/llm/io_processing/lfm2/tool_parser.cpp
przepeck: Added quoteDepth check

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,481 @@
...
+        return true;
+    }
+    return false;
+}
+
+void Lfm2ToolParser::parse(ParsedOutput& parsedOutput, const std::vector<int64_t>& generatedTokens) {
+    std::vector<std::string> tools;
+    std::vector<std::pair<size_t, size_t>> toolCallPositions;
+    size_t pos = 0;
+    int main_guard = 0;
```
mzegla:
```suggestion
    int mainGuard = 0;
```

src/llm/io_processing/lfm2/tool_parser.cpp
```diff
@@ -0,0 +1,481 @@
...
+        auto itArgs = std::find(generatedTokens.begin() + start, generatedTokens.end(), eotTokenId);
+        if (itArgs != generatedTokens.end()) {
+            end = std::distance(generatedTokens.begin(), itArgs);
+        } else {
+            break;
+        }
+
+        std::string toolListStr = tokenizer.decode(std::vector<int64_t>(generatedTokens.begin() + start + 1, generatedTokens.begin() + end), ov::AnyMap{ov::genai::skip_special_tokens(false)});
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Parsed tool list string: {}", toolListStr);
+        int tool_guard = 0;
```
mzegla:
```suggestion
        int toolGuard = 0;
```

src/llm/io_processing/lfm2/tool_parser.cpp
mzegla: Actually I'm not convinced by this naming. It's a bit confusing like how is it different than tool guard if limits are set by `MAX_TOOL_CALLS` and  `MAX_TOOLS_PER_CALL` ? Could we make it more straightforward?

przepeck: LFM returns lists of tools <tool_call_start>[tool1(...), tool2(...)]<tool_call_end>
main loop considers multiple pairs of toolcall start-end tokens
inner loop is responsible for tool1(), tool2() etc. inside the list of tools

maybe we will use MAIN_LOOP_MAX_ITERS and TOOL_CALL_LOOP_MAX_ITERS to fit the convention for guards

mzegla (approved):

src/llm/io_processing/lfm2/tool_parser.cpp
mzegla: I would consider dropping the outer guard. If expected format is
```<tool_calls_start>[<tool_call>, <tool_call>...<tool_call>]<tool_call_end>```
then I think we could limit tool calls on that level only and treat situation like
```<tool_calls_start>[<tool_call>, <tool_call>...<tool_call>]<tool_call_end><tool_calls_start>[<tool_call>, <tool_call>...<tool_call>]<tool_call_end>``` as unexpected. 
@atobiszei what do you think?

**Review by dtrawins — APPROVED** [review]

File: src/llm/io_processing/lfm2/lfm2_tool_parser.cpp [hunk 29]
```diff
@@ -0,0 +1,481 @@
...
+    const char first = normalized.front();
+    const char last = normalized.back();
+    if (first == '{' && last == '}') {
+        normalized = parseObjectParameter(normalized);
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Argument contains is an object, replaced single quotes with double quotes for JSON parsing. Modified string: {}", normalized);
+    }
+
+    if (first == '[' && last == ']') {
+        normalized = parseArrayParameter(normalized);
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Argument is an array, normalized quotes for JSON parsing. Modified string: {}", normalized);
```
atobiszei, line 110: Minor but to your consideration -> since file is named tool_parser that what will appear in logs. We already have a split in naming convention, but with sth like lfm2_tool_parser.cpp its clear from start if user configured tool parser correctly just by looking at logs.

src/test/llm/output_parsers/lfm2_output_parser_test.cpp
```diff
@@ -0,0 +1,809 @@
...
+    for (const auto& [chunk, expectedDelta] : chunkToDeltaVec) {
+        // Second argument is false as we simulate the case where tools have not been provided in the request
+        std::optional<rapidjson::Document> doc = outputParserWithRegularToolParsing->parseChunk(chunk, false, ov::genai::GenerationFinishReason::NONE);
+        assertChunkEqual(doc, expectedDelta, chunk);
+    }
+}
+
+// Malformed tool calls
+
+TEST_F(LFM2OutputParserTest, ParseToolCallWithMissingParentheses) {
```
atobiszei, line 661: Do we have any test that would check if the tool call can return python string with all special characters?
I would recommend having in test sth like this:
https://github.com/openvinotoolkit/model_server/blob/main/src/test/llm/output_parsers/qwen3coder_output_parser_test.cpp#L748

przepeck: the test `ParseToolCallWithStringArgumentsContainingSpecialCharacters` is similar but not as complex as yours. 
I will add another test with similar input

src/llm/io_processing/lfm2/lfm2_tool_parser.cpp
przepeck: it seems like a good practice, I will change that and take it into account when we will be refactoring parsers in the future.

PR #4107, opened 2026-04-01
Hf pull process-wide cancellation API with terminate handling.

rasapala:
### 🛠 Summary
CVS-CVS-184757

Libgit2 Lfs-filter.c patch changes:
1. Process-wide cancellation API. Added an exported atomic flag with git_lfs_cancel_set(int) / git_lfs_cancel_get(void) (declared GIT_LFS_CANCEL_EXPORT, backed by git_atomic32), plus a host-supplied probe git_lfs_shutdown_requested(void) that the filter calls to learn whether the embedding application has requested shutdown. This gives OVMS a reliable way to abort in-flight LFS operations from outside libgit2.
2. Cancellation responsiveness across the LFS pipeline. Cancellation is now polled in the curl progress callbacks and around every long-running step: before/after each curl perform, between resume attempts, and inside the retry sleep loop (so a Ctrl+C during a back-off no longer waits the full interval).
3. Cancellation coverage extended to both LFS HTTP requests. Both the batch metadata request and the object download request are wired to the progress callback and to explicit abort guards, so cancellation aborts the operation regardless of which phase is in flight.
4. Error-focused logging with on-demand log file. Switched the patch to an error-focused logger; lfs_error.txt is now created/truncated only when an actual error is emitted, written under the repository workdir. Successful runs no longer leave a stale file behind, which OVMS relies on as a post-clone error signal.
5. Configurable, interruptible resume. Resume behaviour is preserved but made safer: the number of attempts and the interval between them are configurable via GIT_LFS_RESUME_ATTEMPTS and GIT_LFS_RESUME_INTERVAL_SECONDS (initialized once via INIT_ONCE), and cancel checks are repeated around resume retries and the fallback path.

Ovms Libgit2.cpp Changes:
1. End-to-end cancellation propagation. Server shutdown signaling is plumbed all the way into the clone and LFS download paths via libgit2::isCloneCancellationRequestedFromServer() and the exported git_lfs_shutdown_requested(void) C entry point, so Ctrl+C aborts both git_clone and ongoing LFS transfers (transfer/sideband/update-tips/checkout-notify callbacks all return -1 on cancel via the shared RETURN_IF_OVMS_CLONE_CANCELLED macro).
2. Reliable resume after interruption. Resume logic was redesigned to re-trigger the LFS smudge filter for partially downloaded files: it repairs the index entry from the HEAD tree (an aborted clone may leave the index entry missing), removes the stale worktree pointer file, and forces git_checkout_head with GIT_CHECKOUT_FORCE | GIT_CHECKOUT_RECREATE_MISSING for the affected paths. A complementary path restores tracked non-LFS files that were missing after a crash, then normalizes the index against HEAD.
.lfswip work-in-progress marker. A sibling marker file (<repo>.lfswip) is created before clone/resume and removed only on a clean finish. Its presence on a subsequent run indicates a prior interruption and switches the pull into resume mode (including a HEAD-tree scan for tracked files missing from the worktree).
lfs_error.txt handling. After clone or resume, OVMS reads any lfs_error.txt left by the patch, logs its contents at error level, removes the file, and surfaces a dedicated status to the caller.
3. Shared shutdown-state module (src/shutdown_state.{hpp,cpp}). New module exposes getShutdownRequestValue / setShutdownRequestValue, signal-safe setSignalShutdownRequested plus a main-thread processSignalShutdownRequest, decoupling pull-module cancellation checks from Server internals and satisfying Bazel dependency boundaries (the pull module no longer needs to depend on the server target).
4. New status codes & tests.
HF_GIT_CLONE_CANCELLED and HF_GIT_LIBGIT2_LFS_DOWNLOAD_FAILED added (alongside the existing HF_GIT_LIBGIT2_NOT_INITIALIZED).
5. New tests in pull_hf_model_test.cpp cover resume after a server-shutdown request (HfPull.ResumeShutdown), resume after SIGKILL/TerminateProcess (HfPull.ResumeTerminate), shutdown-driven cancellation (HfPull.CloneCancelledByShutdownRequest), no-op re-pull, partial-resume, user-removed and user-edited file scenarios on a cached repository (HfPullCache.RePull/Resume/UserRemoved/UserEdited), and the LFS-resume option capture tests.
6. New tests in libgit2_test.cpp document the .lfswip marker behavior (LibGit2LfsWipMarker.*).
Third-party libgit2 integration updates. Bumped the upstream libgit2 commit (libgit2_engine.bzl) and significantly expanded lfs.patch.

### 🧪 Checklist

- [ ] Unit tests added.
``


src/pull_module/libgit2.cpp
```diff
@@ -715,6 +724,14 @@ Status HfDownloader::downloadModel() {
         SPDLOG_DEBUG("Model repository status check passed after model download.");
     }
 
+    // Checking if git status is ok but we are left with LFS pointer files - no actual lfs download was triggered because of lfs errors
+    auto matches = libgit2::findLfsLikeFiles(this->downloadPath, true);
+    if (!matches.empty()) {
+        SPDLOG_ERROR("Model repository status check failed after git clone. Lfs pointer files found.");
+        SPDLOG_DEBUG("Review the log for lfs server connectivity errors.");
+        return StatusCode::HF_GIT_LIBGIT2_LFS_DOWNLOAD_FAILED;
+    }
```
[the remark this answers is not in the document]
rasapala: No time to do that and downloading does not work ATM.

atobiszei (approved):

third_party/libgit2/lfs.patch
```diff
@@ -481,16 +462,129 @@ index 000000000..630573c3d
...
++	 * without requiring symbol interposition (-rdynamic). */
++	if (git_lfs_cancel_requested)
++		return 1;
++#if defined(__GNUC__) || defined(__clang__)
++	/* Fallback: weak symbol provided by the host executable. */
++	if (git_lfs_shutdown_requested)
++		return git_lfs_shutdown_requested() != 0;
++#endif
++	return 0;
++}
```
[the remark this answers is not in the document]
rasapala: To be considered.

src/server.cpp
```diff
@@ -255,8 +277,9 @@ void Server::setShutdownRequest(int i) {
         std::this_thread::sleep_for(std::chrono::milliseconds(200));
     }
     if (counter) {
-        shutdown_request = i;
-        SPDLOG_TRACE("Ovms shutdown request set to: {}", shutdown_request);
+        setShutdownRequestValue(i);
+        setLfsCancelRequestedFromSignal(i);
+        SPDLOG_TRACE("Ovms shutdown request set to: {}", getShutdownRequestValue());
```
atobiszei:
```suggestion
        SPDLOG_TRACE("Ovms shutdown request set to: {}", i);
```
?

src/pull_module/libgit2.cpp
```diff
@@ -497,106 +838,304 @@ std::vector<fs::path> findLfsLikeFiles(const std::string& directory, bool recurs
...
+ * 
+ * @return StatusCode::OK on successful download or if model already exists locally,
+ *         StatusCode::HF_GIT_CLONE_CANCELLED if shutdown requested during operation,
+ *         StatusCode::HF_GIT_CLONE_FAILED or StatusCode::HF_GIT_LIBGIT2_LFS_DOWNLOAD_FAILED on errors.
+ * @note Connects to internet: performs git clone from Hugging Face endpoint (requires network).
+ * @note Works on local filesystem and downloads git repository to downloadPath.
+ * @note Implements the full git domain: clone, LFS, checkout, authentication, and repository validation.
+ */
 Status HfDownloader::downloadModel() {
     if (FileSystem::isPathEscaped(this->downloadPath)) {
```
atobiszei: This function should be split up. There's happening too much at different levels of abstractions.
1) CheckIfDownloadPossible/Required? a lot of logic checking if its legal to download, then if libgit is initialized, missing files etc. Some of those can be tested without pulling models at the unit test level I think? Probably up to L1175
2) Check/prepareLFSWipMarkers? up to L1194
3) ResumeDownloadForWIPFiles?
4) PrepareOptions for clone?
Etc

src/pull_module/libgit2.cpp
```diff
@@ -683,6 +1248,13 @@ Status HfDownloader::downloadModel() {
     git_repository* cloned_repo = NULL;
```
atobiszei: PrepareCloneOptions?

src/pull_module/libgit2.cpp
```diff
@@ -51,15 +55,197 @@
...
 std::atomic<int> g_activeLibgit2Guards{0};
+
+/**
+ * Builds the side-marker path used to track interrupted LFS operations.
+ * The marker lives next to the repository directory as <repo>.lfswip.
+ * 
+ * @param repositoryPath Absolute path to the model repository directory.
+ * @return Filesystem path to the .lfswip marker associated with repositoryPath.
+ * @note Pure path computation; does not access filesystem or network.
+ */
```
atobiszei, line 83: Do we really need so much comments for those short functions? I would skip them, naming convention is good so they are self-explanatory I think, and it just bloats the code.

src/pull_module/libgit2.cpp
```diff
@@ -51,15 +55,197 @@
...
+/**
+ * Callback invoked during git clone object transfer to monitor transfer progress.
+ * Checks for cancellation requests from the server shutdown mechanism.
+ * 
+ * @param stats Pointer to git_indexer_progress structure with transfer statistics (unused).
+ * @param payload User-supplied payload pointer (unused).
+ * @return -1 if cancellation requested (aborts transfer), 0 otherwise.
+ * @note Connects to internet during clone operations; works on the clone/fetch domain.
+ */
+int cloneTransferProgressCb(const git_indexer_progress* stats, void* payload) {
```
atobiszei: Cb->Callback?

src/pull_module/libgit2.cpp
```diff
@@ -51,15 +55,197 @@
...
+ * @param len Length of the sideband message (unused).
+ * @param payload User-supplied payload pointer (unused).
+ * @return -1 if cancellation requested (aborts operation), 0 otherwise.
+ * @note Connects to internet; works on the clone/fetch domain.
+ */
+int cloneSidebandProgressCb(const char* str, int len, void* payload) {
+    (void)str;
+    (void)len;
+    (void)payload;
+    if (libgit2::isCloneCancellationRequestedFromServer()) {
```
atobiszei: This is used in several places. Use macro to shorten the code?
#define STOP_IF_OVMS_CANCELLED or sth like that?

src/pull_module/libgit2.cpp
```diff
@@ -51,15 +55,197 @@
...
+        setLfsCancelRequested(1);
+        return -1;
+    }
+    return 0;
+}
+
+/**
+ * Callback for sideband progress messages during git operations (clone, fetch).
```
atobiszei: Here comments again - we don't even use those methods for anythin usefull, so this comments only bloats the code.

src/pull_module/libgit2.cpp
```diff
@@ -497,230 +876,734 @@ std::vector<fs::path> findLfsLikeFiles(const std::string& directory, bool recurs
...
+/**
+ * Checks for LFS download errors recorded by libgit2 patch in the repository root.
+ * Reads and logs error messages from lfs_error.txt, then removes the file for cleanup.
+ * 
+ * @param repositoryRootPath Absolute path to the git repository root directory.
+ * @return true if error file exists (even if unreadable); false if no error file found.
+ * @note Works on specific git repository location; reads and deletes lfs_error.txt.
+ * @note Part of the git LFS domain: detects LFS download failures recorded by patched libgit2.
+ */
+bool hasLfsErrorFileAndLogContent(const std::string& repositoryRootPath) {
```
atobiszei:
```suggestion
bool ifHasLfsErrorFileLogContentAndRemove(const std::string& repositoryRootPath) {
```
or sth like that?
It not only checks but removes as well.

src/pull_module/libgit2.cpp
```diff
@@ -497,230 +876,734 @@ std::vector<fs::path> findLfsLikeFiles(const std::string& directory, bool recurs
...
+/**
+ * Reconciles index with HEAD tree after crash recovery operations.
+ * This removes staged index artifacts before strict cleanliness validation.
+ * 
+ * @param repo Pointer to the git repository object.
+ * @return StatusCode::OK on success, cancellation/error status otherwise.
+ * @note Works on local git index; does not modify tracked file contents in worktree.
+ */
+Status normalizeIndexToHead(git_repository* repo) {
+    auto runGitCall = [&](int rc, const char* callName) -> Status {
```
atobiszei, line 1130: We cannot declare this runGitCall  because we need to pass everything from local function in context as reference correct?

src/pull_module/libgit2.cpp
```diff
@@ -497,230 +876,734 @@ std::vector<fs::path> findLfsLikeFiles(const std::string& directory, bool recurs
...
+        return s;
+
+    return StatusCode::OK;
+}
 
-        for (const auto& p : matches) {
-            std::cout << " Resuming " << p.string() << "...\n";
-            std::string path = p.string();
-            resumeLfsDownloadForFile(repoGuard.get(), path.c_str());
+Status resumeExistingRepository(git_repository* repo,
```
atobiszei, line 1163: Nice - clear logic

src/pull_module/libgit2.cpp
```diff
@@ -497,230 +876,734 @@ std::vector<fs::path> findLfsLikeFiles(const std::string& directory, bool recurs
...
+        if (!restoreMissingStatus.ok()) {
+            return restoreMissingStatus;
         }
-        return StatusCode::OK;
+    } else if (!candidates.missingNonLfsMatches.empty()) {
+        SPDLOG_DEBUG("Skipping non-LFS missing tracked file restoration because .lfswip marker was not found.");
     }
 
-    auto status = IModelDownloader::checkIfOverwriteAndRemove();
+    auto normalizeIndexStatus = normalizeIndexToHead(repo);
```
atobiszei: normalizeIndex->restoreGitIndex?

src/pull_module/libgit2.cpp
atobiszei: Since we don't do a lot in those callbacks maybe name those (
checkIfCancelCloneTransferProgressCallback or sth like that?

src/test/pull_hf_model_test.cpp
```diff
@@ -337,16 +417,360 @@ TEST_F(HfDownloaderPullHfModel, Resume) {
     ASSERT_EQ(expectedDigest, resumedDigest);
 }
 
-TEST_F(HfDownloaderPullHfModel, PositiveDownloadAndStart) {
+// ResumeAfterShutdownRequestAndRerun
+TEST_F(HfPull, ResumeShutdown) {
```
atobiszei, line 423:
```suggestion
TEST_F(HfPull,ResumeAfterShutdownRequestAndRerun) {
```

src/test/pull_hf_model_test.cpp
```diff
@@ -337,16 +417,360 @@ TEST_F(HfDownloaderPullHfModel, Resume) {
...
+    EXPECT_EQ(secondRunCode, EXIT_SUCCESS);
+    EXPECT_FALSE(std::filesystem::exists(removedFilePath));
+    EXPECT_FALSE(std::filesystem::exists(removedFilePath2));
+
+    std::string preservedDigestAfter = sha256File(preservedFilePath, ec);
+    ASSERT_EQ(ec, std::errc());
+    EXPECT_EQ(preservedDigestBefore, preservedDigestAfter);
+}
+
+// PullAfterUserEditedTrackedFileDoesNotOverwriteIt
```
atobiszei, line 541: Put that in test name

src/pull_module/libgit2.cpp
rasapala: This is just convenience. The real reason not to blindly extract it is not “because references are impossible”. It is that the three lambdas are similar but not identical:

src/test/pull_hf_model_test.cpp
rasapala: We cannot have long names in unit tests for lib git 2 clone because on windows we have a path max that can be exceeded.

src/test/pull_hf_model_test.cpp
rasapala: Cant

atobiszei (approved):

dtrawins (approved):

PR #4108, opened 2026-04-01
detect resource constraints to set default parameters

dtrawins:
### 🛠 Summary

Default number of REST workers based only on the number of CPUs might exceed the number of open files limit. It can happen for example in Kubernetes with Xeon CPU with almost 400 CPU cores and ulimit 1024 files.
This change make the default number of REST workers safe in such conditions. It will include open files limit as criteria to allow initialization and open connections. The ration between rest worker could assumes open files for drongon, mediapipe graphs and client connections.
When the number of REST workers is set explicitly, and required open files number exceeds the limit, a clear error will be raised and OVMS will exit without attempt to load the model.

This PR is also detecting CPU constraints inside docker containers. There is checked cpu affinity and quota. They are also used to set default number of threads for modules.

### 🧪 Checklist

- [ ] Unit tests added.
``


src/config.cpp
```diff
@@ -306,6 +334,12 @@ bool Config::validate() {
         std::cerr << "rest_workers is set but rest_port is not set. rest_port is required to start rest servers" << std::endl;
         return false;
     }
+#ifdef __linux__
+    if (restWorkers() > (getMaxOpenFilesLimit() - RESERVED_OPEN_FILES) / 6) {
```
dkalinowski: why is it 6 here, but 7 in getDefaultRestWorkers()?

src/config.cpp
```diff
@@ -37,18 +37,46 @@
...
+    return std::numeric_limits<uint64_t>::max();
+}
+
+const uint64_t RESERVED_OPEN_FILES = 15;  // we need to reserve some file descriptors for other operations, so we don't want to use all of them for drogon workers
+uint64_t getDefaultRestWorkers() {
+    const uint64_t maxOpenFiles = getMaxOpenFilesLimit();
+    if (maxOpenFiles <= RESERVED_OPEN_FILES) {
+        return static_cast<uint64_t>(2);  // minimum functional number
+    }
+    return std::min(static_cast<uint64_t>(AVAILABLE_CORES), (maxOpenFiles - RESERVED_OPEN_FILES) / 7);  // 5x rest_workers to initialize ovms and 2x rest_workers for new connections
```
dkalinowski: can you add log that number of workers has been decreased due to limit of open files?

src/test/ovmsconfig_test.cpp
```diff
@@ -202,6 +206,38 @@ TEST_F(OvmsConfigDeathTest, restWorkersTooLarge) {
...
+    struct rlimit limit;
+    ASSERT_EQ(getrlimit(RLIMIT_NOFILE, &limit), 0);
+    struct rlimit newLimit = {std::min(static_cast<rlim_t>(cpu_cores * 5), limit.rlim_max), limit.rlim_max};
+    std::cout << "Setting open files limit to " << newLimit.rlim_cur << " to test that default rest_workers count is reduced based on open files limit" << std::endl;
+    ASSERT_EQ(setrlimit(RLIMIT_NOFILE, &newLimit), 0);
+
+    char* n_argv[] = {"ovms", "--config_path", "/path1", "--rest_port", "8080", "--port", "8081"};
+    int arg_count = 7;
+    ovms::Config::instance().parse(arg_count, n_argv);
+    EXPECT_TRUE(ovms::Config::instance().validate());
```
dkalinowski, line 237: this test doesnt check if number of workers has been changed

src/config.cpp
dkalinowski: is default rest workers also max possible rest workers?

src/test/ovmsconfig_test.cpp
dkalinowski: never mind, it would require calculation how many cores the machine has

src/config.cpp
```diff
@@ -37,18 +37,46 @@
...
+// on linux, restrict also based on the max allowed number of open files
+#ifdef __linux__
+
+static uint64_t getMaxOpenFilesLimit() {
+    struct rlimit limit;
+    if (getrlimit(RLIMIT_NOFILE, &limit) == 0) {
+        return limit.rlim_cur;
+    }
+    return std::numeric_limits<uint64_t>::max();
+}
```
atobiszei: This looks like utility that could exist in systeminfo.hpp (already included) in L38.

src/test/ovmsconfig_test.cpp
```diff
@@ -202,6 +206,38 @@ TEST_F(OvmsConfigDeathTest, restWorkersTooLarge) {
...
+    // limit allowed number of open files to 1024 to make sure that rest_workers count is too large.
+    struct rlimit limit;
+    ASSERT_EQ(getrlimit(RLIMIT_NOFILE, &limit), 0);
+    struct rlimit newLimit = {1024, limit.rlim_max};
+    std::cout << "Setting open files limit to " << newLimit.rlim_cur << " to test that rest_workers count is too large for the limit based on number of cpu cores alone" << std::endl;
+    ASSERT_EQ(setrlimit(RLIMIT_NOFILE, &newLimit), 0);
+    char* n_argv[] = {"ovms", "--config_path", "/path1", "--rest_port", "8080", "--port", "8081", "--rest_workers", "1000"};
+    int arg_count = 9;
+    EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(OVMS_EX_USAGE), "rest_workers count cannot be larger than 168 due to open files limit. Current open files limit: 1024");
+    ASSERT_EQ(setrlimit(RLIMIT_NOFILE, &limit), 0);
```
atobiszei: If test exits earlier we won't restore original rlimit value. Need either unique_ptr with custom deleter to handle that or helper struct. The same is true for previous test.

src/test/systeminfo_test.cpp
```diff
@@ -31,6 +32,6 @@ using ovms::StatusCode;
 
 TEST(SystemInfo, getCoreCount) {
     uint16_t cpuCount = getCoreCount();
-    EXPECT_GE(cpuCount, 1);
-    EXPECT_LE(cpuCount, std::thread::hardware_concurrency());
+    EXPECT_GE(cpuCount, 2);
+    EXPECT_LE(cpuCount, std::max(static_cast<uint16_t>(2), static_cast<uint16_t>(std::thread::hardware_concurrency())));
```
atobiszei: Why this part of change is needed?

src/config.cpp
```diff
@@ -368,7 +394,25 @@ const std::string Config::restBindAddress() const { return this->serverSettings.
...
+    const uint16_t detectedCores = getCoreCount();
+    SPDLOG_DEBUG("Detected cores: {} (may be constrained by Docker limits), max open files: {}, calculated default rest workers: {}", detectedCores, maxOpenFiles, defaultRestWorkers);
+    if (isRunningInDocker()) {
+        SPDLOG_DEBUG("Docker CPU quota detected: {}, CPU affinity count: {}", getDockerCpuQuota(), getCpuAffinityCount());
+    }
+#else
+    const uint16_t detectedCores = getCoreCount();
+    SPDLOG_DEBUG("Detected cores: {}, calculated default rest workers: {}", detectedCores, defaultRestWorkers);
+#endif
+    return static_cast<uint32_t>(defaultRestWorkers);
```
atobiszei: I don't get it - we have a method getDefaultRestWorkers, already handling part of the core count limitation logic (open files & core count). Why not put all the logic there and here we just stick to old implemetnation with minro adjustement:
```
return this->serverSettings.restWorkers.value_or(getDefaultRestWorkers());
```
Much clearer IMO

src/systeminfo.cpp
```diff
@@ -15,16 +15,139 @@
...
+        const uint16_t affinityCount = getCpuAffinityCount();
+        const uint16_t quotaCount = getDockerCpuQuota();
+        if (quotaCount > 0) {
+            detectedCoreCount = std::min(affinityCount, quotaCount);
+        } else {
+            detectedCoreCount = affinityCount;
+        }
+    }
+#endif
+    return std::max(static_cast<uint16_t>(2), detectedCoreCount);
```
atobiszei: Why return 2 here? This is utility for detecting core count. If it returns 1 (when it fails?)  we should change that in the explicit place that this >1 core is required as it is required by that specific place.

src/config.cpp
dtrawins: this is for debugging purposes. logger is initialized in the config parser but not in systeminfo. For that reason the methods are called again.

src/systeminfo.cpp
dtrawins: this is to avoid situation when only 1 thread is allocated in other modules. Even with fraction of cpu core, there should be at lease 2 threads for the service to function properly.

src/test/systeminfo_test.cpp
```diff
@@ -32,5 +32,4 @@ using ovms::StatusCode;
 TEST(SystemInfo, getCoreCount) {
     uint16_t cpuCount = getCoreCount();
     EXPECT_GE(cpuCount, 1);
```
dtrawins, line 34:
```suggestion
    EXPECT_GE(cpuCount, 1);
    EXPECT_LE(cpuCount, std::thread::hardware_concurrency());
```

dkalinowski (approved):

src/test/systeminfo_test.cpp
atobiszei: @dtrawins

src/systeminfo.cpp
atobiszei: In my opinion its still not the place to change that. This function should return core count. If some place (eg rest workers) require at least >=2 change it there with explanation why.

atobiszei (approved):

PR #4140, opened 2026-04-15
Speech to text streaming support

michalkulakowski:
### 🛠 Summary

CVS-181778
CVS-185048

- Streaming support for speech2text endpoint
- Disconnection support for both speech2text and text2speech
- bugfixing

### 🧪 Checklist

- [x] Unit tests added.
``

### DIFF
--- a/demos/common/export_models/export_model.py
+++ b/demos/common/export_models/export_model.py
@@ -134,14 +134,30 @@ node {
   name: "S2tExecutor"
   input_side_packet: "STT_NODE_RESOURCES:s2t_servable"
   calculator: "S2tCalculator"
+  input_stream: "LOOPBACK:loopback"
   input_stream: "HTTP_REQUEST_PAYLOAD:input"
+  output_stream: "LOOPBACK:loopback"
   output_stream: "HTTP_RESPONSE_PAYLOAD:output"
+  input_stream_info: {
+    tag_index: 'LOOPBACK:0',
+    back_edge: true
+  }
   node_options: {
     [type.googleapis.com / mediapipe.S2tCalculatorOptions]: {
       models_path: "{{model_path}}",
       plugin_config: '{ "NUM_STREAMS": "{{num_streams|default(1, true)}}" }',
       target_device: "{{target_device|default("CPU", true)}}",
       enable_word_timestamps: {% if not enable_word_timestamps %}false{% else %}true{% endif%},
+    }
+  }
+  input_stream_handler {
+    input_stream_handler: "SyncSetInputStreamHandler",
+    options {
+      [mediapipe.SyncSetInputStreamHandlerOptions.ext] {
+        sync_set {
+          tag_index: "LOOPBACK:0"
+        }
+      }
     }
   }
 }
--- a/src/audio/speech_to_text/s2t_calculator.cc
+++ b/src/audio/speech_to_text/s2t_calculator.cc
@@ -209,11 +209,15 @@ private:
             serializeTimestamps(writer, result, config);
         }
         if (endpoint == Endpoint::TRANSLATIONS) {
+            float temperature = 1.0f;
+            auto tempStatus = S2tStreamingHandler::parseTemperature(payload, temperature);
+            if (tempStatus != absl::OkStatus())
+                return tempStatus;
             std::unique_lock lock(pipe->sttPipelineMutex);
             auto disconnectStatus = checkClientDisconnected(payload, cc->NodeName(), "before translation");
             if (!disconnectStatus.ok())
                 return disconnectStatus;
-            std::string generatedText = pipe->sttPipeline->generate(rawSpeech, ov::genai::task("translate"), ov::genai::streamer(disconnectCallback));
+            std::string generatedText = pipe->sttPipeline->generate(rawSpeech, ov::genai::task("translate"), ov::genai::temperature(temperature), ov::genai::streamer(disconnectCallback));
             lock.unlock();
             disconnectStatus = checkClientDisconnected(payload, cc->NodeName(), "after translation");
             if (!disconnectStatus.ok())
--- a/src/audio/speech_to_text/s2t_calculator.cc
+++ b/src/audio/speech_to_text/s2t_calculator.cc
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //*****************************************************************************
+#include <chrono>
 #include <string>
 
 #pragma warning(push)
@@ -114,7 +115,23 @@ public:
 
         // --- LOOPBACK iteration: drain streaming queue ---
         if (!loopbackEmpty) {
-            return streamingHandler_.processIteration(cc, LOOPBACK_TAG_NAME, OUTPUT_TAG_NAME);
+            std::string ssePayload;
+            bool shouldContinueLoopback = false;
+            bool hasOutput = false;
+            auto status = streamingHandler_.processIteration(ssePayload, shouldContinueLoopback, hasOutput);
+            if (status != absl::OkStatus()) {
+                return status;
+            }
+
+            if (hasOutput) {
+                cc->Outputs().Tag(OUTPUT_TAG_NAME).Add(new std::string{std::move(ssePayload)}, cc->InputTimestamp());
+            }
+            if (shouldContinueLoopback) {
+                auto now = std::chrono::system_clock::now();
+                const auto nextTimestamp = ::mediapipe::Timestamp(std::chrono::duration_cast<std::chrono::microseconds>(now.time_since_epoch()).count());
+                cc->Outputs().Tag(LOOPBACK_TAG_NAME).Add(new bool{true}, nextTimestamp);
+            }
+            return absl::OkStatus();
         }
 
         // --- First iteration: new request ---
@@ -164,8 +181,18 @@ public:
             }
 
             if (requestStreaming) {
-                return streamingHandler_.start(cc, pipe, payload, std::move(rawSpeech),
-                    endpoint == Endpoint::TRANSCRIPTIONS, LOOPBACK_TAG_NAME, OUTPUT_TAG_NAME);
+                ov::genai::WhisperGenerationConfig config = pipe->sttPipeline->get_generation_config();
+                auto status = ovms::SttServable::updateTranscriptionConfig(config, pipe, payload);
+                if (status != absl::OkStatus()) {
+                    return status;
+                }
+                config.validate();
+                status = streamingHandler_.start(pipe, payload, std::move(rawSpeech), config);
+                if (status != absl::OkStatus()) {
+                    return status;
+                }
+                cc->Outputs().Tag(LOOPBACK_TAG_NAME).Add(new bool{true}, cc->InputTimestamp());
+                return absl::OkStatus();
             }
 
             return processUnaryRequest(cc, pipe, endpoint, payload, rawSpeech);
@@ -193,9 +220,10 @@ private:
         writer.String("text");
         if (endpoint == Endpoint::TRANSCRIPTIONS) {
             ov::genai::WhisperGenerationConfig config = pipe->sttPipeline->get_generation_config();
-            auto status = ovms::SttServable::applyTranscriptionConfig(config, pipe, payload);
+            auto status = ovms::SttServable::updateTranscriptionConfig(config, pipe, payload);
             if (status != absl::OkStatus())
                 return status;
+            config.validate();
 
             std::unique_lock lock(pipe->sttPipelineMutex);
             auto disconnectStatus = checkClientDisconnected(payload, cc->NodeName(), "before transcription");
@@ -211,19 +239,22 @@ private:
             serializeTimestamps(writer, result, config);
         }
         if (endpoint == Endpoint::TRANSLATIONS) {
-            float temperature = pipe->sttPipeline->get_generation_config().temperature;
-            auto tempStatus = ovms::SttServable::parseTemperature(payload, temperature);
-            if (tempStatus != absl::OkStatus())
-                return tempStatus;
+            ov::genai::WhisperGenerationConfig config = pipe->sttPipeline->get_generation_config();
+            config.task = "translate";
+            auto status = ovms::SttServable::parseTemperature(payload, config);
+            if (status != absl::OkStatus())
+                return status;
+            config.validate();
             std::unique_lock lock(pipe->sttPipelineMutex);
             auto disconnectStatus = checkClientDisconnected(payload, cc->NodeName(), "before translation");
             if (!disconnectStatus.ok())
                 return disconnectStatus;
-            std::string generatedText = pipe->sttPipeline->generate(rawSpeech, ov::genai::task("translate"), ov::genai::temperature(temperature), ov::genai::streamer(disconnectCallback));
+            const ov::genai::WhisperDecodedResults result = pipe->sttPipeline->generate(rawSpeech, config, disconnectCallback);
             lock.unlock();
             disconnectStatus = checkClientDisconnected(payload, cc->NodeName(), "after translation");
             if (!disconnectStatus.ok())
                 return disconnectStatus;
+            const std::string generatedText = result;
             writer.String(generatedText.c_str());
         }
         writer.EndObject();
--- a/src/audio/speech_to_text/s2t_calculator.cc
+++ b/src/audio/speech_to_text/s2t_calculator.cc
@@ -166,7 +166,6 @@ public:
             }
 
             return processUnaryRequest(cc, pipe, endpoint, payload, rawSpeech);
-
         } catch (ov::AssertFailure& e) {
             return absl::InvalidArgumentError(e.what());
         } catch (...) {
--- a/src/audio/speech_to_text/s2t_calculator.cc
+++ b/src/audio/speech_to_text/s2t_calculator.cc
@@ -208,7 +208,7 @@ private:
             serializeTimestamps(writer, result, config);
         }
         if (endpoint == Endpoint::TRANSLATIONS) {
-            float temperature = 1.0f;
+            float temperature = pipe->sttPipeline->get_generation_config().temperature;
             auto tempStatus = ovms::SttServable::parseTemperature(payload, temperature);
             if (tempStatus != absl::OkStatus())
                 return tempStatus;
--- a/src/audio/speech_to_text/s2t_calculator.cc
+++ b/src/audio/speech_to_text/s2t_calculator.cc
@@ -13,10 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //*****************************************************************************
-#include <condition_variable>
-#include <fstream>
-#include <future>
-#include <queue>
+#include <string>
 
 #pragma warning(push)
 #pragma warning(disable : 4005 4309 6001 6385 6386 6326 6011 6246 4456 6246)
@@ -31,24 +28,16 @@
 #include "src/client_connection.hpp"
 #include "src/http_payload.hpp"
 #include "src/logging.hpp"
-#include "src/stringutils.hpp"
-#include <mutex>
-#include <thread>
 
 #pragma warning(push)
 #pragma warning(disable : 6001 4324 6385 6386)
-#include "absl/strings/escaping.h"
 #include "absl/strings/str_cat.h"
 #pragma warning(pop)
 
 #include "src/port/rapidjson_writer.hpp"
 #include "src/port/rapidjson_stringbuffer.hpp"
 #include "s2t_servable.hpp"
-
-#ifdef _WIN32
-#include <fcntl.h>
-#include <io.h>
-#endif
+#include "s2t_streaming_handler.hpp"
 
 using namespace ovms;
 
@@ -62,7 +51,7 @@ enum Endpoint {
     UNSUPPORTED
 };
 
-Endpoint getEndpoint(const std::string& url) {
+static Endpoint getEndpoint(const std::string& url) {
     if (absl::StartsWith(url, "/v3/audio/transcriptions")) {
         return Endpoint::TRANSCRIPTIONS;
     }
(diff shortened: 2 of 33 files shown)


src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -91,12 +100,28 @@ class S2tCalculator : public CalculatorBase {
...
+
+        // --- LOOPBACK iteration: drain streaming queue ---
+        if (hasLoopback_ && !cc->Inputs().Tag(LOOPBACK_TAG_NAME).IsEmpty() && streamingHandler_.isActive()) {
+            return streamingHandler_.processIteration(cc, LOOPBACK_TAG_NAME, OUTPUT_TAG_NAME);
+        }
+
+        // --- First iteration: new request ---
+        if (cc->Inputs().Tag(INPUT_TAG_NAME).IsEmpty()) {
+            return absl::OkStatus();
+        }
```
mzegla, line 140: when does it happen exactly?

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -132,111 +159,119 @@ class S2tCalculator : public CalculatorBase {
...
+            lock.unlock();
+            disconnectStatus = checkClientDisconnected(payload, cc->NodeName(), "after transcription");
+            if (!disconnectStatus.ok())
+                return disconnectStatus;
+            const std::string generatedText = result;
+            writer.String(generatedText.c_str());
+            serializeTimestamps(writer, result, config);
+        }
+        if (endpoint == Endpoint::TRANSLATIONS) {
+            float temperature = 1.0f;
```
mzegla: why is this hardcoded? why this specific value?

src/audio/speech_to_text/s2t_streaming_handler.cpp
```diff
@@ -0,0 +1,77 @@
...
+        if (!temp.has_value()) {
+            temp = ovms::stou32(temperatureStr);
+            if (!temp.has_value())
+                return absl::InvalidArgumentError("Invalid temperature type.");
+        }
+        if (temp.value() < 0.0f || temp.value() > 2.0f)
+            return absl::InvalidArgumentError("Temperature out of range(0.0, 2.0)");
+        temperature = temp.value();
+    } else {
+        temperature = 1.0f;
```
mzegla: why 1.0?

src/audio/speech_to_text/s2t_streaming_handler.hpp
```diff
@@ -0,0 +1,191 @@
...
+// Encapsulates all streaming state and logic for the S2tCalculator.
+// Manages the background generation thread, the text queue, SSE
+// serialization and LOOPBACK signaling.
+class S2tStreamingHandler {
+public:
+    bool isActive() const { return isStreaming_; }
+
+    static absl::Status parseTemperature(const ovms::HttpPayload& payload, float& temperature);
+
+    static std::string serializeTextChunk(const std::string& text) {
```
mzegla: Isn't it duplicated code? I suppose we already have quite a lot of logic for serialization.

src/audio/speech_to_text/s2t_streaming_handler.hpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    absl::Status processIteration(CalculatorContext* cc,
+        const std::string& loopbackTag,
+        const std::string& outputTag) {
+        std::string chunk;
+        bool hasData = streamingQueue_->waitAndPop(chunk);
+
+        if (hasData) {
+            accumulatedText_ += chunk;
```
mzegla: why do we need to accumulate text?

src/audio/speech_to_text/s2t_streaming_handler.hpp
```diff
@@ -0,0 +1,191 @@
...
+#include "s2t_servable.hpp"
+
+namespace mediapipe {
+
+// Encapsulates all streaming state and logic for the S2tCalculator.
+// Manages the background generation thread, the text queue, SSE
+// serialization and LOOPBACK signaling.
+class S2tStreamingHandler {
+public:
+    bool isActive() const { return isStreaming_; }
```
mzegla: do we need that? couldn't we rely on loopback being on/off?

src/audio/speech_to_text/streaming_text_queue.hpp
```diff
@@ -0,0 +1,69 @@
...
+        cv_.wait(lock, [this] { return !queue_.empty() || done_; });
+        if (!queue_.empty()) {
+            out = std::move(queue_.front());
+            queue_.pop();
+            return true;
+        }
+        return false;  // done and empty
+    }
+
+    bool isDone() const {
```
mzegla: The logic of this function does not keep setter/getter pair idea. I mean `setDone` does not guarantee that following `isDone()` will be called. 
Either `setDone` should 
a) clear the queue
b) assert queue is empty and fail if called on non-empty queue

or `isDone` should only check flag set by the setter
or `isDone` should be ranamed to not be considered simply a getter for `done_`

src/drogon_http_server.cpp
```diff
@@ -88,12 +88,13 @@ Status DrogonHttpServer::startAcceptingRequests() {
     drogon::app().disableSigtermHandling();
 
     drogon::app().setDefaultHandler([this](const drogon::HttpRequestPtr& req, std::function<void(const drogon::HttpResponsePtr&)>&& drogonResponseInitializeCallback) {
-        bool isTextGeneration = req->path().find("/completions") != std::string::npos ||
-                                req->path().find("/responses") != std::string::npos;
+        bool isLongRunning = req->path().find("/completions") != std::string::npos ||
+                             req->path().find("/responses") != std::string::npos ||
+                             req->path().find("/audio/") != std::string::npos;
```
mzegla: why `audio` ends with slash and others don't?

src/drogon_http_server.cpp
```diff
@@ -88,12 +88,13 @@ Status DrogonHttpServer::startAcceptingRequests() {
...
 
         // Here we need to schedule the request to the separate thread pool
         // in order to use disconnection callback of drogon.
-        if (isTextGeneration) {
+        if (isLongRunning) {
```
mzegla: Why is this name changed?

mzegla (commented):
We support streaming for stateful LLMs. This new machinery for audio looks different as I recall. Seems quite complex here. Couldn't we reuse the flow of legacy/stateful pipelines?

src/audio/speech_to_text/s2t_streaming_handler.hpp
```diff
@@ -0,0 +1,191 @@
...
+        rapidjson::StringBuffer buffer;
+        rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+        writer.StartObject();
+        writer.String("text");
+        writer.String(text.c_str());
+        writer.EndObject();
+        return buffer.GetString();
+    }
+
+    absl::Status start(CalculatorContext* cc,
```
dkalinowski: why are implementations in header file?

src/audio/speech_to_text/s2t_streaming_handler.hpp
```diff
@@ -0,0 +1,191 @@
...
+        const ovms::HttpPayload& payload,
+        std::vector<float> rawSpeech,
+        bool isTranscription,
+        const std::string& loopbackTag,
+        const std::string& outputTag) {
+        isStreaming_ = true;
+        accumulatedText_.clear();
+        streamingQueue_ = std::make_shared<ovms::StreamingTextQueue>();
+
+        auto queue = streamingQueue_;
```
dkalinowski: why is here a copy of queue shared reference? cant you use streamingQueue_ directly?

src/audio/speech_to_text/s2t_streaming_handler.hpp
```diff
@@ -0,0 +1,191 @@
...
+        };
+
+        if (isTranscription) {
+            ov::genai::WhisperGenerationConfig config = pipe->sttPipeline->get_generation_config();
+            auto status = applyTranscriptionConfig(config, pipe, payload);
+            if (status != absl::OkStatus()) {
+                isStreaming_ = false;
+                return status;
+            }
+            generateFuture_ = std::async(std::launch::async,
```
dkalinowski: Do we really want to create new thread each time? (use of std::async)
Shouldn't we follow ContinuousBatchingPipeline convention and implement add_request, read and read_all in GenAI?

dkalinowski: Maybe we could use threadpool with size 1 to ensure its all sequential and also no new threads are created? Or reuse LLMPipeline generate with streaming logic

src/audio/speech_to_text/s2t_calculator.cc
michalkulakowski: I assume it could happen with invalid graph.pbtxt

src/audio/speech_to_text/s2t_calculator.cc
michalkulakowski: This is default value from genai generation config. Could be probbably changed to comment to make it clear what default value is.

src/drogon_http_server.cpp
michalkulakowski: Because its not related to text generation only :)
But this name is not perfect, i agree

src/mediapipe_internal/mediapipegraphexecutor.hpp
```diff
@@ -303,6 +303,10 @@ class MediapipeGraphExecutor {
 #endif
                 inputSidePackets[LLM_SESSION_SIDE_PACKET_TAG] = mediapipe::MakePacket<GenAiServableMap>(this->sidePacketMaps.genAiServableMap).At(STARTING_TIMESTAMP);
                 inputSidePackets[EMBEDDINGS_SESSION_SIDE_PACKET_TAG] = mediapipe::MakePacket<EmbeddingsServableMap>(this->sidePacketMaps.embeddingsServableMap).At(STARTING_TIMESTAMP);
+                inputSidePackets[RERANK_SESSION_SIDE_PACKET_TAG] = mediapipe::MakePacket<RerankServableMap>(this->sidePacketMaps.rerankServableMap).At(STARTING_TIMESTAMP);
+                inputSidePackets[STT_SESSION_SIDE_PACKET_TAG] = mediapipe::MakePacket<SttServableMap>(this->sidePacketMaps.sttServableMap).At(STARTING_TIMESTAMP);
+                inputSidePackets[TTS_SESSION_SIDE_PACKET_TAG] = mediapipe::MakePacket<TtsServableMap>(this->sidePacketMaps.ttsServableMap).At(STARTING_TIMESTAMP);
+                inputSidePackets[IMAGE_GEN_SESSION_SIDE_PACKET_TAG] = mediapipe::MakePacket<ImageGenerationPipelinesMap>(this->sidePacketMaps.imageGenPipelinesMap).At(STARTING_TIMESTAMP);
```
mzegla: Why is it all added now? Were they missing all along and it's kind of a bugfix?

src/llm/language_model/legacy/legacy_executor.cpp
```diff
@@ -52,50 +46,16 @@ void LegacyExecutor::processRequest() {
...
 
-void LegacyExecutor::waitForRequests(std::atomic<bool>* receivedEndSignal) {
-    std::unique_lock<std::mutex> lock(queueMutex);
-    cv.wait(lock, [this, receivedEndSignal] { return (requests.size() > 0 || *receivedEndSignal); });
-}
-
 void LegacyExecutor::addRequest(std::shared_ptr<LegacyServableExecutionContext> request) {
     std::unique_lock<std::mutex> lock(queueMutex);
     requests.push(request);
     cv.notify_one();
```
mzegla: Couldn;t we use Executor base implementation?

src/audio/speech_to_text/streaming_text_queue.hpp
mzegla: ^^

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -91,12 +100,28 @@ class S2tCalculator : public CalculatorBase {
...
     absl::Status Process(CalculatorContext* cc) final {
         SPDLOG_LOGGER_DEBUG(s2t_calculator_logger, "SpeechToTextCalculator  [Node: {}] Process start", cc->NodeName());
 
+        bool loopbackEmpty = !hasLoopback_ || cc->Inputs().Tag(LOOPBACK_TAG_NAME).IsEmpty();
+        if (cc->Inputs().Tag(INPUT_TAG_NAME).IsEmpty() && loopbackEmpty) {
+            return absl::OkStatus();
+        }
+
+        // --- LOOPBACK iteration: drain streaming queue ---
+        if (hasLoopback_ && !cc->Inputs().Tag(LOOPBACK_TAG_NAME).IsEmpty()) {
```
mzegla: Isn't it the same a `loopbackEmpty`?

src/audio/speech_to_text/s2t_streaming_handler.hpp
```diff
@@ -0,0 +1,77 @@
...
+        const std::string& outputTag);
+
+    absl::Status processIteration(CalculatorContext* cc,
+        const std::string& loopbackTag,
+        const std::string& outputTag);
+
+    // Reused by both streaming start (for config) and unary path.
+    // Kept here to avoid duplicating the parsing logic.
+    static absl::Status applyTranscriptionConfig(ov::genai::WhisperGenerationConfig& config,
+        const std::shared_ptr<ovms::SttServable>& pipe, const ovms::HttpPayload& payload);
```
mzegla: Used for both unary/streaming - should be on servable level then I guess.
Keeping it in streaming handler is confusing also due to the fact it does not have direct connection to streaming (it's input preparation phase)

src/drogon_http_server.cpp
mzegla: What is breaking out of text generation here? Speech generation?

src/executor_base.hpp
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+}
+
+template <typename RequestT>
+struct Executor {
+    std::condition_variable cv;
+    std::queue<RequestT> requests;
+    std::mutex queueMutex;
+
```
mzegla, line 38: Why can't we have add request add this level?

src/audio/speech_to_text/s2t_executor.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include "openvino/genai/whisper_pipeline.hpp"
+#include "src/executor_base.hpp"
+#include "src/logging.hpp"
+
+namespace ovms {
+
+using SttStreamingJob = std::packaged_task<ov::genai::WhisperDecodedResults()>;
+
+struct SttExecutor : public Executor<SttStreamingJob> {
```
dkalinowski: can you move this to cpp file?

src/mediapipe_internal/mediapipegraphexecutor.hpp
michalkulakowski: added by mistake. only  inputSidePackets[STT_SESSION_SIDE_PACKET_TAG] is relevant

src/audio/speech_to_text/s2t_executor.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <mutex>
+#include <utility>
+
+#include "openvino/genai/whisper_pipeline.hpp"
+#include "src/executor_base.hpp"
+#include "src/logging.hpp"
+
+namespace ovms {
+
+using SttStreamingJob = std::packaged_task<ov::genai::WhisperDecodedResults()>;
```
mzegla: looks a bit whisper-dedicated
do we assume there will be no more / no other speech to text pipelines?

src/audio/speech_to_text/s2t_servable.cpp
```diff
@@ -0,0 +1,116 @@
...
+        parsedModelsPath = fsModelsPath;
+    }
+    ov::AnyMap config;
+    auto status = JsonParser::parsePluginConfig(nodeOptions.plugin_config(), config);
+    if (!status.ok()) {
+        SPDLOG_ERROR("Error during llm node plugin_config option parsing to JSON: {}", nodeOptions.plugin_config());
+        throw std::runtime_error("Error during plugin_config option parsing");
+    }
+    enableWordTimestamps = nodeOptions.enable_word_timestamps();
+    if (enableWordTimestamps && nodeOptions.target_device() == "NPU") {
```
mzegla, line 53: doesn't it have to be static always when on NPU?

src/audio/speech_to_text/s2t_servable.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+void SttServable::addRequest(std::shared_ptr<SttServableExecutionContext> executionContext) {
+    if (!streamingExecutor) {
+        throw std::runtime_error("Cannot schedule STT streaming job - executor not initialized");
+    }
+    streamingExecutor->addRequest(std::move(executionContext));
+}
+
+absl::Status SttServable::parseTemperature(const HttpPayload& payload, float& temperature) {
```
mzegla: we shouldn't have such logic here
wrong temperature should be captured in `config.validate()`
maybe we are just missing this call?

src/audio/speech_to_text/s2t_streaming_handler.cpp
```diff
@@ -0,0 +1,195 @@
...
+            isStreaming_ = false;
+            return tempStatus;
+        }
+        auto executionContext = std::make_shared<ovms::SttServableExecutionContext>(ovms::SttServable::StreamingJob(
+            [pipe, rawSpeech = std::move(rawSpeech), streamerCallback, queue = streamingQueue_, temperature]() mutable -> ov::genai::WhisperDecodedResults {
+                try {
+                    std::unique_lock lock(pipe->sttPipelineMutex);
+                    auto result = pipe->sttPipeline->generate(rawSpeech,
+                        ov::genai::task("translate"),
+                        ov::genai::temperature(temperature),
```
mzegla: wouldn't generate accept GenerationConfig? we could wrap those two values in it then

src/audio/speech_to_text/s2t_streaming_handler.cpp
```diff
@@ -0,0 +1,195 @@
...
+            }));
+        try {
+            pipe->addRequest(executionContext);
+            executionContext_ = std::move(executionContext);
+        } catch (const std::exception& e) {
+            isStreaming_ = false;
+            return absl::InternalError(e.what());
+        }
+    } else {
+        float temperature = pipe->sttPipeline->get_generation_config().temperature;
```
mzegla: pipeline level generation config? is it overwritten with every next request?

src/audio/speech_to_text/s2t_streaming_handler.cpp
```diff
@@ -0,0 +1,195 @@
...
+            return ov::genai::StreamingStatus::CANCEL;
+        }
+        if (!text.empty()) {
+            queue->push(std::move(text));
+        }
+        return ov::genai::StreamingStatus::RUNNING;
+    };
+
+    if (isTranscription) {
+        ov::genai::WhisperGenerationConfig config = pipe->sttPipeline->get_generation_config();
```
mzegla: pipeline level generation config? is it overwritten with every next request?

src/executor_base.hpp
```diff
@@ -0,0 +1,120 @@
...
+#include <spdlog/spdlog.h>
+
+namespace ovms {
+
+template <typename ExecutorT>
+void runExecutorLoop(ExecutorT* executor, std::atomic<bool>* receivedEndSignal, const std::shared_ptr<spdlog::logger>& logger) {
+    while (!(*receivedEndSignal)) {
+        SPDLOG_LOGGER_INFO(logger, "All requests: {};", executor->requestsQueueSize());
+        if (executor->hasRequests()) {
+            executor->processRequest();
```
mzegla: Hmm, I can't find the signature for base `processRequest`

src/audio/speech_to_text/s2t_calculator.cc
```diff
@@ -132,111 +162,119 @@ class S2tCalculator : public CalculatorBase {
...
-            }
-            if (endpoint == Endpoint::TRANSLATIONS) {
-                std::unique_lock lock(pipe->sttPipelineMutex);
-                std::string generatedText = pipe->sttPipeline->generate(rawSpeech, ov::genai::task("translate"));
-                lock.unlock();
-                writer.String(generatedText.c_str());
+
+            if (requestStreaming) {
+                return streamingHandler_.start(cc, pipe, payload, std::move(rawSpeech),
+                    endpoint == Endpoint::TRANSCRIPTIONS, LOOPBACK_TAG_NAME, OUTPUT_TAG_NAME);
```
mzegla: We are passing wide scope constants as arguments. If they are needed on streaming handler level, maybe they should just be visible there. If making it visible there does not sound right, then maybe streaming handler should not interact with mediapipe directly

src/audio/speech_to_text/s2t_servable.cpp
```diff
@@ -0,0 +1,116 @@
...
+        throw std::runtime_error("Error during plugin_config option parsing");
+    }
+    enableWordTimestamps = nodeOptions.enable_word_timestamps();
+    if (enableWordTimestamps && nodeOptions.target_device() == "NPU") {
+        config["STATIC_PIPELINE"] = true;
+    }
+    config["word_timestamps"] = enableWordTimestamps;
+    sttPipeline = std::make_shared<ov::genai::WhisperPipeline>(parsedModelsPath.string(), nodeOptions.target_device(), config);
+
+    streamingExecutor = std::make_unique<SttExecutorWrapper>();
```
mzegla: Wouldn't it make sense if streaming executor had reference to sttPipeline as its member?

demos/common/export_models/export_model.py
```diff
@@ -144,6 +150,16 @@ def add_common_arguments(parser):
       enable_word_timestamps: {% if not enable_word_timestamps %}false{% else %}true{% endif%},
     }
   }
+  input_stream_handler {
```
dkalinowski, line 153: @michalkulakowski will old graphs without a loop work? for users that exported model in the past but will use ovms 2026.2 (with a loop input)

src/audio/speech_to_text/streaming_text_queue.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include <utility>
+
+namespace ovms {
+
+// Thread-safe queue for streaming partial results from a background
+// generation thread to the MediaPipe LOOPBACK loop.  Used by S2tCalculator
+// to bridge ov::genai streamer callbacks and the calculator's Process() cycle.
+class StreamingTextQueue {
+public:
+    void push(std::string text) {
```
dkalinowski: Can you move it to cpp?

src/test/audio/speech2text_test.cpp
```diff
@@ -75,6 +75,239 @@ std::unique_ptr<std::thread> Speech2TextHttpTest::t;
...
+        ovms::StatusCode::PARTIAL_END);
+
+    ASSERT_FALSE(receivedChunks.empty());
+
+    const std::string dataPrefix = "data: ";
+    // Validate each delta chunk (all but the last)
+    for (size_t i = 0; i + 1 < receivedChunks.size(); ++i) {
+        const std::string& chunk = receivedChunks[i];
+        ASSERT_GE(chunk.size(), dataPrefix.size());
+        EXPECT_EQ(chunk.substr(0, dataPrefix.size()), dataPrefix) << "Chunk " << i << " missing SSE prefix";
```
dkalinowski: I think we dont need to proceed if  its already broken?

src/test/audio/speech2text_test.cpp
```diff
@@ -75,6 +75,239 @@ std::unique_ptr<std::thread> Speech2TextHttpTest::t;
...
+
+    const std::string dataPrefix = "data: ";
+    auto parseChunkJson = [&dataPrefix](const std::string& chunk) {
+        std::string json = chunk.substr(dataPrefix.size());
+        while (!json.empty() && (json.back() == '\n' || json.back() == '\r'))
+            json.pop_back();
+        return json;
+    };
+
+    // Collect all delta text
```
dkalinowski:
```suggestion
    // Collect all but the last delta text
```

src/audio/speech_to_text/s2t_servable.cpp
```diff
@@ -0,0 +1,116 @@
...
+                return absl::InvalidArgumentError("Invalid temperature type.");
+        }
+        if (temp.value() < 0.0f || temp.value() > 2.0f)
+            return absl::InvalidArgumentError("Temperature out of range(0.0, 2.0)");
+        temperature = temp.value();
+    }
+    return absl::OkStatus();
+}
+
+absl::Status SttServable::applyTranscriptionConfig(ov::genai::WhisperGenerationConfig& config,
```
mzegla:
```suggestion
absl::Status SttServable::updateTranscriptionConfig(ov::genai::WhisperGenerationConfig& config,
```

demos/common/export_models/export_model.py
michalkulakowski: yes, it will work with unary request, for streaming error will be returned 
"{'error': 'Mediapipe execution failed. MP status - INVALID_ARGUMENT: CalculatorGraph::Run() failed: \nCalculator::Process() for node "S2tExecutor" failed: streaming is not supported for this graph configuration (LOOPBACK not configured)'}"

dkalinowski (approved):

src/audio/speech_to_text/s2t_servable.cpp
```diff
@@ -77,6 +77,7 @@ absl::Status SttServable::parseTemperature(const HttpPayload& payload, ov::genai
                 return absl::InvalidArgumentError("Invalid temperature type.");
         }
         config.temperature = temp.value();
+        config.do_sample = true;
```
mzegla: We should only set `do_sample = true`, when temperature is above 0

mzegla (approved):

PR #4143, opened 2026-04-15
server readiness after all models and mp graphs are ready

dtrawins:
### 🛠 Summary

Server readiness when all models and MediaPipe graph are loaded. It corrects the behavior when server was reporting readiness while generative modes were still loading
There is also corrected error status when server is not ready to 503

### 🧪 Checklist

- [ ] Unit tests added.
``


src/server.cpp
```diff
@@ -213,6 +213,9 @@ bool Server::isReady() const {
         return false;
     if (ModuleState::INITIALIZED != it->second->getState())
         return false;
+    const auto& servableManagerModule = static_cast<const ServableManagerModule&>(*it->second);
```
atobiszei, line 216: This is not needed unless we want to ensure that OVMS responds with readiness when all models in config.json added later than initial start, are already loaded. INITIALIZEd status check on servable manager module is sufficient.

PR #4159, opened 2026-04-24
refactor capi stage and targz target

dtrawins:
### 🛠 Summary
This is productization of https://github.com/openvinotoolkit/model_server/pull/3967

CVS-185012 Image building refactor to address issues for RedHat base OS on Konflux system.
Bazel command is in a single layer to avoid issues with passing bazel cache between layers.
It also speeds up build when release docker image is the only needed artefact, skipping capi stage.

### 🧪 Checklist

- [ ] Unit tests added.
``


create_package.sh
```diff
@@ -110,8 +110,7 @@ find /ovms_release/ovms/lib/python/openvino -name *cpython* | grep -vZ $rls_pyth
 
 mkdir -p /ovms_pkg/${BASE_OS}
 cd /ovms_pkg/${BASE_OS}
-tar czf ovms.tar.gz --transform 's/ovms_release/ovms/' /ovms_release/
-sha256sum ovms.tar.gz > ovms.tar.gz.sha256 && \
```
atobiszei, line 112: Why is this removed?

Dockerfile.redhat
```diff
@@ -263,26 +263,17 @@ COPY src/python/binding/tests/requirements.txt /ovms/src/python/binding/tests/re
...
-
 # Mediapipe
 COPY BUILD.bazel /ovms/
 COPY *\.bzl /ovms/
 COPY yarn.lock /ovms/
 COPY package.json /ovms/
 
-# prebuild dependencies before copying sources
-# hadolint ignore=DL3059
-RUN bazel build --jobs=$JOBS ${debug_bazel_flags} //:ovms_dependencies @com_google_googletest//:gtest
```
atobiszei, line 276: Removing all those steps will mean that all those dependencies are not cached in docker layers up until L321:
https://github.com/openvinotoolkit/model_server/pull/4159/changes#diff-507d73bbefd62da3db1eec3a006d6f4bf47433fec017a5299f63921bd0a605b0L321
So basically that means they are never cached in docker. With bazel remote cache that would not be an issue but in every other case its mean recompiling most of ovms dependencies each time anything in our srcs change (L281).

Unless proven empirically that those dependencies recompile anyway, its a no go for me.

create_package.sh
dtrawins: ovms.tar is to be repacked in next stage to add libovms_shared.so

Dockerfile.redhat
mzegla: Yeah, this might be useful to have

dtrawins: this is the evidence https://ci.iotg.sclab.intel.com/job/ovmsc/job/OVMSCOps_SDL_copy/328/console that all the dependencies are included in //src:ovms target

src/config.cpp
```diff
@@ -91,14 +91,12 @@ Config& Config::parse(int argc, char** argv) {
 bool Config::parse(ServerSettingsImpl* serverSettings, ModelsSettingsImpl* modelsSettings) {
     this->serverSettings = *serverSettings;
     this->modelsSettings = *modelsSettings;
-
     static EnvGuard envGuard;
 #if defined(__linux__) || defined(_WIN32)
-    if (this->serverSettings.logLevel == "DEBUG") {
+    if (this->serverSettings.logLevel == "DEBUG" || this->serverSettings.logLevel == "TRACE") {
```
atobiszei, line 96: IMO this log level settings should be set in the same place as for glog:
https://github.com/openvinotoolkit/model_server/blob/main/src/logging.cpp#L134

This can be fixed in other pR.

atobiszei (approved):
PLease check copilot comments.

mzegla (approved):

PR #4162, opened 2026-04-27
add temperature parameter to bfcl streaming

przepeck:
### 🛠 Summary

Fixing BFCL stream option to take into account temperature 

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/demos/continuous_batching/accuracy/gorilla.patch
+++ b/demos/continuous_batching/accuracy/gorilla.patch
@@ -127,7 +127,7 @@ index 10f1a08..b67d39c 100644
          )
  
      #### FC methods ####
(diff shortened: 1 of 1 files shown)


mzegla (approved):
Thanks. Could you also take a quick look at other handlers we use? (like the one for responses etc.) to make sure passing `--temperature` is always effective?

przepeck:
> Thanks. Could you also take a quick look at other handlers we use? (like the one for responses etc.) to make sure passing `--temperature` is always effective?

I' ve checked, in other implementations `temperature` is set

michalkulakowski (approved):

PR #4169, opened 2026-04-29
Mkulakow/legacy pipeline finish reason

michalkulakowski:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -414,7 +414,6 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::Enco
         SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Generated tokens: {}", tokens);
         ParsedOutput parsedOutput = parseOutputIfNeeded(tokens);
         jsonResponse.StartObject();
-        // finish_reason: "stop" in regular scenario, "tool_calls" if output contains tool calls
         if (results.finish_reasons.empty()) {
             throw std::runtime_error("Missing finish reason in unary LM generation result");
         }
@@ -484,7 +483,6 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::VLMD
         SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Generated tokens: {}", generatedTokens);
         ParsedOutput parsedOutput = parseOutputIfNeeded(generatedTokens);
         jsonResponse.StartObject();
-        // finish_reason: "stop" in regular scenario, "tool_calls" if output contains tool calls
         if (results.finish_reasons.empty()) {
             throw std::runtime_error("Missing finish reason in unary VLM generation result");
         }
--- a/src/llm/apis/openai_completions.cpp
+++ b/src/llm/apis/openai_completions.cpp
@@ -409,16 +409,18 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::Enco
 
     // choices: array of size N, where N is related to n request parameter
     jsonResponse.StartArray("choices");
+    if (results.finish_reasons.empty()) {
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Missing finish reason in unary LM generation result, defaulting to STOP for all choices");
+    } else if (results.finish_reasons.size() != results.tokens.size()) {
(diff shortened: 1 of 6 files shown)


src/llm/apis/openai_completions.cpp
```diff
@@ -409,17 +409,18 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::Enco
...
-    int index = 0;
-    for (int i = 0; i < results.tokens.size(); i++) {
+    for (size_t i = 0; i < results.tokens.size(); ++i) {
         const std::vector<int64_t>& tokens = results.tokens[i];
         SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Generated tokens: {}", tokens);
         ParsedOutput parsedOutput = parseOutputIfNeeded(tokens);
         jsonResponse.StartObject();
         // finish_reason: "stop" in regular scenario, "tool_calls" if output contains tool calls
-        auto finishReason = mapFinishReason(ov::genai::GenerationFinishReason::STOP, !parsedOutput.toolCalls.empty());
+        const ov::genai::GenerationFinishReason finishReasonRaw =
```
dkalinowski: is it possible to have results.finish_reasons.empty() ? in which situations?

src/llm/apis/openai_completions.cpp
```diff
@@ -409,17 +409,18 @@ std::string OpenAIChatCompletionsHandler::serializeUnaryResponse(ov::genai::Enco
...
+            (!results.finish_reasons.empty()) ? results.finish_reasons[0] : ov::genai::GenerationFinishReason::STOP;
```
dkalinowski: please add comment that we access [0] because we always use generate with batch=1

src/llm/apis/openai_responses.cpp
```diff
@@ -649,10 +649,17 @@ std::string OpenAIResponsesHandler::serializeUnaryResponse(ov::genai::EncodedRes
     usage.promptTokens = results.perf_metrics.get_num_input_tokens();
     usage.completionTokens = results.perf_metrics.get_num_generated_tokens();
     std::vector<ParsedOutput> parsedOutputs;
+    ov::genai::GenerationFinishReason responsesFinishReason = ov::genai::GenerationFinishReason::STOP;
     for (const auto& tokens : results.tokens) {
         parsedOutputs.push_back(parseOutputIfNeeded(tokens));
     }
-    return serializeUnaryResponseImpl(parsedOutputs);
+    for (const auto& finishReason : results.finish_reasons) {
```
dkalinowski, line 663: why do we have different implementation than chat/completions? cant we just take [0]? I think we also have batch size=1 here always

src/llm/language_model/legacy/servable.cpp
```diff
@@ -229,7 +229,11 @@ absl::Status LegacyServable::preparePartialResponse(std::shared_ptr<GenAiServabl
         if (!executionContext->lastStreamerCallbackOutput.empty()) {
             lastTextChunk = lastTextChunk + executionContext->lastStreamerCallbackOutput;
         }
-        std::string serializedChunk = executionContext->apiHandler->serializeStreamingChunk(lastTextChunk, ov::genai::GenerationFinishReason::STOP);
+        ov::genai::GenerationFinishReason finishReason = ov::genai::GenerationFinishReason::STOP;
+        if (!legacyExecutionContext->results.finish_reasons.empty()) {
+            finishReason = legacyExecutionContext->results.finish_reasons[0];
```
dkalinowski: isnt 0 finish reasons internal error here? should we throw if it happens?

src/llm/apis/openai_responses.cpp
michalkulakowski: For now we could have same implementation for both APIs, but this approach highlights differences (for chat/completions there is one finish reason per output, for response one incomplete details for whole response)

src/llm/apis/openai_completions.cpp
michalkulakowski: rather not, i would say that this is defensive approach for unexpected scenarios

michalkulakowski: but it should be changed to returning internal error as you suggested in other comment

dkalinowski (approved):

atobiszei (approved):

PR #4191, opened 2026-05-07
[WIP] Gemma4 reasoning parser

dkalinowski:


dkalinowski:
To be done: add mapping document which tool parsers are for which models

PR #4208, opened 2026-05-14
Tokenize endpoint  empty array crash fix  

przepeck:
### 🛠 Summary

[CVS-186166](https://jira.devtools.intel.com/browse/CVS-186166)
Adding checks for empty nested array, which was making OVMS crash

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/test/llm/tokenize_endpoint_test.cpp
+++ b/src/test/llm/tokenize_endpoint_test.cpp
@@ -445,7 +445,7 @@ TEST_P(LLMTokenizeTests, tokenizeEmptyNestedArray) {
 
     std::string requestBody = R"(
         {
-            "model": ")"+ params.modelName +
+            "model": ")" + params.modelName +
                               R"(",
             "text": [[]]
         }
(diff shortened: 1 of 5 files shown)


michalkulakowski (approved):

src/test/llm/tokenize_endpoint_test.cpp
```diff
@@ -417,9 +417,6 @@ TEST_P(LLMTokenizeTests, tokenizeArrayOfStringsWithPaddingSideLeft) {
 
 TEST_P(LLMTokenizeTests, tokenizeStringWithAddSpecialTokens) {
     auto params = GetParam();
-    if (params.modelName == "vlm_cb_regular" || params.modelName == "vlm_legacy_regular") {
-        GTEST_SKIP() << "Skipping test for " << params.modelName;
```
dkalinowski, line 421: Is it working for all paths, or only special token path? Do other tests have it unskipped?

dkalinowski (approved):

src/test/llm/tokenize_endpoint_test.cpp
przepeck: This was only the case of special tokens in vlm in other cases tests weren't skipped. I've checked with latest Miłosz's changes according to special tokens in VLM and it seems to pass.

PR #4212, opened 2026-05-15
Log VLM perf metrics at DEBUG level for legacy and continuous batchin…

junruizh2021:
…g pipelines

Add debug-level logging of VLM performance metrics including:
- input_token_count, prepare_embeddings_time_ms, ttft_ms, prefill_speed_tps
- image_slice_count (from GenAI VLMPerfMetrics)
- For CB pipeline: output_token_count, llm_ttft_ms, total_token_count

Legacy VLM pipeline logs metrics after unary and streaming responses. Continuous batching VLM pipeline retrieves request-level metrics via GenerationHandle::get_vlm_perf_metrics() (requires openvinotoolkit/openvino.genai#3860).

CVS-186710

### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/language_model/continuous_batching/servable.cpp
+++ b/src/llm/language_model/continuous_batching/servable.cpp
@@ -42,6 +42,23 @@
 
 namespace ovms {
 
+void ContinuousBatchingServable::logPerfMetrics(ov::genai::PerfMetrics& perfMetrics) {
+    const size_t inputTokenCount = perfMetrics.get_num_input_tokens();
+    const size_t outputTokenCount = perfMetrics.get_num_generated_tokens();
+    // GenerationHandle metrics are scoped to this request; TTFT contains one sample.
+    const double ttftMs = perfMetrics.get_ttft().mean;
+    const double prefillSpeedTps = calculatePrefillSpeed(inputTokenCount, ttftMs);
+
+    SPDLOG_LOGGER_DEBUG(
+        llm_calculator_logger,
+        "Request processing metrics | input_token_count: {} | output_token_count: {} | total_token_count: {} | ttft_ms: {:.3f} | prefill_speed_tps: {:.3f}",
+        inputTokenCount,
+        outputTokenCount,
+        inputTokenCount + outputTokenCount,
+        ttftMs,
+        prefillSpeedTps);
+}
+
 void ContinuousBatchingServable::notifyExecutorThread() {
     SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Notifying executor thread");
     if (properties->llmExecutorWrapper == nullptr) {
@@ -142,6 +159,28 @@ absl::Status ContinuousBatchingServable::readPartialExecutionResults(std::shared
         }
     }
     return absl::OkStatus();
+}
+
+absl::Status ContinuousBatchingServable::prepareCompleteResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto status = GenAiServable::prepareCompleteResponse(executionContext);
+    if (status.ok() && llm_calculator_logger->should_log(spdlog::level::debug)) {
+        auto cbExecutionContext = std::static_pointer_cast<ContinuousBatchingServableExecutionContext>(executionContext);
+        auto perfMetrics = cbExecutionContext->generationHandle->get_perf_metrics();
+        logPerfMetrics(perfMetrics);
+    }
+    return status;
+}
+
+absl::Status ContinuousBatchingServable::preparePartialResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto status = GenAiServable::preparePartialResponse(executionContext);
+    if (status.ok() &&
+        !executionContext->sendLoopbackSignal &&
+        llm_calculator_logger->should_log(spdlog::level::debug)) {
+        auto cbExecutionContext = std::static_pointer_cast<ContinuousBatchingServableExecutionContext>(executionContext);
+        auto perfMetrics = cbExecutionContext->generationHandle->get_perf_metrics();
+        logPerfMetrics(perfMetrics);
+    }
+    return status;
 }
 
 }  // namespace ovms
--- a/src/llm/servable.cpp
+++ b/src/llm/servable.cpp
@@ -45,6 +45,10 @@
 
 namespace ovms {
 
+double calculatePrefillSpeed(size_t inputTokenCount, double ttftMs) {
+    return ttftMs > 0.0 ? (1000.0 * inputTokenCount) / ttftMs : 0.0;
+}
+
 void GenAiServable::determineDecodingMethod() {
     getProperties()->decodingMethod = DecodingMethod::STANDARD;
     auto& pluginConfig = getProperties()->pluginConfig;
--- a/src/llm/servable.hpp
+++ b/src/llm/servable.hpp
@@ -48,6 +48,8 @@ namespace ovms {
 // Some pipelines internals rely on request_id, so for now we provide increasing ID
 static std::atomic<uint64_t> currentRequestId = 0;
 
+double calculatePrefillSpeed(size_t inputTokenCount, double ttftMs);
+
 /*
 GenAiServable support.
 
--- a/src/llm/visual_language_model/continuous_batching/servable.cpp
+++ b/src/llm/visual_language_model/continuous_batching/servable.cpp
@@ -35,6 +35,28 @@
 
 namespace ovms {
 
+void VisualLanguageModelServable::logPerfMetrics(ov::genai::VLMPerfMetrics& perfMetrics) {
+    const size_t inputTokenCount = perfMetrics.get_num_input_tokens();
+    const size_t outputTokenCount = perfMetrics.get_num_generated_tokens();
+    const double prepareEmbeddingsTimeMs = perfMetrics.get_prepare_embeddings_duration().mean;
+    // Continuous batching starts request timing after embeddings preparation.
+    const double llmTtftMs = perfMetrics.get_ttft().mean;
+    const double ttftMs = llmTtftMs + prepareEmbeddingsTimeMs;
+    const double prefillSpeedTps = calculatePrefillSpeed(inputTokenCount, llmTtftMs);
+
+    SPDLOG_LOGGER_DEBUG(
+        llm_calculator_logger,
+        "Request processing metrics | input_token_count: {} | output_token_count: {} | total_token_count: {} | prepare_embeddings_time_ms: {:.3f} | llm_ttft_ms: {:.3f} | ttft_ms: {:.3f} | prefill_speed_tps: {:.3f} | image_slice_count: {}",
+        inputTokenCount,
+        outputTokenCount,
+        inputTokenCount + outputTokenCount,
+        prepareEmbeddingsTimeMs,
+        llmTtftMs,
+        ttftMs,
+        prefillSpeedTps,
+        perfMetrics.get_total_image_slice_count());
+}
+
 absl::Status VisualLanguageModelServable::addRequestToPipeline(std::shared_ptr<ContinuousBatchingServableExecutionContext>& executionContext) {
     auto vlmExecutionContext = std::static_pointer_cast<VisualLanguageModelServableExecutionContext>(executionContext);
     vlmExecutionContext->generationHandle = properties->pipeline->add_request(currentRequestId++,  // to be removed from API?
@@ -56,6 +78,28 @@ std::shared_ptr<GenAiServableExecutionContext> VisualLanguageModelServable::crea
 
 std::shared_ptr<GenAiServableProperties> VisualLanguageModelServable::getProperties() {
     return properties;
+}
+
+absl::Status VisualLanguageModelServable::prepareCompleteResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto status = GenAiServable::prepareCompleteResponse(executionContext);
+    if (status.ok() && llm_calculator_logger->should_log(spdlog::level::debug)) {
+        auto vlmExecutionContext = std::static_pointer_cast<VisualLanguageModelServableExecutionContext>(executionContext);
+        auto perfMetrics = vlmExecutionContext->generationHandle->get_vlm_perf_metrics();
+        logPerfMetrics(perfMetrics);
+    }
+    return status;
+}
+
+absl::Status VisualLanguageModelServable::preparePartialResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    auto status = GenAiServable::preparePartialResponse(executionContext);
+    if (status.ok() &&
+        !executionContext->sendLoopbackSignal &&
+        llm_calculator_logger->should_log(spdlog::level::debug)) {
+        auto vlmExecutionContext = std::static_pointer_cast<VisualLanguageModelServableExecutionContext>(executionContext);
+        auto perfMetrics = vlmExecutionContext->generationHandle->get_vlm_perf_metrics();
+        logPerfMetrics(perfMetrics);
+    }
+    return status;
 }
 
 }  // namespace ovms
--- a/src/llm/language_model/continuous_batching/servable.hpp
+++ b/src/llm/language_model/continuous_batching/servable.hpp
@@ -40,6 +40,7 @@ class ContinuousBatchingServable : public GenAiServable {
 protected:
     std::shared_ptr<ContinuousBatchingServableProperties> properties;
     void notifyExecutorThread();
+    void logPerfMetrics(ov::genai::PerfMetrics& perfMetrics);
 
 public:
     ContinuousBatchingServable() {
@@ -56,5 +57,7 @@ public:
     absl::Status scheduleExecution(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
     absl::Status readCompleteExecutionResults(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
     absl::Status readPartialExecutionResults(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
+    absl::Status prepareCompleteResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
+    absl::Status preparePartialResponse(std::shared_ptr<GenAiServableExecutionContext>& executionContext) override;
 };
 }  // namespace ovms
--- a/src/llm/language_model/legacy/servable.cpp
+++ b/src/llm/language_model/legacy/servable.cpp
@@ -46,6 +46,22 @@
 
 namespace ovms {
 
+void LegacyServable::logPerfMetrics(ov::genai::PerfMetrics& perfMetrics) {
+    const size_t inputTokenCount = perfMetrics.get_num_input_tokens();
+    const size_t outputTokenCount = perfMetrics.get_num_generated_tokens();
+    const double ttftMs = perfMetrics.get_ttft().mean;
+    const double prefillSpeedTps = calculatePrefillSpeed(inputTokenCount, ttftMs);
+
+    SPDLOG_LOGGER_DEBUG(
+        llm_calculator_logger,
+        "Request processing metrics | input_token_count: {} | output_token_count: {} | total_token_count: {} | ttft_ms: {:.3f} | prefill_speed_tps: {:.3f}",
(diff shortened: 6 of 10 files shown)


mzegla:
If we want to push this PR forward:
- looks like it requires GenAI PR with metrics merged first, so it must happen
- there has been a refactor recently, so a rebase is required
- I would like to go further - if we merge such change with logging more per request stats, let's do it for LLMs too

junruizh2021:
@mzegla Thanks, agreed on all three points.
- The GenAI dependency is tracked in openvinotoolkit/openvino.genai#3860. It is still open, so I will get that change reviewed and merged first.
- I have rebased this PR onto the current main (43e825e1); the updated PR head is b1105d99.
- I agree that request-level performance logging should be consistent for LLM and VLM. I will extend the OVMS change to cover both VLM & LLM paths as well

junruizh2021:
The OpenVINO GenAI dependency PR [openvinotoolkit/openvino.genai#3860](https://github.com/openvinotoolkit/openvino.genai/pull/3860) has been merged.
I have rebased this PR onto the latest model_server/main and updated the implementation to use the finalized OpenVINO GenAI APIs:
- GenerationHandle::get_perf_metrics()
- GenerationHandle::get_vlm_perf_metrics()
- VLMPerfMetrics::get_total_image_slice_count()
Request-level metrics logging now covers both LLM and VLM paths, including legacy and continuous batching pipelines, for unary and final streaming responses.

I also updated the OpenVINO GenAI dependency to the 2026.4.0.0.dev20260722 nightly build and added unit tests for the shared request metrics calculations.
Validation completed:
- //src/llm:genai_servables builds successfully against the updated GenAI package.
- All 4 RequestPerfMetricsTest tests pass.

versions.mk
mzegla, line 1: Please drop that change and align to main. It's already has GenAI version with required changes.

src/llm/servable.cpp
```diff
@@ -45,6 +46,85 @@
 
 namespace ovms {
 
+static constexpr double MIN_PREFILL_DURATION_MS = 1e-9;
```
mzegla: What's the purpose of that const?

src/llm/servable.cpp
```diff
@@ -45,6 +46,85 @@
...
+
+static const char* getPipelineTypeName(GenAiPipelineType pipelineType) {
+    switch (pipelineType) {
+    case GenAiPipelineType::LEGACY:
+        return "legacy";
+    case GenAiPipelineType::CONTINUOUS_BATCHING:
+        return "continuous_batching";
+    }
+    return "unknown";
+}
```
mzegla: I don't think we need it. I would suggest to remove this method and modify the solution. 
Instead of having logLLMPerfMetrics and logVLMPerfMetrics here, we could attach that logic to proper servables making them simply have `logPerfMetrics` method with respective types. Then we would call them like:
```
if (llm_calculator_logger->should_log(spdlog::level::debug)) {
	logPerfMetrics(metrics);
}
```

src/llm/servable.hpp
```diff
@@ -48,6 +50,30 @@ namespace ovms {
...
+};
+
+struct RequestPerfMetrics {
+    size_t inputTokenCount;
+    size_t outputTokenCount;
+    size_t totalTokenCount;
+    double llmTtftMs;
+    double ttftMs;
+    double prefillSpeedTps;
+};
```
mzegla: Can we drop this struct and rely on PerfMetrics directly in the log method?

src/llm/servable.hpp
```diff
@@ -48,6 +50,30 @@ namespace ovms {
...
+
+RequestPerfMetrics getRequestPerfMetrics(
+    ov::genai::PerfMetrics& perfMetrics,
+    double prepareEmbeddingsTimeMs = 0.0,
+    bool ttftIncludesPrepareEmbeddings = false);
```
mzegla: I don't think we need those parameters. Prepare embeddings time should already be included in the perfMetrics object and the flag ttftIncludesPrepareEmbeddings is always true if process request with images.
Perhaps if no embeddings were computed the perfMetrics.get_prepare_embeddings_duration().mean would be zero? If that's the case we can unconditionally add or subtract this time to get isolated LLM prefill processing time. Can you confirm it and if that's the case make a change here?

versions.mk
junruizh2021: Done. I rebased the branch onto the latest `main` and dropped the `versions.mk` changes entirely. The final PR diff now uses the GenAI version already provided by `main`.

src/llm/servable.hpp
junruizh2021: Confirmed and updated. The extra parameters and flag have been removed. One nuance from checking the GenAI implementation: `prepare_embeddings_duration` is also recorded for text-only VLM requests, because the VLM path prepares input embeddings even when there are no images, so its mean is not expected to be zero. Also, legacy VLM measures TTFT from the start of `generate()` and therefore includes embeddings preparation, while continuous batching starts request timing after embeddings preparation. The calculations now live in the respective servables: legacy subtracts the preparation duration to obtain isolated LLM TTFT, while continuous batching adds it to obtain total request TTFT.

src/llm/servable.cpp
junruizh2021: Removed. Instead of using the epsilon constant to avoid division by zero, the servable-specific calculations now explicitly check whether TTFT is greater than zero and report `prefill_speed_tps` as `0` otherwise.

src/llm/servable.hpp
junruizh2021: Done. `RequestPerfMetrics` and `getRequestPerfMetrics` have been removed. Each `logPerfMetrics(...)` method now reads and calculates the logged values directly from `PerfMetrics` or `VLMPerfMetrics`.

src/llm/servable.cpp
junruizh2021: Done. I removed the shared `logLLMPerfMetricsDebug` / `logVLMPerfMetricsDebug` functions. Each concrete LLM/VLM legacy/continuous-batching servable now has its own typed `logPerfMetrics(...)` method, and the call sites first check the debug log level as suggested.

src/llm/language_model/continuous_batching/servable.cpp
```diff
@@ -42,6 +42,22 @@
...
 
+void ContinuousBatchingServable::logPerfMetrics(ov::genai::PerfMetrics& perfMetrics) {
+    const size_t inputTokenCount = perfMetrics.get_num_input_tokens();
+    const size_t outputTokenCount = perfMetrics.get_num_generated_tokens();
+    const double ttftMs = perfMetrics.get_ttft().mean;
+    const double prefillSpeedTps = ttftMs > 0.0 ? (1000.0 * inputTokenCount) / ttftMs : 0.0;
+
+    SPDLOG_LOGGER_DEBUG(
+        llm_calculator_logger,
+        "LLM continuous_batching request metrics | input_token_count: {} | output_token_count: {} | total_token_count: {} | ttft_ms: {:.3f} | prefill_speed_tps: {:.3f}",
```
mzegla: For now I would drop first part. If we have just one pipeline deployed - it's irrelevant. If we have more - it's inconclusive, so let's just go with the generic message.
```suggestion
        "Request processing metrics | input_token_count: {} | output_token_count: {} | total_token_count: {} | ttft_ms: {:.3f} | prefill_speed_tps: {:.3f}",
```

src/llm/language_model/continuous_batching/servable.cpp
```diff
@@ -42,6 +42,22 @@
 
 namespace ovms {
 
+void ContinuousBatchingServable::logPerfMetrics(ov::genai::PerfMetrics& perfMetrics) {
+    const size_t inputTokenCount = perfMetrics.get_num_input_tokens();
+    const size_t outputTokenCount = perfMetrics.get_num_generated_tokens();
+    const double ttftMs = perfMetrics.get_ttft().mean;
```
mzegla, line 49: It's a `mean`. That would suggest `perfMetrics` averages TTFT across multiple requests. This can work if it only holds metrics for one result only, so the mean is effectively a direct value of TTFT from that one request. Can you confirm the `perfMetrics` we see here are exclusive for the request we are logging for?

src/llm/visual_language_model/continuous_batching/servable.cpp
```diff
@@ -35,6 +35,28 @@
...
+    const size_t outputTokenCount = perfMetrics.get_num_generated_tokens();
+    const double prepareEmbeddingsTimeMs = perfMetrics.get_prepare_embeddings_duration().mean;
+    // Continuous batching starts request timing after embeddings preparation.
+    const double llmTtftMs = perfMetrics.get_ttft().mean;
+    const double ttftMs = llmTtftMs + prepareEmbeddingsTimeMs;
+    const double prefillSpeedTps = llmTtftMs > 0.0 ? (1000.0 * inputTokenCount) / llmTtftMs : 0.0;
+
+    SPDLOG_LOGGER_DEBUG(
+        llm_calculator_logger,
+        "VLM continuous_batching request metrics | input_token_count: {} | output_token_count: {} | total_token_count: {} | prepare_embeddings_time_ms: {:.3f} | llm_ttft_ms: {:.3f} | ttft_ms: {:.3f} | prefill_speed_tps: {:.3f} | image_slice_count: {}",
```
mzegla: Same as for method in LLM. Let's go with plain `Request processing metrics`. The log will appear in the context so the part `VLM continuous batching` feels redundant.

mzegla:
Also, CI failed due to linter issue:
`[2026-07-31T03:07:58.413Z] src/llm/visual_language_model/legacy/servable.cpp(61): error cpplint: [build/include_what_you_use] Add #include <algorithm> for max [4]`

src/llm/language_model/continuous_batching/servable.cpp
junruizh2021: Confirmed. The `PerfMetrics` returned by this `GenerationHandle` is scoped to the current request. In GenAI continuous batching, every request owns a separate `SequenceGroup` with its own `m_start_time` and `m_perf_metrics`. The pipeline updates each scheduled `SequenceGroup` separately and, on completion, stores that group metrics in that request `GenerationStream`; `GenerationHandle::get_perf_metrics()` reads only from that stream.

`evaluate_statistics(m_start_time)` also clears `m_times_to_first_token` and inserts exactly one value, `first_token_timestamp - request_start_time`, before calculating `ttft.mean`. Therefore this `mean` is the mean of a single TTFT sample for this request, not an average across concurrent requests. The `MeanStdPair` is the generic PerfMetrics API return type. I added a source comment in `e1d1d8ed` to make this request-level scope explicit.

src/llm/visual_language_model/continuous_batching/servable.cpp
junruizh2021: Done in `e1d1d8ed`. The VLM log now uses the plain `Request processing metrics` prefix. I also standardized the legacy LLM/VLM paths for consistency.

src/llm/language_model/continuous_batching/servable.cpp
junruizh2021: Done in `e1d1d8ed`. I changed the prefix to `Request processing metrics` and applied the same generic prefix consistently across all four LLM/VLM legacy and continuous-batching servables.

junruizh2021:
@mzegla The reported cpplint failure is fixed in `e1d1d8ed` by adding the explicit `<algorithm>` include required by `std::max`. `make cpplint` now passes.

One dependency consistency issue surfaced during validation: current `main` pins `OV_GENAI_BRANCH` to `bd8d6542`, whose `GenerationHandle` header does not expose `get_perf_metrics()` or `get_vlm_perf_metrics()`; those APIs are present in the merged GenAI PR #3860, and the commits currently have diverged histories. The prebuilt package may contain the APIs, but a source build using the pinned commit appears inconsistent. I kept `versions.mk` unchanged as requested and aligned this PR with `main`. Could you confirm whether the GenAI source pin is expected to be updated separately?

src/llm/language_model/continuous_batching/servable.cpp
```diff
@@ -42,6 +42,23 @@
 
 namespace ovms {
 
+void ContinuousBatchingServable::logPerfMetrics(ov::genai::PerfMetrics& perfMetrics) {
+    const size_t inputTokenCount = perfMetrics.get_num_input_tokens();
+    const size_t outputTokenCount = perfMetrics.get_num_generated_tokens();
+    // GenerationHandle metrics are scoped to this request; TTFT contains one sample.
+    const double ttftMs = perfMetrics.get_ttft().mean;
+    const double prefillSpeedTps = ttftMs > 0.0 ? (1000.0 * inputTokenCount) / ttftMs : 0.0;
```
michalkulakowski: This could be moved to utility function and used across all servables

michalkulakowski (changes requested):

src/llm/language_model/continuous_batching/servable.cpp
junruizh2021: Done in `847d84cd`. I extracted the calculation, including zero-TTFT handling, into the shared `calculatePrefillSpeed(...)` utility and now use it across all four LLM/VLM legacy and continuous-batching servables. The servable-specific TTFT calculations and logging remain in their respective servables.

michalkulakowski (approved):

mzegla (approved):

junruizh2021:
> ### 🟡 Not ready to approve
> The new perf-metrics helpers and call sites introduce avoidable copying and lack const-correctness (non-const references / non-const methods) that should be corrected before merging.
> 
> _Once you've addressed the issues Copilot identified, you can request another Copilot review._
> 
> _This review doesn't count toward merge requirements. [Sign up for the private preview](https://forms.cloud.microsoft/r/zLCqnkB1FJ) to control whether Copilot approvals count._
> 
> Review details
> ### Suppressed comments (12)
> **src/llm/visual_language_model/continuous_batching/servable.cpp:101**
> 
> * Same as unary path: prefer binding perf metrics to `const auto&` to avoid copying if GenAI returns a reference-like object.
> 
> ```
>     if (status.ok() &&
>         !executionContext->sendLoopbackSignal &&
>         llm_calculator_logger->should_log(spdlog::level::debug)) {
>         auto vlmExecutionContext = std::static_pointer_cast<VisualLanguageModelServableExecutionContext>(executionContext);
>         auto perfMetrics = vlmExecutionContext->generationHandle->get_vlm_perf_metrics();
>         logPerfMetrics(perfMetrics);
>     }
> ```
> 
> **src/llm/language_model/continuous_batching/servable.cpp:182**
> 
> * Same as unary path: prefer `const auto&` to avoid copying perf metrics when the GenAI API returns a reference-like object.
> 
> ```
>     if (status.ok() &&
>         !executionContext->sendLoopbackSignal &&
>         llm_calculator_logger->should_log(spdlog::level::debug)) {
>         auto cbExecutionContext = std::static_pointer_cast<ContinuousBatchingServableExecutionContext>(executionContext);
>         auto perfMetrics = cbExecutionContext->generationHandle->get_perf_metrics();
>         logPerfMetrics(perfMetrics);
>     }
> ```
> 
> **src/llm/visual_language_model/legacy/servable.hpp:58**
> 
> * logPerfMetrics does not modify the metrics object and doesn’t appear to mutate servable state; taking a non-const reference prevents calling it with temporaries/const metrics and makes intent less clear. Prefer a const-correct signature (and make the method const).
> 
> ```
> class VisualLanguageModelLegacyServable : public GenAiServable {
>     std::shared_ptr<VisualLanguageModelLegacyServableProperties> properties;
>     void logPerfMetrics(ov::genai::VLMPerfMetrics& perfMetrics);
> ```
> 
> **src/llm/visual_language_model/continuous_batching/servable.hpp:39**
> 
> * logPerfMetrics takes a non-const reference even though it only reads metrics; using const reference (and making the method const) improves const-correctness and allows binding to temporaries returned from GenAI APIs.
> 
> ```
> class VisualLanguageModelServable : public ContinuousBatchingServable {
>     void logPerfMetrics(ov::genai::VLMPerfMetrics& perfMetrics);
> ```
> 
> **src/llm/visual_language_model/legacy/servable.cpp:56**
> 
> * logPerfMetrics doesn’t modify perfMetrics or class state; make it const-correct to avoid requiring a mutable metrics instance and to better document intent.
> 
> ```
> void VisualLanguageModelLegacyServable::logPerfMetrics(ov::genai::VLMPerfMetrics& perfMetrics) {
> ```
> 
> **src/llm/visual_language_model/continuous_batching/servable.cpp:38**
> 
> * logPerfMetrics only reads metrics; prefer const-correct signature (const ref + const method) to avoid forcing callers to have a mutable metrics instance.
> 
> ```
> void VisualLanguageModelServable::logPerfMetrics(ov::genai::VLMPerfMetrics& perfMetrics) {
> ```
> 
> **src/llm/visual_language_model/continuous_batching/servable.cpp:89**
> 
> * get_vlm_perf_metrics() may return a reference-like object or a potentially expensive-to-copy struct; using `auto` here forces a copy when it returns by reference. Binding to `const auto&` avoids an unnecessary copy while still working if the API returns by value (lifetime extension).
> 
> This issue also appears on line 95 of the same file.
> 
> ```
>     if (status.ok() && llm_calculator_logger->should_log(spdlog::level::debug)) {
>         auto vlmExecutionContext = std::static_pointer_cast<VisualLanguageModelServableExecutionContext>(executionContext);
>         auto perfMetrics = vlmExecutionContext->generationHandle->get_vlm_perf_metrics();
>         logPerfMetrics(perfMetrics);
>     }
> ```
> 
> **src/llm/language_model/legacy/servable.hpp:55**
> 
> * logPerfMetrics is read-only but takes a non-const reference; prefer `const ov::genai::PerfMetrics&` and make the method const to improve const-correctness and usability.
> 
> ```
> class LegacyServable : public GenAiServable {
>     std::shared_ptr<LegacyServableProperties> properties;
>     void logPerfMetrics(ov::genai::PerfMetrics& perfMetrics);
> ```
> 
> **src/llm/language_model/legacy/servable.cpp:49**
> 
> * logPerfMetrics only reads metrics; make it `const` and take a `const` reference to avoid requiring a mutable PerfMetrics instance.
> 
> ```
> void LegacyServable::logPerfMetrics(ov::genai::PerfMetrics& perfMetrics) {
> ```
> 
> **src/llm/language_model/continuous_batching/servable.hpp:44**
> 
> * logPerfMetrics reads but does not mutate metrics; using a non-const reference is unnecessarily restrictive. Prefer a const-correct signature and make the method const.
> 
> ```
> class ContinuousBatchingServable : public GenAiServable {
> protected:
>     std::shared_ptr<ContinuousBatchingServableProperties> properties;
>     void notifyExecutorThread();
>     void logPerfMetrics(ov::genai::PerfMetrics& perfMetrics);
> ```
> 
> **src/llm/language_model/continuous_batching/servable.cpp:45**
> 
> * logPerfMetrics only reads PerfMetrics; make it const-correct (const ref + const method) to clarify intent and avoid requiring a mutable metrics object.
> 
> ```
> void ContinuousBatchingServable::logPerfMetrics(ov::genai::PerfMetrics& perfMetrics) {
> ```
> 
> **src/llm/language_model/continuous_batching/servable.cpp:170**
> 
> * Using `auto` here copies PerfMetrics if get_perf_metrics() returns by reference-like type; binding to `const auto&` avoids an unnecessary copy while remaining safe if the API returns by value.
> 
> This issue also appears on line 176 of the same file.
> 
> ```
>     if (status.ok() && llm_calculator_logger->should_log(spdlog::level::debug)) {
>         auto cbExecutionContext = std::static_pointer_cast<ContinuousBatchingServableExecutionContext>(executionContext);
>         auto perfMetrics = cbExecutionContext->generationHandle->get_perf_metrics();
>         logPerfMetrics(perfMetrics);
>     }
> ```
> 
> * **Files reviewed:** 8/8 changed files
> * **Comments generated:** 0 new
> * **Review effort level:** Lite
> 
> We're testing this review assessment. Please use 👍 or 👎 to tell us if it's correct.

The GenAI get_perf_metrics() and get_vlm_perf_metrics() APIs return metrics by value, so the local auto materializes the returned object without an additional copy. Also, the PerfMetrics getters used here are currently non-const because they lazily evaluate and cache statistics. Consequently, accepting const PerfMetrics& or binding the result as const auto& would prevent those getters from being called. The current mutable local object and non-const reference are therefore required by the GenAI API.

PR #4216, opened 2026-05-18
Chat template application fixes

dkalinowski:


src/llm/io_processing/qwen3/reasoning_parser.cpp
```diff
@@ -38,6 +38,19 @@ void Qwen3ReasoningParser::parse(ParsedOutput& parsedOutput, const std::vector<i
         parsedOutput.reasoning = reasoningText;
         // Remove reasoning from content
         parsedOutput.content.erase(startPos, endPos - startPos + endReasoningTag.length());
+        return;
+    }
+
+    // Implicit-start mode: the chat template already emitted the start tag as the prompt
+    // suffix, so the model output begins inside the reasoning segment.
+    if (implicitStart && startPos == std::string::npos) {
```
mzegla: 1. I would move this block up - I feel like implicitStart should have priority here. 
2. `startPos == npos` - is that a right thing to pair it with implicitStart? If model emits `<think>` somewhere in the output, then `implicitStart` is not honored and I suppose it's not right.

src/llm/io_processing/output_parser.hpp
```diff
@@ -79,6 +79,16 @@ class OutputParser {
     bool isReasoningParserAvailable() const;
     std::string getToolParserStartTag() const;
 
+    // Configure parser to treat the output as already-in-reasoning from the first token.
+    // Used when the chat template appends the reasoning start tag (e.g. "<think>\n") as
+    // the prompt suffix - the model then emits only the reasoning body and the closing tag.
+    // No-op when no reasoning parser is configured.
+    void setImplicitReasoningStart(bool value);
```
mzegla: Could be private?

src/llm/io_processing/output_parser.hpp
```diff
@@ -79,6 +79,16 @@ class OutputParser {
...
+
+    // Auto-detect and apply implicit reasoning start based on the prompt produced by
+    // the chat template. Returns true if implicit start was activated.
+    bool detectAndSetImplicitReasoningStart(const std::string& renderedPrompt);
```
mzegla: Return value is never checked.

src/llm/io_processing/qwen3/reasoning_parser.cpp
dkalinowski: If model emits <think>, it entire response is already compromised and irrecoverable
I dont mind changing order

dkalinowski: @mzegla

src/llm/io_processing/output_parser.hpp
dkalinowski: done

src/llm/io_processing/output_parser.hpp
dkalinowski: done

src/test/llm/output_parsers/qwen3_output_parser_test.cpp
```diff
@@ -542,3 +542,197 @@ TEST_F(Qwen3OutputParserTest, ToolCallsDataAfterToolCall) {
...
+    ParsedOutput parsedOutput = outputParser->parse(generatedTokens, false);
+    EXPECT_EQ(parsedOutput.reasoning, "let me think");
+    EXPECT_EQ(parsedOutput.content, "final answer");
+}
+
+TEST_F(Qwen3OutputParserTest, ImplicitStart_UnaryExplicitThinkInOutputStillHonored) {
+    // If implicit start was detected but the model also emitted an explicit <think> (unusual
+    // but legal), the explicit-tag branch wins and behaves like the no-implicit-start case.
+    outputParser->detectAndSetImplicitReasoningStart("<|im_start|>assistant\n<think>\n");
+    std::string input = "prefix<think>inner</think>suffix";
```
dkalinowski, line 650: this should be changed to not honored

extras/chat_template_examples/chat_template_qwen36.jinja
```diff
@@ -0,0 +1,167 @@
...
+                    {%- else %}
+                        {{- '<tool_call>\n<function=' + tool_call.name + '>\n' }}
+                    {%- endif %}
+                {%- else %}
+                    {{- '\n<tool_call>\n<function=' + tool_call.name + '>\n' }}
+                {%- endif %}
+                {%- if tool_call.arguments is defined %}
+                    {%- for args_name, args_value in tool_call.arguments|items %}
+                        {{- '<parameter=' + args_name + '>\n' }}
+                        {%- if args_value is string %}
```
atobiszei, line 129: Leave comment in chat template what is changed? # BEGIN_OF_CHANGE / #END_OF_CHANGE?

src/llm/apis/openai_completions.cpp
```diff
@@ -180,8 +180,8 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages(std::optional<std::stri
                 request.chatHistory.last()[memberName] = member->value.GetString();
                 continue;
             }
-            // Handle tool_call_id and name string fields for tool calling
-            if (member->value.IsString() && (memberName == "tool_call_id" || memberName == "name")) {
+            // Handle tool_call_id and name string fields for tool calling, and reasoning_content for assistant messages
```
atobiszei, line 183: General comment but outside of this PR scope I think, there's a lot of logic here requiring loading models, and probably not testable at this level.

src/test/llm/output_parsers/qwen3_output_parser_test.cpp
atobiszei: So what happens in this case? It looks like nested "reasoning"

dkalinowski: We agreed with @mzegla this is already corrupted model response, but the output should be:
- reasoning: `prefix<think>inner`
- content: `suffix`

atobiszei (approved):

mzegla (approved):

extras/chat_template_examples/chat_template_qwen36.jinja
dkalinowski: done

PR #4223, opened 2026-05-19
avoid too high number of threads on multi socket system and with HT

dtrawins:
### 🛠 Summary

Problems to solve:
on dual socket host with HT, number of threads was too high in latency model. Optimal is number of physical cores on one socket.
Without docker on linux, OV should apply all defaults
When container has quota, number of threads should be like number of allocated cores.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/ov_utils.cpp
+++ b/src/ov_utils.cpp
@@ -186,8 +186,7 @@ Status applyDefaultCpuProperties(ov::AnyMap& properties) {
             if (isThroughput) {
                 numThreads = static_cast<int>(coreCount);
             } else {
-                const uint16_t sockets = getNumberOfSockets();
-                numThreads = std::max(1, std::min(static_cast<int>(coreCount), static_cast<int>(getNumberOfPhysicalCores() / sockets)));
+                numThreads = std::min(static_cast<int>(coreCount), static_cast<int>(getPhysicalCoresPerSocket()));
             }
             properties[ov::inference_num_threads.name()] = numThreads;
             SPDLOG_DEBUG("applyDefaultCpuProperties: setting inference_num_threads to {}", numThreads);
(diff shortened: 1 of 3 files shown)


src/ov_utils.cpp
```diff
@@ -151,28 +151,22 @@ Status validatePluginConfiguration(const plugin_config_t& pluginConfig, const st
...
-        const uint16_t coreCount = getCoreCount();
-        const uint16_t sanitizedCoreCount = coreCount > 0 ? coreCount : 1;
-
-        if (properties.find(ov::inference_num_threads.name()) == properties.end()) {
-            properties[ov::inference_num_threads.name()] = static_cast<int>(sanitizedCoreCount);
-            SPDLOG_DEBUG("applyDefaultCpuProperties: setting inference_num_threads to {}", sanitizedCoreCount);
+        if (!isRunningInDocker()) {
+            return StatusCode::OK;
         }
+        const uint16_t coreCount = getCoreCount();
```
mzegla, line 159: No need for >0 check?

src/systeminfo.cpp
```diff
@@ -152,6 +153,24 @@ uint16_t getDockerCpuQuota() {
...
 }
 
+uint16_t getPhysicalCoresPerSocket() {
+    std::set<std::string> uniqueCores;
+    std::ifstream cpuInfo("/proc/cpuinfo");
+    if (!cpuInfo.is_open()) {
+        return std::max<uint16_t>(static_cast<uint16_t>(std::thread::hardware_concurrency()), 1);
+    }
+    std::string line;
+    while (std::getline(cpuInfo, line)) {
```
mzegla, line 163: Do we treat cpu info file as trusted? 
We read, iterate and load lines to memory so potentially malicious cpu info could overflow memory.

mzegla (approved):

dkalinowski (approved):

PR #4229, opened 2026-05-21
Fix race condition on files in libgit2

rasapala:
### 🛠 Summary

Fixes a race condition in the LFS shutdown path and hardens the interrupt-and-resume test coverage for the HF model pull module.

Production fix:

Adds clearStaleErrorFile() in libgit2.cpp to remove stale lfs_error.txt artifacts left by a previous interrupted pull. Without this, a false error file caused subsequent resume attempts to misidentify the repository state and fail.
Adds a shutdown check inside the LFS write callback so that an in-progress LFS download honours a cancellation request promptly rather than completing the current chunk before noticing the shutdown.

Test improvements:

Refactors HfPull and HfPullCache fixtures to centralize repeated model path setup into shared members, eliminating per-test duplication.
Removes the deterministic fallback helper that masked the real SIGKILL / SIGINT code paths; both ResumeTerminate and ResumeCtrlC now require observing genuine in-progress LFS activity before signalling.
Replaces fragile fixed-sleep interruption gating with a polling loop that checks for main plus LFS artifacts or an in-flight model file, making the tests reliable across both fast and slow network conditions.

### 🧪 Checklist

- [ ] Unit tests added.
``


ngrozae (approved):

third_party/libgit2/lfs.patch
```diff
@@ -2258,30 +2310,37 @@ index 000000000..c0e97d408
...
-+		if (p_unlink(la->full_path) < 0 && errno != ENOENT) {
-+			lfs_log_error_with_state(
-+			        &la->log_state,
-+			        "\n[ERROR] failed to delete existing destination file '%s' (errno=%d: %s)\n",
-+			        la->full_path, errno, strerror(errno));
-+			/* Ignore here and let rename produce the definitive failure. */
-+		}
-+	}
-+
 +	if (p_rename(tmp_out_file, la->full_path) < 0) {
```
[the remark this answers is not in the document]
rasapala: We use safe windows implementation.

src/pull_module/libgit2.cpp
```diff
@@ -1193,6 +1193,10 @@ Status resumeExistingRepository(git_repository* repo,
         SPDLOG_WARN("Failed to create .lfswip marker before pull resume at {}", libgit2::getLfsWipMarkerPath(downloadPath).string());
     }
 
+    // Clear any stale lfs_error.txt from previous failed attempts to avoid false error reports
+    std::error_code ec;
+    fs::remove(fs::path(downloadPath) / "lfs_error.txt", ec);
```
mzegla: Is it safe if the file does not exist?

third_party/libgit2/lfs.patch
```diff
@@ -424,10 +424,10 @@ index 58cb4b424..00ddee9f3 100644
...
 diff --git a/src/libgit2/lfs_filter.c b/src/libgit2/lfs_filter.c
 new file mode 100644
-index 000000000..c0e97d408
+index 000000000..794a22bba
 --- /dev/null
 +++ b/src/libgit2/lfs_filter.c
-@@ -0,0 +1,1952 @@
+@@ -0,0 +1,2009 @@
 +﻿/*
 +/ Copyright 2025 Intel Corporation
```
mzegla, line 432: Shouldn't be `2026` copyright?

third_party/libgit2/lfs.patch
```diff
@@ -1046,6 +1046,58 @@ index 000000000..c0e97d408
...
 +}
 +
++static const char *lfs_last_path_sep(const char *path)
++{
++	const char *slash = strrchr(path, '/');
++
++#ifdef _WIN32
++	const char *backslash = strrchr(path, '\\');
++	if (!slash || (backslash && backslash > slash))
++		slash = backslash;
```
mzegla: Ambiguous naming in my opinion. `slash` and `backslash` could be replaced with something like `last_sep_position`

third_party/libgit2/lfs.patch
```diff
@@ -1046,6 +1046,58 @@ index 000000000..c0e97d408
...
++	if (sep == path)
++		return (p_stat("/", &st) == 0 && S_ISDIR(st.st_mode));
++
++	dir_len = (size_t)(sep - path);
++
++#ifdef _WIN32
++	/* For drive-root paths like "C:\\file", keep the trailing separator so
++	 * p_stat checks "C:\\" instead of "C:". */
++	if (dir_len == 2 && isalpha((unsigned char)path[0]) && path[1] == ':' &&
++	    (*sep == '/' || *sep == '\\'))
```
mzegla: Isn't last part `(*sep == '/' || *sep == '\\')` already guaranteed by `lfs_last_path_sep` ?

third_party/libgit2/lfs.patch
```diff
@@ -1955,7 +2007,7 @@ index 000000000..c0e97d408
 +			return CURLE_OK;
 +		}
 +
-+		lfs_log_error_with_state(ftpfile->log_state, "[WARN] Resume attempt %d/%d failed: %s\n",
++		printf("[WARN] Resume attempt %d/%d failed: %s\n",
```
mzegla, line 2015: Why `printf` and not some built in logging function? Isn't there any?

src/pull_module/libgit2.cpp
rasapala: Yes its ok. But I wil ladd file exists first, beacause on some VS code 2022 implementations it might throw.

third_party/libgit2/lfs.patch
rasapala: Not part of this change.

third_party/libgit2/lfs.patch
rasapala: There are but those are messages we want to always see and the git_trace uses different values of level logging. I do not want to debug and this functionality in this PR.

third_party/libgit2/lfs.patch
rasapala: No the file is from 2025.

third_party/libgit2/lfs.patch
mzegla: What do you mean? GitHub shows this part as a new code added in this PR.
Was it copied from somewhere that I don't see?

third_party/libgit2/lfs.patch
mzegla: So regular stdout printing is our only option here?

third_party/libgit2/lfs.patch
rasapala: Ok I missed that, fixed.

src/test/pull_hf_model_test.cpp
```diff
@@ -654,6 +633,11 @@ TEST_F(HfPullCache, PullNonGit) {
     ASSERT_EQ(ec, std::errc()) << "Failed to remove .git from cached repository: " << ec.message();
     ASSERT_FALSE(std::filesystem::exists(gitDir));
 
+#ifdef _WIN32
+    // On Windows, gtest stdout capture can conflict with SPDLOG stdout sink.
```
atobiszei: We could use log file with ovms to check, unless sth we expect does not appear in spdlong but only in stdout?

src/test/pull_hf_model_test.cpp
```diff
@@ -1079,7 +1067,8 @@ TEST_F(HfPull, ResumeCtrlC) {
     EXPECT_NE(WEXITSTATUS(childStatus), EXIT_SUCCESS);
 #endif
 
-    EXPECT_TRUE(observedPartialDownload);
+    SPDLOG_INFO("ResumeCtrlC test state: observedPartialDownload={}, interruptionSent={}",
```
atobiszei: DEBUG?

src/test/pull_hf_model_test.cpp
```diff
@@ -423,13 +428,6 @@ TEST_F(HfPull, ResumeShutdown) {
 #ifdef _WIN32
     SKIP_AND_EXIT_IF_NOT_RUNNING_UNSTABLE();
 #endif
```
atobiszei: Left skipped?

src/test/pull_hf_model_test.cpp
atobiszei: What are those env variables? Why are they needed?
Also couldn't we have static utility funcitons that would do proper sending signals in tests? Now test logic is containing both setup how the initial state is prepared, then details how to properly send signals depending on OS.

Its not easily readable at first glance what is actual test logic, what is preparation, what is mechanism to simulate ctrl c etc.

Also I see magic constants:
    ASSERT_EQ(std::filesystem::file_size(modelPath), 52417240);

have const size_t MAIN_MODEL_SIZE
TOKENIZER_MODEL_SIZE

src/test/pull_hf_model_test.cpp
```diff
@@ -849,48 +819,50 @@ TEST_F(HfPull, ResumeTerminate) {
...
-    auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(60);
+    bool interruptionSent = false;
+    const std::string mainRefPath = ovms::FileSystem::appendSlash(modelBasePath) + ".git/refs/heads/main";
+    auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(180);
     while (std::chrono::steady_clock::now() < deadline) {
+        // Wait until pull is in progress and repository is already resumable.
+        std::error_code ec;
+        const bool hasMainRef = std::filesystem::exists(mainRefPath, ec);
+        ec.clear();
+        const bool modelExists = std::filesystem::exists(modelPath, ec);
```
mzegla: no `ec` check after this call?

src/test/pull_hf_model_test.cpp
```diff
@@ -849,48 +819,50 @@ TEST_F(HfPull, ResumeTerminate) {
...
     bool observedPartialDownload = false;
-    auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(60);
+    bool interruptionSent = false;
+    const std::string mainRefPath = ovms::FileSystem::appendSlash(modelBasePath) + ".git/refs/heads/main";
+    auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(180);
     while (std::chrono::steady_clock::now() < deadline) {
+        // Wait until pull is in progress and repository is already resumable.
+        std::error_code ec;
+        const bool hasMainRef = std::filesystem::exists(mainRefPath, ec);
+        ec.clear();
```
mzegla: error code cleared without checking? why is it even cleared? won't the last call before check from line 841 override it anyway before check?

src/test/pull_hf_model_test.cpp
```diff
@@ -849,48 +819,50 @@ TEST_F(HfPull, ResumeTerminate) {
...
+            }
+        }
         auto lfsCandidates = ovms::libgit2::findLfsLikeFiles(downloadPath, true);
-        auto hasOpenvinoModelPointer = std::find_if(lfsCandidates.begin(), lfsCandidates.end(),
-                                           [](const std::filesystem::path& p) { return p.filename() == "openvino_model.bin"; }) != lfsCandidates.end();
-        const std::string partPath = ovms::FileSystem::appendSlash(basePath) + "openvino_model.binlfs_part";
-        const bool hasPartFile = std::filesystem::exists(partPath);
-        if (hasOpenvinoModelPointer || hasPartFile) {
+        const bool hasLfsArtifacts = !lfsCandidates.empty();
+        const bool modelInFlight = modelExists && (modelSize > 0) && (modelSize < 52417240);
```
mzegla: hardcoded limits - why this upper bound has been chosen?
please move to const

src/test/pull_hf_model_test.cpp
```diff
@@ -849,48 +819,50 @@ TEST_F(HfPull, ResumeTerminate) {
...
     }
 
+    ASSERT_TRUE(observedPartialDownload)
+        << "Did not observe in-progress pull within timeout, cannot verify SIGKILL interruption path.";
+
 #ifdef _WIN32
     ASSERT_TRUE(TerminateProcess(pi.hProcess, 1));
-    ASSERT_EQ(WaitForSingleObject(pi.hProcess, 10000), WAIT_OBJECT_0);
+    interruptionSent = true;
+    ASSERT_EQ(WaitForSingleObject(pi.hProcess, 120000), WAIT_OBJECT_0);
```
mzegla: move value to const variable so we know the context

src/test/pull_hf_model_test.cpp
```diff
@@ -1031,33 +1003,48 @@ TEST_F(HfPull, ResumeCtrlC) {
...
+        ec.clear();
+        const bool modelExists = std::filesystem::exists(modelPath, ec);
+        std::uintmax_t modelSize = 0;
+        if (modelExists) {
+            ec.clear();
+            modelSize = std::filesystem::file_size(modelPath, ec);
+            if (ec) {
+                modelSize = 0;
+            }
+        }
```
mzegla: Duplicate of part of logic above (same comments apply here as well). 
Could we have it in a function?

src/test/pull_hf_model_test.cpp
```diff
@@ -1031,33 +1003,48 @@ TEST_F(HfPull, ResumeCtrlC) {
...
         auto lfsCandidates = ovms::libgit2::findLfsLikeFiles(downloadPath, true);
-        const bool hasOpenvinoModelPointer = std::find_if(lfsCandidates.begin(), lfsCandidates.end(),
-                                                 [](const std::filesystem::path& p) { return p.filename() == "openvino_model.bin"; }) != lfsCandidates.end();
-        const std::string partPath = ovms::FileSystem::appendSlash(basePath) + "openvino_model.binlfs_part";
-        const bool hasPartFile = std::filesystem::exists(partPath);
-        if (hasOpenvinoModelPointer || hasPartFile) {
+        const bool hasLfsArtifacts = !lfsCandidates.empty();
+        const bool modelInFlight = modelExists && (modelSize > 0) && (modelSize < 52417240);
```
mzegla: duplicate use of hardcoded values

third_party/libgit2/lfs.patch
rasapala: We can switch when we productize the patch for upstream.

src/test/pull_hf_model_test.cpp
rasapala: Logger configuration is process-global and initialized earlier in the test
binary, so a per-test --log_path does not reliably capture this warning.

src/test/test_utils.cpp
```diff
@@ -745,11 +760,37 @@ void EnsureServerStartedWithTimeout(ovms::Server& server, int timeoutSeconds) {
...
+    auto elapsed = [&] {
+        return std::chrono::duration_cast<std::chrono::seconds>(
+            std::chrono::high_resolution_clock::now() - start)
+            .count();
+    };
+
+    // Phase 1: yield until the HF pull module appears in the server's module map.
+    // std::this_thread::yield() is required — a bare busy-spin starves the server thread
+    // (especially on machines with few cores or under load) and the module never starts.
+    while (server.getModuleState(ovms::HF_MODEL_PULL_MODULE_NAME) == ovms::ModuleState::NOT_INITIALIZED &&
```
atobiszei: You use timeout seconds twice so it effectively becomes x2. Either add 2nd parameter or subtract from first time.

atobiszei: Also add yield + micro sleep,t o avoid busy wait.

src/test/pull_hf_model_test.cpp
```diff
@@ -61,29 +62,439 @@
...
+    if (!probeError || probeError == std::errc::no_such_file_or_directory) {
+        tracker.consecutiveErrors = 0;
+        tracker.lastError.clear();
+        tracker.lastOperation.clear();
+        tracker.lastPath.clear();
+        return false;
+    }
+
+    if ((probeError == tracker.lastError) && (tracker.lastOperation == operation) && (tracker.lastPath == path)) {
+        ++tracker.consecutiveErrors;
```
atobiszei, line 93: Since we now moved this out of single test, maybe use more meaningful names like detokenizerModelPath instead of model2Path?

atobiszei: @rasapala

src/test/test_utils.cpp
```diff
@@ -769,7 +816,24 @@ void SetUpServerForDownload(std::unique_ptr<std::thread>& t, ovms::Server& serve
         EXPECT_EQ(expected_code, server.start(argc, argv));
     }));
 
-    EnsureServerModelDownloadFinishedWithTimeout(server, timeoutSeconds);
+    EnsureServerModelDownloadFinishedWithTimeout(server, timeoutSeconds, timeoutSeconds);
+}
+
+void SetUpServerForDownload(std::unique_ptr<std::thread>& t, ovms::Server& server, std::string& source_model, std::string& download_path, std::string& task, const std::string& logPath, int expected_code, int timeoutSeconds) {
```
atobiszei, line 826: User of this method passes 1 timeout -then its 2x that timeout. Either use 2nd timeout field or subtract. I am not sure why why need 2 timeouts here.

src/test/test_utils.cpp
```diff
@@ -743,13 +758,45 @@ void EnsureServerStartedWithTimeout(ovms::Server& server, int timeoutSeconds) {
...
+    while (state != ovms::ModuleState::SHUTDOWN &&
+           state != ovms::ModuleState::NOT_INITIALIZED &&
+           (std::chrono::high_resolution_clock::now() - phase2Start) < completionTimeout) {
+        std::this_thread::yield();
+        std::this_thread::sleep_for(pollInterval);
+        state = server.getModuleState(ovms::HF_MODEL_PULL_MODULE_NAME);
+    }
+
+    const auto finalState = state;
+    ASSERT_TRUE(finalState == ovms::ModuleState::SHUTDOWN || finalState == ovms::ModuleState::NOT_INITIALIZED)
```
atobiszei, line 802:
```suggestion
	EXPECT_NE(ovms::ModuleState::NOT_INITIALIZED, finalState) << "HF pull module never started within " << startupTimeoutSeconds << "s. Check machine load and network load";
    ASSERT_TRUE(finalState == ovms::ModuleState::SHUTDOWN || finalState == ovms::ModuleState::NOT_INITIALIZED)
```
wouldn';t that eliminate need for first loop? Unless we know that first timeout can be more aggresive (much shorter one).

rasapala: Two whiles needed for tests stability and failure root causing.

src/test/test_utils.cpp
```diff
@@ -758,8 +758,8 @@ void EnsureServerStartedWithTimeout(ovms::Server& server, int timeoutSeconds) {
     ASSERT_EQ(server.getModuleState(ovms::SERVABLE_MANAGER_MODULE_NAME), ovms::ModuleState::INITIALIZED) << "OVMS did not fully load until allowed time:" << timeoutSeconds << "s. Check machine load";
 }
 
-void EnsureServerModelDownloadFinishedWithTimeout(ovms::Server& server, int startupTimeoutSeconds, int completionTimeoutSeconds) {
-    const auto startupTimeout = std::chrono::seconds(startupTimeoutSeconds);
+void EnsureServerModelDownloadFinishedWithTimeout(ovms::Server& server, int completionTimeoutSeconds) {
+    const auto startupTimeout = std::chrono::seconds(10);
```
atobiszei, line 762: Not it is 10s+completionTimeoutSeconds

atobiszei (approved):

PR #4235, opened 2026-05-22
VLM chat aplication via python jinja / graph minja/jinja switch

dkalinowski:
This change:
- switches chat template application from GenAI to Python's Jinja for VLM & VLM_CB pipelines.
- allows runtime MINJA/JINJA switch based on graph.pbtxt parameter

MMMU_VAL benchmark results

| model | current application via minja | chat application via python jinja | llama.cpp |
| --- | --- | --- | --- |
| Qwen3-VL-8B-Instruct | 0.51 | 0.5033 | 0.5122 |
| Phi-3.5-vision-instruct-int8-ov | 0.4322 | 0.4333 | |
| Qwen3.6-35B-A3B-int4-ov | 0.226 | 0.2567 | | 

BFCL with new chat templates (multi-turn)

| model | current application via minja | chat application via python jinja |
| --- | --- | --- | 
| Qwen3-VL-8B-Instruct | 0.53 | 0.515 |
| Qwen3.6-35B-A3B-int4-ov | 0.226 | TBD  |
| Qwen3.5-35B-A3B-int4-ov | ? |  TBD |


src/llm/visual_language_model/continuous_batching/servable.cpp
```diff
@@ -22,6 +22,16 @@
...
+#pragma warning(push)
+#pragma warning(disable : 4005 4309 6001 6385 6386 6326 6011 4005 4456 6246)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#include <rapidjson/document.h>
+#include <rapidjson/stringbuffer.h>
+#include <rapidjson/writer.h>
+#pragma GCC diagnostic pop
+#pragma warning(pop)
+
```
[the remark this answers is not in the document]
dkalinowski: fixed

src/llm/visual_language_model/continuous_batching/servable.cpp
```diff
@@ -94,6 +104,40 @@ absl::Status VisualLanguageModelServable::prepareInputs(std::shared_ptr<GenAiSer
...
+        // Inject image tags into the JSON messages for Python Jinja template processing
+        if (!imageTags.empty()) {
+            rapidjson::Document jsonDoc;
+            jsonDoc.Parse(jsonForTemplate.c_str());
+            if (!jsonDoc.HasParseError() && jsonDoc.HasMember("messages") && jsonDoc["messages"].IsArray()) {
+                auto& messages = jsonDoc["messages"];
+                for (const auto& [chatTurnIndex, imageTagString] : imageTags) {
+                    if (chatTurnIndex < messages.Size()) {
+                        auto& msg = messages[chatTurnIndex];
+                        if (msg.HasMember("content") && msg["content"].IsString()) {
```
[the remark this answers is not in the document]
dkalinowski: fixed

src/llm/llm_calculator.proto
```diff
@@ -137,4 +137,15 @@ message LLMCalculatorOptions {
     optional SparseAttentionConfig sparse_attention_config = 24;
 
     optional uint64 cache_interval_multiplier = 25;
+
+    enum ChatTemplateMode {
+      // Use GenAI's apply_chat_template (minja-based).
+      // Default for builds without Python. Also selectable in Python builds.
```
dtrawins: Without python it is the only option, not default

src/llm/llm_calculator.proto
```diff
@@ -137,4 +137,15 @@ message LLMCalculatorOptions {
...
 
     optional uint64 cache_interval_multiplier = 25;
+
+    enum ChatTemplateMode {
+      // Use GenAI's apply_chat_template (minja-based).
+      // Default for builds without Python. Also selectable in Python builds.
+      MINJA = 0;
+      // Use Python Jinja2 module for chat template processing.
+      // Default for builds with Python enabled.
+      JINJA = 1;
```
dtrawins, line 149: Till we have complete refactor for chat template processing, maybe we should have a logic to use by default Minja for VLM and Jinja for LLM? I guess it would be 100% compatible with existing chat templates.

src/llm/visual_language_model/continuous_batching/servable.cpp
```diff
@@ -95,26 +102,72 @@ absl::Status VisualLanguageModelServable::prepareInputs(std::shared_ptr<GenAiSer
...
+                if (!jsonDoc.HasParseError() && jsonDoc.IsObject() && jsonDoc.HasMember("messages") && jsonDoc["messages"].IsArray()) {
+                    auto& messages = jsonDoc["messages"];
+                    for (const auto& [chatTurnIndex, imageTagString] : imageTags) {
+                        if (chatTurnIndex < messages.Size()) {
+                            auto& msg = messages[chatTurnIndex];
+                            if (msg.IsObject() && msg.HasMember("content") && msg["content"].IsString()) {
+                                std::string newContent = imageTagString + msg["content"].GetString();
+                                msg["content"].SetString(newContent.c_str(), newContent.length(), jsonDoc.GetAllocator());
+                            }
+                        }
```
mzegla, line 126: Two things here:
1. Don't we break ordering here? It might not be visible in accuracy checks but don't we reverse tags here? For example if we have 3 images in the same chat turn - [ov_genai_image_1, ov_genai_image_2, ov_genai_image_3], then at the end of the loop new content would be `<ov_genai_image_3><ov_genai_image_2><ov_genai_image_1>content...`. Unless imageTagString has that concatenated already...
2. In such case we also set string in the JSON multiple times. If the content is large it might generate unnecessary overhead - maybe we could optimize it? Again, unless it's concatenated before.

src/llm/visual_language_model/continuous_batching/servable.cpp
```diff
@@ -95,26 +102,72 @@ absl::Status VisualLanguageModelServable::prepareInputs(std::shared_ptr<GenAiSer
...
+            }
+        } else
+#endif
+        {
+            constexpr bool addGenerationPrompt = true;  // confirm it should be hardcoded
+            auto toolsStatus = vlmExecutionContext->apiHandler->parseToolsToJsonContainer();
+            if (!toolsStatus.ok()) {
+                return toolsStatus.status();
+            }
+            const auto& tools = toolsStatus.value();
```
mzegla: `toolsStatus.value()` giving real tools does not look good to me. 
maybe something like `toolParsingResult`? then both `ok()` and `value()` looks good in my opinion

src/llm/visual_language_model/continuous_batching/servable.cpp
```diff
@@ -95,26 +102,72 @@ absl::Status VisualLanguageModelServable::prepareInputs(std::shared_ptr<GenAiSer
...
+            auto chatTemplateKwargsStatus = vlmExecutionContext->apiHandler->parseChatTemplateKwargsToJsonContainer();
+            if (!chatTemplateKwargsStatus.ok()) {
+                return chatTemplateKwargsStatus.status();
+            }
+            const auto& chatTemplateKwargs = chatTemplateKwargsStatus.value();
```
mzegla: same as for tools

src/llm/llm_calculator.proto
```diff
@@ -137,4 +137,15 @@ message LLMCalculatorOptions {
     optional SparseAttentionConfig sparse_attention_config = 24;
 
     optional uint64 cache_interval_multiplier = 25;
+
+    enum ChatTemplateMode {
```
mzegla, line 141: I know it has been discussed but just to note - I don't like it. 
Honestly, if we have to expose this I think I would do it in `generation_config.json`...
But I will not block the merge because of this.

src/llm/servable.cpp
```diff
@@ -188,53 +188,21 @@ absl::Status GenAiServable::prepareInputs(std::shared_ptr<GenAiServableExecution
     switch (executionContext->endpoint) {
     case Endpoint::CHAT_COMPLETIONS: {
 #if (PYTHON_DISABLE == 0)
-        bool success;
```
mzegla, line 191: Is it fine to remove this whole part from here? This servable has logic for all types of pipelines. Won't we lose some functionality for non visual language pipelines?

src/llm/visual_language_model/continuous_batching/servable.cpp
mzegla: Same applies to legacy

src/llm/llm_calculator.proto
dkalinowski: This is very good idea

src/llm/visual_language_model/continuous_batching/servable.cpp
dkalinowski: The ordering issue does not occur because imageTagString has already multiple tags concatenated beforehand.
I agree that this adds overhead - again parsing json and serializing json. I think it would be good idea to re-organize everything so that rapidjson parsing happens once, I think this is already being made in refactoring PR?

For now, this path is not the default one anyway, since MINJA is default for VLM

src/llm/servable.cpp
dkalinowski: We remove alot but we also add alot, the only thing that differs here, is that we add support for `getProperties()->chatTemplateMode` - for both CHAT_COMPLETIONS switch and RESPONSES switch

dtrawins (approved):

mzegla (approved):

PR #4271, opened 2026-06-02
Remove stateful model feature support

atobiszei:
- Removed stateful model instance, sequence manager, and all related code
- Removed CLI options: --stateful, --idle_sequence_cleanup, --low_latency_transformation, --max_sequence_number, --sequence_cleaner_poll_wait_minutes
- Removed config parameters: stateful, idle_sequence_cleanup, low_latency_transformation, max_sequence_number
- Kept CAPI function OVMS_ServerSettingsSetSequenceCleanerPollWaitMinutes as deprecated no-op returning NOT_IMPLEMENTED
- Removed docs/stateful_models.md and all documentation references
- Updated tests to remove stateful test cases and fixtures


rasapala (approved):

docs/legacy.md
```diff
@@ -5,15 +5,9 @@
 maxdepth: 1
 hidden:
 ---
-ovms_docs_stateful_models
 ovms_docs_dag
 ```
 
-## Stateful models
-Implement any CPU layer, that is not supported by OpenVINO yet, as a shared library.
```
dkalinowski, line 13: This relates to Custom Op in OpenVINO, not stateful models. Is it mistake?

src/test/c_api_tests.cpp
```diff
@@ -127,7 +122,6 @@ TEST(CAPIConfigTest, MultiModelConfiguration) {
     ASSERT_CAPI_STATUS_NULL(OVMS_ServerSettingsSetGrpcMaxThreads(_serverSettings, 100));
     ASSERT_CAPI_STATUS_NULL(OVMS_ServerSettingsSetGrpcMemoryQuota(_serverSettings, (size_t)1000000));
     ASSERT_CAPI_STATUS_NULL(OVMS_ServerSettingsSetFileSystemPollWaitSeconds(_serverSettings, 2));
-    ASSERT_CAPI_STATUS_NULL(OVMS_ServerSettingsSetSequenceCleanerPollWaitMinutes(_serverSettings, 3));
```
dkalinowski, line 130: I think its missing unit test for new behavior, NOT_IMPLEMENTED

src/http_server.cpp
```diff
@@ -108,15 +108,6 @@ static const ovms::HTTPStatusCode http(const ovms::Status& status) {
         {StatusCode::CANNOT_COMPILE_MODEL_INTO_TARGET_DEVICE, ovms::HTTPStatusCode::PRECOND_FAILED},
 
         // Sequence management
```
dkalinowski:
```suggestion
```

src/modelinstance.cpp
```diff
@@ -67,12 +68,6 @@
 #include "vaapitensorfactory.hpp"
 #endif
 
-#ifdef _WIN32
-namespace ovms {
-bool malloc_trim_win();
```
dkalinowski, line 73: why are you removing malloc trim?

src/status.hpp
```diff
@@ -79,22 +79,22 @@ enum class StatusCode {
...
-    SEQUENCE_ALREADY_EXISTS,         /*!< Sequence with provided ID already exists */
-    SEQUENCE_ID_NOT_PROVIDED,        /*!< Sequence ID has not been provided in request inputs */
-    SEQUENCE_ID_BAD_TYPE,            /*!< Wrong sequence ID type */
-    INVALID_SEQUENCE_CONTROL_INPUT,  /*!< Unexpected value of sequence control input */
-    SEQUENCE_CONTROL_INPUT_BAD_TYPE, /*!< Sequence control input in bad type */
-    SEQUENCE_TERMINATED,             /*!< Sequence last request is being processed and it's not available anymore */
-    SPECIAL_INPUT_NO_TENSOR_SHAPE,   /*!< Special input proto does not contain tensor shape information */
-    MAX_SEQUENCE_NUMBER_REACHED,     /*!< Model handles maximum number of sequences and will not accept new ones */
+    REQUESTED_DYNAMIC_PARAMETERS_ON_STATEFUL_MODEL,    /*!< Deprecated/reserved to preserve StatusCode numeric values */
+    REQUESTED_STATEFUL_PARAMETERS_ON_SUBSCRIBED_MODEL, /*!< Deprecated/reserved to preserve StatusCode numeric values */
```
dkalinowski: why do we need to preserve numeric values?

dkalinowski: is it due to C-API? I dont think we adhered to it in the past

dkalinowski: Example: https://github.com/openvinotoolkit/model_server/commit/8b8400a67ba5759d75ea17f880fa92a2d95608ef

but probably many other

docs/legacy.md
atobiszei: And it references stateful_models.md. I think this was mistake from the begining

src/modelinstance.cpp
atobiszei: Its just declaration. Definition still exists in cleaner utils

src/status.hpp
atobiszei: Yes it is due to C-API to preserve numeric values.

dkalinowski (approved):

src/status.hpp
atobiszei: Since some status were already removed in between releases we decided to remove those as well.

PR #4295, opened 2026-06-17
Gorilla BFCL patch update to automatize root cause detection and divergence in output between servings

dkalinowski:


dtrawins:
rename this patch as gorilla_debug.patch - that would be optional one for advanced debugging purposes

PR #4307, opened 2026-06-22
Reasoning and tool parser for LFM2.5 

przepeck:
### 🛠 Summary

[CVS-188930](https://jira.devtools.intel.com/browse/CVS-188930)
Reasoning Parser for LFM2.5 
### 🧪 Checklist

- [ ] Unit tests added.
``


mzegla (commented):
I think we mention available parsers in the docs. Please update documentation as well.

src/llm/io_processing/lfm2/lfm25_tool_parser.hpp
```diff
@@ -0,0 +1,46 @@
...
+//*****************************************************************************
+
+#include "./lfm2_tool_parser.hpp"
+
+namespace ovms {
+class Lfm25ToolParser : public Lfm2ToolParser {
+protected:
+    static const int64_t toolCallStartTokenId = 124905; // <|tool_call_start|>
+    static const int64_t toolCallEndTokenId = 124906;   // <|tool_call_end|>
+    static const int64_t reasoningEndTokenId = 124902;   // </think>
```
mzegla: Why do we need it? Feels like we are leaking responsibilities here.

src/llm/io_processing/lfm2/lfm2_tool_parser.hpp
```diff
@@ -51,8 +52,8 @@ class Lfm2ToolParser : public BaseOutputParser {
         std::string value;
     };
     Lfm2ToolParser() = delete;
-    explicit Lfm2ToolParser(ov::genai::Tokenizer& tokenizer) :
-        BaseOutputParser(tokenizer) {}
+    explicit Lfm2ToolParser(ov::genai::Tokenizer& tokenizer, int64_t botTokenId = 10, int64_t eotTokenId = 11) :
+        BaseOutputParser(tokenizer), botTokenId(botTokenId), eotTokenId(eotTokenId) {}
```
mzegla: We are bending the LFM2 parser beyond the standard with those changes. Now some internal consts are static and some are not + the constructor is different. Not in line with other parsers. 
For now I see two possibilities: 
- have full implementation for LFM2.5, maybe we some extracted functions shared with LFM2, but not exact calls from parent
- make such change for all parsers (drop static fields, accept reading it from somewhere instead of hardcoded in the class)

I think option 2 would be worth considering during refactor, but since it would likely require broad correctness checks it might be quicker to go with option 1 for now.

przepeck: I will go with first option then

src/llm/io_processing/lfm2/lfm25_tool_parser.hpp
przepeck: As I did in gemma's parser, when I was using only tokens in unary I needed to remove reasoning from content at the end of the process. Base output parser's parse() doesn't allow us changing generatedTokens vector.

extras/chat_template_examples/chat_template_lfm25.jinja
```diff
@@ -0,0 +1,120 @@
+{#-
+  Modifications to original chat template:
```
dkalinowski: did you test it both: with jinja and minja? @przepeck  the new requirement is to have all new models supported in both ways. @dtrawins 

to test it, change `chat_template_mode` to either JINJA/MINJA in graph.pbtxt

przepeck: this template won't be working with minja, it uses from_json which is not supported, but I think this PR will be merged after workarounds auto detection, that means I can remove this template from the PR

src/llm/io_processing/lfm2/lfm25_tool_parser.cpp
```diff
@@ -0,0 +1,108 @@
...
+    }
+    case State::ToolCallStarted: {
+        auto wasParsedCorrectly = parseInToolCallState(this->streamingContent, this->toolCall, this->streamingPosition, this->currentState);
+        if (wasParsedCorrectly) {
+            this->toolCallIndex++;
+        }
+        return wasParsedCorrectly;
+    }
+    case State::ToolCallParameters: {
+        return parseToolCallParametersState(this->streamingContent, this->toolCall, this->streamingPosition, this->currentState);
```
mzegla: Very minor, but there is a naming inconsistency here. Other methods start with `parseInXXXState`, but this one is `parseXXXState`. 
It is obvious when read in the context like here in this switch, but out of context it suggest that some tool call parameters state will be parsed to something.

src/llm/io_processing/output_parser.cpp
```diff
@@ -193,7 +195,14 @@ OutputParser::OutputParser(ov::genai::Tokenizer& tokenizer, const std::string to
...
     } else if (toolParserName == "lfm2") {
-        toolParser = std::make_unique<Lfm2ToolParser>(tokenizer);
+        auto vocab = tokenizer.get_vocab();
+        auto token = vocab.find(Lfm25ToolParser::TOOL_CALL_START_TAG);
+        auto tokenId = token != vocab.end() ? token->second : -1;
+        if (tokenId == Lfm25ToolParser::toolCallStartTokenId) {
+            toolParser = std::make_unique<Lfm25ToolParser>(tokenizer);
+        } else {
+            toolParser = std::make_unique<Lfm2ToolParser>(tokenizer);
+        }
```
mzegla, line 207: Smooth selection 😄 but hopefully we can drop it after refactor and unify both parsers

mzegla (approved):

src/llm/io_processing/lfm2/lfm25_tool_parser.cpp
przepeck: done

przepeck:
Current result on GPU for LFM2.5 with reasoning:
image

przepeck:
image

accuracy with preserve_thinking=true and workaround applied

demos/common/export_models/export_model.py
```diff
@@ -54,8 +54,8 @@ def add_common_arguments(parser):
 parser_text.add_argument('--max_prompt_len', required=False, type=int, default=None, help='Sets NPU specific property for maximum number of tokens in the prompt. '
                          'Not effective if target device is not NPU', dest='max_prompt_len')
 parser_text.add_argument('--prompt_lookup_decoding', action='store_true', help='Set pipeline to use prompt lookup decoding', dest='prompt_lookup_decoding')
-parser_text.add_argument('--reasoning_parser', choices=["qwen3", "gptoss"], help='Set the type of the reasoning parser for reasoning content extraction', dest='reasoning_parser')
-parser_text.add_argument('--tool_parser', choices=["llama3", "phi4", "hermes3", "mistral", "qwen3coder", "gptoss", "devstral", "lfm2"], help='Set the type of the tool parser for tool calls extraction', dest='tool_parser')
+parser_text.add_argument('--reasoning_parser', choices=["qwen3", "gptoss", "lfm2.5", "gemma4"], help='Set the type of the reasoning parser for reasoning content extraction', dest='reasoning_parser')
```
dtrawins: for consistency it might be better to name reasoning and tool parsers with the same name lfm2

przepeck: done

src/test/llm/chat_template_end_to_end_minja_test.cpp
```diff
@@ -388,13 +399,35 @@ TEST_F(ChatTemplateEndToEndMinjaTest, LFM2_ToolCallWithStringArgs) {
...
+    EXPECT_FALSE(caps.supportsToolCalls);
+    EXPECT_FALSE(caps.requiresObjectArguments);
+    EXPECT_TRUE(caps.missnamedReasoningField.empty());
+
+    std::string expectedOutput = R"(</s><|im_start|>user
+What's the weather in Paris?<|im_end|>
+<|im_start|>assistant
+{
+  "tool_calls": [
+    {
```
dkalinowski, line 420: this is minja polyfill, maybe we can fix that, same as we did with lfm2.5?

dkalinowski: im not sure whats the point of supporting Lfm2 tool parser if we dont have any working scenario with this model

src/test/llm/chat_template_analyzer_test.cpp
```diff
@@ -204,4 +206,5 @@ TEST_F(ChatTemplateAnalyzerTest, defaultCapsValues) {
     ChatTemplateCaps caps;
     EXPECT_FALSE(caps.supportsToolCalls);
     EXPECT_FALSE(caps.requiresObjectArguments);
+    EXPECT_TRUE(caps.missnamedReasoningField.empty());
```
dkalinowski, line 209: add it to other tests too?

src/test/llm/chat_template_end_to_end_minja_test.cpp
przepeck: I think we could, but in another PR, I will probably fix chat template for this model, but it's prority is low

src/test/llm/chat_templates/chat_template_lfm25.jinja
```diff
@@ -34,11 +38,15 @@
     {%- for tool_call in tool_calls -%}
         {%- set func_name = tool_call["function"]["name"] -%}
         {%- set func_args = tool_call["function"]["arguments"] -%}
-        {%- set args_ns = namespace(arg_strings=[]) -%}
-        {%- for arg_name, arg_value in func_args.items() -%}
-            {%- set args_ns.arg_strings = args_ns.arg_strings + [arg_name + "=" + format_arg_value(arg_value)] -%}
-        {%- endfor -%}
-        {%- set tool_calls_ns.tool_calls = tool_calls_ns.tool_calls + [func_name + "(" + (args_ns.arg_strings | join(", ")) + ")"] -%}
+        {%- if func_args is string -%}
+            {%- set tool_calls_ns.tool_calls = tool_calls_ns.tool_calls + [func_name + "(" + func_args + ")"] -%}
```
dkalinowski, line 42: put this chat template in extras directory as well, next to other chat templates 
please also ensure this template is published to OV HF together with the model - @przepeck  @skuros

src/test/llm/chat_template_analyzer_test.cpp
przepeck: done

src/test/llm/chat_templates/chat_template_lfm25.jinja
przepeck: done

przepeck:
current bfcl results:
image

dkalinowski (approved):

src/llm/io_processing/chat_template/probe.cpp
```diff
@@ -129,6 +132,56 @@ static bool analyzeProbeResults(bool strOk, const std::string& strOut,
     return true;
 }
 
+bool probeChatTemplateReasoning(ov::genai::Tokenizer& tokenizer, ChatTemplateCaps& caps) {
```
mzegla, line 135: General question about probing I guess. What is the interpretation of the return value? 
I wonder why this returns `false`:
`SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Reasoning probe: '{}' not supported", field);`

While this returns `true`:
`SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Reasoning probe: no reasoning field supported by template");`

src/llm/io_processing/input_processors/chat_template_adapter.cpp
```diff
@@ -60,11 +60,27 @@ void funcArgsToObjectHistory(ov::genai::ChatHistory& chatHistory) {
...
+void injectReasoningIntoMissnamedSection(ov::genai::ChatHistory& chatHistory, const std::string& templateReasoningFieldName) {
+    for (size_t msgIdx = 0; msgIdx < chatHistory.size(); ++msgIdx) {
+        auto message = chatHistory[msgIdx];
+        if (!message.contains("reasoning_content")) {
+            continue;
+        }
+        auto reasoning = message["reasoning_content"];
+        std::string contentStr = reasoning.get_string();
+
+        message[templateReasoningFieldName.c_str()] = contentStr;
```
mzegla: 1. Is it guaranteed that `get_string()` call is valid? Don't we have to check the type first?
2. If we merged those two lines wouldn't we avoid additional, intermediate copy?

przepeck: you are right, I will correct that

src/llm/io_processing/chat_template/probe.cpp
przepeck: it should return false if the reasoning is not supported or reasoning field indicated to replace reasonig_content field doesn't render it. I've changed last return to be false.

PR #4312, opened 2026-06-22
Auto parser detect / auto input-workaround detection / --parser none

dkalinowski:
- auto parser detection
- auto input-workaround/template adapter detection
- support for --tool_parser none / --reasoning_parser none
- switch VLM pipelines to JINJA by default, making it aligned with LLM pipelines (jinja has broader support chat template application)


src/llm/io_processing/chat_template/analyzer.cpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    // LFM2 detection
+    if (contains(templateSource, "<|assistant_tool_call|>") || contains(templateSource, "<|tool_call_start|>")) {
+        result.detectedModelFamily = "lfm2";
+        result.detectedToolParser = "lfm2";
+        result.caps.supportsToolCalls = true;
+        result.caps.supportsTools = true;
+        result.caps.supportsToolResponses = true;
+        return result;
```
dkalinowski, line 64: TODO: Add reasoning parser once @przepeck implements it

src/llm/chat_template_analyzer.cpp
```diff
@@ -0,0 +1,136 @@
...
+        result.caps.supportsTools = true;
+        result.caps.supportsToolResponses = true;
+        return result;
+    }
+
+    // Llama3 detection — <|python_tag|>
+    if (contains(templateSource, "<|python_tag|>")) {
+        result.detectedToolParser = "llama3";
+        result.caps.supportsToolCalls = true;
+        result.caps.supportsTools = true;
```
dkalinowski: dont set capabilties in this step

src/test/llm/chat_template_compatibility.md
```diff
@@ -0,0 +1,21 @@
+# Chat Template Engine Compatibility
+
+| Model | Family Detection | Auto Tool Parser Detection | Auto Reasoning Parser Detection | Minja Mode | Jinja Mode | Notes |
+|-------|--------|-----------------|-------------|-------|-------|--------|
+| [gpt-oss-20b](https://huggingface.co/OpenVINO/gpt-oss-20b-int4-ov) | gptoss | gptoss | gptoss | ✅ Tools work | ✅ Tools work | The chat template we have uploaded contain many workarounds, including one that relates to `string2obj`. Without workaround, it translates to `{"key":"val"}`. To keep supporting that, we dont run `str2obj` workaround. |
```
dkalinowski: Add that to unit test comments

src/llm/servable_initializer.cpp
```diff
@@ -83,9 +116,53 @@ void GenAiServableInitializer::loadChatTemplate(std::shared_ptr<GenAiServablePro
...
+            analysisResult.caps.supportsToolCalls,
+            analysisResult.caps.requiresObjectArguments,
+            analysisResult.caps.requiresNonNullContent);
+        // Auto-detect or report manually configured tool parser
+        if (!properties->toolParserName.empty()) {
+            SPDLOG_LOGGER_INFO(llm_calculator_logger, "Using manually configured tool_parser: {}", properties->toolParserName);
+        } else if (analysisResult.detectedToolParser.has_value()) {
+            properties->toolParserName = analysisResult.detectedToolParser.value();
+            SPDLOG_LOGGER_INFO(llm_calculator_logger, "Auto-detected tool_parser: {}", properties->toolParserName);
+        }
```
przepeck, line 133: What if user for some reason don't want to use toolparser with model?
I remember, we were talking about CLI option --tool_parser none, don't we want to do it in this PR?

src/llm/io_processing/input_processors/input_workarounds_processor.hpp
```diff
@@ -0,0 +1,35 @@
...
+#pragma once
+
+#include "../../chat_template_caps.hpp"
+#include "../base_input_processor.hpp"
+
+namespace ovms {
+
+// Applies input workarounds to ChatHistory based on detected ChatTemplateCaps.
+// Runs before ChatTemplateProcessor so the template receives corrected input.
+class InputWorkaroundsProcessor : public BaseInputProcessor {
```
mzegla: As we discussed, thinking about another name for those since "workarounds" sound like temporary, not comprehensive solution and I'm not sure if this will be only a temporary logic.

How about something like `ChatTemplateAdapter`? To indicate input payload will be adapted to match chat template capabilities?

src/llm/visual_language_model/legacy/servable.cpp
```diff
@@ -14,6 +14,7 @@
 // limitations under the License.
 //*****************************************************************************
 
+#include <chrono>
```
mzegla, line 17: Why we need this include here now?

src/llm/BUILD
```diff
@@ -136,7 +136,8 @@ ovms_cc_library(
 
 ovms_cc_library(
     name = "io_processing_input_processor_context",
-    hdrs = ["io_processing/input_processor_context.hpp"],
+    hdrs = ["io_processing/input_processor_context.hpp",
+            "chat_template_caps.hpp"],
```
mzegla: Doesn't chat_template_caps belong to io_processing as well?

src/llm/chat_template_analyzer.cpp
```diff
@@ -0,0 +1,117 @@
...
+        result.caps.supportsToolCalls = true;
+        return result;
+    }
+
+    // Gemma4 detection
+    if (contains(templateSource, "'<|tool_call>call:'") || contains(templateSource, "<|tool_call>call:")) {
+        result.detectedToolParser = "gemma4";
+        result.detectedReasoningParser = "gemma4";  // gemma is always tied to its own parser for reasoning
+        result.caps.supportsToolCalls = true;
+        // result.caps.requiresObjectArguments = true;  // we know that gemma4 requires it, but let dry-run do its job
```
mzegla: Do we want to keep this comment?

src/llm/io_processing/chat_template/analyzer.cpp
```diff
@@ -0,0 +1,117 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+#include "chat_template_analyzer.hpp"
+
+#include <string>
+
+namespace ovms {
+
+static bool contains(const std::string& haystack, const std::string& needle) {
```
mzegla, line 22: Do we have to implement it ourselves? Isn't such method available in the standard library?

src/llm/chat_template_analyzer.cpp
```diff
@@ -0,0 +1,117 @@
...
+        result.detectedToolParser = "mistral";
+        result.caps.supportsToolCalls = true;
+        return result;
+    }
+
+    // Llama3 detection — <|python_tag|>
+    if (contains(templateSource, "<|python_tag|>")) {
+        result.detectedToolParser = "llama3";
+        result.caps.supportsToolCalls = true;
+        // result.caps.requiresNonNullContent = true;  // TODO: we know that llama3 requires it, but let dry-run do its job
```
mzegla: Do we want to keep that comment?

src/llm/chat_template_caps.hpp
```diff
@@ -0,0 +1,35 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+#pragma once
+
+namespace ovms {
+
+struct ChatTemplateCaps {
+    // TODO: Do we keep it?
```
mzegla: How do we resolve this TODO? Is this member not used/needed?

src/llm/io_processing/chat_template/probe.cpp
```diff
@@ -0,0 +1,194 @@
...
+#endif
+
+namespace ovms {
+
+bool probeChatTemplateBasicRenderMinja(ov::genai::Tokenizer& tokenizer) {
+    if (tokenizer.get_chat_template().empty()) {
+        return true;
+    }
+
+    static const std::string contentNeedle = "probe_basic_7Qw2";
```
mzegla, line 37: what is this value? is it just randomly picked string?

src/llm/io_processing/chat_template/probe.cpp
```diff
@@ -0,0 +1,194 @@
...
+            SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Basic render probe output: {}", output);
+            return false;
+        }
+
+        // Check for raw JSON dump markers — if minja silently failed, it dumps
+        // the message object as JSON containing "content": and "role":
+        if (output.find("\"content\": \"" + contentNeedle + "\"") != std::string::npos ||
+            output.find("\"content\":\"" + contentNeedle + "\"") != std::string::npos) {
+            SPDLOG_LOGGER_WARN(llm_calculator_logger, "Basic render probe: output contains raw JSON dump of message object. "
+                                                      "Template is not supported by minja.");
```
mzegla, line 64: Is it always true? Could template actually render `{"content": "blabla"}` as expected behavior?
Also, the condition doesn't match the comment . There is no "role" check and there is `||` not `&&`.

src/llm/io_processing/chat_template/probe.cpp
```diff
@@ -0,0 +1,194 @@
...
+    } catch (const std::exception& e) {
+        SPDLOG_LOGGER_WARN(llm_calculator_logger, "Basic render probe: exception: {}", e.what());
+        return false;
+    } catch (...) {
+        SPDLOG_LOGGER_WARN(llm_calculator_logger, "Basic render probe: unknown exception");
+        return false;
+    }
+}
+
+static const std::string PROBE_NEEDLE = "probe_needle_xK9m";
```
mzegla, line 79: different naming convention - previous need was name in camelCase

src/llm/io_processing/chat_template/probe.cpp
```diff
@@ -0,0 +1,194 @@
...
+    bool strArgsRendersNative = strOk && rendersNativeArgs(strOut);
+    bool objArgsRendersNative = objOk && rendersNativeArgs(objOut);
+
+    SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Dry-run probe: strRendersNative={}, objRendersNative={}",
+        strArgsRendersNative, objArgsRendersNative);
+    SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Dry-run probe strArgs output: {}", strOut);
+    SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Dry-run probe objArgs output: {}", objOut);
+
+    // Detect silent failure: output contains raw JSON dump of tool_calls
+    static const std::string silentFailureMarker = "\"tool_calls\": [";
```
mzegla, line 109: Are we sure it is never expected and always a failure marker?

src/llm/io_processing/chat_template/probe.cpp
```diff
@@ -0,0 +1,194 @@
...
+    bool strArgsFailed = strOk && strOut.find(silentFailureMarker) != std::string::npos;
+    bool objArgsFailed = objOk && objOut.find(silentFailureMarker) != std::string::npos;
+
+    if (strArgsFailed || objArgsFailed) {
+        SPDLOG_LOGGER_WARN(llm_calculator_logger, "Dry-run probe: template silently failed to render tool calls");
+        caps.supportsToolCalls = false;
+        return false;
+    }
+
+    if (strArgsRendersNative || objArgsRendersNative) {
```
mzegla, line 119: Are they mutually exclusive? Is it impossible for both those values be either true or false?

src/llm/io_processing/chat_template/probe.cpp
```diff
@@ -0,0 +1,194 @@
...
+        SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Dry-run probe: template does not render tool_call arguments natively");
+    }
+    return true;
+}
+
+bool probeChatTemplateCapsMinja(ov::genai::Tokenizer& tokenizer, ChatTemplateCaps& caps) {
+    if (tokenizer.get_chat_template().empty()) {
+        return true;
+    }
+    if (!caps.supportsToolCalls) {
```
mzegla, line 151: Early exit here. Does it mean this method is dedicated only for tool calling templates?
Would be worth noting, so if someone extends caps detection logic for template that does not support tool calling, they won't do it here.

src/llm/chat_template_probe.cpp
```diff
@@ -0,0 +1,194 @@
...
+        try {
+            ov::genai::ChatHistory history;
+            history.push_back(ov::genai::JsonContainer::from_json_string(R"({"role":"user","content":"Hello"})"));
+            history.push_back(ov::genai::JsonContainer::from_json_string(
+                R"({"role":"assistant","content":"","tool_calls":[{"id":"call_0_ab","type":"function","function":{"name":"probe_fn","arguments":{")" + PROBE_NEEDLE + R"(":"val"}}}]})"));
+            std::string output = tokenizer.apply_chat_template(history, true);
+            return {true, std::move(output)};
+        } catch (...) {
+            return {false, ""};
+        }
```
mzegla: big duplication of above code - could be extract it, so the code is shared?

src/llm/io_processing/chat_template/probe.cpp
```diff
@@ -0,0 +1,194 @@
...
+    auto [strOk, strOut] = strArgsFuture.get();
+    auto [objOk, objOut] = objArgsFuture.get();
+
+    return analyzeProbeResults(strOk, strOut, objOk, objOut, caps);
+}
+
+#if (PYTHON_DISABLE == 0)
+
+bool probeChatTemplateCapsJinja(PyJinjaTemplateProcessor& templateProcessor, ChatTemplateCaps& caps) {
+    if (!caps.supportsToolCalls) {
```
mzegla, line 174: Same as for Minja. If someone doesn't see the early exit here and implements caps detection for non function calling template it will be a dead code.

src/llm/io_processing/chat_template/probe.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    std::string strArgsJson = R"({"messages":[{"role":"user","content":"Hello"},{"role":"assistant","content":"","tool_calls":[{"id":"call_0_ab","type":"function","function":{"name":"probe_fn","arguments":"{\")" + PROBE_NEEDLE + R"(\":\"val\"}"}}]}]})";
+    std::string objArgsJson = R"({"messages":[{"role":"user","content":"Hello"},{"role":"assistant","content":"","tool_calls":[{"id":"call_0_ab","type":"function","function":{"name":"probe_fn","arguments":{")" + PROBE_NEEDLE + R"(":"val"}}}]}]})";
+
+    std::string strOut, objOut;
+    bool strOk = false, objOk = false;
+
+    try {
+        strOk = PyJinjaTemplateProcessor::applyChatTemplate(templateProcessor, strArgsJson, strOut);
+    } catch (...) {
```
mzegla, line 188: Is it always safe to just move on from exception? Could error in applyChatTemplate corrupt the state and second call that is done below would crash?

src/llm/chat_template_probe.hpp
```diff
@@ -0,0 +1,49 @@
...
+// Probes whether a chat template can render a basic user/assistant exchange.
+// Returns false if minja silently fails.
+// Should be called before tool-specific probing — if this fails,
+// the template is completely broken even for non-agentic use cases.
+bool probeChatTemplateBasicRenderMinja(ov::genai::Tokenizer& tokenizer);
+
+// Probes a chat template by dry-running it with synthetic inputs
+// to empirically detect whether the template requires special input workarounds.
+// Updates caps parameter based on probe results.
+// Only performs probing if caps.supportsToolCalls is true. TODO: do we need supportsToolCalls?
```
mzegla: Let's resolve TODO before merge, can we?

src/llm/chat_template_adapter.cpp
```diff
@@ -0,0 +1,89 @@
...
+            auto args = function["arguments"];
+            if (!args.is_string()) {
+                continue;
+            }
+            std::string argsStr = args.get_string();
+            // Parse and replace string arguments with the parsed JSON object
+            try {
+                function["arguments"] = ov::genai::JsonContainer::from_json_string(argsStr);
+            } catch (...) {
+                // If parsing fails, leave as-is
```
mzegla: Some logging on trace level?

src/llm/servable_initializer.cpp
dkalinowski: it should be possible to set --tool_parser ""
i will verify and add tests, good catch

src/llm/servable_initializer.cpp
```diff
@@ -83,9 +116,53 @@ void GenAiServableInitializer::loadChatTemplate(std::shared_ptr<GenAiServablePro
...
+    // Analyze the chat template to detect capabilities and model family
+    std::string templateSource = properties->tokenizer.get_chat_template();
+    if (!templateSource.empty()) {
+        auto analysisResult = ChatTemplateAnalyzer::analyze(templateSource);
+        properties->chatTemplateCaps = analysisResult.caps;
+        SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Chat template capabilities: supportsToolCalls={}, "
+                                                   "requiresObjectArguments={}, requiresNonNullContent={}",
+            analysisResult.caps.supportsToolCalls,
+            analysisResult.caps.requiresObjectArguments,
+            analysisResult.caps.requiresNonNullContent);
```
mzegla: Maybe caps object could provide a string representation? This way we would avoid necessity of updates here whenever we add a new capability.

src/llm/servable_initializer.cpp
```diff
@@ -83,9 +116,53 @@ void GenAiServableInitializer::loadChatTemplate(std::shared_ptr<GenAiServablePro
...
+#endif
+            if (!probeChatTemplateBasicRenderMinja(properties->tokenizer)) {
+                SPDLOG_LOGGER_ERROR(llm_calculator_logger, "Chat template is not compatible with minja — basic rendering failed. "
+                                                           "Disabling /chat/completions endpoint for this model.");
+                properties->tokenizer.set_chat_template("");
+                return;
+            }
+#if (PYTHON_DISABLE == 0)
+        }
+#endif
```
mzegla, line 159: IMHO it's not a very readable construction with those ifdefs. Maybe we could determine if minja should be used earlier and just rely on the simple flag?

mzegla: btw. what about the jinja path? Is it used here?

src/llm/visual_language_model/legacy/servable.cpp
dkalinowski: chrono is actually being used in this file, we forgot to include it in earlier commits

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: yes

src/llm/BUILD
dkalinowski: entire set of new files landed in src/llm:
- chat template probe cpp/hpp
- chat template analzer cpp/hpp

I will move all 5 of these files to io_processing

src/llm/io_processing/chat_template/analyzer.cpp
dkalinowski: It's in standard library of C++23 and onwards, so it is not available in our setup

src/llm/chat_template_analyzer.cpp
dkalinowski: will remove

src/llm/chat_template_analyzer.cpp
dkalinowski: will remove

src/llm/chat_template_caps.hpp
dkalinowski: it is used to detect if input_workarounds::apply capabilities need to be probed in first place, maybe we dont need it, we can just probe it anyway, and for regular non-agentic chat templates report that it is bugged and tools will not work (dry-runs wont work)

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: Yes, i didnt see such chat templates, but theoretically it could render it like that, why not. I also dont like it, but i'd like to keep it in minja path since minja is just bugged and not reliable

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: one is used in multiple functions, thats why it is outside of function, thats why it is full caps
another one is only used in 1 function, therefore it is static inside a function thats why camel case

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: same as above - not. but until minja is not reliable, we need to detect abnormal behaviors in some way
another option is to drop minja support

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: no, they are not mutually exclusive, some chat templates even handle it in a way that the type is checked before use

if string -> treat as string
if dictlike -> iterate over key,vals

some chat templates are not that smart and thas why we need those caps

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: For now it is, is adding comment here enough?
In future if workarounds are also non-tool related, we can remove this if statement.
As seen in llama/minja workarounds are only for tool topic

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: answered above

src/llm/io_processing/chat_template/probe.cpp
mzegla: I see. How about the second part? The condition itself - is it correct?
`output.find("\"content\": \"" + contentNeedle + "\"") != std::string::npos || output.find("\"content\":\"" + contentNeedle + "\"") != std::string::npos`
And no role check. The comment above mentioned it though.

src/llm/io_processing/chat_template/probe.cpp
mzegla: I think a comment would be enough. We kind of expect this code to evolve, so it will come up sooner or later I guess.

src/llm/io_processing/chat_template/probe.cpp
mzegla: So we treat objArgs render as higher priority? I mean if both are true we set requiresObjectArguments.

dkalinowski: Yes, otherwise some models would not have capabilities detected correctly, just check gemma, it would fall into string category:

image

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: Yes, one condition checks fo `"key": "val"` and another for `"key":"val"`, any of this indicates minja is failing

Yes, i will fix the comment

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: Unit tests document it very well, changing the logic to `if (!strArgsRendersNative && objArgsRendersNative) {`:

```
    FAILED  ChatTemplateEndToEndJinjaTest.Gemma4_ToolCallWithStringArgs (0.2s)
    FAILED  ChatTemplateEndToEndJinjaTest.Qwen3_ToolCallWithStringArgs (0.1s)
    FAILED  ChatTemplateEndToEndJinjaTest.Qwen3VL_ToolCallWithStringArgs (0.1s)
    FAILED  ChatTemplateEndToEndJinjaTest.Qwen3_30B_ToolCallWithStringArgs (0.1s)
    FAILED  ChatTemplateEndToEndMinjaTest.Qwen36_ToolCallWithStringArgs (0.5s)
    FAILED  ChatTemplateEndToEndMinjaTest.Gemma4_ToolCallWithStringArgs (1.1s)
    FAILED  ChatTemplateEndToEndMinjaTest.Qwen3_ToolCallWithStringArgs (0.3s)
    FAILED  ChatTemplateEndToEndMinjaTest.Mistral7B_ToolCallWithStringArgs (0.3s)
    FAILED  ChatTemplateEndToEndMinjaTest.Qwen3VL_ToolCallWithStringArgs (0.4s)
    FAILED  ChatTemplateEndToEndMinjaTest.Qwen3_30B_ToolCallWithStringArgs (0.3s)
```

src/llm/chat_template_caps.hpp
dkalinowski: Removed todo since we keep it, look https://github.com/openvinotoolkit/model_server/pull/4312#discussion_r3520428885

src/llm/io_processing/chat_template/probe.cpp
dkalinowski: Neither of workarounds rely on exception check from jinja. Jinja doesnt throw an exception if it cannot apply the template. Either way, I added try catch. Will add error logs here just in case it changes in the future

src/llm/servable_initializer.cpp
dkalinowski: > IMHO it's not a very readable construction with those ifdefs. Maybe we could determine if minja should be used earlier and just rely on the simple flag?

I think @rasapala is working on that with runtime python check

> btw. what about the jinja path? Is it used here?

No, we dont need basic probe for jinja, jinja doesnt silently fail

src/test/llm/chat_template_end_to_end_jinja_test.cpp
```diff
@@ -0,0 +1,546 @@
...
+<|im_start|>assistant
+<|tool_call_start|>[get_weather(location='Paris', unit='celsius')]<|tool_call_end|><|im_end|>
+<|im_start|>assistant
+)";
+    EXPECT_EQ(appliedOutput, expectedOutput);
+}
+
+// =============================================================================
+// Same story as Qwen3-8B.
+// TODO(mzeglars): Do we keep it?
```
mzegla: Don't think so. If qwen3-8B has the same type of template (in terms of needed adaptations) and we have separate tests in chat templates dir, then I suppose we can drop it here in the end to end test. Unless we wanted to check some VLM specific capabilities.

przepeck (approved):
could you add to the docs/parameters.md information about --tool_parser none option?

src/llm/io_processing/input_processors/chat_template_adapter.cpp
```diff
@@ -0,0 +1,101 @@
...
+        auto message = chatHistory[msgIdx];
+        if (!message.contains("tool_calls")) {
+            continue;
+        }
+        if (!message.contains("content")) {
+            message["content"] = "";
+        } else if (message["content"].is_null()) {
+            message["content"] = "";
+        }
+    }
```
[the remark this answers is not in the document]
dkalinowski: ov::genai::JsonContainer uses reference semantics internally — operator[] returns a lightweight proxy/handle that refers back to the same underlying JSON data, not a deep copy.

src/llm/io_processing/input_processors/chat_template_adapter.cpp
```diff
@@ -0,0 +1,101 @@
...
+        for (size_t i = 0; i < toolCalls.size(); ++i) {
+            auto toolCall = toolCalls[i];
+            if (!toolCall.is_object() || !toolCall.contains("function")) {
+                continue;
+            }
+            auto function = toolCall["function"];
+            if (!function.is_object() || !function.contains("arguments")) {
+                continue;
+            }
+            auto args = function["arguments"];
```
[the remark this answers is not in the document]
dkalinowski: ov::genai::JsonContainer uses reference semantics internally — operator[] returns a lightweight proxy/handle that refers back to the same underlying JSON data, not a deep copy.

src/test/llm/chat_templates/chat_template_devstral.jinja
```diff
@@ -0,0 +1,107 @@
...
+
+{{- bos_token }}
+
+\nWhen generating the tool call, follow the format: [TOOL_CALLS]tool_name[ARGS]arguments as json\n\n
```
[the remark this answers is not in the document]
dkalinowski: this is chat template we actually use

src/test/llm/chat_templates/chat_template_phi4_mini.jinja
```diff
@@ -0,0 +1,63 @@
...
+
+{%- if messages %}
+<|system|>
+{{ system_message }}
+{%- if tools %}
+In addition to plain text responses, you can chose to call one or more of the provided functions.
+
+Use the following rule to decide when to call a function:
+  * if the response can be generated from your internal knowledge (e.g., as in the case of queries like "What is the capital of Poland?"), do so
+  * if you need external information that can be obtained by calling one or more of the provided functions, generate a function calls
```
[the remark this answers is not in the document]
dkalinowski: this is chat template we actually use

src/test/llm/chat_templates/chat_template_phi4_mini.jinja
```diff
@@ -0,0 +1,63 @@
...
+  * if you receive the tool response, interpret the response and try to answer user's question or command
+
+If you decide to call functions:
+  * prefix function calls with functools marker (no closing marker required)
+  * all function calls should be generated in a single JSON list formatted as functools[{"name": [function name], "arguments": [function arguments as JSON]}, ...]
+  * follow the provided JSON schema. Do not hallucinate arguments or values. Do to blindly copy values from the provided samples
+  * respect the argument type formatting. E.g., if the type if number and format is float, write value 7 as 7.0
```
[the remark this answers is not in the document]
dkalinowski: this is chat template we actually use

mzegla (approved):

src/llm/servable_initializer.cpp
dkalinowski: Added support for --tool_parser none and --reasoning_parser none

PR #4317, opened 2026-06-22
default task detection based on model config

dtrawins:
### 🛠 Summary

CVS-188024
task parameter will be determined automatically based on model config.json, model_index.json and in special cases based on model name pattern. 
It simplifies model deployment from HF hub. 
While using --source_model - it will always create or update graph.pbtxt
While using --model_path - it will using params from graph.pbtxt if exist and no overwrite params are passed. When graph.pbtxt is missing, it will determine task automatically
In case task can't be determined like for unknows architectures - no default task will be set to be provided by the user.

### 🧪 Checklist

- [ ] Unit tests added.
``


src/cli_parser.cpp
```diff
@@ -335,12 +535,62 @@ std::variant<bool, std::pair<int, std::string>> CLIParser::parse(int argc, char*
...
+            !result->count("source_model") &&
+            result->count("model_path") &&
+            !isConfigManagementFlow) {
+            const std::optional<std::string> modelPath = std::make_optional(result->operator[]("model_path").as<std::string>());
+            const auto configPath = std::filesystem::path(*modelPath) / MODEL_CONFIG_FILENAME;
+            if (std::filesystem::exists(configPath)) {
+                // Check if task-specific parameters are provided or if graph.pbtxt is missing
+                bool hasUnmatchedOptions = ::ovms::hasTaskSpecificParameters(result->unmatched());
+                bool graphExists = ::ovms::graphPbtxtExists(*modelPath);
+
```
dtrawins:
```suggestion
```

src/cli_parser.cpp
```diff
@@ -335,12 +535,62 @@ std::variant<bool, std::pair<int, std::string>> CLIParser::parse(int argc, char*
...
             std::vector<std::string> unmatchedOptions;
             GraphExportType task;
-            if (result->count("task")) {
-                task = stringToEnum(result->operator[]("task").as<std::string>());
+            std::string taskValue;
+            if (!result->count("task")) {
+                const std::optional<std::string> modelPath = result->count("model_path") ? std::make_optional(result->operator[]("model_path").as<std::string>()) : std::nullopt;
+                const std::optional<std::string> sourceModel = result->count("source_model") ? std::make_optional(result->operator[]("source_model").as<std::string>()) : std::nullopt;
+                const std::optional<std::string> modelRepositoryPath = result->count("model_repository_path") ? std::make_optional(result->operator[]("model_repository_path").as<std::string>()) : std::nullopt;
+
```
dtrawins:
```suggestion
```

src/cli_parser.cpp
```diff
@@ -335,12 +535,62 @@ std::variant<bool, std::pair<int, std::string>> CLIParser::parse(int argc, char*
...
+                    shouldInferTask = true;
+                } else if (modelPath.has_value() && !modelPath->empty()) {
+                    // For local model_path, infer task if:
+                    // 1. Unmatched options (task-specific parameters) are present, OR
+                    // 2. graph.pbtxt doesn't exist (need to create in-memory graph)
+                    bool hasUnmatchedOptions = ::ovms::hasTaskSpecificParameters(result->unmatched());
+                    bool graphExists = ::ovms::graphPbtxtExists(*modelPath);
+                    shouldInferTask = hasUnmatchedOptions || !graphExists;
+                }
+
```
dtrawins:
```suggestion
```

src/test/models_config_json/xlm_roberta/config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "architectures": ["XLMRobertaForSequenceClassification"],
```
dkalinowski, line 2: I see its possible to have more than 1 architecture (its a list). How would OVMS behave? Can you add test for that behavior?

src/test/models_config_json/stable_diffusion/config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "architectures": ["UNet2DConditionModel"],
```
dkalinowski, line 2: Please add these files to data field in unit test build target. If you dont do that, the unit tests will not re-run once this file change.

Example: https://github.com/openvinotoolkit/model_server/blob/main/src/BUILD#L2337-L2364

src/test/ovmsconfig_test.cpp
```diff
@@ -2822,4 +2850,105 @@ TEST(OvmsConfigManipulationTest, positiveDisableModel) {
...
+    ASSERT_EQ(config.getServerSettings().serverMode, ovms::HF_PULL_AND_START_MODE);
+}
+
+// Scenario 2: --model_path points to a directory that has config.json but NO graph.pbtxt.
+// The task must be inferred and the server must start in IN_MEMORY_GRAPH_MODE.
+TEST_F(OvmsInferredTaskTest, positiveModelPathNoGraphPbtxtInferTask) {
+    const std::string modelPath = resolveTestModelPath("llama");
+    const std::filesystem::path configJson = std::filesystem::path(modelPath) / "config.json";
+    if (!std::filesystem::exists(configJson)) {
+        GTEST_SKIP() << "Test prerequisite missing: " << configJson.string();
```
dkalinowski: Why skipping? Shouldnt it error?

src/BUILD
```diff
@@ -2158,6 +2161,19 @@ cc_binary(
     linkstatic = True,
 )
 
+cc_binary(
+    name = "num_streams_repro",
+    srcs = [
+        "num_streams_repro.cpp",
```
dkalinowski: ?

src/cli_parser.cpp
```diff
@@ -33,13 +49,147 @@
...
+    {"SeamlessM4Tv2Model", "speech2text"},
+    {"SpeechT5ForTextToSpeech", "text2speech"},
+    {"T5EncoderModel", "embeddings"},
+    {"UNet2DConditionModel", "image_generation"},
+    {"WhisperForConditionalGeneration", "speech2text"},
+    {"XLMRobertaForSequenceClassification", "rerank"},
+    {"XLMRobertaModel", "embeddings"},
+};
+
+std::string getEnvOrDefault(const char* envName, const std::string& defaultValue = "") {
```
dkalinowski: these functions should be static

src/cli_parser.cpp
```diff
@@ -33,13 +49,147 @@
...
+        if (!resolvedTask.has_value()) {
+            resolvedTask = task;
+            continue;
+        }
+        if (resolvedTask.value() != task) {
+            throw std::logic_error("config.json architectures map to multiple default tasks");
+        }
+    }
+    if (!resolvedTask.has_value()) {
+        throw std::logic_error("config.json architectures do not map to a supported default task");
```
dkalinowski: please add test for that

src/cli_parser.cpp
```diff
@@ -335,12 +534,70 @@ std::variant<bool, std::pair<int, std::string>> CLIParser::parse(int argc, char*
 
         result = std::make_unique<cxxopts::ParseResult>(options->parse(argc, argv));
 
+        const bool isConfigManagementFlow =
```
dkalinowski: could be a CLIParser method

src/test/models_config_json/xlm_roberta/config.json
dtrawins: In theory this is possible but rarely if ever happens. Added tests to cover it. First architecture will be used.

src/cli_parser.cpp
dtrawins: why? They are already inside a namespace block?

src/cli_parser.cpp
```diff
@@ -33,13 +49,206 @@
...
+        return "";
+    }
+    const auto& [defaultTask, patternRules] = architectureRules->second;
+    for (const auto& [task, keyword] : patternRules) {
+        if (normalizedModelIdentifier.find(keyword) != std::string::npos) {
+            return task;
+        }
+    }
+    return defaultTask;
+}
```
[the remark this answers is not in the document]
dtrawins: currently the only questionable architecture is Qwen3ForCausalLM which is used most frequently for text generation like Qwen3-4B, Qwen3-8B etc. There is no unique pattern specific for text generation. The pattern for rerank and embed should help in proper task identification. For potential other questionable architectures it would be possible to set empty default task.

src/pull_module/hf_env_vars.hpp
```diff
@@ -0,0 +1,22 @@
+//****************************************************************************
+// Copyright 2025 Intel Corporation
```
mzegla: 2026

mzegla (commented):
I'm not a fan of putting entire auto detection logic directly to cli_parser file, especially if we expect it to grow.
From the architecture point of view I would encourage the following:
1. Keep CLI parser clean - on the parser level just parse input.
2. Have a separate file with a class `AutoTaskDetector` or something similar
3. Instead of multiple mapping we could have object representations like "Qwen3ForCausalLM" maps to Qwen3ForCausalLM class which internally has logic to check further (since its ambiguous and we can't directly tell if it's for generation, embedding, reranker etc.). Missing or null architecture could have their own hirarchy.

I know it would expand the code significantly, but such approach would look more scalable and easier to manage to me. As there are tons of models and multiple selection criteria, having multiple mapping and top-level chain of ifs directly in CLI parser does not look like a good design to me.

src/BUILD
```diff
@@ -2273,6 +2274,7 @@ cc_test(
         "test/tensor_conversion_test.cpp",
         "test/tensorinfo_test.cpp",
         "test/tensorutils_test.cpp",
+        "test/task_determine_test.cpp",
```
atobiszei, line 2164: Add as separate test library target and make those new config files as data field there.

src/cli_parser.cpp
```diff
@@ -15,15 +15,31 @@
...
 #include <variant>
 
+#ifdef _MSC_VER
+#pragma warning(push)
+// Suppress warning originating from third-party RapidJSON internals on MSVC.
+#pragma warning(disable : 6313)
+#endif
+#include <rapidjson/document.h>
+#include <rapidjson/error/en.h>
+#include <rapidjson/istreamwrapper.h>
```
atobiszei: include src/port/rapidjson*hpp instead to avoid spilling pragmas/ifdefs.

src/cli_parser.cpp
```diff
@@ -731,10 +1063,16 @@ void CLIParser::prepareGraph(ServerSettingsImpl& serverSettings, HFSettingsImpl&
...
         // use model_path as the model location (no HF download needed)
         if (!result->count("pull") && !result->count("source_model") && result->count("model_path")) {
-            hfSettings.exportSettings.modelPath = result->operator[]("model_path").as<std::string>();
+            const auto configuredModelPath = std::filesystem::path(result->operator[]("model_path").as<std::string>());
+            hfSettings.exportSettings.modelPath = std::filesystem::absolute(configuredModelPath).lexically_normal().string();
+            SPDLOG_DEBUG("Using local absolute model path for graph export: {}", hfSettings.exportSettings.modelPath);
+        }
+        const std::string taskValue = getEffectiveTaskParameter();
+        if (inferredTaskParameter.has_value()) {
+            SPDLOG_INFO("Identified default task '{}' from model config", inferredTaskParameter.value());
```
atobiszei: We did not use spdlog in cli_parser on purpose. Logging could not be setup yet. Use cout. We also did use cout for model pulling.

src/cli_parser.cpp
```diff
@@ -15,15 +15,31 @@
...
+#endif
+#include <rapidjson/document.h>
+#include <rapidjson/error/en.h>
+#include <rapidjson/istreamwrapper.h>
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif
+
 #include "capi_frontend/server_settings.hpp"
+#include "logging.hpp"
```
atobiszei, line 30: spdlog was not used in cli_parser on purpose - during cli parsing spdlog could be not initialized yet.

src/cli_parser.cpp
atobiszei: Why we do not use utils from tils/env_guard.hpp?

src/BUILD
```diff
@@ -373,6 +374,8 @@ ovms_cc_library(
         "//src/graph_export:t2s_graph_cli_parser",
         "//src/graph_export:s2t_graph_cli_parser",
         "//src/graph_export:image_generation_graph_cli_parser",
+        "//src/pull_module:curl_downloader",
```
atobiszei: Why we do need to add those includes here? We are spilling dependencies.

src/BUILD
```diff
@@ -363,6 +363,7 @@ ovms_cc_library(
     srcs = ["cli_parser.cpp"],
     deps = [
         "@com_github_jarro2783_cxxopts//:cxxopts",
+        "@com_github_tencent_rapidjson//:rapidjson",
```
atobiszei: Depend on src/port/rapidjson*

src/cli_parser.cpp
```diff
@@ -53,6 +262,65 @@ std::string getConfigPath(const std::string& configPath) {
...
+                return determineTaskFromModelIndex(indexFile, indexPath.string());
+            }
+            throw std::logic_error("failed to open model config file: " + configPath.string() + " or " + indexPath.string());
+        }
+    }
+
+    std::string responseBody;
+    const std::string hfEndpoint = ensureTrailingSlash(getEnvOrDefault(HF_ENDPOINT_ENV_VAR, DEFAULT_HF_ENDPOINT));
+    const std::string configUrl = hfEndpoint + *sourceModel + "/resolve/main/" + MODEL_CONFIG_FILENAME;
+    const auto status = fetchUrlToString(configUrl, getEnvOrDefault(HF_TOKEN_ENV_VAR), responseBody);
```
atobiszei: What happens if someone launches OVMS with local model without access to network?

src/cli_parser.cpp
```diff
@@ -53,6 +262,65 @@ std::string getConfigPath(const std::string& configPath) {
     return configPath;
 }
 
+std::string CLIParser::determineDefaultTaskParameter(const std::optional<std::string>& modelPath, const std::optional<std::string>& sourceModel, const std::optional<std::string>& modelRepositoryPath) {
+    if (modelPath.has_value() && !modelPath->empty()) {
```
atobiszei: I think most logic for determining the task doesn't depend a lot on cli_parser and could be separated into new file, which wouldn't be rebuild with each cli_parser.cpp change.

src/pull_module/hf_env_vars.hpp
mzegla: @dtrawins

src/test/ovmsconfig_test.cpp
```diff
@@ -403,7 +403,7 @@ TEST_F(OvmsConfigDeathTest, hfNoTaskParameter) {
         "/some/path",
     };
     int arg_count = 6;
-    EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(OVMS_EX_USAGE), "error parsing options - --task parameter wasn't passed");
+    EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(OVMS_EX_USAGE), "--task parameter wasn't passed");
```
mzegla: This change creates and outsider among others. See previous tests. Their expected message includes "error parsing options". I think we should keep it aligned, one way or another.

src/cli_parser.cpp
```diff
@@ -299,7 +317,7 @@ std::variant<bool, std::pair<int, std::string>> CLIParser::parse(int argc, char*
 
         options->add_options("generative task (applies to: pull hf model, single model)")
             ("task",
-                "Specifies the generative task for the local model. It should be followed by task specific parameters. Supported tasks: text_generation, embeddings, rerank, image_generation, text2speech, speech2text. It creates the pipeline graph in memory based on the provided task-specific options.",
+                "Specifies the generative task for the local model. If not provided, default task value is inferred from model config.json architectures. It should be followed by task specific parameters. Supported tasks: text_generation, embeddings, rerank, image_generation, text2speech, speech2text. It creates the pipeline graph in memory based on the provided task-specific options.",
```
mzegla: Is it entirely true? I suppose we infer task based on more than config.json. Maybe just say that tasked is inferred from model specification or something like that to not point to a single file.

src/test/ovmsconfig_test.cpp
```diff
@@ -2823,6 +2851,222 @@ TEST(OvmsConfigManipulationTest, positiveDisableModel) {
     ASSERT_EQ(modelSettings.configPath, configPath);
 }
 
+// ===================== Inferred Default Task Tests =====================
+// Fixture providing resolved paths to pre-built test model directories.
+class OvmsInferredTaskTest : public ::testing::Test {
+public:
+    static std::string resolveTestModelPath(const std::string& modelDirName) {
+        const std::string relPath = std::string("src/test/models_config_json/") + modelDirName;
```
mzegla, line 2859: Does it all work on Windows as well? I don't see path adjustment calls here?

src/test/task_determine_test.cpp
```diff
@@ -0,0 +1,116 @@
+//*****************************************************************************
+// Copyright 2020-2021 Intel Corporation
```
mzegla, line 2:
```suggestion
// Copyright 2026 Intel Corporation
```

src/cli_parser.cpp
```diff
@@ -335,12 +353,73 @@ std::variant<bool, std::pair<int, std::string>> CLIParser::parse(int argc, char*
...
 
+        const bool isConfigManagementFlow =
+            result->count("add_to_config") || result->count("remove_from_config") || result->count("list_models");
+        if (!result->count("task") &&
+            !result->count("pull") &&
+            !result->count("source_model") &&
+            result->count("model_path") &&
+            !isConfigManagementFlow &&
+            !result->count("help") &&
+            !result->count("version")) {
```
mzegla: Could we make this condition more meaningful? Like use more of `isConfigManagementFlow` and less `result->count{"param")` to help understand when we need to enter that block. More like:
`if (taskNotDefined && deployingFromLocalDir && !isConfigManagementFlow)`

src/default_task.cpp
```diff
@@ -0,0 +1,278 @@
+//*****************************************************************************
+// Copyright 2025 Intel Corporation
```
mzegla:
```suggestion
// Copyright 2026 Intel Corporation
```

src/default_task.cpp
```diff
@@ -0,0 +1,278 @@
...
+    {"SpeechT5ForTextToSpeech", "text2speech"},
+    {"T5EncoderModel", "embeddings"},
+    {"UNet2DConditionModel", "image_generation"},
+    {"WhisperForConditionalGeneration", "speech2text"},
+    {"XLMRobertaForSequenceClassification", "rerank"},
+    {"XLMRobertaModel", "embeddings"},
+};
+
+// architecture: {default task, {task, pattern}}
+static const std::map<std::string, std::pair<std::string, std::vector<std::pair<std::string, std::string>>>> questionableArchitectureTaskKeywords = {
```
mzegla: 1. I think we could change the wording here and in all other places from `questionable` to `ambiguous`. 
Questionable sounds more like we have some doubts about the architecture itself. Ambiguous indicates we do not have enough information to make a decision.
2. I don't like this type. It's hard to grasp all that nesting. 
a) simpler option would be to use named types like
b) a bit bigger option would be to represent each architecture as a struct/class

src/default_task.cpp
```diff
@@ -0,0 +1,278 @@
...
+    path.push_back('/');
+    return path;
+}
+
+std::string getTaskForArchitecture(const std::string& architecture) {
+    const auto exactMatch = architectureToTask.find(architecture);
+    if (exactMatch != architectureToTask.end()) {
+        return exactMatch->second;
+    }
+    if (architecture == "WhisperForConditionalGeneration" || architecture.rfind("SeamlessM4T", 0) == 0) {
```
mzegla: I wonder if looking from this side is extensible. For example how about `https://huggingface.co/Qwen/Qwen3-ASR-1.7B/blob/main/config.json#L3` ? 
Did you consider per servable type detector? like:
```
std::string modelPath; // path to model catalog
for (servableTypeDetector : servableTypeDetectors) {
	if(servableTypeDetector.scan(modelPath))
		return servableTypeDector.get_name()
}
```

Then we close conditions to be checked and that have to match in separate modules - each responsible for answering the question: "Is my servable type right for that model?". It's a general suggestion, not just for this line.

What do you think?

src/pull_module/hf_env_vars.hpp
dtrawins:
```suggestion
// Copyright 2026 Intel Corporation
```

src/test/ovmsconfig_test.cpp
```diff
@@ -881,7 +881,35 @@ TEST_F(OvmsConfigDeathTest, hfSourceModelWithoutTask) {
         "/some/path",
     };
     int arg_count = 5;
-    EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(OVMS_EX_USAGE), "--source_model should be used combined with --task");
+    EXPECT_EXIT(ovms::Config::instance().parse(arg_count, n_argv), ::testing::ExitedWithCode(OVMS_EX_USAGE), "--task parameter wasn't passed");
```
mzegla: make this error more explicit

src/default_task.cpp
```diff
@@ -0,0 +1,105 @@
...
+// limitations under the License.
+//*****************************************************************************
+#include "default_task.hpp"
+
+#include <filesystem>
+#include <optional>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <vector>
```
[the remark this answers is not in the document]
rasapala: Add.

src/default_task.cpp
```diff
@@ -0,0 +1,105 @@
...
+    return !unmatchedOptions.empty();
+}
+
+std::string determineDefaultTaskParameter(const std::optional<std::string>& modelPath, const std::optional<std::string>& sourceModel, const std::optional<std::string>& modelRepositoryPath) {
+    DefaultTaskDetector detector;
+
+    if (modelPath.has_value() && !modelPath->empty()) {
+        const std::filesystem::path pathFs(*modelPath);
+        ModelCatalogContext ctx(pathFs, *modelPath);
+        const std::string task = detector.detect(ctx);
```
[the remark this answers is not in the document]
rasapala: Add.

src/default_task.cpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    ModelCatalogContext ctx(std::filesystem::path{}, *sourceModel);
+    ctx.addContent("config.json", std::move(responseBody));
+    const std::string task = detector.detect(ctx);
+    if (task.empty()) {
+        throw std::logic_error("cannot determine default task for source model: " + *sourceModel);
+    }
+    return task;
+}
```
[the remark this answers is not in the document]
rasapala: ??

src/cli_parser.cpp
```diff
@@ -335,12 +350,73 @@ std::variant<bool, std::pair<int, std::string>> CLIParser::parse(int argc, char*
...
+                // Infer task if:
+                // 1. Task-specific parameters are provided (unmatched options), OR
+                // 2. graph.pbtxt doesn't exist (need to create in-memory graph)
+                // Otherwise, if graph.pbtxt exists and no task parameters, use the filesystem graph
+                if (hasUnmatchedOptions || !graphExists) {
+                    try {
+                        inferredTaskParameter = determineDefaultTaskParameter(modelPath, std::nullopt, std::nullopt);
+                    } catch (const std::exception& e) {
+                        SPDLOG_INFO("Default task inference skipped for model_path '{}': {}", modelPath.value_or(""), e.what());
+                    }
```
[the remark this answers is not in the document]
rasapala: Add.

src/cli_parser.cpp
```diff
@@ -53,6 +58,16 @@ std::string getConfigPath(const std::string& configPath) {
...
 }
 
+std::string CLIParser::getEffectiveTaskParameter() const {
+    if (result->count("task")) {
+        return result->operator[]("task").as<std::string>();
+    }
+    if (inferredTaskParameter.has_value()) {
+        return inferredTaskParameter.value();
+    }
+    throw std::logic_error("error parsing options - --task parameter wasn't passed");
```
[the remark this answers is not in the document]
rasapala: Add.

src/cli_parser.cpp
```diff
@@ -379,12 +455,12 @@ std::variant<bool, std::pair<int, std::string>> CLIParser::parse(int argc, char*
                         break;
                     }
                     case UNKNOWN_GRAPH: {
-                        ss << "error parsing options - --task parameter unsupported value: " + result->operator[]("task").as<std::string>();
+                        ss << "error parsing options - --task parameter unsupported value: " + taskValue;
                         return std::make_pair(OVMS_EX_USAGE, ss.str());
                     }
```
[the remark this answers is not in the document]
rasapala: Add.

src/default_task.cpp
```diff
@@ -0,0 +1,105 @@
...
+        ModelCatalogContext ctx(pathFs, *modelPath);
+        const std::string task = detector.detect(ctx);
+        if (task.empty()) {
+            throw std::logic_error("cannot determine default task for model path: " + *modelPath);
+        }
+        return task;
+    }
+
+    if (!sourceModel.has_value() || sourceModel->empty()) {
+        throw std::logic_error("cannot determine default --task without model_path or source_model");
```
rasapala: Switch flow control from throwing exceptions to logging and returning empty task handling.

rasapala (commented):
Copilot had good suggestions plus exceptions control flow refactor.

mzegla (approved):

rasapala (approved):

PR #4329, opened 2026-06-26
Propagate global --cache_dir to continuous batching pipeline (#4230)

### LINKED ISSUE #4230: --cache_dir not propagated to LLM continuous batching pipeline (regression vs 2025.4 promise)
**Component:** LLM continuous batching, `src/llm/language_model/continuous_batching/servable_initializer.cpp`

**OVMS version:** 2026.1.0.72cc0624 (OpenVINO backend 2026.1.0, OpenVINO GenAI backend 2026.1.0.0)

**Platform:** Windows 11, Intel Arc 140T iGPU (`OPTIMIZATION_CAPABILITIES` includes `EXPORT_IMPORT`)

**Summary**

The 2025.4 release notes state:

> `--cache_dir` now enables compilation caching for both classic models and generative pipelines.

In 2026.1 this no longer holds for generative pipelines. When the server is started with `--cache_dir <path>`, the log line `Model cache is enabled: <path>` is printed, but the `CACHE_DIR` property is empty in every OpenVINO Core device-plugin dump that follows, and no `.cl_cache` / `.blob` files are written for `HttpLLMCalculator` nodes. The model is fully recompiled on every restart.

This reproduces both with and without speculative decoding, so the gap is not specific to drafts — it affects any `LLMCalculatorOptions` node.

**Reproduction A — minimal single-model LLM**

`graph.pbtxt`:

    node: {
      calculator: "HttpLLMCalculator"
      ...
      node_options: {
          max_num_seqs: 256,
          device: "GPU",
          models_path: "./",
          enable_prefix_caching: true,
          cache_size: 0,
        }
      }
      ...
    }

Start:

    ovms.exe --config_path config.json --rest_port 8001 --cache_dir C:/some/empty/dir --log_level DEBUG

Send one request, stop. The cache directory remains empty. Log shows `Model cache is enabled: <path>` but `OpenVINO Core plugin: GPU; plugin configuration: { ... CACHE_DIR: , ... }`.

**Reproduction B — same plus speculative decoding**

Add to the same node:

    draft_models_path: "OpenVINO-Qwen3-0.6B-int8-ov",
    draft_device: "CPU",

Same outcome: log says cache enabled, plugin `CACHE_DIR:` is empty, directory stays empty.

**Diagnosis**

For classic models, `--cache_dir` flows through `src/modelinstance.cpp::setCacheOptions` → `ieCore.set_property(ov::cache_dir(...))`. The LLM continuous-batching path bypasses `ModelInstance` and builds a `ContinuousBatchingPipeline` directly in `src/llm/language_model/continuous_batching/servable_initializer.cpp`. Grepping that file (and its neighbours under `src/llm/`) finds zero references to `cache_dir`, `CACHE_DIR`, `ov::cache_dir`, or `ServerSettings.cache_dir`. The global flag value is never read on this code path.

The same pattern was reported for the underlying GenAI pipeline in issue openvinotoolkit/openvino.genai#1992. OVMS LLM nodes route through the same GenAI `ContinuousBatchingPipeline`, so the surface area overlaps.

**Workaround**

Setting `CACHE_DIR` inside `plugin_config` of the LLM node propagates correctly:

```
    node_options: {
        device: "GPU"
        models_path: "./"
        plugin_config: '{"CACHE_DIR": "C:/some/empty/dir"}'
        ...
      }
    }
```

After this, `.cl_cache` files appear in the directory and subsequent starts compile materially faster.

Two notes for anyone hitting this:
- Forward slashes are required in the JSON value on Windows. With backslashes, pbtxt unescapes them once and the JSON parser then fails: `[servable_initializer.cpp:203] Error during llm node plugin_config option parsing to JSON: ...`.
- For speculative decoding, the same `plugin_config` likely needs to be set for the draft model as well (not yet verified empirically).

**Proposed fix**

In the LLM `servable_initializer.cpp` path, when the global `--cache_dir` is set and `plugin_config["CACHE_DIR"]` is not already provided, inject it into the `pluginConfig` map before constructing the `ContinuousBatchingPipeline`. This would restore the 2025.4 contract for generative pipelines while keeping explicit `plugin_config` user overrides authoritative.

exzile:
## Summary

Fixes #4230.

The continuous batching servable initializer constructs the GenAI `ContinuousBatchingPipeline` directly and never applied the server-level `--cache_dir` (`ServerSettings.cacheDir`). Unlike the non-CB path — which applies it via `ModelInstance::setCacheOptions` (`ieCore.set_property(ov::cache_dir(...))`) — the CB path left model-compilation caching disabled unless the user manually duplicated the value into the node's `plugin_config` as `CACHE_DIR`. As a result, `.blob`/`.cl_cache` artifacts were never persisted and every server restart fully recompiled the model.

## Fix

In `ContinuousBatchingServableInitializer::initialize`, inject the global `cache_dir` into the pipeline `pluginConfig` (keyed by `ov::cache_dir.name()` == `CACHE_DIR`) right after parsing the node `plugin_config` and before constructing the pipeline. An explicit `CACHE_DIR` in the node's `plugin_config` remains authoritative.

```cpp
const std::string& globalCacheDir = Config::instance().cacheDir();
if (!globalCacheDir.empty()) {
    if (properties->pluginConfig.find(ov::cache_dir.name()) == properties->pluginConfig.end()) {
        properties->pluginConfig[ov::cache_dir.name()] = globalCacheDir;
    } // else: explicit node CACHE_DIR wins
}
```

This also applies to the VLM continuous-batching servable, which reuses this initializer.

## Testing

Added a regression test `LLMNodeOptionsCacheDirPropagation` (run for both the CB and VLM fixtures) covering:
1. The global `--cache_dir` is propagated into `pluginConfig["CACHE_DIR"]` when the node does not set it.
2. An explicit `CACHE_DIR` in the node `plugin_config` takes precedence over the global value.

Built and ran locally on Windows (MSVC) against a real `facebook/opt-125m` continuous-batching pipeline:

```
[       OK ] LLMOptionsHttpTest.LLMNodeOptionsCheckPluginConfig
[       OK ] LLMOptionsHttpTest.LLMNodeOptionsCacheDirPropagation
[  PASSED  ] 2 tests.
```


exzile:
### End-to-end verification on GPU

Beyond the unit test, I verified the actual caching behavior on an Intel Arc Pro B70 dGPU (`EXPORT_IMPORT` capable) serving `facebook/opt-125m` via continuous batching with `device: "GPU"` and a `--cache_dir` pointing at an initially-empty directory:

**Run 1 (empty cache):**
- Log confirms the fix path: `servable_initializer.cpp] Applying global cache_dir to continuous batching pipeline: .../cachedir`
- Cache dir went from empty → `*.cl_cache` files + a 130 MB `.blob` (compiled model)
- Pipeline init → AVAILABLE: ~1.6 s

**Run 2 (restart, populated cache):**
- Same fix path fires
- No recompile: blob is reused (file unchanged), no new artifacts written
- Pipeline init → AVAILABLE: ~0.24 s (~6.7× faster cold start)

Without the change the cache directory stays empty on the CB path and every restart recompiles. With it, blobs persist and are reused across restarts as intended.

src/test/llm/llmnode_test.cpp
```diff
@@ -4425,6 +4426,201 @@ TEST_F(LLMVLMOptionsHttpTest, LLMVLMNodeOptionsCheckPluginConfig) {
...
+        // The test harness may rewrite the path for the target platform, so match
+        // on substrings: the explicit node value must win over the global one.
+        std::string nodeCacheDir = properties->pluginConfig["CACHE_DIR"].as<std::string>();
+        ASSERT_NE(nodeCacheDir.find("ovms_node_cache"), std::string::npos) << "Explicit node CACHE_DIR should be used, got: " << nodeCacheDir;
+        ASSERT_EQ(nodeCacheDir.find("ovms_global_cache"), std::string::npos) << "Global cache_dir must not override explicit node CACHE_DIR, got: " << nodeCacheDir;
+    }
+
+    // Restore the global cache_dir so the singleton does not leak into other tests.
+    char* reset_argv[] = {(char*)"ovms", (char*)"--model_path", (char*)"/path/to/model", (char*)"--model_name", (char*)"some_name", (char*)"--rest_port", (char*)"8080"};
+    ovms::Config::instance().parse(7, reset_argv);
```
dkalinowski: we probably need this restore also after any ASSERT in this suite. check how other tests do that via guard patter or SetUp/TearDown mechanic (some tests recover env variables this way)

src/llm/language_model/continuous_batching/servable_initializer.cpp
```diff
@@ -227,6 +227,21 @@ Status ContinuousBatchingServableInitializer::initialize(std::shared_ptr<GenAiSe
         return status;
     }
 
+    // Propagate the global --cache_dir (ServerSettings) into the continuous batching
+    // pipeline plugin config. Unlike the non-CB path (ModelInstance::setCacheOptions),
+    // this initializer constructs the pipeline directly, so the server-level cache_dir
+    // is otherwise never applied. An explicit CACHE_DIR in the node's plugin_config
+    // remains authoritative.
+    const std::string& globalCacheDir = Config::instance().cacheDir();
```
dkalinowski: This fixes it for language_model + continuous_batching

what about: 
- visual_language_model + legacy
- visual_language_model + continuous_batching
- language_model + legacy
?

src/test/llm/llmnode_test.cpp
exzile: @dkalinowski Added a GlobalCacheDirGuard RAII struct that restores the Config singleton (and removes the temp cache dir) on scope exit, so a failed ASSERT_* mid-test can no longer leak the modified --cache_dir into later tests. Used it in both cache tests; removed the manual end-of-test restores.
Let me know if you want to change this or go a different route.

src/llm/language_model/continuous_batching/servable_initializer.cpp
exzile: Investigated all four paths. VLM_CB already shares the same ContinuousBatchingServableInitializer, so it was already covered.
**The two legacy paths (LM and VLM) construct GenAI pipelines directly and had the same bug — global --cache_dir was never applied.

Extracted the propagation into a shared GenAiServableInitializer::applyGlobalCacheDir() helper (matching the existing loadChatTemplate static pattern), now called by all three initializers (base, CB, LM-legacy, VLM-legacy). This eliminates the duplication and covers every path through one tested code path.

This should suffice, if not we can rewire it a different way. @dkalinowski

rasapala:
Fixed based on the contribution https://github.com/openvinotoolkit/model_server/pull/4329 in  https://github.com/openvinotoolkit/model_server/pull/4378. I am closing the PR. @exzile  Thank you for the contribution.

Issue #4221, opened 2026-05-19
Allow server-side default for num_assistant_tokens in LLMCalculatorOptions


korund:
## Summary

Make the default value of `num_assistant_tokens` (used in speculative-decoding pipelines when the client omits the field) configurable per-deployment in `graph.pbtxt`, instead of always falling back to the hardcoded `5` in `src/llm/io_processing/base_generation_config_builder.cpp`.

## Motivation

The optimal speculative token count depends on the specific main + draft model pair, the target hardware, and the workload profile. This is a **per-deployment** value: once an operator measures the sweet spot for their pair, it stays stable until they change the model lineup. Persisting it in the same configuration that already pins models, devices, and other LLMCalculator parameters is the natural location.

The current model — "either accept the hardcoded `5`, or send `num_assistant_tokens` on every request" — has two practical problems:

1. **Separation of concerns.** From a client's perspective, `/v3/chat/completions` is a chat completion request; the existence of a draft model and the count of speculative tokens is server-side topology the client shouldn't need to know. Asking every client to advertise this parameter leaks deployment internals upward.

2. **Tooling friction.** `num_assistant_tokens` is an OVMS extension over the OpenAI API. Standard OpenAI SDKs (`openai-python`, `openai-node`, LangChain, most IDE plugins) either drop unknown body fields or require a fork to add them. So "configure your client" is, in practice, "fork the SDK." This is disproportionate cost for a value that doesn't change across requests in a given deployment.

## Counter to the existing position

PR #2777 review explicitly declined to add server-side defaults with the reasoning: *"I didn't specify any default because it's hard to recommend a single value that would work well for different combinations of main and draft models.."*

That reasoning argues against a **hardcoded universal default** — and we agree, no single value fits all combinations. But the project already ships exactly such a hardcoded value (`5` in `base_generation_config_builder.cpp:34`). The current state combines the worst of both: a non-overridable value that the maintainers themselves don't endorse as a recommendation.

A per-deployment configurable default doesn't contradict the maintainers' position; it implements it. The right value is determined by whoever runs the server, once, against their actual model pair — not by upstream and not by every client downstream.

## Proposed change

Add an optional field to `LLMCalculatorOptions` in `src/llm/llm_calculator.proto`, for example:

```proto
optional uint32 default_num_assistant_tokens = 25;
```

Resolution order in `base_generation_config_builder.cpp` becomes:

1. Value from request body (`num_assistant_tokens`), if present.
[...]

PR #4330, opened 2026-06-26
Add per-LLM-node generation_config.json path override (#4233)

### LINKED ISSUE #4233: generation_config.json path override per LLM node
**Component:** LLM continuous batching, `LLMCalculatorOptions` / mediapipe registration
**OVMS version:** 2026.1.0.72cc0624 (OpenVINO backend 2026.1.0, OpenVINO GenAI backend 2026.1.0.0)

**Context**

When several deployments share the same on-disk model directory but need different generation defaults (e.g. different `num_assistant_tokens`, `temperature`, or sampling settings per served endpoint), the only current option is to duplicate the model directory — including the weights — because OVMS reads `generation_config.json` from a fixed name inside `models_path`. For multi-gigabyte LLMs this is impractical.
The same problem exists for `graph.pbtxt`, but is already solved there: `graph_path` in the mediapipe config entry lets one model directory back several deployments with different graphs. There is no equivalent for `generation_config.json`.

Related to #4221 

**Question**

Would it be feasible to add a per-LLM-node override for the generation-config file path — analogous to `graph_path`? A natural shape would be either:
- a `generation_config_path` field in `LLMCalculatorOptions` (next to `models_path`), absolute or relative to `models_path`; or
- a sibling field at the mediapipe config-entry level (next to `graph_path`).
From a quick read of openvino.genai, `ContinuousBatchingPipeline` accepts an optional `GenerationConfig` at construction and exposes `set_config()` post-construction, so the underlying mechanism appears to be already in place. The work seems contained within `src/llm/language_model/continuous_batching/servable_initializer.cpp` on the OVMS side.

**Use case**

Multiple served names backed by the same model weights, each with its own generation defaults. Without per-entry generation-config selection, each variant requires a full copy of the model directory on disk.

**Open questions**

- Is there a reason this hasn't been exposed yet — for example, a planned different mechanism (per-deployment overrides through some other channel), or an interaction with model auto-detection/conversion that I'm missing?
- Is one of the placement options preferred from the architecture side?

exzile:
## Summary

Closes #4233.

Adds an optional `generation_config_path` field to `LLMCalculatorOptions`, so several deployments backed by the same model weights can use different generation defaults without duplicating the model directory. This mirrors how `graph_path` already lets one model directory back several deployments with different graphs.

## Behavior

- **Unset** → `generation_config.json` from `models_path`, exactly as before (no change).
- **Set** → absolute, or relative to `models_path` (resolved against the model directory, or its parent when `models_path` points at a file such as a GGUF).
- An explicit path that does not exist is a load error (fail fast rather than silently falling back).

A shared `resolveGenerationConfigPath` helper is used by both the continuous-batching and legacy initializers, so behavior is identical across pipeline types (and the VLM CB servable, which reuses the CB initializer).

## Placement rationale

Per the discussion in #4233 (and the config taxonomy from #4221): `graph_path` already lets one model directory back several deployments with different node options, so putting the override in the node options keeps the whole per-deployment configuration in `graph.pbtxt`.

## Testing

Added `LLMGenerationConfigPath.ResolveGenerationConfigPath` covering default, absolute, relative, and missing-path cases. Built and ran locally on Windows (MSVC).


exzile:
@dkalinowski rebased and conflict-free now (additive test collision with main's cache_dir tests — both kept). Adds a per-LLM-node `generation_config.json` override (#4233). Ready for review when you get a chance.

src/llm/servable_initializer.cpp
```diff
@@ -443,6 +443,30 @@ Status parseModelsPath(std::string& outPath, std::string modelsPath, std::string
...
+    // Explicit per-node override. A relative path is resolved against models_path
+    // (its parent directory when models_path points at a file, e.g. a GGUF).
+    std::filesystem::path overridePath(nodeOptions.generation_config_path());
+    if (overridePath.is_relative()) {
+        std::filesystem::path base(parsedModelsPath);
+        if (!std::filesystem::is_directory(base)) {
+            base = base.parent_path();
+        }
+        overridePath = base / overridePath;
+    }
```
mzegla: This looks like something we could extract to a separate utility function as it serves quite generic purpose. 
There is already another open PR that adds similar logic:
https://github.com/openvinotoolkit/model_server/pull/4385/changes#diff-610bd2c9c70ee9b3c3a838abdcc1bf18dcf9141481dcbd3d0e10f531b8d79efeR245.

mzegla: @exzile - can we do such extraction to utils file? 

Also, tests failed on:
```
[ RUN      ] LLMGenerationConfigPath.ResolveGenerationConfigPath
src/test/llm/llmnode_test.cpp(4535): error: Expected equality of these values:
  std::filesystem::path(outPath)
    Which is: "C:\\Users\\<username>\\AppData\\Local\\Temp\\ovms_gencfg_test\\model\\generation_config.json"
  overrideDir / "custom_generation_config.json"
    Which is: "C:\\Users\\<username>\\AppData\\Local\\Temp\\ovms_gencfg_test\\overrides\\custom_generation_config.json"
[  FAILED  ] LLMGenerationConfigPath.ResolveGenerationConfigPath (2 ms)
```

Looks like this function does not update `outPath` on override.

exzile: Will do.

mzegla (commented):
@exzile 
Generation config is relevant also for VLM and Omni pipelines (@dkalinowski ).
I believe their respective servable initializers should also include the generation config path resolution logic so we are not creating a gap.

PR #4331, opened 2026-06-26
Support add_generation_prompt request parameter for chat completions (#3877)

### LINKED ISSUE #3877: Feature Request: Allow Assistant Messages to Act as Generation Prefix (Prefill)
This issue proposes adding support for assistant prefill: allowing an `assistant` message to be provided as a prefix that the model should continue generating from, rather than treating it as a completed assistant turn.

This capability enables deterministic continuation, structured output anchoring, and server-controlled tool or schema-guided generation.

## Current Behavior

When passing messages like:

```python
messages = [
    {"role": "user", "content": "Write a short apology email to a customer for a delayed shipment."},
    {"role": "assistant", "content": "Hi John,\n\nI'm sorry for the delay with your order. "}
]
```

the model server currently serializes the assistant message as a completed turn (`constexpr bool add_generation_prompt = true;`) and then starts a new assistant turn:

```
Pipeline input text: <|im_start|>user
Write a short apology email to a customer for a delayed shipment.<|im_end|>
<|im_start|>assistant
Hi John,

I'm sorry for the delay with your order. <|im_end|>
<|im_start|>assistant
```

As a result, the provided assistant content cannot be used as the active generation prefix.

## Expected Behavior

The assistant message should be treated as a partial prefix, and generation should continue immediately after it:

```
<|im_start|>user
Write a short apology email to a customer for a delayed shipment.<|im_end|>
<|im_start|>assistant
Hi John,

I'm sorry for the delay with your order. 
```

### Structured Output Prefill

```python
messages = [
    {
        "role": "user",
        "content": "Is the customer satisfied? Respond in JSON with fields \"reasoning\" and \"answer\"."
    },
    {
        "role": "assistant",
        # Prefills the JSON shape and anchors a concise reasoning style.
        "content": '{\n  "reasoning": "Based on the user\'s tone and wording, '
    }
]
```

This allows the server to enforce output structure while still letting the model complete the response naturally.

### Tool-Guided Generation

```python
messages = [
    {
        "role": "user",
        "content": "Tell the user that their package is delayed by 2 days."
    },
    {
        "role": "assistant",
        # Tool name and immutable arguments are injected by the system.
        # The model only needs to generate the remaining message content.
        "content": '{"tool": "send_notification", "args": {"user_id": "uid_541", "message": "'
    }
]
```

This pattern enables server-controlled tool configuration, while allowing the model to complete the remaining schema-constrained fields.

exzile:
## Summary

Closes #3877.

The chat template was always rendered with `add_generation_prompt=true`, hardcoded in every servable. This exposes an optional `add_generation_prompt` field (bool, default `true`) on `/v3/chat/completions`, matching HF transformers and vLLM. When `false`, the trailing generation prompt is omitted — the building block for assistant prefill.

## Changes

- Parse and validate `add_generation_prompt` in the request handler; store it on the request struct.
- Honor it at every chat-template application site: the MINJA path (LLM and VLM continuous batching, legacy) and the Python-Jinja path (read from the request body and passed into the template render).
- Replaces the previously hardcoded `add_generation_prompt = true`.

## Testing

Added tests covering default (generation prompt added) and `false` (generation prompt omitted, assistant message preserved).

Verified end-to-end on the MINJA path with `HuggingFaceTB/SmolLM2-360M-Instruct`:
- default → prompt ends with `<|im_start|>assistant`
- `add_generation_prompt: false` → that trailing generation prompt is omitted

## Scope note

This implements `add_generation_prompt` only. Full assistant *prefill* — continuing from the final assistant message without closing it (`continue_final_message` in transformers/vLLM) — is a separate control that the genai C++ `apply_chat_template` does not currently expose, and is left as a follow-up.


src/llm/apis/openai_api_handler.cpp
```diff
@@ -664,6 +664,17 @@ absl::Status OpenAIApiHandler::parseCommonPart(std::optional<uint32_t> maxTokens
         request.ignoreEOS = it->value.GetBool();
     }
 
+    // add_generation_prompt: bool; optional - defaults to true
+    // Extension, unsupported by OpenAI API, however supported by HF transformers and vLLM.
+    // When false, the chat template is rendered without a trailing generation prompt
+    // so a final assistant message can be continued as a prefix (assistant prefill).
+    it = doc.FindMember("add_generation_prompt");
```
mzegla: Is it valid and functional for Responses API? 
If so, please also update docs for responses endpoint.

exzile: Yes — it's valid and functional for the Responses API. `add_generation_prompt` is parsed in the common request path (no `RESPONSES` guard), and the `Endpoint::RESPONSES` branch in `src/llm/servable.cpp` reads `request.addGenerationPrompt` and passes it to `apply_chat_template` (lines 243/255), same as chat completions; the Python-Jinja path picks it up from the request JSON in `py_jinja_template_processor.cpp` as well. Updated the responses endpoint docs (`docs/model_server_rest_api_responses.md`) to document it. Pushed in 8c850e71.

mzegla:
Looks like a rebase with conflict resolution is needed.

src/llm/io_processing/input_request.hpp
```diff
@@ -37,6 +37,7 @@ using InputPayload = std::variant<ov::genai::ChatHistory, std::string>;
 struct InputRequest {
     InputPayload input;                            // set in parseRequest()
     ov::genai::GenerationConfig generationConfig;  // set in parseRequest()
+    bool addGenerationPrompt = true;               // set in parseRequest(); from the request's add_generation_prompt (default true)
```
mzegla: Actually, now that I'm looking at it - do we really need to have `addGenerationPrompt` as a separate member anywhere?
Since it's only used in chat template application why couldn't we treat it as part of `chat_template_kwargs`?

exzile: @mzegla  I had claude look into to see if your point stood. It looks like its ok to stay:

1. Correct default requires an explicit field. Jinja templates gate on {% if add_generation_prompt %}; if the variable is absent, that's falsy. So we need to always pass add_generation_prompt=true by default. If it only existed inside optional chat_template_kwargs, omitting it would silently flip the default to false — a behavior change for every existing caller, not just an API nicety.


2.  The MINJA path doesn't take it via kwargs at all. For the Python-Jinja path it's a render() keyword, but MINJA's apply_chat_template takes add_generation_prompt as its own dedicated argument, separate from the extra-context/kwargs map (chatHistory.get_extra_context()). A single normalized bool on InputRequest lets both template paths consume it uniformly, instead of MINJA re-extracting it out of a generic kwargs map each time.
  
3. Collision risk if merged. If add_generation_prompt could also be supplied inside chat_template_kwargs, the Python-Jinja call render(..., add_generation_prompt=X, **chat_template_kwargs) would raise "got multiple values for keyword argument" whenever a user set both. Keeping it a distinct top-level field avoids that ambiguity entirely.
 
4. It matches the OpenAI/vLLM/TGI convention — add_generation_prompt is a top-level chat-completions parameter in that ecosystem (and is what #3877 specifically asked for), not something nested under a generic kwargs bag. Keeping it top-level here matches user expectations from other servers.

So functionally it's used only in chat-template application, but I think it deserves to stay a named field on InputRequest rather than being routed through chat_template_kwargs — happy to reconsider if you see a cleaner way to guarantee the default-true behavior without a dedicated field, though.

mzegla: Logically, `add_generation_prompt` belongs to chat template arguments in that case since it's only consumer is chat template rendering logic. Treating it as a separate field comes from how vLLM does it. If I'm not missing anything OpenAI does not specify it at all, and solutions like Llama.cpp or SGLang do not allow request level setting for this value. 

I don't think we have to blindly follow implementations from other projects if it's not a standard across the industry. 
From my point of view chat template kwargs map is a good place for this parameter even if some implementations for apply_chat_template require it as a separate argument.

Re-extracting this value for minja is fine by me as it's the one time, small operation with negligible impact and it allows us to keep, in my opinion, more consistent logic.

exzile: Makes sense — updated. `add_generation_prompt` is now folded into `chat_template_kwargs` at parse time (default `true` when unset); the MINJA path extracts it back out for genai's dedicated argument, and the Python-Jinja path pops it from the dict before splatting into `render()` to avoid a duplicate-keyword collision. This also surfaced a latent bug — the Python-Jinja path had never actually wired `add_generation_prompt` through at all before this change, since it was only ever read from a top-level field that `serializeForPyJinja` never emitted. Fixed, and added/updated tests (`ChatTemplateAddGenerationPromptDefaultsTrue`/`False` in `llmtemplate_test.cpp`, which also had a stale 4-arg signature that never compiled).

src/llm/io_processing/input_request.hpp
```diff
@@ -37,6 +37,8 @@ using InputPayload = std::variant<ov::genai::ChatHistory, std::string>;
 struct InputRequest {
     InputPayload input;                            // set in parseRequest()
     ov::genai::GenerationConfig generationConfig;  // set in parseRequest()
+    // add_generation_prompt is folded into ChatHistory's extra_context (chat_template_kwargs)
+    // by extractInputRequest() since it is only ever consumed by chat template rendering.
```
mzegla:
```suggestion
```

src/llm/io_processing/input_processors/chat_template_processor.cpp
```diff
@@ -82,9 +81,16 @@ absl::Status ChatTemplateProcessor::process(InputRequest& req) {
...
-        constexpr bool addGenerationPrompt = true;
         const auto& tools = chatHistory.get_tools();
-        const auto& kwargs = chatHistory.get_extra_context();
+        // add_generation_prompt lives inside chat_template_kwargs; MINJA's apply_chat_template
+        // takes it as a dedicated argument, so extract it here and drop it from the kwargs map
+        // passed alongside so it isn't supplied twice.
+        ov::genai::JsonContainer kwargs = chatHistory.get_extra_context();
+        bool addGenerationPrompt = true;
+        if (kwargs.contains("add_generation_prompt")) {
+            addGenerationPrompt = kwargs["add_generation_prompt"].get_bool();
```
mzegla: Is it safe to assume that value will be `boolean`? 
If request has `"add_generation_prompt": "some text"` will it be rejected earlier? Do we have such check?

src/llm/apis/openai_api_handler.cpp
```diff
@@ -525,6 +528,17 @@ absl::Status OpenAIApiHandler::parseCommonPart(std::optional<uint32_t> maxTokens
...
+    if (it != doc.MemberEnd() && !it->value.IsNull()) {
+        if (!it->value.IsBool())
+            return absl::InvalidArgumentError("add_generation_prompt accepts values true or false");
+        request.addGenerationPrompt = it->value.GetBool();
+    }
```
mzegla: Is it still needed now that we expect `add_generation_prompt` to live in `chat_template_kwargs`?
It will be loaded via `parseChatTemplateKwargsToJsonContainer()` and land in `ChatHistory`.

src/llm/apis/openai_api_handler.cpp
```diff
@@ -358,9 +358,12 @@ absl::StatusOr<InputRequest> OpenAIApiHandler::extractInputRequest(GenerationCon
...
         }
-        if (kwargsResult.value().has_value()) {
-            chatHistory.set_extra_context(kwargsResult.value().value());
-        }
+        // add_generation_prompt is only ever consumed by chat template rendering, so it is
+        // folded into chat_template_kwargs rather than kept as a separate field, and both the
+        // MINJA and Python-Jinja rendering paths read it from this single source.
+        ov::genai::JsonContainer kwargs = kwargsResult.value().value_or(ov::genai::JsonContainer::object());
+        kwargs["add_generation_prompt"] = request.addGenerationPrompt.value_or(true);
+        chatHistory.set_extra_context(kwargs);
```
mzegla: We run `parseChatTemplateKwargsToJsonContainer` before which is supposed to read all chat template kwargs. 
This logic seems like a leftover / mix of the two solutions 
load_kwargs -> read_kwargs -> overwrite add_generation_prompt with the value extracted in another place -> save kwargs

I would say it is not needed anymore.

docs/model_server_rest_api_chat.md
```diff
@@ -222,6 +222,7 @@ Some parameters, especially related to sampling (like `temperature`, `top_p` etc
...
 | skip_special_tokens | ✅ | ❌ | ✅ | bool (default: `true`) | Whether to remove special tokens (e.g. `<\|endoftext\|>`, `<\|im_end\|>`) from the generated output. Set to `false` to include them, which is useful when the model uses special tokens to encode structured information (e.g. bounding boxes, reasoning markers). When `false`, any tool or reasoning parser configured on the endpoint is silently disabled for the request, so the raw token stream is returned. This option works with most detokenizers exported with OpenVINO Tokenizers 2024.5 or later, unless they are based on custom ops. |
+| add_generation_prompt | ✅ | ❌ | ✅ | bool (default: `true`) | Whether to append the chat template's generation prompt (the marker that signals the model to start a new assistant turn). Set to `false` to render the conversation without a trailing generation prompt — useful for assistant prefill where the final `assistant` message should be continued rather than treated as a completed turn. Applies to both the Python-Jinja and MINJA chat template paths. |
```
mzegla: Since we are bringing this parameter under chat template kwargs, they should not be treated separately here in the documentation. Please remove the entry and optionally mention it in `chat_template_kwargs`.

docs/model_server_rest_api_responses.md
```diff
@@ -106,6 +106,7 @@ curl http://localhost/v3/responses \
...
 | skip_special_tokens | ✅ | ❌ | bool (default: `true`) | Whether to remove special tokens (e.g. `<\|endoftext\|>`, `<\|im_end\|>`) from the generated output. Set to `false` to include them, which is useful when the model uses special tokens to encode structured information (e.g. bounding boxes, reasoning markers). When `false`, any tool or reasoning parser configured on the endpoint is silently disabled for the request, so the raw token stream is returned. This option works with most detokenizers exported with OpenVINO Tokenizers 2024.5 or later, unless they are based on custom ops. |
+| add_generation_prompt | ✅ | ❌ | bool (default: `true`) | Whether to append the chat template's generation prompt (the marker that signals the model to start a new assistant turn). Set to `false` to render the conversation without a trailing generation prompt — useful for assistant prefill where the final `assistant` message should be continued rather than treated as a completed turn. Applies to both the Python-Jinja and MINJA chat template paths. |
```
mzegla: Since we are bringing this parameter under chat template kwargs, they should not be treated separately here in the documentation. Please remove the entry and optionally mention it in `chat_template_kwargs`.

mzegla (commented):
Please have a look at those few comments. Once we have it resolved I think we will be good to run it through the CI and merge.

exzile:
@mzegla tended to all those comments in the latest commit.
Removed the leftover top-level add_generation_prompt field/parsing (now lives solely in chat_template_kwargs).
Fixed a real bug: the MINJA path used get_bool() outside its try/catch, which would throw an unhandled ov::Exception on a non-boolean value instead of a clean error — now uses as_bool() and returns InvalidArgumentError.
Trimmed the leftover comment line in input_request.hpp per the inline suggestion.
Updated both REST docs pages to document add_generation_prompt under chat_template_kwargs instead of as a separate parameter.
Added two unit tests (false-case and invalid-type rejection).

src/llm/apis/openai_api_handler.cpp
```diff
@@ -359,11 +359,11 @@ absl::StatusOr<InputRequest> OpenAIApiHandler::extractInputRequest(GenerationCon
             return kwargsResult.status();
         }
         // add_generation_prompt is only ever consumed by chat template rendering, so it is
-        // folded into chat_template_kwargs rather than kept as a separate field, and both the
-        // MINJA and Python-Jinja rendering paths read it from this single source.
-        ov::genai::JsonContainer kwargs = kwargsResult.value().value_or(ov::genai::JsonContainer::object());
-        kwargs["add_generation_prompt"] = request.addGenerationPrompt.value_or(true);
-        chatHistory.set_extra_context(kwargs);
+        // read directly out of chat_template_kwargs (both the MINJA and Python-Jinja rendering
+        // paths default it to true when absent) rather than kept as a separate request field.
```
mzegla: I don't really like that `add_generation_prompt` stands out here. Kwargs can carry more parameters and I would not specifically mention any of them in the comments here. I think we can remove this comment entirely.

src/llm/io_processing/input_request.hpp
```diff
@@ -37,8 +37,7 @@ using InputPayload = std::variant<ov::genai::ChatHistory, std::string>;
 struct InputRequest {
     InputPayload input;                            // set in parseRequest()
     ov::genai::GenerationConfig generationConfig;  // set in parseRequest()
-    // add_generation_prompt is folded into ChatHistory's extra_context (chat_template_kwargs)
-    // by extractInputRequest() since it is only ever consumed by chat template rendering.
+    // add_generation_prompt is folded into ChatHistory's extra_context (chat_template_kwargs).
```
mzegla: drop this comment entirely - only explicit metion of add_generation_prompt could be in chat kwargs description in the docs that you already added

src/llm/io_processing/input_processors/chat_template_processor.cpp
```diff
@@ -82,9 +81,21 @@ absl::Status ChatTemplateProcessor::process(InputRequest& req) {
...
+        bool addGenerationPrompt = true;
+        if (kwargs.contains("add_generation_prompt")) {
+            const auto asBool = kwargs["add_generation_prompt"].as_bool();
+            if (!asBool.has_value()) {
+                return absl::Status(absl::StatusCode::kInvalidArgument,
+                    "add_generation_prompt accepts values true or false");
+            }
+            addGenerationPrompt = asBool.value();
+            kwargs.erase("add_generation_prompt");
+        }
```
dkalinowski: please separate it to method
```suggestion
        // add_generation_prompt lives inside chat_template_kwargs; MINJA's apply_chat_template
        // takes it as a dedicated argument, so extract it here and drop it from the kwargs map
        // passed alongside so it isn't supplied twice.
        bool addGenerationPrompt = true;
        auto status = adjustExtraContext(chatHistory, addGenerationPrompt);
        if (!status.ok())
            return status;
```

dkalinowski:
@exzile Do you know about any existing app that uses this feature (add_generation_prompt set to false)? If so, it would be worth aligning to the API - it could point us if this param should be in chat_template_extra_kwargs field or level above.

exzile:
Checked the two serving frameworks that actually implement this (checked their current source, not just docs):

**vLLM** ([`vllm/entrypoints/openai/chat_completion/protocol.py`](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/chat_completion/protocol.py#L282-L335)) declares both as **top-level** `ChatCompletionRequest` fields, siblings to `chat_template_kwargs`, not nested inside it:
```python
add_generation_prompt: bool = Field(default=True, ...)
continue_final_message: bool = Field(default=False, ...)  # the actual prefill toggle; mutually exclusive with add_generation_prompt
...
chat_template_kwargs: dict[str, Any] | None = Field(default=None, ...)
```
It only merges them into the kwargs dict internally, right before calling the Jinja renderer — that's an implementation detail of how vLLM talks to the template, not the client-facing contract:
```python
chat_template_kwargs=merge_kwargs(
    self.chat_template_kwargs,
    dict(add_generation_prompt=self.add_generation_prompt, continue_final_message=self.continue_final_message, ...),
)
```

**SGLang** ([`sglang/srt/entrypoints/openai/protocol.py`](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/entrypoints/openai/protocol.py)) follows the same shape: `continue_final_message: bool = False` is a top-level field, separate from `chat_template_kwargs: Optional[Dict] = None`. SGLang doesn't expose `add_generation_prompt` at all — `continue_final_message` is its only prefill control.

**llama.cpp server** doesn't expose `add_generation_prompt` (or an equivalent) as a chat-completions request parameter at all — only the generic `chat_template_kwargs` passthrough exists there, so no data point either way.

So the two real implementations of this feature both keep it as a top-level sibling to `chat_template_kwargs`, and only fold it in internally at render time. That's actually what this PR had before this round moved it under `chat_template_kwargs` — worth reconsidering going back to a top-level field to match the existing convention, if that's the direction you'd like. Happy to make that change if so.

mzegla (approved):

dkalinowski (approved):

Issue #4178, opened 2026-05-05
OVMS 2026.1-gpu fails to load OpenVINO/gemma-4-E4B-it-int4-ov with Unsupported 'gemma4' VLM model type


jthakare17:
**Describe the bug**
OVMS fails to load the VLM model `OpenVINO/gemma-4-E4B-it-int4-ov` and exits during initialization with:

```text
Unsupported 'gemma4' VLM model type
```

This happens when running `openvino/model_server:2026.1-gpu`.

The behavior appears inconsistent with the current `openvinotoolkit/openvino.genai` source, where `gemma4` is explicitly recognized in VLM config parsing.

**To Reproduce**
1. Prepare a model repository containing:
   - `models/OpenVINO/gemma-4-E4B-it-int4-ov/`
   - The model used is `OpenVINO/gemma-4-E4B-it-int4-ov`

2. Launch OVMS:
   ```bash
   docker run -it --rm \
     -p 8000:8000 \
     --device /dev/dri \
     --group-add=$(stat -c "%g" /dev/dri/render* | head -n 1) \
     -v $(pwd)/models:/models:ro \
     openvino/model_server:2026.1-gpu \
     --rest_port 8000 \
     --model_name OpenVINO/gemma-4-E4B-it-int4-ov \
     --model_path ./models/OpenVINO/gemma-4-E4B-it-int4-ov/
   ```

3. No client request is needed. The failure happens during model initialization at server startup.

4. Observe the error:
   ```text
   Error during llm node initialization ...
   Unsupported 'gemma4' VLM model type
   ```

**Expected behavior**
OVMS should recognize `model_type: "gemma4"` and initialize the model successfully, or the image/docs should clearly state that Gemma4 is not supported by this OVMS image.

**Logs**
Startup version info:
```text
[2026-05-05 08:15:34.612][1][serving][info][server.cpp:88] OpenVINO Model Server 2026.1.0.72cc06244
[2026-05-05 08:15:34.612][1][serving][info][server.cpp:89] OpenVINO backend 2026.1.0-21367-63e31528c62-releases/2026/1
[2026-05-05 08:15:34.612][1][serving][info][server.cpp:92] OpenVINO GenAI backend 2026.1.0.0-2957-1dabb8c2255
```

Relevant failure logs:
```text
[2026-05-05 08:06:22.232][1][modelmanager][info][servable_initializer.cpp:448] Initializing Visual Language Model Continuous Batching servable
[...]

PR #4336, opened 2026-06-28
Fall back to legacy pipeline when CB model lacks PagedAttention (e.g. Gemma)

exzile:
### Problem

Models whose OpenVINO IR exports without a `ScaledDotProductAttention` op — e.g. **Gemma / gemma4** (`Gemma4ForConditionalGeneration`) — cannot undergo the `SDPAToPagedAttention` transformation required by the continuous batching engine.

With the default `pipeline_type: AUTO`, OVMS selects a continuous-batching pipeline (`LM_CB`/`VLM_CB`) and then fails hard during LLM node initialization, taking the **whole server down** with a cryptic error:

```
Check 'ov::op::util::has_op_with_type<ov::op::v13::ScaledDotProductAttention>(model)' failed at
src/core/src/pass/sdpa_to_paged_attention.cpp:81:
No ScaledDotProductAttention operation observed in the graph, cannot perform the SDPAToPagedAttention transformation.
```

This is the trap behind #4178 — the documented workaround is to manually set `pipeline_type: VLM`, but a user following defaults just gets a dead server.

### Change

- Add a dedicated status `LLM_NODE_PAGED_ATTENTION_NOT_SUPPORTED`.
- In the continuous batching initializer, detect this specific failure (matching the `ScaledDotProductAttention` signal) and return the new status, logged at **warning** level instead of error.
- In `initializeGenAiServable`, when the pipeline type was **auto-selected**, fall back to the legacy pipeline (`LM_CB`→`LM`, `VLM_CB`→`VLM`).
- An **explicitly** requested `*_CB` pipeline still surfaces the error rather than being silently downgraded.

### Behavior after fix

```
[warning] Model at models_path: .../gemma-4-E4B-it-int4-ov does not support PagedAttention required by continuous batching: ...
[warning] Model does not support PagedAttention, falling back to Visual Language Model Legacy servable
[info]    Mediapipe: gemma4 state changed to: AVAILABLE
```

### Testing

Manually verified with `OpenVINO/gemma-4-E4B-it-int4-ov` on OVMS built from this branch (OpenVINO/GenAI 2026.3): with default `AUTO`, the server now falls back to the legacy VLM servable and serves `/v3/chat/completions` successfully, instead of failing to start.

No automated test is included — exercising the fallback needs a no-SDPA (Gemma-class) model in the test model set, which `windows_prepare_llm_models.bat` / the export tooling does not currently produce. Happy to add coverage if a suitable small fixture model can be agreed on.

### Notes

This restores a working default path; it does **not** enable continuous batching for Gemma — that remains gated on OpenVINO core adding PagedAttention support for the Gemma architecture.

Related: #4178


src/llm/language_model/continuous_batching/servable_initializer.cpp
```diff
@@ -253,7 +253,16 @@ Status ContinuousBatchingServableInitializer::initialize(std::shared_ptr<GenAiSe
             properties->pluginConfig, properties->tokenizerPluginConfig);
         properties->tokenizer = properties->pipeline->get_tokenizer();
     } catch (const std::exception& e) {
-        SPDLOG_ERROR("Error during llm node initialization for models_path: {} exception: {}", parsedModelsPath, e.what());
+        // Some architectures (e.g. Gemma) export without a ScaledDotProductAttention op, so the
+        // SDPAToPagedAttention transformation required by continuous batching cannot be applied.
+        // Report this distinctly (at warning level) so callers can fall back to the legacy pipeline
+        // when appropriate, rather than logging it as a hard error on an otherwise-recoverable path.
+        const std::string error = e.what();
+        if (error.find("ScaledDotProductAttention") != std::string::npos) {
```
mzegla, line 267: Not sure if error message based fallback decision is a good path. More suitable solution would be some model analysis I suppose. Such mechanism could be created and exposed on GenAI level.

mzegla:
What hardware was used for testing? GPU or CPU? What models and specs?

exzile:
@mzegla 
Testing details:

Hardware: CPU only — AMD Ryzen 9 9950X3D (16C/32T), 128 GB RAM, Windows. (The box also has an Arc Pro B70 GPU, but this change was not exercised on GPU — see the note below on why it shouldn't matter.)

Software: OVMS built from main (2026.3.0, commit https://github.com/openvinotoolkit/model_server/commit/9ddd45c9a587ddb29a3c0ea21dd613906ebdfd5c), OpenVINO / GenAI 2026.3 (genai 2026.3.0.0-3173).

Model: OpenVINO/gemma-4-E4B-it-int4-ov (INT4, ~6.1 GB) — the model from https://github.com/openvinotoolkit/model_server/issues/4178.

What I verified end-to-end:

--pipeline_type AUTO (default) selects VLM_CB and, before this change, hard-fails with: No ScaledDotProductAttention operation observed in the graph, cannot perform the SDPAToPagedAttention transformation (sdpa_to_paged_attention.cpp). That's the gap this PR handles.
With the change, AUTO logs falling back to Visual Language Model Legacy servable, reaches state AVAILABLE, and successfully served a /v3/chat/completions request (prompt 23 / completion 28 tokens).
Explicit --pipeline_type VLM also serves successfully; explicit *_CB still surfaces the error (no silent override).
Caveats (being upfront):

One model, CPU only. The SDPAToPagedAttention transformation runs on the model graph before device compilation, so the "no SDPA op → fall back" decision is device-independent; I'd expect identical behavior on GPU, but I did not verify it there.
No unit test yet — it needs a small no-SDPA fixture model. Happy to add one if useful, alongside whichever detection approach we land on from the previous comment.

src/llm/language_model/continuous_batching/servable_initializer.cpp
exzile: Agreed — matching on the exception text is brittle (an upstream message change would silently break the fallback). I went with it as a minimal, well-contained guard (it only triggers when pipeline_type is AUTO, so an explicit *_CB request still surfaces the error), but a structural check is clearly better. Two ways to do the model analysis you're describing:

A. OVMS-side structural check (doable in this PR). Before constructing the CB pipeline, read the model and look for the op the SDPAToPagedAttention transformation needs:

auto model = core.read_model(parsedModelsPath / "openvino_model.xml");
bool hasSDPA = std::any_of(model->get_ops().begin(), model->get_ops().end(),
[](const auto& op){ return ov::is_typeov::op::v13::ScaledDotProductAttention(op); });
If absent → fall back without attempting CB. OVMS already reads models this way (sidepacket_servable.cpp) and scans ops (embeddings_servable.cpp). Cost is one extra model read at init. The only caveat is keeping the op/opset check in sync with what the transformation actually requires.

B. GenAI-level capability query (cleaner, your suggestion). A helper such as supports_paged_attention(const std::shared_ptrov::Model&) would be authoritative and would live naturally next to the existing has_linear_attention_states() / explicitly_requires_paged_attention() analysis in openvino.genai. OVMS (and any other consumer) would just query it. This is the right long-term home but is cross-repo.

I'm happy to do either: switch this PR to (A) now as a robust interim, and/or open a genai PR adding (B) and have OVMS consume it once available. Which do you prefer? If you're good with (A) as the near-term fix I'll update the PR; if you'd rather wait for (B), I can drive that upstream first.

mzegla: I would prefer going for B immediately - we do not need to rush with temporary A.

exzile: @mzegla  Sounds good — went with B. Opened https://github.com/openvinotoolkit/openvino.genai/pull/4060, which adds a supports_paged_attention(const std::shared_ptr<ov::Model>&) capability query (structural check for ScaledDotProductAttention ops, i.e. exactly what SDPAToPagedAttention needs) next to has_linear_attention_states(), plus a public OPENVINO_GENAI_EXPORTS wrapper so OVMS can query it.

Once that merges and the bundled GenAI version is bumped here, I'll update this PR to call ov::genai::supports_paged_attention(model) before constructing the CB pipeline and drop the exception-text match entirely. Happy to keep this PR open as the OVMS-side consumer, or close it and open a fresh one against the new GenAI once available — whichever you prefer.

exzile:
@mzegla friendly ping on this one. This is the OVMS side of the PagedAttention-capability discussion — the cross-repo helper you preferred is now up as openvinotoolkit/openvino.genai#4060 (`supports_paged_attention()`), freshly rebased clean on master. Happy to wire OVMS to consume it once the direction there settles. Anything you'd like changed here in the meantime?

mzegla:
@exzile I don't think so. The PA support detection must live in GenAI and OVMS should be only a consumer.
The detection mechanism might evolve as more architectures are enabled or PagedAttention operation implementation changes. If such logic is placed on serving level we would risk getting in outdated or incompatible with newer versions of OV and GenAI. 
If it's added in GenAI we would be happy to wire it in. Thanks for tracking this.

exzile:
Thanks @mzegla, that's clear and makes sense — detection belongs in GenAI, OVMS as consumer. I'll treat openvinotoolkit/openvino.genai#4060 as the home for it. There's an open design question there (whether to report support from *inside* `SDPAToPagedAttention` per @CuriousPanCake, vs. a standalone helper) — once the GenAI team settles that and it lands, I'll rework this PR to drop the error-message fallback and simply call the GenAI API. I'll mark this one as depends-on #4060 in the meantime. Appreciate the guidance.

Issue #2337, opened 2024-02-19
vehicle-detection-0202 responds to queries when running on CPU, gives an error when running on GPU


lugi0:
I am using the model `vehicle-detection-0202` in its openvino_ir format running in a OVMS environment, however when I try to send requests to the model running on a CPU I can get back the expected response, e.g.
```
{
    "model_name": "my-model",
    "model_version": "1",
    "outputs": [{
            "name": "detection_out",
            "shape": [1, 1, 200, 7],
            "datatype": "FP32",
            "data": [0.0, ..., data, ..., 0.0]
        }]
}
```
If I instead try to send a request to the model running on a GPU I keep getting error messages like the following:
```
{"code":2,"message":"json: unsupported value: -Inf"}
```
(Sometimes instead of -Inf the value is NaN). I don't quite understand what I'm doing wrong, as the requests are in the same format and the data has all the same attributes. Any idea what's going on?

exzile:
I tried to reproduce this on current OpenVINO and it no longer occurs — sharing the method and results in case it helps close this out.

**Setup**
- OpenVINO 2026.3 (`2026.3.0-22291`)
- GPU: Intel® Arc™ Pro B70 (dGPU), plus CPU as reference
- Model: `vehicle-detection-0202` IR from Open Model Zoo, both **FP16** and **FP32**

**Method**
Compiled the model on CPU and GPU and ran inference with four input types covering normal and stress cases — all-zeros, all-255 (saturated), uniform random in `[0, 255]` (image-like), and standard-normal noise — then checked the `detection_out` output for any non-finite (`NaN`/`±Inf`) values and compared GPU vs CPU.

**Results**

| precision | device | non-finite values | output range |
|---|---|---|---|
| FP16 | CPU | 0 | [-1, ~1.04] |
| FP16 | **GPU** | 0 | [-1, ~1.04] |
| FP32 | CPU | 0 | [-1, ~1.04] |
| FP32 | **GPU** | 0 | [-1, ~1.04] |

Every output is finite and bounded, and GPU matches CPU closely across all inputs — no `-Inf`/`NaN` on GPU.

**Conclusion**
The `-Inf`/`NaN` GPU behavior reported here (Feb 2024) does not reproduce on current OpenVINO. It appears to have been resolved by GPU-plugin fixes in the intervening releases (or was specific to the original, unspecified GPU/driver). 

[...]

PR #4337, opened 2026-06-28
Return clear error for non-finite (NaN/Inf) floats in KServe REST output

exzile:
### Summary
The REST output serializers silently produced **malformed JSON** when a model output contained non-finite floats (`NaN`/`±Inf`). Floats were written via rapidjson `Writer::Double()` while ignoring its return value — rapidjson writes nothing for non-finite values (they are invalid per the JSON spec and the default writer has no `kWriteNanAndInfFlag`), so the affected array element was dropped, corrupting the response. There was no `isfinite` guard anywhere in the REST output path.

This now returns a clear `JSON_SERIALIZATION_ERROR` with a descriptive message instead of emitting broken JSON, on **both** the KServe v2 and the TFS REST paths.

### Change
- Added an `isJsonRepresentable` helper (overloaded: integral types always true; `float`/`double` use `std::isfinite`).
- **KServe v2 path:** guarded both write paths of the `PARSE_OUTPUT_DATA` macro (val-field and raw-output) to return `JSON_SERIALIZATION_ERROR` with `Output "<name>" contains a non-finite (NaN/Inf) value that cannot be serialized to JSON`.
- **TFS path:** applied the same finiteness check to `makeJsonFromPredictResponse` (`PredictResponse`), validating `float`/`double` outputs before `MakeJsonFromTensors` and returning the same error.
- Integer outputs are unaffected (helper resolves to the always-true template; negligible cost).
- Added unit tests for `Inf`, `NaN`, and the unaffected-integer case on both paths.

### Validation
Built `ovms_test` (Windows, py-off) and ran the serialization suite — all pass:
- Full `*MakeJsonFromPredictResponse*` suite: **85/85** (KFS + TFS, raw + val-field) — no regression.
- New non-finite tests: 3 KFS + 4 TFS (Inf/NaN float, double, int-unaffected) all pass.

### Context
Found while triaging #2337 (a model emitting `-Inf`/`NaN` on GPU). That issue's root cause is upstream/model-side and appears resolved on current OpenVINO; this PR is the independent OVMS-side hardening so that *any* model producing non-finite values fails cleanly over REST instead of returning corrupt JSON.

Relates to #2337.


exzile:
Update: extended the same non-finite (NaN/Inf) guard to the TFS REST serialization path as well (it was originally flagged as a possible follow-up). Both REST paths are now covered, and the description/tests have been updated accordingly — full `*MakeJsonFromPredictResponse*` suite now 85/85 (KFS + TFS).

exzile:
@dkalinowski rebased onto main and resolved conflicts. Note: main's #4349 removed the TFS frontend, so this is now **KFS-only** — the TFS guard/tests are dropped as obsolete. Net change is the `isJsonRepresentable()` NaN/Inf guard on the KServe REST path + tests. Ready for review when you have a moment.

Issue #4268, opened 2026-06-01
Add tool_parser minicpm5 for MiniCPM5-1B XML tool calls


DmytroMelnyk:
# [Feature request] Add `tool_parser: minicpm5` for MiniCPM5-1B XML tool calls

## Summary

MiniCPM5-1B uses XML-style function calls in model output. OVMS does not expose a matching `tool_parser`, so `/v3/chat/completions` responses leave tool XML in `message.content` with `tool_calls: []`. OpenAI-compatible agent clients then never execute tools.

## Expected model output (MiniCPM5)

Documented by OpenBMB: https://github.com/OpenBMB/MiniCPM#with-tool-calling

```xml
<function name="get_weather"><param name="city">Beijing</param></function>
```

The shipped `chat_template.jinja` uses the same `<function>` / `<param>` format.

## Current OVMS behavior

- Supported parsers include `lfm2`, `hermes3`, `llama3`, etc. (see [LLM reference](https://github.com/openvinotoolkit/model_server/blob/main/docs/llm/reference.md) and `--tool_parser` in export/CLI).
- There is **no** `minicpm5` parser.
- Using `--tool_parser lfm2` with MiniCPM5-1B is incorrect: LFM2 expects `<|tool_call_start|>[...]<|tool_call_end|>` (Pythonic/JSON), not MiniCPM5 XML.
- Observed with OVMS 2026.2-gpu: model may emit plausible XML in `content`, but API returns `"tool_calls": []`.

## Reference implementations

| Project | Parser | Notes |
|---------|--------|--------|
| [SGLang](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/function_call/minicpm5_detector.py) | `--tool-call-parser minicpm5` | Recommended by OpenBMB; merged in [sgl-project/sglang#25600](https://github.com/sgl-project/sglang/pull/25600) |
| [OpenBMB MiniCPM](https://github.com/OpenBMB/MiniCPM/blob/main/tool_parsers/minicpm5xml_tool_parser.py) | vLLM-style XML parser | Reference for regex/XML extraction |

SGLang does **not** enable structural-tag guided generation for MiniCPM5 (`supports_structural_tag() -> False`); post-generation extraction is the primary path.

## Requested change

1. Add **`minicpm5`** to supported `tool_parser` values (CLI, `graph.pbtxt`, `export_model.py`, docs).
[...]

PR #4339, opened 2026-06-29
Add minicpm5 tool_parser for MiniCPM5 XML tool calls (draft, #4268)

exzile:
### Summary
Adds a `minicpm5` tool parser so MiniCPM5 XML-style tool calls are surfaced as OpenAI-style `tool_calls` on `/v3/chat/completions`, instead of being left as raw XML in `message.content` with `tool_calls: []`.

Addresses #4268.

> **Note:** #4268 is assigned to @przepeck. I opened this as a **draft** so it can either be reviewed/merged or used as a starting point — whatever is least disruptive to any work already underway. Happy to hand it off, iterate on review, or close it if it duplicates effort.

### Approach
MiniCPM5's format (`<function name="get_weather"><param name="city">Beijing</param></function>`) is structurally almost identical to `qwen3coder`, so the parser is modeled directly on it:

- New `src/llm/io_processing/minicpm5/minicpm5_tool_parser.{hpp,cpp}` — `Minicpm5ToolParserImpl` state machine + `Minicpm5ToolParser : BaseOutputParser`.
- Unary `parse()` and streaming `parseChunk()` (same first-delta / full-delta protocol as qwen3coder).
- Attribute-style tag parsing (`name="..."`, also tolerates `'...'`), `<think>...</think>` stripping before parsing, and schema-driven argument typing (string vs number/bool) reusing the existing `ToolsSchemas_t` mechanism.
- Registered `minicpm5` in `output_parser.cpp` and `getSupportedToolParserNames()`; bazel target wired into `output_parsers`.

### Tests (`src/test/llm/output_parsers/minicpm5_output_parser_test.cpp`)
- Unary `OutputParser::parse` cases (single/multiple calls, integer-typed params, mixed types, `<think>` stripping, content preservation, content-around-call removal) — these run in CI (tokenizer-gated).
- Impl-level state-machine cases: single/multiple calls, integer typing, newline trimming, content removal, single-quote attributes.
- **Streaming robustness:** char-by-char and **every fragment size 1..N** must yield the same calls as a single-chunk parse; two-calls-char-by-char; leading content + `<think>` char-by-char.
- **Edge cases (no crash / no spurious calls):** unterminated `<function>`, empty `<param>` value, stray `<` in plain prose.

All 15 impl/streaming tests pass locally (Windows build). The `OutputParser`-level tests are tokenizer-gated and run in CI (same as the existing qwen3coder tests).

### Not yet included (intentional, for v1 follow-up / reviewer input)
- `export_model.py` + `docs/llm/reference.md` updates for the new `--tool_parser minicpm5` value.
- End-to-end validation against a real `openbmb/MiniCPM5-1B` IR (the acceptance test in #4268).
- Tool-guided / XGrammar generation for MiniCPM5 (explicitly out of scope per the issue).

Glad to add the docs/export wiring and run the end-to-end check once direction is confirmed.


src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+bool Minicpm5ToolParserImpl::parseUntilStateChange(ToolCalls_t& toolCalls) {
+    SPDLOG_TRACE("Minicpm5: state: {}", this->currentState);
+    auto previousState = this->currentState;
+
+    switch (this->currentState) {
+    case State::Content: {
+        // Look for either <think> or <function, whichever comes first.
+        auto posThink = this->streamContent.find(Minicpm5ToolParser::THINK_START_TAG, this->lastProcessedPosition);
```
przepeck: Reasoning handling should be extracted to reasoning parser to be consistent with other parsers

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+// Helper macro (mirrors qwen3coder)
+#define MINICPM5_FIND_TAG_OR_BREAK(TAG)                                                     \
+    auto pos = this->streamContent.find(TAG, this->lastProcessedPosition);                  \
+    if (pos == std::string::npos) {                                                         \
+        SPDLOG_TRACE("Minicpm5: did not find: {}", TAG);                                    \
+        break;                                                                              \
+    }
+
+bool Minicpm5ToolParserImpl::parseUntilStateChange(ToolCalls_t& toolCalls) {
```
przepeck, line 259: This method is needlessly long, it should be separated.

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,449 @@
...
+    rapidjson::Document doc = wrapFirstDelta(toolCallName, toolCallId);
+    this->currentJson.CopyFrom(doc, this->currentJson.GetAllocator());
+    this->returnedFirstDeltas.insert(toolCallId);
+    SPDLOG_DEBUG("Minicpm5ToolParser: first delta: {}", documentToString(doc));
+    return doc;
+}
+
+std::optional<rapidjson::Document> Minicpm5ToolParser::parseChunk(
+        const std::string& newChunk,
+        const std::vector<int64_t>& /*tokens*/,
```
przepeck, line 364: MiniCPM is using special tokens as key tags, it should either override requiresStreamingWithSpecialTokens with true returned or preferably base parsing on tokens vector.

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,449 @@
...
+            this->currentState = State::InsideFunctionName;
+        }
+        break;
+    }
+    case State::InsideThink: {
+        // Skip until </think>
+        MINICPM5_FIND_TAG_OR_BREAK(Minicpm5ToolParser::THINK_END_TAG);
+        this->lastProcessedPosition = pos + Minicpm5ToolParser::THINK_END_TAG.size();
+        this->currentState = State::Content;
+        break;
```
przepeck: reasoning should be preserved, sending it back to the user but in the filed "reasoning_content"

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,449 @@
...
+        trimNewline(parameterValueAsString);
+
+    auto paramIt = this->toolsParametersTypeMap.find(this->currentFunction.name);
+    auto& currentFunctionArgsDoc = this->currentFunction.argumentsAsDocument;
+    auto& allocator = currentFunctionArgsDoc.GetAllocator();
+    auto& key = this->currentParameterName;
+    rapidjson::Value keyVal(key.c_str(), allocator);
+
+    rapidjson::Document temp;
+    // Boolean normalisation (same as qwen3coder)
```
przepeck: If it's the same as qwen3coders, maybe it's better to extract it to some utility file, I can see a lot of code which can be shared between these 2 parsers

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
```diff
@@ -0,0 +1,401 @@
...
+    Minicpm5ToolParserImpl parser(minicpm5TypeMap);
+    auto callsOpt = parser.parseChunk(content);
+    ASSERT_TRUE(callsOpt.has_value());
+    auto& calls = callsOpt.value();
+    ASSERT_EQ(calls.size(), 1u);
+    EXPECT_EQ(calls[0].name, "get_weather");
+    EXPECT_EQ(calls[0].arguments, R"({"city":"Sydney"})");
+}
+
+// ---- Streaming robustness: feed input in fragments and aggregate ----
```
przepeck: This approach doesn't test tool calls are returned to the user after destined token, but only a final result of parsing.

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
przepeck, line 1: I cannot see any complex output tests. There aren't any tests with code, special chars and escaping. We will need some.

exzile:
Thanks for the detailed review @przepeck — pushed updates addressing all of the comments:

- **Reasoning extracted to a reasoning parser + preserved as `reasoning_content`** (`:222`, `:244`): the tool parser no longer handles `<think>` (dropped the `InsideThink` state/tags). MiniCPM5 uses the same `<think>`/`</think>` as Qwen3, so reasoning is handled by a reasoning parser and emitted as `reasoning_content`, consistent with the other parsers.
- **Long method split** (`:215`): `parseUntilStateChange` is now a small dispatcher delegating to per-state handlers.
- **Shared code with qwen3coder** (`:148`): `trimNewline`, `jsonTypeOf`, `enforceStringValue` and Python-bool normalization moved to `io_processing/utils`; `parseToolSchema`/`createToolsParametersTypesMap` moved to `base_output_parser` (which owns the `ParameterType`/`ToolsSchemas` types). Both parsers use the shared versions; qwen3coder behavior is unchanged.
- **Tests** (`:304`, `:1`): added a test driving the **public** `parseChunk` streaming path (name delta once, arguments delta after the function closes), plus complex-content cases — embedded quotes/backslashes, code with braces/newlines, JSON-looking string values, and special/unicode characters.
- **Special-token tags** (`:433`): confirmed and fixed. In the `openbmb/MiniCPM5-1B` tokenizer the tool tags **are special tokens** — `<function`=18, `</function>`=19, `<param`=20, `</param>`=21 (all `special: true`) — so the default streamer skips them and the tag-scanning parser never saw them while streaming. The tool parser now overrides `requiresStreamingWithSpecialTokens()` to return `true`. Because the framework requires the paired reasoning parser to agree on that flag, and MiniCPM5's `<think>`/`</think>` are **non-special** tokens (so the Qwen3 reasoning logic still applies), I added a thin `Minicpm5ReasoningParser` subclass that flips the flag and registered `minicpm5` as a reasoning parser. A deployment uses `tool_parser: minicpm5` + `reasoning_parser: minicpm5`.

All changes compile-verified locally (parser libraries + `output_parsers` build clean); the full gtest run is via CI here, as my local `ovms_test` build is blocked by an unrelated genai-version skew in the audio module.

exzile:
**Local verification update**

Built `//src:ovms_test` and ran the minicpm5 suite locally — **30 tests, 22 passed, 8 failed**, where the 8 failures are purely an environmental missing-model issue, not a code problem:

- The 8 failing tests (`Minicpm5OutputParserTest.Parse*`) call `tokenizer->encode(...)` and fail with `Tokenizer::encode is not available — openvino_tokenizer.xml was not provided or it was not loaded correctly` because the tokenizer IR isn't present in my local checkout. These are the tests that need a real model and will run under CI.
- **All 22 model-independent tests pass**, including:
  - the `Minicpm5ToolParserImplTest` core + edge-case + streaming-robustness tests,
  - the public streaming-delta test (name delta once, arguments delta after `</function>`),
  - the complex-content tests (embedded quotes/backslashes, code with braces/newlines, JSON-looking string values, unicode/special chars),
  - and a `RequiresStreamingWithSpecialTokens` test confirming the tool parser and the new `Minicpm5ReasoningParser` both report `true` and the `minicpm5` + `minicpm5` pairing constructs without tripping the `requiresStreamingWithSpecialTokens()` consistency check.

So the refactor and the special-token handling are behaviorally verified for everything testable without a downloaded model; the end-to-end/encode-dependent cases are left to CI.

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
```diff
@@ -0,0 +1,631 @@
...
+#include "src/llm/io_processing/minicpm5/minicpm5_reasoning_parser.hpp"
+#include "src/test/platform_utils.hpp"
+
+using namespace ovms;
+
+#ifdef _WIN32
+// Reuse an existing tokenizer from the test assets (content doesn't matter for these unary tests)
+const std::string minicpm5TokenizerPath = getWindowsRepoRootPath() + "\\src\\test\\llm_testing\\Qwen\\Qwen3-8B";
+#else
+const std::string minicpm5TokenizerPath = "/ovms/src/test/llm_testing/Qwen/Qwen3-8B";
```
przepeck, line 39: qwen3 tokenizer?

src/llm/io_processing/minicpm5/minicpm5_reasoning_parser.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace ovms {
+
+// MiniCPM5 uses the same <think>...</think> reasoning tags as Qwen3, so the reasoning logic is
+// reused verbatim. The only difference is that the MiniCPM5 tool parser requires the stream to
+// retain special tokens (its tool tags are special tokens); the framework requires the paired
+// reasoning parser to agree on requiresStreamingWithSpecialTokens(), so this subclass returns
+// true. The <think> tags themselves are ordinary (non-special) tokens, so the inherited
+// text-based reasoning parsing is unaffected by preserving special tokens.
+class Minicpm5ReasoningParser : public Qwen3ReasoningParser {
```
przepeck, line 38: I think it should be using it's own reasoning parser implementation, using tokens for parsing reasoning. We would like to smoothly end using requriresStreamingWithSpecialTokens().

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
exzile: Good catch on the ambiguity - there's no MiniCPM5 tokenizer in the test assets, so this reuses the existing Qwen3-8B one rather than adding a new asset. It's safe here because the Minicpm5 parser operates on OutputParser::parse's decoded text (tokenizer.decode(generatedTokens)) via plain string/XML-tag matching - it never reads token ids or vocab-specific special tokens. Every test string is round-tripped through encode() then decode() with this same tokenizer instance, so the parser sees back the exact text it was given regardless of whose vocabulary produced the token ids.

Pushed a commit expanding the comment above minicpm5TokenizerPath to spell this out, so it's not confusing to future readers.

src/llm/io_processing/minicpm5/minicpm5_reasoning_parser.hpp
exzile: Agreed, and confirmed it's feasible: <think> and </think> are real dedicated special tokens in MiniCPM5's tokenizer (added_tokens_decoder ids 8 and 9), not incidental text. Pushed a native implementation that mirrors the existing Gemma4ReasoningParser pattern -- parse() now scans generatedTokens for those two token ids directly to find the reasoning boundaries, instead of inheriting Qwen3ReasoningParser's text-based substring search. requiresStreamingWithSpecialTokens() stays true, for the same reason Gemma4's does: MiniCPM5's tool tags (<function>, <param>, ...) are also special tokens, so the paired tool/reasoning parser still has to agree on that setting.

On "smoothly end using requiresStreamingWithSpecialTokens()" - if you mean moving away from each parser hardcoding that boolean (e.g. deriving it structurally, maybe from whether a parser's tags are registered special tokens in the tokenizer) that sounds like a good direction, but a bigger cross-parser change than this PR. Happy to pick that up separately if there's appetite, or take direction if you already have a specific design in mind.

One thing this surfaced: the existing reasoning-parser test built its token sequence by calling encode() on literal "<think>...</think>" text with the Qwen3-8B tokenizer this test file uses as a stand-in (there's no MiniCPM5 tokenizer test asset). Qwen3-8B has its own different ids for those strings (151667/151668), so with the new token-id-based parse(), that test would have silently stopped exercising the boundary check at all. Fixed it to assemble the token sequence directly with the real ids (8/9) as delimiters, and use the stand-in tokenizer only to encode/decode the free-form text around them, which is unaffected by whose vocab it is (same reasoning as your earlier "qwen3 tokenizer?" thread). Long-term, adding a real MiniCPM5 tokenizer test asset (like Devstral and Gemma4 have their own) would let this be tested end-to-end via encode() instead - let me know if you'd like that as a follow-up.

I haven't been able to run a full C++ build in this environment, so please treat this as needing a CI pass / your review rather than as already-verified.

exzile: Self-review follow-up on the native reasoning parser I just pushed: found and fixed a real bug in my own implementation before it caused trouble.

parse() was decoding the content spans (outside the <think>...</think> boundary) with skip_special_tokens(true), copied from Gemma4ReasoningParser's pattern. That's fine for Gemma4 because its tool parser reads raw token ids directly, but MiniCPM5's tool parser is text-based - Minicpm5ToolParserImpl matches the literal "<function"/"<param" substrings in decoded content. Those tags are themselves special tokens in MiniCPM5's tokenizer (<function = id 18, <param = id 20, both "special": true), so skip_special_tokens(true) was silently erasing them from content before the tool parser ever ran - breaking tool-call extraction whenever a reasoning block preceded a function call, which is exactly the scenario the ParseWithThinkBlockHandledByReasoningParser test exercises.

Fixed: content spans now decode with skip_special_tokens(false), matching the framework's own default full-decode behavior. Only the extracted reasoning text keeps skip_special_tokens(true), since it's a human-facing string, not matched against tags downstream. Pushed as a follow-up commit. Also documented the existing find-first-occurrence limitation shared with Gemma4ReasoningParser (a stray </think> before any <think> would hide a later well-formed pair) as a code comment rather than leaving it silent.

Still haven't been able to run a full build/test pass in this environment - flagging so this gets a careful look in CI/review rather than assumed correct.

exzile: Verified locally, end to end: all 30 Minicpm5 tests pass, including ParseWithThinkBlockHandledByReasoningParser (the one covering the skip_special_tokens fix from my last comment).

[==========] 30 tests from 4 test suites ran. (436 ms total)
[  PASSED  ] 30 tests.

For the record on how, since the full ovms_test target doesn't build in my environment (an unrelated, pre-existing header mismatch in the audio/speech_to_text module, nothing to do with this PR): I built a small isolated cc_test depending only on the output-parsers library, and I had to convert a real Qwen3-8B tokenizer locally first (openvino-tokenizers' convert_tokenizer, tokenizer-only, no model weights) since that test asset wasn't present on this machine at all. Both were temporary/local-only; nothing about that is part of this PR's diff.

This replaces my earlier "haven't been able to run a full build/test pass" caveat - the fix is now confirmed working, not just reasoned through.

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
exzile: Good point, and correct: this Minicpm5ToolParserImplStreamTest section (streamInFragments) only asserts the AGGREGATED final result is fragmentation-invariant - it doesn't test what a client actually receives mid-stream or when.

That user-facing concern is covered separately by Minicpm5PublicStreamTest further down (added in an earlier review round), which drives the real production entrypoint (Minicpm5ToolParser::parseChunk, the one OutputParser actually calls) and asserts on the emitted deltas themselves: that a name delta is surfaced exactly once, as soon as the name is known, and the full arguments delta only once the function completes - i.e. testing the destined-token timing you're asking about, not just the end state.

I've pushed two changes to make this clearer and close a real gap I found while looking at it:
1. Retitled/expanded the comment above this section to state explicitly that it tests state-machine fragment-boundary robustness, not the delta-timing contract, and points to Minicpm5PublicStreamTest for that.
2. That public-delta coverage was single-call only. Added EmitsDeltasForTwoSequentialCalls, which verifies two sequential tool calls each get exactly one name delta and one arguments delta, in order, with correct/distinct tool_call indices - so the second call's deltas aren't emitted early or mixed with the first's.

All 31 Minicpm5 tests pass locally (built and run directly, not just reasoned through).

Let me know if you think there's still a specific scenario the public-delta tests don't cover.

exzile:
Closing as superseded: the MiniCPM5 tool parser landed in main via #4373 ("MiniCPM5 tool parser", based on this PR), which is a more complete implementation (chat-template analyzer/probe support and a MiniCPM5 jinja template that this draft did not include). No unique changes here remain unmerged, so closing to keep the queue clean. Thanks!

PR #4359, opened 2026-07-08
Accept empty content array - transform it to null

mzegla:


src/llm/io_processing/input_processors/empty_content_normalization_processor.cpp
```diff
@@ -0,0 +1,38 @@
...
+// limitations under the License.
+//*****************************************************************************
+
+#include "empty_content_normalization_processor.hpp"
+
+#include <variant>
+
+namespace ovms {
+
+absl::Status EmptyContentNormalizationProcessor::process(InputRequest& req) {
```
dkalinowski: EmptyContentArrayNormalizationProcessor

src/llm/io_processing/input_processors/empty_content_array_normalization_processor.cpp
```diff
@@ -0,0 +1,38 @@
...
+namespace ovms {
+
+absl::Status EmptyContentNormalizationProcessor::process(InputRequest& req) {
+    if (!std::holds_alternative<ov::genai::ChatHistory>(req.input)) {
+        return absl::Status(absl::StatusCode::kInternal,
+            "EmptyContentNormalizationProcessor received input that is not a ChatHistory");
+    }
+    ov::genai::ChatHistory& chatHistory = std::get<ov::genai::ChatHistory>(req.input);
+    for (size_t i = 0; i < chatHistory.size(); i++) {
+        const auto content = chatHistory[i]["content"];
```
dkalinowski, line 30: shouldnt we first check if "content" key exists?

mzegla: it is always added, if request does not have content field, api handler adds it with empty string
not sure if that's how it should look like, but it's out of scope for this PR

dkalinowski (approved):

michalkulakowski (approved):

PR #4363, opened 2026-07-09
Bundle libxml2 + ICU in the baremetal package so it runs on newer distros (fixes #4362)

### LINKED ISSUE #4362: Baremetal ovms_ubuntu24 binary fails on newer distros — libxml2.so.2 (+ ICU 74) is a runtime dep but not bundled
## Summary
The baremetal Linux release tarball (`ovms_ubuntu24_<ver>_python_{off,on}.tar.gz`) bundles ~82 shared libraries under `ovms/lib/`, but **not `libxml2`** — even though the `ovms` binary links it directly (`ldd ovms → libxml2.so.2 => not found`). On Ubuntu 22.04/24.04 this is masked because the host provides `libxml2.so.2` via the distro `libxml2` package. On any newer distro that ships libxml2 under a different soname (and a newer ICU), the binary won't start.

## Version
- OVMS **2026.2.1** (`PROJECT_VERSION: 2026.2.1.1122f03bf`, `BASE_IMAGE: ubuntu:24.04`)
- Both `python_off` and `python_on` tarballs affected (same missing dependency)

## Environment where it fails
- **Ubuntu 26.04 LTS**, glibc 2.43
- Ships `libxml2.so.16` (no `libxml2.so.2`); the `libxml2` apt package has no candidate
- Ships **ICU 78** (`libicuuc.so.78`); no ICU 74

Generally: reproducible on any distro where the system libxml2 soname ≠ `.so.2`, or where ICU ≠ 74.

## Symptom
```
$ LD_LIBRARY_PATH=/opt/ovms/lib /opt/ovms/bin/ovms --version
/opt/ovms/bin/ovms: error while loading shared libraries: libxml2.so.2: cannot open shared object file: No such file or directory
```

## Root cause / dependency chain
1. `ovms` binary → needs `libxml2.so.2` (not in `ovms/lib/`, not on the host).
2. Supplying `libxml2.so.2` (2.9.14) then reveals → needs `libicuuc.so.74` (host has ICU 78).

So the full unmet chain is `libxml2.so.2 → libicuuc.so.74 → libicudata.so.74`.

## Steps to reproduce
```bash
# On Ubuntu 26.04 (or any distro without libxml2.so.2 / ICU 74):
curl -sL -O https://github.com/openvinotoolkit/model_server/releases/download/v2026.2.1/ovms_ubuntu24_2026.2.1_python_off.tar.gz
sudo tar -xzf ovms_ubuntu24_2026.2.1_python_off.tar.gz -C /opt
LD_LIBRARY_PATH=/opt/ovms/lib ldd /opt/ovms/bin/ovms | grep 'not found'
# -> libxml2.so.2 => not found
```

## Workaround
Dropping compatible libs into OVMS's own `lib/` (isolated; does not touch the system) resolves it:
```bash
# libxml2 2.9.14 + ICU 74 (any ubuntu24-era source works)
cp libxml2.so.2.9.14   /opt/ovms/lib/libxml2.so.2
cp libicuuc.so.74.2    /opt/ovms/lib/libicuuc.so.74
cp libicudata.so.74.2  /opt/ovms/lib/libicudata.so.74
cp libicui18n.so.74.2  /opt/ovms/lib/libicui18n.so.74
# ldd now clean; ovms starts and serves models on the Intel Arc GPU.
```

## Suggested fix (any one)
1. **Bundle `libxml2` (and its ICU deps) in `ovms/lib/`** like the other ~82 libraries — the tarball is already designed to be self-contained via its bundled libs.
2. **Statically link libxml2** into the binary.
3. At minimum, **document the exact external runtime dependencies** (`libxml2.so.2` + ICU 74) and the supported distro range for the baremetal package — the `ubuntu24` filename implies broader compatibility than the artifact actually has.

## Notes
GPU serving itself works fine on Ubuntu 26.04 once the libs are supplied (Intel Arc Pro B60; the bundled `libopenvino_intel_gpu_plugin.so` + `libOpenCL.so` plus the distro `intel-opencl-icd`). This is purely a packaging/portability gap in the baremetal artifact, not a runtime GPU issue.

exzile:
## Problem
The baremetal Linux release tarball fails to start on distros newer than the build base, because it does not bundle `libxml2` — a runtime dependency of the `ovms` binary.

`ldd /opt/ovms/bin/ovms` reports `libxml2.so.2 => not found`. `create_package.sh` bundles the bazel-out artifacts plus the OpenVINO/OpenCV/TBB/OpenCL libraries, but `libxml2` is pulled in from the base image (`Dockerfile.ubuntu` `apt-get install ... libxml2`), so the portable tarball never included it. On Ubuntu 22.04/24.04 the host provides `libxml2.so.2`, masking the gap. On a host whose `libxml2`/ICU soname differs it breaks — e.g. **Ubuntu 26.04** ships `libxml2.so.16` + ICU 78 (no `libxml2.so.2`, no ICU 74):

```
ovms: error while loading shared libraries: libxml2.so.2: cannot open shared object file: No such file or directory
```

The full unmet chain (via `readelf -d`) is `ovms → libxml2.so.2 → libicuuc.so.74 → libicudata.so.74`.

Reported in #4362.

## Fix
Bundle `libxml2` and its ICU dependency chain into `/ovms_release/lib/` in `create_package.sh`, alongside the existing `libOpenCL` copy and **before** the `patchelf --set-rpath '$ORIGIN/../lib'` pass, so the copied libraries get an rpath that resolves them within the package.

```sh
if [[ "$BASE_OS" =~ "ubuntu" ]] ; then
	cp -P /usr/lib/x86_64-linux-gnu/libxml2.so.2* /ovms_release/lib/
	cp -P /usr/lib/x86_64-linux-gnu/libicuuc.so.* /ovms_release/lib/
	cp -P /usr/lib/x86_64-linux-gnu/libicudata.so.* /ovms_release/lib/
fi
```

Only the minimal chain is bundled — `libz`, `liblzma`, `libstdc++`, `libm`, `libc` have stable sonames present on every target and are intentionally left to the host.

## Verification
On **Ubuntu 26.04** (glibc 2.43), with the current `v2026.2.1` `ovms_ubuntu24` tarball the binary won't start (`libxml2.so.2: cannot open shared object file`). After adding these three libraries to `ovms/lib/`, `ldd ovms` is fully resolved and OVMS starts and serves an LLM on the GPU (Intel Arc Pro B60) end-to-end via `/v3/chat/completions`. This PR makes `create_package.sh` produce that self-contained tarball directly.

## Notes / scope
- **ubuntu x86_64 only.** This is the reported and verified case. The `redhat` package (and aarch64) likely need the analogous copy from their own libdirs, but I could only verify ubuntu x86_64, so I left them untouched rather than ship unverified paths.
- **Size:** adds ~30 MB, almost entirely `libicudata` (the ICU data blob `libxml2` requires). If that's a concern, an alternative is to link a libxml2 built `--without-icu` (or statically), which would avoid bundling ICU — happy to pursue that direction instead if maintainers prefer.
- **Licenses:** `libxml2` (MIT) and ICU (Unicode license) are now shipped in the tarball; `thirdparty-licenses/` may want their texts added.

### DIFF
--- a/create_package.sh
+++ b/create_package.sh
@@ -115,6 +115,23 @@ fi
 if [ "$BASE_OS" == "redhat" ] ; then cp -P /usr/lib64/libOpenCL.so* /ovms_release/lib/ ; fi
 if [[ "$BASE_OS" =~ "ubuntu" ]] ; then cp -P /usr/lib/x86_64-linux-gnu/libOpenCL.so* /ovms_release/lib/ ; fi
 
+# libxml2 is a runtime dependency of the ovms binary (Dockerfile.ubuntu apt-installs
+# it into the runtime image), but it is provided by the base image rather than by
+# bazel-out, so the portable tarball never bundled it. The Docker image is therefore
+# fine while the baremetal package fails to start on any host whose system libxml2 /
+# ICU soname differs from the ubuntu build base -- e.g. Ubuntu 26.04 ships
+# libxml2.so.16 + ICU 78, not the libxml2.so.2 + ICU 74 this build links against:
+#   ovms: error while loading shared libraries: libxml2.so.2: cannot open shared object file
+# Bundle libxml2 and its ICU dependency chain (ovms -> libxml2.so.2 -> libicuuc ->
+# libicudata) so the tarball is self-contained. The copies land before the patchelf
+# rpath pass below, so they get '$ORIGIN/../lib' and resolve one another in the bundle.
+# See https://github.com/openvinotoolkit/model_server/issues/4362
+if [[ "$BASE_OS" =~ "ubuntu" ]] ; then
+	cp -P /usr/lib/x86_64-linux-gnu/libxml2.so.2* /ovms_release/lib/
+	cp -P /usr/lib/x86_64-linux-gnu/libicuuc.so.* /ovms_release/lib/
+	cp -P /usr/lib/x86_64-linux-gnu/libicudata.so.* /ovms_release/lib/
+fi
+
 if [ "$FUZZER_BUILD" == "0" ]; then find /ovms/bazel-bin/src -name 'ovms' -type f -exec cp -v {} /ovms_release/bin \; ; fi;
 cd /ovms_release/bin
 if [ "$FUZZER_BUILD" == "0" ]; then patchelf --remove-rpath ./ovms && patchelf --set-rpath '$ORIGIN/../lib/' ./ovms; fi;


exzile:
@rasapala this one's been sitting mergeable without a review — bundles libxml2 + ICU into the baremetal package so it runs on newer distros (fixes #4362). Could you take a look or route it to the right owner? Happy to address anything.

PR #4367, opened 2026-07-10
setting default recommended device

dtrawins:
### 🛠 Summary

Apply a naive approach for setting the default target device:
use discrete GPU if present
if several discrete GPUs present, use the one with the most free VRAM
if iGPU present and no discrete - use it
fallback to CPU if no GPU is present

### 🧪 Checklist

- [ ] Unit tests added.
``


src/llm/language_model/continuous_batching/servable_initializer.cpp
```diff
@@ -190,6 +190,10 @@ Status ContinuousBatchingServableInitializer::initialize(std::shared_ptr<GenAiSe
     }
 
     properties->device = nodeOptions.device();
+    if (properties->device.empty()) {
+        properties->device = recommendTargetDevice();
+        SPDLOG_INFO("No device specified for LLM CB model, using recommended device: {}", properties->device);
```
mzegla: Not sure if we should specify CB here. The log informing about type of servable will be there anyway, so this information will be embedded in the context and if we decide to change naming convention it's another place to change it.

src/llm/language_model/legacy/servable_initializer.cpp
```diff
@@ -85,6 +86,10 @@ Status LegacyServableInitializer::initialize(std::shared_ptr<GenAiServable>& ser
     properties->schedulerConfig.enable_prefix_caching = nodeOptions.enable_prefix_caching();
 
     properties->device = nodeOptions.device();
+    if (properties->device.empty()) {
+        properties->device = recommendTargetDevice();
+        SPDLOG_INFO("No device specified for LLM legacy model, using recommended device: {}", properties->device);
```
mzegla: Not sure if we should specify "legacy" here. The log informing about type of servable will be there anyway, so this information will be embedded in the context and if we decide to change naming convention it's another place to change it.

src/llm/visual_language_model/legacy/servable_initializer.cpp
```diff
@@ -84,6 +85,10 @@ Status VisualLanguageModelLegacyServableInitializer::initialize(std::shared_ptr<
     properties->schedulerConfig.enable_prefix_caching = nodeOptions.enable_prefix_caching();
 
     properties->device = nodeOptions.device();
+    if (properties->device.empty()) {
+        properties->device = recommendTargetDevice();
+        SPDLOG_INFO("No device specified for VLM model, using recommended device: {}", properties->device);
```
mzegla, line 90: I can't see similar log for VLM CB model. It's possible it would be printed in LLM CB part if both pipelines share the same initializer. I guess that only emphasizes comments in other servables. Mentioning servable type here brings more trouble than benefits.

src/test/graph_export_test.cpp
```diff
@@ -48,7 +48,7 @@ const std::string expectedOneSettingPluginGraphContents = R"(
     node_options: {
         [type.googleapis.com / mediapipe.LLMCalculatorOptions]: {
             max_num_seqs:256,
-            device: "CPU",
+            device: "",
```
mzegla: Could we remove that field here completely? Would closer mimic the real life graphs. Users will not specify empty device usually.

src/ov_utils.cpp
```diff
@@ -150,6 +150,93 @@ Status validatePluginConfiguration(const plugin_config_t& pluginConfig, const st
...
+            uint64_t totalMem = core.get_property(device, "GPU_DEVICE_TOTAL_MEM_SIZE").as<uint64_t>();
+            uint64_t usedMem = 0;
+            try {
+                auto memStats = core.get_property(device, "GPU_MEMORY_STATISTICS").as<std::map<std::string, uint64_t>>();
+                auto it = memStats.find("usm_device");
+                if (it != memStats.end()) {
+                    usedMem = it->second;
+                }
+            } catch (...) {
+                // Memory statistics may not be available before any model is loaded
```
mzegla, line 224: Any warning log here?

src/ov_utils.cpp
```diff
@@ -150,6 +150,93 @@ Status validatePluginConfiguration(const plugin_config_t& pluginConfig, const st
...
 }
 
+std::string recommendTargetDevice() {
+    static ov::Core core;
+    auto availableDevices = core.get_available_devices();
+
+    struct GpuDeviceInfo {
+        std::string name;
+        bool isDiscrete = false;
+        int64_t freeMemBytes = 0;
```
mzegla: Feels more intuitive and safe if `freeMemBytes` is an unsigned type.

src/ov_utils.cpp
```diff
@@ -150,6 +150,93 @@ Status validatePluginConfiguration(const plugin_config_t& pluginConfig, const st
...
+
+    std::vector<GpuDeviceInfo> discreteGpus;
+    std::vector<GpuDeviceInfo> integratedGpus;
+    for (const auto& gpu : gpuDevices) {
+        if (gpu.isDiscrete) {
+            discreteGpus.push_back(gpu);
+        } else {
+            integratedGpus.push_back(gpu);
+        }
+    }
```
mzegla: Do we need `gpuDevices` actually?
If we need such separation we could start with empty separated vectors for  `discreteGpus` and `integratedGpus` and in line 199 `push_back` depending on `isDiscrete`. Less unnecessary data movement and code, but it's a minor since there will usually be likely 2-3 elements.

mzegla (approved):

michalkulakowski (approved):

PR #4369, opened 2026-07-12
watcher started only when needed

dtrawins:
### 🛠 Summary

Now the design is:
versionPollingNeeded — member variable re-evaluated by evaluatePollingState() after every config load
Watcher starts if either config file monitoring is needed OR version polling is needed
updateConfigurationWithoutConfigFile() only runs when versionPollingNeeded is true
After each loadConfig(), evaluatePollingState() is called again — so if the new config adds a versioned model, version polling activates dynamically

### 🧪 Checklist

- [ ] Unit tests added.
``


przepeck:
Could you add unit tests for this change?

src/modelmanager.cpp
```diff
@@ -212,6 +212,38 @@ void ModelManager::startWatcher(bool watchConfigFile) {
...
+void ModelManager::evaluatePollingState() {
+    static const std::set<std::string> singleFileExtensions = {
+        ".xml", ".onnx", ".pdmodel", ".pdiparams", ".pb", ".tflite"};
+
+    versionPollingNeeded = false;
+    for (const auto& [name, config] : servedModelConfigs) {
+        const std::string& basePath = config.getBasePath();
+        auto ext = std::filesystem::path(basePath).extension().string();
+        if (singleFileExtensions.count(ext) == 0) {
+            SPDLOG_LOGGER_TRACE(modelmanager_logger, "Model {} with path {} requires version polling", name, basePath);
```
mzegla, line 224: I would put it on debug or even info if this is a one time call during model loading phase

src/modelmanager.cpp
```diff
@@ -212,6 +212,38 @@ void ModelManager::startWatcher(bool watchConfigFile) {
...
+    bool needWatcher = versionPollingNeeded || isStartedWithConfigFile();
+    if (needWatcher && watcherIntervalMillisec > 0) {
+        if (versionPollingNeeded) {
+            SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Filesystem version polling enabled - at least one model uses version directories");
+        } else {
+            SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Filesystem version polling disabled - no model uses version directories");
+        }
+        startWatcher(isStartedWithConfigFile());
+    } else {
+        SPDLOG_LOGGER_DEBUG(modelmanager_logger, "Filesystem version polling disabled - no model uses version directories");
```
mzegla, line 243: maybe it could be on INFO level logging

src/modelmanager.cpp
```diff
@@ -212,6 +212,38 @@ void ModelManager::startWatcher(bool watchConfigFile) {
     }
 }
 
+void ModelManager::evaluatePollingState() {
```
mzegla, line 215: The name of this method hides the fact that it may launch watcher thread. I would either rename it or break it in two, so we have evaluate method that returns information if we need to start watcher or not and the second one that starts it.

src/modelmanager.cpp
```diff
@@ -1047,18 +1079,22 @@ Status ModelManager::configFileReloadNeeded(bool& isNeeded) {
...
     while (exitSignal.wait_for(std::chrono::milliseconds(this->watcherIntervalMillisec)) == std::future_status::timeout) {
-        SPDLOG_LOGGER_TRACE(modelmanager_logger, "Models configuration and filesystem check cycle begin");
         std::unique_lock<std::recursive_mutex> loadingLock(configMtx);
         if (watchConfigFile) {
+            SPDLOG_LOGGER_TRACE(modelmanager_logger, "Checking if config file changed");
             bool isNeeded;
             configFileReloadNeeded(isNeeded);
             if (isNeeded) {
                 loadConfig();
+                evaluatePollingState();
```
mzegla, line 1089: In it's current implementation, that method can start a watcher thread. Does not look like something that should be run in a while loop.

PR #4373, opened 2026-07-13
MiniCPM5 tool parser 

przepeck:
### 🛠 Summary

https://jira.devtools.intel.com/browse/CVS-187940
Adding tool parser for MiniCPM5, based on [PR#4339](https://github.com/openvinotoolkit/model_server/pull/4339)

### MiniCPM5 - int4
unary results:
image

streaming results:
image

### MiniCPM5 - int8
unary results(without reasoning):
image

unary results(with reasoning):
image
image

Adding autodetection for this parser and reasoning parser.
Minja is currently unsupported for this model.

### 🧪 Checklist

- [x] Unit tests added.
``


przepeck:
image

Model's BFCL accuracy, without reasoning.
With reasoning it's ~ 40%, it has tendency to loop and returning arguments as object with the function name of "arguments".

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,463 @@
...
+    return previousState != this->currentState;
+}
+
+std::optional<ToolCalls_t> Minicpm5ToolParserImpl::parseChunk(const std::string& chunk) {
+    if (chunk.empty())
+        return std::nullopt;
+    ToolCalls_t toolCalls;
+    this->streamContent += chunk;
+    while (parseUntilStateChange(toolCalls)) {
+    }
```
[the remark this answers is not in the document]
przepeck: This is created per request, there is no need to prune it

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,468 @@
...
+        SPDLOG_DEBUG("Minicpm5ToolParser: parse done, removing tool calls from content");
+        auto status = this->streamParser.removeToolCallsFromContentIfNeeded(contentWithSpecialTokens);
+        if (!status.ok()) {
+            SPDLOG_DEBUG("Minicpm5ToolParser: failed to remove tool calls from content: {}", status.string());
+        }
+        parsedOutput.content = std::move(contentWithSpecialTokens);
+        return;
+    }
+    SPDLOG_DEBUG("Minicpm5ToolParser: parse done, no tool calls found");
+}
```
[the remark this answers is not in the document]
przepeck: can't do that, we are using special tokens here non existing in `parsedOutput.content`

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,468 @@
...
+    this->filledParametersTypesMap = true;
+    SPDLOG_DEBUG("Minicpm5ToolParser: created with {} tools", this->toolsParametersTypes.size());
+}
+
+Minicpm5ToolParser::Minicpm5ToolParser(ov::genai::Tokenizer& tokenizer, const ToolsSchemas_t& toolSchemas) :
+    BaseOutputParser(tokenizer),
+    toolSchemas(toolSchemas),
+    streamParser(this->toolsParametersTypes) {}
+
+const std::vector<int64_t> Minicpm5ToolParser::removeReasoningTokens(const std::vector<int64_t>& generatedTokens) {
```
atobiszei, line 308: Why we need to remove reasoning tokens from nside tool parser?

przepeck: We are using tokens, from const  std::vector<int64_t>& generatedTokens, it's const so we cannot remove reasoning tokens from this vector in reasoning parser, that means we need to remove it in tool parser's content. It's smelly WA, but I cannot think about anything better then refactor of BaseOutputParser. Miłosz knows about this issue and it will be probably solved in next iteration of our parser's refactor

src/llm/io_processing/minicpm5/minicpm5_reasoning_parser.cpp
```diff
@@ -0,0 +1,81 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+#include <openvino/genai/tokenizer.hpp>
+#include <string>
+#include <vector>
+
+#include "src/port/rapidjson_document.hpp"
+
+#include "../../../logging.hpp"
```
atobiszei:
```suggestion
#include "src/logging.hpp"
```

src/llm/io_processing/minicpm5/minicpm5_reasoning_parser.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "minicpm5_reasoning_parser.hpp"
+#include "../utils.hpp"
```
atobiszei:
```suggestion
#include "src/llm/io_processing/utils.hpp"
```

src/llm/io_processing/minicpm5/minicpm5_reasoning_parser.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace ovms {
+class Minicpm5ReasoningParser : public BaseOutputParser {
+protected:
+    const std::string parsingStartTag = "<think>";
+    const std::string parsingEndTag = "</think>";
+
+    const int64_t reasoningStartTokenId = 8;  // <think>
+    const int64_t reasoningEndTokenId = 9;    // </think>
+
```
atobiszei: Nitpick but maybe more consistent naming?

```suggestion
    const std::string reasoningStartTag = "<think>";
    const std::string reasoningEndTag = "</think>";

    const int64_t reasoningStartTagTokenId = 8;
    const int64_t reasoningEndTagTokenId = 9;

```
Then comments are not needed anyway

src/test/llm/chat_template_end_to_end_minja_test.cpp
```diff
@@ -596,6 +596,33 @@ What's the weather in Paris?<|im_end|>
     EXPECT_EQ(appliedOutput, expectedOutput);
 }
 
+// =============================================================================
+// MiniCPM5 chat template is not compatible with minja, therefore the test is expected to fail.
+// =============================================================================
+TEST_F(ChatTemplateEndToEndMinjaTest, MiniCPM5_ToolCallWithStringArgs) {
```
atobiszei: Remove comment - change name of test to suggest that. Eg.

Also just by looking at 1 line of code difference between run() and run(true) is not explainable from just 1/2 lines. Add temp bool variable what that bool switch does to improve readability.

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
```diff
@@ -0,0 +1,859 @@
...
+    void SetUp() override {
+        outputParser = std::make_unique<OutputParser>(*minicpm5Tokenizer, "minicpm5", "minicpm5", minicpm5ToolsSchemas);
+    }
+
+    std::vector<int64_t> encodeInput(const std::string& input) {
+        if (input == "<think>") {
+            return {8};  // <think> token ID
+        }
+        if (input == "</think>") {
+            return {9};  // </think> token ID
```
atobiszei: Replace comment and value with sth like
```suggestion
        if (input == Minicpm5ReasoningParser::parsingStartTag) {
            return {Minicpm5ReasoningParser::parsingStartTagTokenId};
        }
        if (input ==Minicpm5ReasoningParser::parsingEndTag) {
            return {Minicpm5ReasoningParser::parsingEndTagTokenId};
```

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
```diff
@@ -0,0 +1,859 @@
...
+        return outputParser->parse(generatedTokens, true);
+    }
+
+    void assertReasoningVec(const std::vector<std::tuple<std::string, ov::genai::GenerationFinishReason, std::optional<std::string>>>& chunkToDeltaVec) {
+        for (const auto& [chunk, finishReason, expectedDelta] : chunkToDeltaVec) {
+            std::vector<int64_t> tokens = {};
+            if (chunk == "<think>") {
+                tokens = {8};  // <think> token ID
+            } else if (chunk == "</think>") {
+                tokens = {9};  // </think> token ID
```
atobiszei: same here with

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
```diff
@@ -0,0 +1,859 @@
...
+}
+
+// This scenario will be handled only in unary, in streaming it's not possible to parse reasoning without the starting tag
+TEST_F(Minicpm5OutputParserTest, ParseReasoningWithoutStartingTag) {
+    const std::string input = "This is my internal reasoning about what to call.</think>";
+    ParsedOutput parsedOutput = generateParsedOutput(input);
+
+    EXPECT_EQ(parsedOutput.toolCalls.size(), 0u);
+    EXPECT_NE(parsedOutput.reasoning.find("internal reasoning"), std::string::npos);
+    EXPECT_EQ(parsedOutput.content, "");
```
atobiszei, line 276: In this case we expect empty reasoning but still do cut out reasoning from model generated content? Looks like are possibly hiding accuracy issue in model.

Either cut out up to </think> and then treat that part as reasoning or do not cut out this part at all and post all as content.

Also if its not clear what is expected from test write it in test name so just by looking at failed test case name you can have an idea what went wrong, even if the test names are longer then,

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
```diff
@@ -0,0 +1,859 @@
...
+}
+
+TEST_F(Minicpm5OutputParserTest, ParseWithThinkBlockHandledByReasoningParser) {
+    constexpr int64_t thinkStartTokenId = 8;  // <think>, per Minicpm5ReasoningParser
+    constexpr int64_t thinkEndTokenId = 9;    // </think>, per Minicpm5ReasoningParser
```
atobiszei: Reuse constants from reasoning parser.

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,468 @@
...
+    if (tagEnd == std::string::npos) {
+        SPDLOG_TRACE("Minicpm5: waiting for '>' of <param> tag");
+        return;
+    }
+    auto nameAttr = this->streamContent.find(Minicpm5ToolParser::NAME_ATTR_PREFIX, this->lastProcessedPosition);
+    if (nameAttr == std::string::npos || nameAttr >= tagEnd) {
+        SPDLOG_DEBUG("Minicpm5: <param> tag has no name= attribute — skipping");
+        this->lastProcessedPosition = tagEnd + 1;
+        this->currentState = State::InsideFunction;
+        return;
```
atobiszei: Is it expected for minicpm to skip name of parameter? We are here changing state even tough we failed to find param name - that's a big difference in comparison to Qwen3Coder and I would like to know if this is intentional. If it is we need specific test that would expose such behavior. It would be good also to have in comments allowed state transitions.

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -0,0 +1,468 @@
...
+    if (this->currentFunction.name.empty()) {
+        SPDLOG_DEBUG("Minicpm5: could not extract function name — skipping tag");
+        this->lastProcessedPosition = tagEnd + 1;
+        this->currentState = State::Content;
+        this->toolCallPositions.begin.pop();
+        return;
+    }
+    SPDLOG_TRACE("Minicpm5: function name: {}", this->currentFunction.name);
+    this->lastProcessedPosition = tagEnd + 1;  // past '>'
+    this->currentState = State::InsideFunction;
```
atobiszei: Why we need such complicated handling with comparison to qwen3 coder?

In coder it;'s just:
```
    case State::InsideFunctionName: {
        DEFINE_TAG_POSITION_AND_BREAK_IF_NOT_FOUND(Qwen3CoderToolParser::XML_TAG_END);
        this->currentFunction.name = streamContent.substr(this->lastProcessedPosition, pos - this->lastProcessedPosition);
        this->lastProcessedPosition = pos + Qwen3CoderToolParser::XML_TAG_END.length();
        this->currentState = State::InsideFunction;
        break;
    }
```

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
przepeck: non empty, we are expecting that reasoning.find(internal reasoning") != std::string::npos 
so we are expecting that we have reasoning and cutting it from content. This model has a tendention to not generate start reasoning token. This isn't problem in unary, only in streaming

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
przepeck: I think this case handles function without parameters, something like  e. g. `pwd()`, I will add the test for that

dtrawins (approved):

atobiszei (approved):

PR #4377, opened 2026-07-16
Fix windows cache_dir path 

rasapala:
### 🛠 Summary

JIRA CVS-187603
Normalize path and add test.

### 🧪 Checklist

- [ ] Unit tests added.
``


src/graph_export/graph_export.cpp
```diff
@@ -98,6 +98,16 @@ static std::string constructModelsPath(const std::string& modelPath, const std::
...
 
+static std::string normalizeWindowsPathSeparators(const std::string& path) {
+#ifdef _WIN32
+    std::string normalized = path;
+    std::replace(normalized.begin(), normalized.end(), '\\', '/');
+    return normalized;
+#else
+    return path;
+#endif
+}
```
mzegla: Don't we have some utils function for that already? Looks like a fairly common logic.

src/graph_export/graph_export.cpp
```diff
@@ -662,6 +672,13 @@ std::variant<std::optional<std::string>, Status> GraphExport::createPluginString
         if (d.Parse(stringPluginConfig.value().c_str()).HasParseError()) {
             return StatusCode::PLUGIN_CONFIG_WRONG_FORMAT;
         }
+        // plugin_config is injected into pbtxt and later parsed as JSON again.
+        // Normalize Windows path separators to avoid backslash escape ambiguity.
+        auto cacheDirIt = d.FindMember("CACHE_DIR");
+        if (cacheDirIt != d.MemberEnd() && cacheDirIt->value.IsString()) {
+            auto normalizedCacheDir = normalizeWindowsPathSeparators(cacheDirIt->value.GetString());
+            cacheDirIt->value.SetString(normalizedCacheDir.c_str(), static_cast<rapidjson::SizeType>(normalizedCacheDir.size()), d.GetAllocator());
+        }
```
mzegla, line 662: Why are we doing it here and not through `pluginConfig.cacheDir.has_value()` like for `kvCachePrecision` below?

atobiszei:
The --cache_dir would propagate to classic models when using config.json if I recall correctly. Fix in graph_export would resolve the issue with OVMS started with single GenAI model. It won't work with --config_path i guess?
Is the fix with plugin config handling generic (ie does the issue exist also for non text GenAI models? Speech/Image gen).

src/graph_export/graph_export.cpp
rasapala: If we did normalization only in the pluginConfig.cacheDir.has_value() branch, we would miss CACHE_DIR coming from manualString.

rasapala:
@atobiszei  So yes: graph_export fix helps generated-graph flows (single task / pull flows), but not automatic injection for pre-existing graph.pbtxt under config_path - this is fixed in other PR.

About genericity across GenAI types:
Yes, the runtime plugin_config normalization fix is generic for graph-based GenAI models, not only text.
All these parse via JsonParser.

src/graph_export/graph_export.cpp
```diff
@@ -652,6 +652,16 @@ std::variant<std::optional<std::string>, Status> GraphExport::createPluginString
...
             return StatusCode::PLUGIN_CONFIG_WRONG_FORMAT;
         }
+        // plugin_config is injected into pbtxt and later parsed as JSON again.
+        // Normalize Windows path separators to avoid backslash escape ambiguity.
+        auto cacheDirIt = d.FindMember("CACHE_DIR");
+        if (cacheDirIt != d.MemberEnd() && cacheDirIt->value.IsString()) {
+            auto normalizedCacheDir = std::string(cacheDirIt->value.GetString());
+#ifdef _WIN32
+            normalizedCacheDir = std::filesystem::path(normalizedCacheDir).generic_string();
+#endif
```
mzegla: Does it have to be windows specific part? Maybe we can do it for both and avoid ifdef?

src/graph_export/graph_export.cpp
```diff
@@ -703,8 +713,12 @@ std::variant<std::optional<std::string>, Status> GraphExport::createPluginString
         }
     }
     if (exportSettings.pluginConfig.cacheDir.has_value()) {
+        auto normalizedCacheDir = exportSettings.pluginConfig.cacheDir.value();
+#ifdef _WIN32
+        normalizedCacheDir = std::filesystem::path(normalizedCacheDir).generic_string();
+#endif
```
mzegla: same as above - maybe we could always go through filesystem::path

src/json_parser.cpp
```diff
@@ -57,6 +58,10 @@ std::string numericValueToString(const rapidjson::Value& v) {
     return "";
 }
 
+bool isPathLikePluginKey(const std::string& key) {
+    return key == "CACHE_DIR";
+}
```
mzegla: Odd naming. Quite generic name for very strict condition. Can we name this method to instantly indicate the purpose - so checking cache dir property?

src/json_parser.cpp
```diff
@@ -128,7 +141,15 @@ Status JsonParser::parsePluginConfig(const rapidjson::Value& node, plugin_config
...
-                    pluginConfig[it->name.GetString()] = it->value.GetString();
+                    if (isPathLikePluginKey(topKey)) {
+                        auto normalizedValue = std::string(it->value.GetString());
+#ifdef _WIN32
+                        normalizedValue = std::filesystem::path(normalizedValue).generic_string();
+#endif
+                        pluginConfig[topKey] = normalizedValue;
+                    } else {
+                        pluginConfig[topKey] = it->value.GetString();
+                    }
```
mzegla: second appearance of the very similar logic - if isPathLikePluginKey is only checking cache dir then maybe we could have function like `normalizeCacheDirPropertyIfNeeded` that would accept pluginConfig/properties and depending on cache dir existance it would normalize it or not. This way here and above we could simply call 
`normalizeCacheDirPropertyIfNeeded(properties/pluginConfig)`

src/json_parser.cpp
```diff
@@ -58,8 +58,11 @@ std::string numericValueToString(const rapidjson::Value& v) {
     return "";
 }
 
-bool isPathLikePluginKey(const std::string& key) {
-    return key == "CACHE_DIR";
+void normalizeCacheDirPropertyIfNeeded(plugin_config_t& config, const std::string& key, const rapidjson::Value& value) {
+    if (key != "CACHE_DIR" || !value.IsString()) {
+        return;
+    }
+    config[key] = std::filesystem::path(value.GetString()).generic_string();
```
mzegla: use new function toGenericPath here?

src/json_parser.cpp
```diff
@@ -93,13 +96,8 @@ Status JsonParser::parsePluginConfig(const rapidjson::Value& node, plugin_config
...
-                                auto normalizedValue = std::string(propertyIt->value.GetString());
-#ifdef _WIN32
-                                normalizedValue = std::filesystem::path(normalizedValue).generic_string();
-#endif
-                                properties[propertyKey] = normalizedValue;
-                            } else {
+                            normalizeCacheDirPropertyIfNeeded(properties, propertyKey, propertyIt->value);
+                            if (propertyKey != "CACHE_DIR") {
                                 properties[propertyKey] = propertyIt->value.GetString();
                             }
```
mzegla: If we have to check for CACHE_DIR here anyway maybe we should go with other approach like:
```
normalizeCacheDirPropertyIfNeeded(properties, propertyKey, propertyIt->value);
if (propertyKey == "CACHE_DIR") {
	properties[propertyKey] = toGenericPath(propertyIt->value.GetString());
} else {
	properties[propertyKey] = propertyIt->value.GetString();
}
```

Then we don't even need specialized method for this property: `normalizeCacheDirPropertyIfNeeded`.
Would it work?

atobiszei (approved):

mzegla (approved):

PR #4378, opened 2026-07-16
Cache dir propagation fix for llms, image_gen and audio.

rasapala:
### 🛠 Summary

Issue https://github.com/openvinotoolkit/model_server/issues/4230

Cherry-pick https://github.com/openvinotoolkit/model_server/pull/4329

Align cache_dir behavior across classic models and GenAI pipelines, and document the effective precedence that is now covered by tests.

Classic ModelInstance-based models use the ModelManager cache directory unconditionally. The global cache_dir value applied by the manager overrides any per-model cache_dir that may come from config loading.
GenAI graph nodes keep an explicit CACHE_DIR from graph.pbtxt plugin_config. That node-level value wins over the global cache_dir.
GenAI graph nodes without CACHE_DIR in plugin_config inherit the global cache_dir from CLI or server settings as a fallback.
The config.json path for multiple GenAI models is covered as well: one graph with explicit CACHE_DIR keeps its own value, while another graph in the same config without CACHE_DIR receives the global fallback.
The classic config.json model-loading path is covered separately to confirm that the manager-level cache_dir is propagated to loaded ModelInstance configs.
Tests cover both ModelManager cache_dir propagation and GenAI precedence for single-node, multi-node, and multi-model config.json scenarios.

### 🧪 Checklist

- [ ] Unit tests added.
``


src/test/llm/llmnode_test.cpp
```diff
@@ -4501,6 +4502,222 @@ TEST_F(LLMVLMOptionsHttpTest, LLMVLMNodeOptionsCheckPluginConfig) {
...
+        }
+        }
+    )";
+    adjustConfigForTargetPlatform(testPbtxt);
+    ::mediapipe::CalculatorGraphConfig config;
+    ASSERT_TRUE(::google::protobuf::TextFormat::ParseFromString(testPbtxt, &config));
+    std::shared_ptr<GenAiServable> servable;
+    ASSERT_EQ(initializeGenAiServable(servable, config.node(0), ""), StatusCode::OK);
+
+    bool foundCacheArtifact = false;
```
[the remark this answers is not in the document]
rasapala: OpenVINO writes cache artifacts at ov::Core::compile_model() time, which happens synchronously inside the ContinuousBatchingPipeline] constructor that initializeGenAiServable calls. Cache writing is a side effect of construction, not of inference.

atobiszei:
I think we are missing handling at least for image gen cases:
https://github.com/openvinotoolkit/model_server/blob/0447498ea4633a136afcb13e97df5aa52f590735/src/image_gen/imagegen_init.cpp#L280
What about audio?

src/llm/servable_initializer.cpp
```diff
@@ -36,6 +37,7 @@
 #pragma GCC diagnostic pop
 #pragma warning(pop)
 
+#include "../config.hpp"
```
atobiszei, line 40:
```suggestion
#include "src/config.hpp"
```

atobiszei: And thats probably why bazel didnt catch this when theres no change in BUILD file.

rasapala: Lets leave it this way.

mzegla (approved):

atobiszei (approved):

PR #4379, opened 2026-07-16
Omni legacy pipeline

dkalinowski:
CVS-191153


src/llm/omni_model/legacy/legacy_executor.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include "servable.hpp"
+
+namespace ovms {
+OmniModelLegacyExecutor::OmniModelLegacyExecutor(std::shared_ptr<ov::genai::OmniPipeline> pipe) {
+    this->pipe = std::move(pipe);
+}
+
+bool OmniModelLegacyExecutor::hasRequests() {
+    return (requests.size() > 0);
+}
```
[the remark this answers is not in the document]
dkalinowski: I dont think we need to add lock here. Similarly to other legacy servables, we rely on a fact that hasRequest can only be called from executor thread and addRequest from calculator thread. HasRequest reads stale true/false value and if we read it wrong, we wil get it right in the next iteration anyway

src/llm/omni_model/legacy/legacy_executor.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+size_t OmniModelLegacyExecutor::requestsQueueSize() {
+    return requests.size();
+}
```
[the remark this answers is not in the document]
dkalinowski: No control flow depends on it, it is purely for logging. Same as in visual_language_model legacy pipeline as well

src/llm/omni_model/legacy/legacy_executor.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+void OmniModelLegacyExecutor::processRequest() {
+    OVMS_PROFILE_FUNCTION();
+    auto& requestExecutionContext = requests.front();
+    if (requestExecutionContext->clientDisconnected) {
```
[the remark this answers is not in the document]
dkalinowski: pushing to the back does not invalidate references to existing elements, this is a guaranteed property of std::deque

src/audio/audio_utils.cpp
```diff
@@ -300,3 +300,82 @@ void validateAudioFileSize(
     expectedSize *= bytesPerSample;
     validateAudioFileSizeAgainstMaxValue(expectedSize);
 }
+
+std::vector<float> readWithoutResample(const std::string_view& audioData, const std::string& format) {
```
dkalinowski, line 307: merge with readMp3/readWav and optional parameter to fix sample rate?

src/llm/apis/openai_api_handler.cpp
```diff
@@ -762,6 +762,39 @@ absl::Status OpenAIApiHandler::parseCommonPart(std::optional<uint32_t> maxTokens
 
     request.maxModelLength = maxModelLength;
 
+    // modalities: array of strings; optional — controls output modalities
+    it = doc.FindMember("modalities");
```
dkalinowski, line 766: Per OpenAI's behavior with gpt-4o-audio: when modalities: ["audio"] is set without "text", the response still includes a transcript in audio.transcript but message.content is null/empty.

Currently we don't distinguish this - text is always included in the response regardless. To be fully compliant, when "text" is NOT in modalities but "audio" is, we should:

Set message.content to null (or omit it)
Still include audio.transcript with the text

src/llm/io_processing/input_processors/image_decoding_processor.cpp
```diff
@@ -81,16 +84,35 @@ absl::Status ImageDecodingProcessor::process(InputRequest& req) {
...
-        if (!imageTags.empty() || !textContent.empty()) {
+        if (hasOtherParts) {
+            // Rebuild content array preserving original order: replace image_url entries
+            // with image tag text parts, keep text and other parts (e.g. input_audio) in place.
+            auto newContent = ov::genai::JsonContainer::array();
+            for (size_t j = 0; j < content.size(); j++) {
+                const auto part = content[j];
+                const auto type = part["type"].as_string().value_or("");
+                if (type == "image_url") {
+                    ov::genai::JsonContainer textEntry({{"type", "text"}, {"text", imageTagByPart[j]}});
```
dkalinowski: it will work only for omni i think, needs to be changed

demos/omni/chat_completions.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2024 Intel Corporation
```
mzegla:
```suggestion
# Copyright (c) 2026 Intel Corporation
```

demos/omni/responses.py
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2024 Intel Corporation
```
mzegla:
```suggestion
# Copyright (c) 2026 Intel Corporation
```

docs/omni.md
```diff
@@ -0,0 +1,201 @@
+# Omni Model Support
+
+OpenVINO Model Server supports **Qwen3-Omni** models — multimodal models that accept text, images, and audio as input and produce text and speech as output.
+
+## Supported Models
+
+- Qwen3-Omni Dense (INT4/INT8/FP16)
+- Qwen3-Omni MoE (INT4/INT8/FP16) - **does not work yet**
```
mzegla: Let's include FP32 to have full set. Unless it's not supported, but that would be a bit suspicious.

docs/omni.md
```diff
@@ -0,0 +1,201 @@
...
+## Quick Start
+
+### Model Setup
+
+The model directory must contain `openvino_talker_model.xml` — OVMS auto-detects it as an Omni pipeline.
+
+```
+models/qwen3-omni/
+├── openvino_model.xml          # thinker (VLM)
+├── openvino_talker_model.xml   # talker (speech)
```
mzegla: Why the comments? If we add comments to those two models, why not for other XMLs?

docs/omni.md
```diff
@@ -0,0 +1,201 @@
...
+├── openvino_code_predictor_model.xml
+├── openvino_code2wav_model.xml
+├── config.json
+├── generation_config.json
+├── tokenizer.xml / tokenizer.json
+└── ...
+```
+
+### Graph Configuration
+It should be inside model folder, or simply run OVMS with `--task text_generation` param so it will automatically create one in memory.
```
mzegla: Can you try without the `--task`? It should be auto detected now.

mzegla: Also since it's omni, is text_generation proper task?

src/audio/audio_utils.cpp
mzegla: I think so

src/llm/apis/openai_request.hpp
```diff
@@ -81,6 +81,11 @@ struct OpenAIRequest {
 
     bool skipSpecialTokens{true};
 
+    // Audio output (Omni pipeline)
+    bool audioOutputRequested{false};
+    std::string audioVoice;          // maps to OmniTalkerSpeechConfig::speaker
+    std::string audioFormat{"wav"};  // "wav" or "pcm16"
```
mzegla: Could be some oneOf/variant type? that would handle validation naturally?

src/llm/io_processing/input_processors/audio_decoding_processor.cpp
```diff
@@ -0,0 +1,105 @@
...
+            const auto part = content[j];
+            const auto type = part["type"].as_string().value_or("");
+
+            if (type == "input_audio") {
+                hasAudio = true;
+                const auto data = part["input_audio"]["data"].as_string().value_or("");
+                const auto format = part["input_audio"]["format"].as_string().value_or("wav");
+
+                if (data.empty()) {
+                    return absl::InvalidArgumentError("input_audio data field is empty");
```
mzegla, line 61: Wonder if this validation is not too strict. Can user send other valid modalities and input_audio with no data?
Thinking about some existing frameworks and integrations that might tear it loosely. Is OpenAI API spec strict about it?

src/llm/io_processing/input_processors/audio_decoding_processor.cpp
```diff
@@ -0,0 +1,105 @@
...
+                    std::memcpy(audioTensor.data<float>(), pcm.data(), pcm.size() * sizeof(float));
+                    req.inputAudios.push_back(std::move(audioTensor));
+                } catch (const std::exception& e) {
+                    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Audio decoding failed: {}", e.what());
+                    return absl::InvalidArgumentError(std::string("Audio decoding failed: ") + e.what());
+                }
+            }
+        }
+
+        // Remove input_audio parts from the content array after extracting tensor data.
```
mzegla, line 84: Just a note - in the future we would likely want to keep some part of input audio content for the sake of building the prompt.

src/llm/io_processing/input_processors/image_decoding_processor.cpp
```diff
@@ -66,9 +66,12 @@ absl::Status ImageDecodingProcessor::process(InputRequest& req) {
             continue;
         }
 
-        // Accumulate image tags and text parts from a single message's content array.
+        // First pass: decode images and detect non-image/non-text parts.
+        // Track image tag per content part index for the rebuild.
+        std::vector<std::string> imageTagByPart(content.size());
         std::string imageTags;
         std::string textContent;
+        bool hasOtherParts = false;
```
mzegla: What is "other parts"? Too wide and generic to me. Let's be more specific here.

src/llm/io_processing/input_processors/image_decoding_processor.cpp
```diff
@@ -66,9 +66,12 @@ absl::Status ImageDecodingProcessor::process(InputRequest& req) {
             continue;
         }
 
-        // Accumulate image tags and text parts from a single message's content array.
+        // First pass: decode images and detect non-image/non-text parts.
```
mzegla: First pass?

src/llm/io_processing/input_processors/image_decoding_processor.cpp
```diff
@@ -81,16 +84,35 @@ absl::Status ImageDecodingProcessor::process(InputRequest& req) {
...
                 }
                 textContent += part["text"].as_string().value_or("");
+            } else {
+                hasOtherParts = true;
             }
         }
 
-        if (!imageTags.empty() || !textContent.empty()) {
+        if (hasOtherParts) {
+            // Rebuild content array preserving original order: replace image_url entries
```
mzegla: This will cost as we would potentially copy big text content elements.  But perhaps its acceptable as a temporary solution.

src/llm/io_processing/input_processors/image_decoding_processor.cpp
```diff
@@ -81,16 +84,35 @@ absl::Status ImageDecodingProcessor::process(InputRequest& req) {
                     return imageResult.status();
                 }
                 req.inputImages.push_back(std::move(imageResult).value());
-                imageTags += "<ov_genai_image_" + std::to_string(imageIndex++) + ">\n";
+                imageTagByPart[j] = "<ov_genai_image_" + std::to_string(imageIndex) + ">\n";
+                imageTags += imageTagByPart[j];
```
mzegla: is `imageTags` needed anymore now?

src/llm/omni_model/legacy/legacy_executor.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+bool OmniModelLegacyExecutor::hasRequests() {
+    return (requests.size() > 0);
+}
+
+size_t OmniModelLegacyExecutor::requestsQueueSize() {
+    return requests.size();
+}
+
+void OmniModelLegacyExecutor::processRequest() {
```
mzegla, line 38: This method mixes logging levels heavily. Please use DEBUG/TRACE only for per request info.

src/llm/omni_model/legacy/legacy_executor.cpp
```diff
@@ -0,0 +1,197 @@
...
+    OVMS_PROFILE_FUNCTION();
+    auto& requestExecutionContext = requests.front();
```
mzegla, line 40: Can we move some of this logic to the servable like `prepareInputs` etc.?
I feel like at least creating generation config belongs there. Ideally this function would go straight to `generate()`. Notice that executor logic is one threaded and queued, so if we can do something on the server worker (servable/calculator), let's do it there.

Also:
- are you sure we don't need to lock queue mutex here?
- is it guaranteed that we have valid `front` element here?

src/llm/omni_model/legacy/servable.cpp
```diff
@@ -0,0 +1,402 @@
...
+               payload.uri == "/v1/responses") {
+        executionContext->endpoint = Endpoint::RESPONSES;
+    } else if (TokenizeParser::isTokenizeEndpoint(payload.uri)) {
+        executionContext->endpoint = Endpoint::TOKENIZE;
+    } else {
+        return absl::InvalidArgumentError("Wrong endpoint. Omni Servable allowed only on /v[13]/chat/completions, /v[13]/responses, /v[13]/tokenize");
+    }
+    executionContext->payload = payload;
+    return absl::OkStatus();
+}
```
mzegla: This seriously needs to be reworked. That logic does nothing more than endpoint selection and copying payload into the context right? 
If the only specific thing here is mentioning Omni Servable in the log I don't think it's needed. 

Could you check if all servables have method that looks like this? If so, maybe we can reduce it to one implementation and add to core GenAIServableProperties member like `supportedEndpoints` that initializers would fill?

src/llm/omni_model/legacy/servable.cpp
```diff
@@ -0,0 +1,402 @@
...
+    if (!inputRequestResult.ok()) {
+        return inputRequestResult.status();
+    }
+    omniExecutionContext->inputRequest = std::move(*inputRequestResult);
+
+    // Copy audio output config from parsed request to execution context
+    const auto& req = omniExecutionContext->apiHandler->getRequest();
+    omniExecutionContext->audioOutputRequested = req.audioOutputRequested;
+    omniExecutionContext->audioVoice = req.audioVoice;
+    omniExecutionContext->audioFormat = req.audioFormat;
```
mzegla, line 174: Is that the only omni specific part in this method? If so, could we call base class parseRequest on this and then cast context to omni and fill remaining fields?

src/llm/omni_model/legacy/servable.cpp
```diff
@@ -0,0 +1,402 @@
...
+    auto omniExecutionContext = std::static_pointer_cast<OmniModelLegacyServableExecutionContext>(executionContext);
+    if (omniExecutionContext->payload.client->isDisconnected()) {
+        return absl::CancelledError();
+    }
+    omniExecutionContext->finished.wait();
+    if (!omniExecutionContext->success) {
+        return absl::InvalidArgumentError("Request processing failed, check its correctness.");
+    }
+    return absl::OkStatus();
+}
```
mzegla, line 244: Do we need implementation of that? Looks kind of generic - wouldn't base class be able to handle this?

src/llm/omni_model/legacy/servable.cpp
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Complete unary response: {}", executionContext->response);
+    return absl::OkStatus();
+}
+
+absl::Status OmniModelLegacyServable::readPartialExecutionResults(std::shared_ptr<GenAiServableExecutionContext>& executionContext) {
+    executionContext->deltaChannel.waitForData();
+    return absl::OkStatus();
+}
```
mzegla, line 406: Same, looks very generic, not sure if we need implementation here

src/llm/omni_model/legacy/servable.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <openvino/genai/omni/pipeline.hpp>
+#include <openvino/genai/omni/decoded_results.hpp>
+
+#include "../../servable.hpp"
+#include "legacy_executor.hpp"
+#include "src/llm/llm_calculator.pb.h"
+
+namespace ovms {
+
+struct OmniModelLegacyServableExecutionContext : public GenAiServableExecutionContext {
```
mzegla, line 34: How about VLM servable execution context as base? Have you considered that?
Same for all other classes. Maybe it would allow us to reuse some more generic methods?

new_class_diagram.png
mzegla, line 1: to be removed?

docs/omni.md
dkalinowski: I wanted to showcase whats new, it is not really user facing documentation yet I would say, updated to this for now:

image

but after we merge it we should add sphinx doc and actual demo similar to our other demos

docs/omni.md
dkalinowski: yes, it is proper task. it works without --task as well, i think i will completely remove this paragraph since its optional

src/llm/apis/openai_request.hpp
dkalinowski: Good point - will add enum

src/llm/io_processing/input_processors/audio_decoding_processor.cpp
dkalinowski: It is struct but it follows OpenAI spec. I think we should make it more forgiving once we see there is some "bugged" application out there, let's keep it correct for now?

src/llm/omni_model/legacy/legacy_executor.cpp
dkalinowski: I dont see any input preparation code in this function. Maybe you are pointing at audio streamer definition? It contains audio operations. However, it is not executed before ::generate. It is only streamer definition.

Regarding locks - it is pure copy of VLM servable so I dont think we need to re-investigate it over again

src/llm/omni_model/legacy/servable.cpp
```diff
@@ -0,0 +1,402 @@
...
+            auto& alloc = responseDoc.GetAllocator();
+            if (responseDoc.HasMember("choices") && responseDoc["choices"].IsArray() && responseDoc["choices"].Size() > 0) {
+                // Chat Completions format: inject into choices[0].message.audio
+                auto& message = responseDoc["choices"][0]["message"];
+                rapidjson::Value audioObj(rapidjson::kObjectType);
+                audioObj.AddMember("data", rapidjson::Value(audioBase64.c_str(), alloc), alloc);
+                std::string transcript = message.HasMember("content") && message["content"].IsString() ? message["content"].GetString() : "";
+                audioObj.AddMember("transcript", rapidjson::Value(transcript.c_str(), alloc), alloc);
+                message.AddMember("audio", audioObj, alloc);
+            } else if (responseDoc.HasMember("output") && responseDoc["output"].IsArray()) {
```
dkalinowski: probably base on apihandler->api type or something like that

src/llm/apis/openai_responses.cpp
```diff
@@ -1449,8 +1480,23 @@ std::string OpenAIResponsesHandler::serializeStreamingChunk(rapidjson::Document
 
     // parsedDelta is a pre-parsed Document produced by OVMSTextStreamer::flushChunk.
     // Shape: {"delta":{...}} for content/reasoning/tool_calls, or an empty Document{}
-    // for finish-only chunks. Inspect the delta directly — no parsing needed here.
-    if (parsedDelta.HasMember("delta") && parsedDelta["delta"].IsObject()) {
+    // for finish-only chunks, or {"_audio_delta":"<base64>"} for audio chunks.
```
dkalinowski, line 1495: the comment should be moved below

src/llm/omni_model/legacy/servable.cpp
```diff
@@ -0,0 +1,392 @@
...
```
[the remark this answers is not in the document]
dkalinowski: it will base on executionContext->apiHandler->getEndpoint()

src/audio/audio_utils.cpp
dkalinowski: done

src/llm/apis/openai_responses.cpp
dkalinowski: it still applies in general, so i extended it

src/llm/omni_model/legacy/legacy_executor.cpp
dkalinowski: done

src/audio/audio_utils.cpp
```diff
@@ -300,3 +300,66 @@ void validateAudioFileSize(
...
+            throw std::runtime_error("WAV audio must be mono or stereo");
+        }
+        if (wav.bitsPerSample < 8 || (wav.bitsPerSample % 8) != 0) {
+            drwav_uninit(&wav);
+            throw std::runtime_error("WAV audio has unsupported bits per sample");
+        }
+        const uint64_t n = wav.totalPCMFrameCount;
+        std::vector<int16_t> pcm16(n * wav.channels);
+        drwav_read_pcm_frames_s16(&wav, n, pcm16.data());
+        drwav_uninit(&wav);
```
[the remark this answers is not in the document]
dkalinowski: done

src/audio/audio_utils.cpp
```diff
@@ -300,3 +300,66 @@ void validateAudioFileSize(
...
+    } else if (format == "mp3") {
+        drmp3 mp3;
+        if (!drmp3_init_memory(&mp3, audioData.data(), audioData.size(), nullptr)) {
+            throw std::runtime_error("MP3 audio parsing failed");
+        }
+        if (mp3.channels != 1 && mp3.channels != 2) {
+            drmp3_uninit(&mp3);
+            throw std::runtime_error("MP3 audio must be mono or stereo");
+        }
+        constexpr size_t CHUNK_FRAMES = 1152;
```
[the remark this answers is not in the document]
dkalinowski: done

src/llm/apis/openai_api_handler.cpp
dkalinowski: done

src/llm/io_processing/input_processors/image_decoding_processor.cpp
dkalinowski: done

src/llm/apis/openai_request.hpp
dkalinowski: added

src/llm/io_processing/input_processors/image_decoding_processor.cpp
dkalinowski: fixed it entirely after Junrui's PR it looks different now. please take a look

src/llm/io_processing/input_processors/image_decoding_processor.cpp
dkalinowski: fixed it entirely after Junrui's PR it looks different now. please take a look

it is: 
ImageDecodingProcessor -> only decodes images and converts image content part to text with tag
AudioDecodingProcessor -> only decodes audio and removes audio part
Flattener flattens every text remaining to single text with new lines

src/llm/io_processing/input_processors/image_decoding_processor.cpp
dkalinowski: same as above

src/llm/io_processing/input_processors/image_decoding_processor.cpp
dkalinowski: same as above

src/llm/omni_model/legacy/servable.cpp
dkalinowski: done, moved to base class now

src/llm/omni_model/legacy/servable.cpp
dkalinowski: I think we should do it in next PR - it is related to VLMPipeline & OmniPipeline cleanup

src/llm/omni_model/legacy/servable.cpp
dkalinowski: I think we should do it in next PR - it is related to VLMPipeline & OmniPipeline cleanup

src/llm/omni_model/legacy/servable.cpp
dkalinowski: I think we should do it in next PR - it is related to VLMPipeline & OmniPipeline cleanup

src/llm/omni_model/legacy/servable.cpp
dkalinowski: done

src/audio/audio_utils.hpp
```diff
@@ -21,10 +21,24 @@
 #include <stdint.h>
 #include <string_view>
 
+namespace ovms::audio_utils {
+
+static constexpr uint32_t PIPELINE_SUPPORTED_SAMPLE_RATE = 16000;
```
mzegla: Will read methods work with custom sampling rate?
This name is also quite generic - what pipeline?

src/audio/audio_utils.cpp
```diff
@@ -300,3 +303,15 @@ void validateAudioFileSize(
...
+
+std::vector<float> readWithoutResample(const std::string_view& audioData, const std::string& format) {
+    if (format == "wav") {
+        return readWav(audioData, 0);
+    } else if (format == "mp3") {
+        return readMp3(audioData, 0);
+    } else {
+        throw std::runtime_error("Unsupported audio format: " + format);
+    }
+}
```
mzegla, line 315: If 0 is a special value to disable resampling, maybe we should make it a const and somehow document (like explicitly in the comment)?

src/audio/audio_utils.cpp
```diff
@@ -107,7 +106,9 @@ std::vector<float> readWav(const std::string_view& wavData) {
         throw std::runtime_error("WAV file header claims more frames than possible from data chunk size");
     }
     // Validate decoded buffer size before resampling (float32 mono output)
-    validateAudioFileSize(wav.totalPCMFrameCount, wav.sampleRate, PIPELINE_SUPPORTED_SAMPLE_RATE, /*will be downmixed to mono*/ 1, sizeof(float));
+    if (targetSampleRate > 0) {
```
mzegla, line 109: This comment goes for mp3 method as well. 
Do we need validation of `targetSampleRate` for upper boundaries?
Especially if it's possible to specify from the request level, I would question if allowing max uint32 through is safe.

src/llm/apis/openai_api_handler.cpp
```diff
@@ -762,6 +762,61 @@ absl::Status OpenAIApiHandler::parseCommonPart(std::optional<uint32_t> maxTokens
...
+                return absl::InvalidArgumentError("modalities array must contain strings");
+            const std::string modality = mod.GetString();
+            if (modality == "audio") {
+                hasAudio = true;
+            } else if (modality == "text") {
+                hasText = true;
+            }
+        }
+        request.audioOutputRequested = hasAudio;
+        // When modalities is explicitly provided and "text" is absent, suppress text in response
```
mzegla:
```suggestion
        // When "modalities" is explicitly provided and "text" is absent, suppress text in response
```

src/llm/apis/openai_api_handler.cpp
```diff
@@ -762,6 +762,61 @@ absl::Status OpenAIApiHandler::parseCommonPart(std::optional<uint32_t> maxTokens
...
+        bool hasAudio = false;
+        for (const auto& mod : it->value.GetArray()) {
+            if (!mod.IsString())
+                return absl::InvalidArgumentError("modalities array must contain strings");
+            const std::string modality = mod.GetString();
+            if (modality == "audio") {
+                hasAudio = true;
+            } else if (modality == "text") {
+                hasText = true;
+            }
```
mzegla, line 780: No images/video modalities?

src/llm/apis/openai_api_handler.cpp
```diff
@@ -762,6 +762,61 @@ absl::Status OpenAIApiHandler::parseCommonPart(std::optional<uint32_t> maxTokens
...
+            std::string fmt = formatIt->value.GetString();
+            if (fmt == "wav") {
+                request.audioFormat = OpenAIRequest::AudioFormat::WAV;
+            } else if (fmt == "pcm16") {
+                request.audioFormat = OpenAIRequest::AudioFormat::PCM16;
+            } else {
+                return absl::InvalidArgumentError("audio.format must be \"wav\" or \"pcm16\"");
+            }
+        }
+        // This is not OpenAI standard, but we allow it since it is configuratble in OpenVINO GenAI
```
mzegla:
```suggestion
        // This is not OpenAI standard, but we allow it since it is configurable in OpenVINO GenAI
```

src/llm/apis/openai_api_handler.cpp
```diff
@@ -762,6 +762,61 @@ absl::Status OpenAIApiHandler::parseCommonPart(std::optional<uint32_t> maxTokens
...
+        auto chunkFramesIt = audioObj.FindMember("chunk_frames");
+        if (chunkFramesIt != audioObj.MemberEnd() && chunkFramesIt->value.IsUint()) {
+            request.audioChunkFrames = chunkFramesIt->value.GetUint();
+            if (request.audioChunkFrames < 1) {
+                return absl::InvalidArgumentError("audio.chunk_frames must be >= 1");
+            }
```
mzegla: Since `chunk_frames` is not in OpenAI API spec and is unique to OV GenAI I think it belongs to extra params.
Including it in here might pose a risk of integration with external tools and frameworks and would be less straightforward to remove when it's no longer needed.

src/llm/apis/openai_completions.cpp
```diff
@@ -219,6 +219,14 @@ absl::Status OpenAIChatCompletionsHandler::parseMessages(std::optional<std::stri
                         if (!imageUrl.HasMember("url") || !imageUrl["url"].IsString()) {
                             return absl::InvalidArgumentError("Invalid message structure - image_url does not have url field");
                         }
+                    } else if (entryType == "input_audio") {
+                        if (!entry.HasMember("input_audio") || !entry["input_audio"].IsObject()) {
+                            return absl::InvalidArgumentError("Invalid message structure - input_audio object missing");
+                        }
+                        const auto inputAudio = entry["input_audio"].GetObject();
+                        if (!inputAudio.HasMember("data") || !inputAudio["data"].IsString()) {
+                            return absl::InvalidArgumentError("Invalid message structure - input_audio does not have data field");
```
mzegla:
```suggestion
                            return absl::InvalidArgumentError("Invalid message structure - input_audio does not have a valid data field");
```
With that it would cover case when there is delta, but it's not a string.

src/llm/apis/openai_request.hpp
```diff
@@ -81,6 +81,15 @@ struct OpenAIRequest {
...
     bool skipSpecialTokens{true};
 
+    // Audio output (Omni pipeline)
+    enum class AudioFormat { WAV,
+        PCM16 };
+    bool audioOutputRequested{false};
+    bool textOutputRequested{true};  // false when modalities is ["audio"] without "text"
+    std::string audioVoice;          // maps to OmniTalkerSpeechConfig::speaker
+    AudioFormat audioFormat{AudioFormat::WAV};
+    size_t audioChunkFrames{4};  // Number of codec frames per streaming audio chunk (each frame = 80ms @ 24kHz)
```
mzegla: Why 4? Could we name it in a const var?

src/llm/apis/openai_responses.cpp
```diff
@@ -1447,10 +1478,25 @@ std::string OpenAIResponsesHandler::serializeStreamingChunk(rapidjson::Document
...
+        rapidjson::StringBuffer buf;
+        rapidjson::Writer<rapidjson::StringBuffer> writer(buf);
+        writer.StartObject();
+        writer.Key("type");
+        writer.String("response.audio.delta");
+        writer.Key("delta");
+        writer.String(audioB64.c_str());
+        writer.Key("sequence_number");
+        writer.Uint64(responsesState.sequenceNumber);
+        writer.EndObject();
```
mzegla: Would we wrap this in a tiny method accpeting audio string and sequence number and returning json object ref?
Ideally delta should be already with proper structure like in `OVMSTextStreamer` path, but I get it that we don't want our own implementation of another streamer...

src/llm/io_processing/input_processors/audio_decoding_processor.cpp
```diff
@@ -0,0 +1,105 @@
...
+            const auto type = part["type"].as_string().value_or("");
+
+            if (type == "input_audio") {
+                hasAudio = true;
+                const auto data = part["input_audio"]["data"].as_string().value_or("");
+                const auto format = part["input_audio"]["format"].as_string().value_or("wav");
+
+                if (data.empty()) {
+                    return absl::InvalidArgumentError("input_audio data field is empty");
+                }
```
mzegla, line 62: Do we allow to pass such data through request level validation?

src/llm/io_processing/input_processors/audio_decoding_processor.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+        bool hasAudio = false;
+        for (size_t j = 0; j < content.size(); j++) {
+            const auto part = content[j];
+            const auto type = part["type"].as_string().value_or("");
+
+            if (type == "input_audio") {
+                hasAudio = true;
+                const auto data = part["input_audio"]["data"].as_string().value_or("");
+                const auto format = part["input_audio"]["format"].as_string().value_or("wav");
```
mzegla, line 58: Not very sure about `value_or` here. I would say if we are here, input_audio should be valid and defined on earlier stages. I worry about multiple source of defaults and some implicit issues that would be hard to find (we think we take default from one place, but in reality its overwrote with some other value). What do you think?

src/llm/io_processing/input_processing_config.hpp
```diff
@@ -23,6 +23,9 @@ struct InputProcessingConfig {
     // still runs (to populate inputIds for usage statistics and max-length checks)
     // but inputIds is not passed to the VLM pipeline for inference.
     bool isVLM = false;
+    // True for Omni servables. Enables AudioDecodingProcessor in addition to
+    // ImageDecodingProcessor (implies isVLM-like behavior for images).
+    bool isOmni = false;
```
mzegla, line 28: I can see possibility that we will serve models that will accept audio input and will not be Omni. Like text, image, audio on the input and just text on the output. That's not Omni I suppose, but still requires audio processing. I'm thinking about renaming `isVLM`, `isOmni` to something explicit like `supportsVisualInput`, `supportsAudioInput`. 
That naming would remove pipeline type awareness from the scope and focus on input data types.

src/llm/io_processing/input_processor.cpp
```diff
@@ -42,17 +43,19 @@ InputProcessor::InputProcessor(InputProcessorContext& context,
...
             processors.emplace_back(std::make_unique<ImageDecodingProcessor>(
                 settings.allowedLocalMediaPath,
                 settings.allowedMediaDomains));
         }
 
+        if (context.config.isOmni) {
+            processors.emplace_back(std::make_unique<AudioDecodingProcessor>());
+        }
+
+        processors.emplace_back(std::make_unique<TextContentNormalizationProcessor>());
```
mzegla, line 57: We have a comment in input processor header I believe where we mention input processor ordering. Could you update it as well to reflect the new ordering?

src/llm/llm_calculator.proto
```diff
@@ -80,6 +80,8 @@ message LLMCalculatorOptions {
       VLM_CB = 3;
       // Default option. If selected, pipeline type will be inferred by the serving
       AUTO = 4;
+      // Omni model (text, image, audio input; text + speech output) based on OmniPipeline
+      OMNI = 5;
```
mzegla: I would prefer the default option to be first or last so it stands out among others. Would it be possible to for example move AUTO to the top with `0` identifier? Or would we experience some compatibility issues?

src/llm/servable_initializer.cpp
```diff
@@ -477,20 +479,25 @@ Status determinePipelineType(PipelineType& pipelineType, const mediapipe::LLMCal
...
 
     std::filesystem::path parsedModelsPathFs(parsedModelsPath);
+    // Existence of talker model indicates omni pipeline
+    bool hasTalkerModel = std::filesystem::exists(parsedModelsPathFs / "openvino_talker_model.xml");
     // Existence of embeddings models indicates we are dealing with VLM pipeline
-    bool isVLM = (std::filesystem::exists(parsedModelsPathFs / "openvino_text_embeddings_model.xml") &&
-                  std::filesystem::exists(parsedModelsPathFs / "openvino_vision_embeddings_model.bin"));
+    bool hasEmbeddingsModels = (std::filesystem::exists(parsedModelsPathFs / "openvino_text_embeddings_model.xml") &&
+                                   std::filesystem::exists(parsedModelsPathFs / "openvino_vision_embeddings_model.bin")) &&
+                               !hasTalkerModel;
```
mzegla: Does talker model existence automatically means model cannot have embedding models?

src/llm/servable_initializer.cpp
```diff
@@ -575,6 +585,14 @@ Status initializeGenAiServable(std::shared_ptr<GenAiServable>& servable, const :
                 SPDLOG_LOGGER_ERROR(modelmanager_logger, "Error during LLM node resources initialization: {}", status.string());
                 return status;
             }
+        } else if (pipelineType == PipelineType::OMNI) {
```
mzegla, line 592: Don't we need similar validation as for VLM to match behavior? For example if someone sets Omni explicitly, but there is no talker model in the catalog.

dkalinowski: When someone forces --pipeline_type VLM but the model is LLM:
> Models directory content indicates non-VLM pipeline, but pipeline type is set to VLM type.

When someone forces --pipeline_type OMNI but the model is VLM:
> Models directory content indicates VLM pipeline, but pipeline type is set to non-VLM type.

When someone forces --pipeline_type VLM but the model is OMNI:
> Models directory content indicates non-VLM pipeline, but pipeline type is set to VLM type.

I think it works as expected

src/llm/servable_initializer.cpp
dkalinowski: No, I will change the comment to:

```
// Existence of embeddings models indicates we are dealing with VLM pipeline
    // But if it has talker model, it means it is Omni pipeline which is built out of VLM Pipeline and Talker
```

src/llm/llm_calculator.proto
dkalinowski: I dont think it will introduce any compatibility issues - I will change it

src/llm/io_processing/input_processor.cpp
dkalinowski: done

src/llm/io_processing/input_processing_config.hpp
dkalinowski: This is good direction, but lets do it once we have actual models that are capable of that

src/llm/apis/openai_api_handler.cpp
dkalinowski: These modalities determine what model is capable of generating. We dont know any llm model that is capable of generating images. And it is not supported by OpenAI API anyway

src/llm/io_processing/input_processors/audio_decoding_processor.cpp
dkalinowski: At this moment it is the only source of default. It is optional param and only existence of data is validated beforehand. What would be your suggestion for placing defaults?

src/llm/io_processing/input_processors/audio_decoding_processor.cpp
dkalinowski: No we dont, but I thought it would be safer to check it anyway

src/llm/apis/openai_responses.cpp
dkalinowski: done

src/llm/apis/openai_request.hpp
dkalinowski: On B70 card 4 had better results and achieved real time streaming, where as GenAI's default 1 was slightly below that threshold meaning there were gaps between chunks during live playback.
Will add const var for magic 4 number.

src/llm/apis/openai_api_handler.cpp
dkalinowski: Good catch, moved it to top-level key

michalkulakowski (approved):

src/audio/audio_utils.cpp
dkalinowski: For now, readWav/readMp3 has always the same fixed target sample rate hardcoded: 16000
It is required for kokoro. I did not change behavior, only refactored it.
Omni does not need resampling on our level, it is done internally by genai.
Target sample rate is not user-facing param, it is our hardcoded constant.

src/audio/audio_utils.cpp
dkalinowski: It is documented in code:

```
// Decode WAV data into mono float32 PCM samples.
// If targetSampleRate > 0, resamples to that rate. Otherwise returns at native sample rate.
std::vector<float> readWav(const std::string_view& wavData, uint32_t targetSampleRate = PIPELINE_SUPPORTED_SAMPLE_RATE);

// Decode MP3 data into mono float32 PCM samples.
// If targetSampleRate > 0, resamples to that rate. Otherwise returns at native sample rate.
std::vector<float> readMp3(const std::string_view& mp3Data, uint32_t targetSampleRate = PIPELINE_SUPPORTED_SAMPLE_RATE);
```

but I will make 0 a constexpr RESAMPLE_DISABLED

src/llm/servable_initializer.cpp
mzegla: Still, the comment does not resolve the way `hasEmbeddingsModels` means. With that change it is dependent of `hasTalkerModel` which blurs the meaning. 
Maybe we need another bool here? Or change conditions to explicitly `if(hasEmbeddingsModels && !hasTalkerModel) then VLM`

src/config.cpp
```diff
@@ -198,9 +198,9 @@ bool Config::validate() {
             }
             const auto& exportSettings = this->serverSettings.hfSettings.exportSettings;
             auto textGenSettings = std::get<TextGenGraphSettingsImpl>(this->serverSettings.hfSettings.graphSettings);
-            std::vector allowedPipelineTypes = {"LM", "LM_CB", "VLM", "VLM_CB", "AUTO"};
+            std::vector allowedPipelineTypes = {"LM", "LM_CB", "VLM", "VLM_CB", "AUTO", "OMNI"};
```
mzegla: Can we keep the ordering as in the graph? So AUTO at the end?

src/test/llm/input_processing/audio_decoding_processor_test.cpp
```diff
@@ -0,0 +1,308 @@
...
+    ov::genai::ChatHistory history;
+    history.push_back({{"role", "system"}, {"content", "You are helpful."}});
+    history.push_back({{"role", "user"}, {"content", "What is OpenVINO?"}});
+
+    InputRequest req = makeChatRequest(history);
+    AudioDecodingProcessor processor;
+    const auto status = processor.process(req);
+
+    EXPECT_TRUE(status.ok());
+    EXPECT_TRUE(req.inputAudios.empty());
```
mzegla, line 89: can we also check that processor did not remove anything, so the history is intact

src/test/llm/input_processing/input_processing_integration_test.cpp
mzegla, line 1: Have you considered a test with image, audio and normalization processor working together?

dkalinowski: added

src/test/llm/input_processing/audio_decoding_processor_test.cpp
dkalinowski: done

src/config.cpp
dkalinowski: done

src/llm/omni_model/legacy/legacy_executor.cpp
dkalinowski: Moved all preparation to prepareInput, processRequest consists only of generate call now

mzegla (approved):

PR #4385, opened 2026-07-21
Add automatic pooling type recognition

michalkulakowski:
### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/embeddings/embeddings_calculator_ov.proto
+++ b/src/embeddings/embeddings_calculator_ov.proto
@@ -34,6 +34,6 @@ message EmbeddingsCalculatorOVOptions {
       LAST = 1;
       MEAN = 2;
     }
-    optional Pooling pooling = 5 [default = CLS];
+    optional Pooling pooling = 5;
     optional bool truncate = 6 [default = false];
 }
--- a/src/embeddings/embeddings_node_initializer.cpp
+++ b/src/embeddings/embeddings_node_initializer.cpp
@@ -15,18 +15,151 @@
 //*****************************************************************************
 #include <memory>
 #include <string>
+#include <sstream>
 #include <utility>
+#include <filesystem>
+#include <fstream>
+#include <optional>
 
 #include "src/mediapipe_internal/graph_side_packets.hpp"
 #include "src/mediapipe_internal/node_initializer.hpp"
 #include "src/stringutils.hpp"
+#include "embeddings_node_initializer_utils.hpp"
 #include "embeddings_servable.hpp"
 #include "mediapipe/framework/calculator.pb.h"
 #include "src/embeddings/embeddings_calculator_ov.pb.h"
+#include "src/port/rapidjson_document.hpp"
+#include "src/port/rapidjson_istreamwrapper.hpp"
+#include "src/port/rapidjson_error.hpp"
 
 #include "src/logging.hpp"
 
 namespace ovms {
+namespace {
+
+std::filesystem::path resolveModelsPath(const std::string& modelsPath, const std::string& basePath) {
+    auto fsModelsPath = std::filesystem::path(modelsPath);
+    if (fsModelsPath.is_relative()) {
+        return std::filesystem::path(basePath) / fsModelsPath;
+    }
+    return fsModelsPath;
+}
+
+}  // namespace
+
+std::optional<mediapipe::EmbeddingsCalculatorOVOptions_Pooling> detectEmbeddingsPoolingFromConfig(
+    const std::filesystem::path& modelsPath) {
+    const auto poolingConfigPath = modelsPath / "1_Pooling" / "config.json";
+    if (!std::filesystem::exists(poolingConfigPath)) {
+        return std::nullopt;
+    }
+
+    std::ifstream ifs(poolingConfigPath);
+    if (!ifs.is_open()) {
+        SPDLOG_WARN("Failed to open pooling config file: {}", poolingConfigPath.string());
+        return std::nullopt;
+    }
+
+    rapidjson::Document poolingConfig;
+    rapidjson::IStreamWrapper isw(ifs);
+    rapidjson::ParseResult parseResult = poolingConfig.ParseStream(isw);
+    if (parseResult.Code()) {
+        SPDLOG_WARN("Failed to parse {}: {}",
+            poolingConfigPath.string(),
+            rapidjson::GetParseError_En(parseResult.Code()));
+        return std::nullopt;
+    }
+
+    if (!poolingConfig.IsObject()) {
+        SPDLOG_WARN("Invalid pooling config root in {}. Expected a JSON object.", poolingConfigPath.string());
+        return std::nullopt;
+    }
+
+    const bool useLastToken = poolingConfig.HasMember("pooling_mode_lasttoken") &&
+                              poolingConfig["pooling_mode_lasttoken"].IsBool() &&
+                              poolingConfig["pooling_mode_lasttoken"].GetBool();
+    const bool useMean = poolingConfig.HasMember("pooling_mode_mean_tokens") &&
+                         poolingConfig["pooling_mode_mean_tokens"].IsBool() &&
+                         poolingConfig["pooling_mode_mean_tokens"].GetBool();
+    const bool useCls = poolingConfig.HasMember("pooling_mode_cls_token") &&
+                        poolingConfig["pooling_mode_cls_token"].IsBool() &&
+                        poolingConfig["pooling_mode_cls_token"].GetBool();
+    const bool useMax = poolingConfig.HasMember("pooling_mode_max_tokens") &&
+                        poolingConfig["pooling_mode_max_tokens"].IsBool() &&
+                        poolingConfig["pooling_mode_max_tokens"].GetBool();
+    const bool useMeanSqrtLen = poolingConfig.HasMember("pooling_mode_mean_sqrt_len_tokens") &&
+                                poolingConfig["pooling_mode_mean_sqrt_len_tokens"].IsBool() &&
+                                poolingConfig["pooling_mode_mean_sqrt_len_tokens"].GetBool();
+    const bool useWeightedMean = poolingConfig.HasMember("pooling_mode_weightedmean_tokens") &&
+                                 poolingConfig["pooling_mode_weightedmean_tokens"].IsBool() &&
+                                 poolingConfig["pooling_mode_weightedmean_tokens"].GetBool();
+
+    if (useMax || useMeanSqrtLen || useWeightedMean) {
+        std::ostringstream unsupportedModes;
+        bool first = true;
+        auto appendMode = [&unsupportedModes, &first](const char* mode) {
+            if (!first) {
+                unsupportedModes << ", ";
+            }
+            unsupportedModes << mode;
+            first = false;
+        };
+        if (useMax) {
+            appendMode("pooling_mode_max_tokens");
+        }
+        if (useMeanSqrtLen) {
+            appendMode("pooling_mode_mean_sqrt_len_tokens");
+        }
+        if (useWeightedMean) {
+            appendMode("pooling_mode_weightedmean_tokens");
+        }
+
+        SPDLOG_WARN("Unsupported pooling mode(s) requested in {}: {}. OVMS currently supports only pooling_mode_cls_token, pooling_mode_mean_tokens and pooling_mode_lasttoken.",
+            poolingConfigPath.string(),
+            unsupportedModes.str());
+        return std::nullopt;
+    }
+
+    const size_t enabledModes = static_cast<size_t>(useLastToken) +
+                                static_cast<size_t>(useMean) +
+                                static_cast<size_t>(useCls);
+    if (enabledModes == 0) {
+        return std::nullopt;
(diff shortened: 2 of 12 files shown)


src/embeddings/embeddings_node_initializer.cpp
```diff
@@ -15,18 +15,151 @@
...
 namespace ovms {
+namespace {
+
+std::filesystem::path resolveModelsPath(const std::string& modelsPath, const std::string& basePath) {
+    auto fsModelsPath = std::filesystem::path(modelsPath);
+    if (fsModelsPath.is_relative()) {
+        return std::filesystem::path(basePath) / fsModelsPath;
+    }
+    return fsModelsPath;
+}
```
mzegla, line 38: Looks like quite common utility function. Don't we already have something like that in utils? If not, maybe we could have it in the higher level utils file, so other features can use it as well.

src/embeddings/embeddings_node_initializer.cpp
```diff
@@ -15,18 +15,151 @@
...
+                        poolingConfig["pooling_mode_cls_token"].GetBool();
+    const bool useMax = poolingConfig.HasMember("pooling_mode_max_tokens") &&
+                        poolingConfig["pooling_mode_max_tokens"].IsBool() &&
+                        poolingConfig["pooling_mode_max_tokens"].GetBool();
+    const bool useMeanSqrtLen = poolingConfig.HasMember("pooling_mode_mean_sqrt_len_tokens") &&
+                                poolingConfig["pooling_mode_mean_sqrt_len_tokens"].IsBool() &&
+                                poolingConfig["pooling_mode_mean_sqrt_len_tokens"].GetBool();
+    const bool useWeightedMean = poolingConfig.HasMember("pooling_mode_weightedmean_tokens") &&
+                                 poolingConfig["pooling_mode_weightedmean_tokens"].IsBool() &&
+                                 poolingConfig["pooling_mode_weightedmean_tokens"].GetBool();
```
mzegla: Maybe we could have enum `PoolingMode` and a function like `getPoolingModeFromConfig(poolingConfig)` that would return it here. 
We would extract those ~20 lines and have a clean one-liner.

src/embeddings/embeddings_node_initializer.cpp
```diff
@@ -15,18 +15,151 @@
...
+        };
+        if (useMax) {
+            appendMode("pooling_mode_max_tokens");
+        }
+        if (useMeanSqrtLen) {
+            appendMode("pooling_mode_mean_sqrt_len_tokens");
+        }
+        if (useWeightedMean) {
+            appendMode("pooling_mode_weightedmean_tokens");
+        }
```
mzegla: Is it possible that we will get into more than one mode here? Are they not exclusive? Can we be at multiple modes at once? 
If they are exclusive then we could simplify this logic.

src/embeddings/embeddings_node_initializer.cpp
```diff
@@ -15,18 +15,151 @@
...
+                                static_cast<size_t>(useCls);
+    if (enabledModes == 0) {
+        return std::nullopt;
+    }
+
+    if (enabledModes > 1) {
+        SPDLOG_WARN("Ambiguous pooling config in {}. Exactly one of pooling_mode_lasttoken, pooling_mode_mean_tokens, pooling_mode_cls_token should be true.",
+            poolingConfigPath.string());
+        return std::nullopt;
+    }
```
mzegla: I guess it answers the question above. If we follow the idea of get pooling mode method, it could have that checking logic and return information about problematic config that we would log.

src/embeddings/embeddings_node_initializer.cpp
```diff
@@ -15,18 +15,151 @@
...
+
+    if (useLastToken) {
+        return mediapipe::EmbeddingsCalculatorOVOptions_Pooling_LAST;
+    }
+    if (useMean) {
+        return mediapipe::EmbeddingsCalculatorOVOptions_Pooling_MEAN;
+    }
+    return mediapipe::EmbeddingsCalculatorOVOptions_Pooling_CLS;
```
mzegla: Maybe we could have a map? like `poolingModeToMediapipeParam` that would hold that info and here we would just
`return poolingModeToMediapipeParam.get(<pooling_mode>, <default>)`
or something like that.

src/embeddings/embeddings_node_initializer.cpp
```diff
@@ -15,18 +15,151 @@
...
+    }
+
+    if (const auto detectedPooling = detectEmbeddingsPoolingFromConfig(modelsPath)) {
+        SPDLOG_DEBUG("Detected pooling '{}' from {}",
+            mediapipe::EmbeddingsCalculatorOVOptions_Pooling_Name(*detectedPooling),
+            (modelsPath / "1_Pooling" / "config.json").string());
+        return *detectedPooling;
+    }
+
+    return graphPooling;
```
mzegla: returning graphPooling even if `hasGraphPooling==false`? 
if it's true we have early return in L150, yet here we return again in a situation where this flag is false.

src/embeddings/embeddings_node_initializer_utils.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace ovms {
+
+std::optional<mediapipe::EmbeddingsCalculatorOVOptions_Pooling> detectEmbeddingsPoolingFromConfig(
+    const std::filesystem::path& modelsPath);
+
+mediapipe::EmbeddingsCalculatorOVOptions_Pooling resolveEmbeddingsPooling(
+    const std::filesystem::path& modelsPath,
+    bool hasGraphPooling,
+    mediapipe::EmbeddingsCalculatorOVOptions_Pooling graphPooling);
```
mzegla: Implementations are in embeddings_node_initalizer.cpp. Why we create separate header that is tied to a cpp file that is not his own? Maybe we could have cpp file for those two utils functions or move them to initializer header?

src/embeddings/embeddings_node_initializer.cpp
```diff
@@ -15,18 +15,151 @@
...
+    }
+
+    if (!poolingConfig.IsObject()) {
+        SPDLOG_WARN("Invalid pooling config root in {}. Expected a JSON object.", poolingConfigPath.string());
+        return std::nullopt;
+    }
+
+    const bool useLastToken = poolingConfig.HasMember("pooling_mode_lasttoken") &&
+                              poolingConfig["pooling_mode_lasttoken"].IsBool() &&
+                              poolingConfig["pooling_mode_lasttoken"].GetBool();
```
dkalinowski:
```suggestion
    bool getIfExists(const rapidjson::Document& doc, const std::string& name) {
      return poolingConfig.HasMember(name) && poolingConfig[name].IsBool() && poolingConfig[name].GetBool();
    }

    const bool useLastToken = getIfExists("pooling_mode_lasttoken");
    const bool ...
```

src/embeddings/embeddings_node_initializer.cpp
```diff
@@ -15,18 +15,151 @@
...
+            appendMode("pooling_mode_weightedmean_tokens");
+        }
+
+        SPDLOG_WARN("Unsupported pooling mode(s) requested in {}: {}. OVMS currently supports only pooling_mode_cls_token, pooling_mode_mean_tokens and pooling_mode_lasttoken.",
+            poolingConfigPath.string(),
+            unsupportedModes.str());
+        return std::nullopt;
+    }
+
+    const size_t enabledModes = static_cast<size_t>(useLastToken) +
```
dkalinowski: do we want to rely on a fact that bool casted to size_t is usually 1?

src/embeddings/embeddings_node_initializer.cpp
```diff
@@ -56,12 +189,16 @@ class EmbeddingsNodeInitializer : public NodeInitializer {
         }
         mediapipe::EmbeddingsCalculatorOVOptions nodeOptions;
         nodeConfig.node_options(0).UnpackTo(&nodeOptions);
+
+        const auto modelsPath = resolveModelsPath(nodeOptions.models_path(), basePath);
+        const auto pooling = resolveEmbeddingsPooling(modelsPath, nodeOptions.has_pooling(), nodeOptions.pooling());
```
dkalinowski: wouldnt it be simpler to pass nodeOptions to resolveEmbeddingsPooling? 1 param less and it can use has_pooling/pooling inside the resolveEmbeddingsPooling

src/test/embeddingsnode_test.cpp
```diff
@@ -18,17 +18,191 @@
...
+}  // namespace
+
+class EmbeddingsPoolingConfigTest : public ::testing::TestWithParam<EmbeddingsPoolingCase> {
+protected:
+    TempDir tempDir;
+
+    std::filesystem::path writePoolingConfig(const std::string& content) {
+        const auto poolingDir = tempDir.dir / "1_Pooling";
+        mkdirs(poolingDir);
+        writeFile(poolingDir, "config.json", content);
```
dkalinowski, line 65: is anything cleaning this up? why cant we use TestWithTempDir?

src/embeddings/embeddings_calculator_ov.proto
```diff
@@ -34,6 +34,6 @@ message EmbeddingsCalculatorOVOptions {
       LAST = 1;
       MEAN = 2;
     }
-    optional Pooling pooling = 5 [default = CLS];
+    optional Pooling pooling = 5;
```
mzegla, line 37: Won't this change impact backward compatibility?

src/embeddings/embeddings_node_initializer_utils.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+mediapipe::EmbeddingsCalculatorOVOptions_Pooling resolveEmbeddingsPooling(
+    const std::filesystem::path& modelsPath,
+    std::optional<mediapipe::EmbeddingsCalculatorOVOptions_Pooling> graphPooling) {
+    if (graphPooling.has_value()) {
+        return *graphPooling;
+    }
+
+    if (const auto detectedPooling = detectEmbeddingsPoolingFromConfig(modelsPath)) {
+        SPDLOG_DEBUG("Detected pooling '{}' from {}",
```
dkalinowski: Please check how we log automatic output parsers - it is on info level. Maybe we should do the same for pooling?

src/embeddings/embeddings_node_initializer_utils.cpp
```diff
@@ -0,0 +1,109 @@
...
+            mediapipe::EmbeddingsCalculatorOVOptions_Pooling_Name(*detectedPooling),
+            (modelsPath / "1_Pooling" / "config.json").string());
+        return *detectedPooling;
+    }
+
+    SPDLOG_WARN("No pooling type specified in graph config and no 1_Pooling/config.json found in {}. "
+        "Defaulting to CLS pooling.",
+        modelsPath.string());
```
mzegla: I would make the first time of the log more generic. Like: "Pooling mode was not specified and could not be inferred. Defaulting to..."

The part with "no 1_Pooling/config.json found in " might many times not be true as `detectEmbeddingsPoolingFromConfig` has many points where nullopt is returned and it's not always due to missing pooling config.json

src/embeddings/embeddings_node_initializer_utils.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+mediapipe::EmbeddingsCalculatorOVOptions_Pooling resolveEmbeddingsPooling(
+    const std::filesystem::path& modelsPath,
+    std::optional<mediapipe::EmbeddingsCalculatorOVOptions_Pooling> graphPooling) {
+    if (graphPooling.has_value()) {
+        return *graphPooling;
+    }
+
+    if (const auto detectedPooling = detectEmbeddingsPoolingFromConfig(modelsPath)) {
```
mzegla: How about we make `poolingConfigPath` here and pass it to this method. Then we can have it in one place and uniformly log. Now we compose it below in logging and inside detect method which only uses modelsPath to create path to pooling json

src/embeddings/embeddings_calculator_ov.proto
michalkulakowski: No, default pooling is still CLS but its set when there is no pooling config file and no pooling mode set in graph.pbtxt

mzegla (approved):

src/test/embeddingsnode_test.cpp
```diff
@@ -18,17 +18,188 @@
...
+    return info.param.testName;
+}
+
+std::string getUnsupportedPoolingCaseName(const testing::TestParamInfo<UnsupportedEmbeddingsPoolingCase>& info) {
+    return info.param.testName;
+}
+
+}  // namespace
+
+class EmbeddingsPoolingConfigTest : public ::testing::TestWithParam<EmbeddingsPoolingCase> {
```
dtrawins, line 58: wouldn;t it be easier just to add several test folder with 1_Pooling/config.json to check if function gets expected option including negative structure?

dkalinowski (approved):

PR #4403, opened 2026-07-24
Servable framework refactor - phase 2

mzegla:
To be added...


src/llm/apis/openai_api_handler.cpp
```diff
@@ -395,15 +382,61 @@ void OpenAIApiHandler::incrementProcessedTokens(size_t numTokens) {
...
-        parsedOutput.content = this->tokenizer.decode(generatedIds, ov::genai::skip_special_tokens(request.skipSpecialTokens));
-    } else {
-        parsedOutput = outputParser->parse(generatedIds, this->areToolsAvailable());
+std::string OpenAIApiHandler::serializeUnaryResponse(
+    const std::vector<std::vector<rapidjson::Document>>& allDeltas,
+    const std::vector<ov::genai::GenerationFinishReason>& finishReasons) {
+    return serializeUnaryResponse(allDeltas, finishReasons, {});
+}
+
+ParsedOutput OpenAIApiHandler::parsedOutputFromDeltas(const std::vector<rapidjson::Document>& deltas) {
```
przepeck: I wonder if the it wouldn't be better to have some struct containing all of these fields instead of assemble it from rapidjson deletas

src/llm/io_processing/devstral/tool_parser.cpp
```diff
@@ -175,6 +131,20 @@ std::optional<rapidjson::Document> DevstralToolParser::parseChunk(const std::str
...
+                if (endPos != std::string::npos) {
+                    this->streamContent = this->streamContent.substr(0, endPos);
+                }
+                if (!this->streamContent.empty()) {
+                    return parseContentChunk();
+                }
+            }
+            return std::nullopt;
+        }
         if (pos != std::string::npos) {
```
przepeck: `else` for clarity

src/llm/io_processing/base_output_parser.hpp
```diff
@@ -69,22 +70,53 @@ ToolsParameterTypeMap_t createToolsParametersTypesMap(const ToolsSchemas_t& tool
...
-    // append "<think>\n" at the end of the prompt when thinking is enabled.
     bool implicitStart = false;
 
+    // Resolve tokenIdStartTags → resolvedStartTokenToTag using the tokenizer.
+    // Called once from constructors that set parsingConfig.tokenIdStartTags.
+    void resolveSpecialTokenIds() {
+        for (const auto& tag : parsingConfig.tokenIdStartTags) {
+            if (tag.empty())
+                continue;
+            const auto tensor = tokenizer.encode(tag, ov::genai::add_special_tokens(false)).input_ids;
```
przepeck, line 91: OutputParser instance is created per request, do we want to tokenize tags every request? 
Maybe we will use tokenIds in `OutputParsingConfig`?

src/llm/io_processing/output_parser.cpp
```diff
@@ -231,12 +223,25 @@ OutputParser::OutputParser(ov::genai::Tokenizer& tokenizer, const std::string to
...
+    // <|channel|>... throughout the stream; devstral's [TOOL_CALLS] tag and minicpm5's <s>/<|im_end|>
+    // must be visible in the content phase). For all other parser combinations the content phase
+    // decodes with skip_special_tokens=true (the default, lower noise).
+    defaultDecodingWithSpecialTokens =
+        toolParserName == "gptoss" ||
+        toolParserName == "devstral" ||
+        toolParserName == "minicpm5" ||
+        reasoningParserName == "gptoss" ||
+        reasoningParserName == "lfm2" ||
+        reasoningParserName == "minicpm5";
```
przepeck: Isn't it better to have this info in tool parser's class?

src/llm/io_processing/llama3/tool_parser.cpp
```diff
@@ -122,13 +49,17 @@ static inline void changeParametersToArguments(rapidjson::Document& json) {
...
+        jsonHasArgumentsOrParameters(lastJson);
+    if (chunk.empty() && !hasPendingState) {
         SPDLOG_LOGGER_DEBUG(llm_calculator_logger, "Received empty chunk for Llama3ToolParser");
         return std::nullopt;
     }
 
-    // <|python_tag|> appears
-    if (chunk.find(parsingStartTag) != std::string::npos) {
+    // <|python_tag|> boundary text (synthesised by OutputParser on token-ID detection)
+    if (chunk.find("<|python_tag|>") != std::string::npos) {
```
przepeck: maybe use `parsingConfig.startTags[0]`

src/llm/io_processing/llama3/tool_parser.hpp
```diff
@@ -28,13 +29,7 @@
...
 class Llama3ToolParser : public BaseOutputParser {
 protected:
-    const std::string parsingStartTag = "<|python_tag|>";
-    // Tools calls are expected to be the last part of the content and there is no unique separator between tools, so we do not specify an end tag.
-    const std::string parsingEndTag = "";
-
-    // Id of the <|python_tag|> which is a special token used to indicate the start of a tool calls
-    int64_t botTokenId = 128010;
-    // ";" is used as a separator between tool calls in the response
+    // "" separator between tool calls
```
przepeck:
```suggestion
    // ";" separator between tool calls
```

src/llm/io_processing/lfm2/lfm25_reasoning_parser.hpp
przepeck, line 1: don't we want to use qwen3 reasoning parser for this model?

src/llm/io_processing/onyx/onyx_reasoning_parser.cpp
przepeck, line 1: same as for LFM2.5, can't we use qwen3 reasoning parser?

src/llm/io_processing/minicpm5/minicpm5_tool_parser.cpp
```diff
@@ -301,7 +301,8 @@ void Minicpm5ToolParser::lazyFillInitToolParametersTypesMap() {
 }
 
 Minicpm5ToolParser::Minicpm5ToolParser(ov::genai::Tokenizer& tokenizer, const ToolsSchemas_t& toolSchemas) :
-    BaseOutputParser(tokenizer),
+    BaseOutputParser(tokenizer,
+        defaultParsingConfig(FUNCTION_START_TAG, SOS_TOKEN_STR, EOS_TOKEN_STR)),
```
przepeck: Why do we parametrize this function here?

src/llm/io_processing/qwen3/reasoning_parser.cpp
```diff
@@ -25,59 +25,54 @@
...
+    rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+    writer.StartObject();
+    writer.String("delta");
+    writer.StartObject();
+    writer.String("reasoning_content");
+    writer.String(text.c_str());
+    writer.EndObject();
+    writer.EndObject();
+    rapidjson::Document doc;
+    doc.Parse(buffer.GetString());
```
przepeck: This code has been present in multiple reasoning parsers. We should extract it. This PR seems perfect for that

src/llm/io_processing/utils.cpp
```diff
@@ -86,7 +91,30 @@ size_t findInStringRespectingSpecialChars(const std::string& str, const std::str
...
+                }
+
+                size_t j = i + 1;
+                while (j < str.size() && std::isspace(static_cast<unsigned char>(str[j])) != 0) {
+                    ++j;
+                }
+                if (j == str.size() || str[j] == ',' || str[j] == ':' || str[j] == ']' || str[j] == '}' || str[j] == ')') {
+                    singleQuoteDepth = 0;
+                }
+            }
```
przepeck, line 117: I am not sure why do you change this function here. What case is it handling that previous behavior wasn't

src/llm/omni_model/legacy/servable.cpp
```diff
@@ -111,9 +111,10 @@ absl::Status OmniModelLegacyServable::parseRequest(std::shared_ptr<GenAiServable
 
     ov::AnyMap streamerConfig;
     if (omniExecutionContext->apiHandler->isStream()) {
-        if ((omniExecutionContext->apiHandler->getOutputParser() != nullptr &&
-                omniExecutionContext->apiHandler->getOutputParser()->requiresStreamingWithSpecialTokens()) ||
-            !omniExecutionContext->apiHandler->getRequest().skipSpecialTokens) {
+        const bool userWantsSpecial = !omniExecutionContext->apiHandler->getRequest().skipSpecialTokens;
+        const bool parserNeedsSpecial = omniExecutionContext->apiHandler->getOutputParser() != nullptr &&
```
przepeck:
```suggestion
        const bool userWantsSpecialTokens = !omniExecutionContext->apiHandler->getRequest().skipSpecialTokens;
        const bool parserNeedsSpecialTokens = omniExecutionContext->apiHandler->getOutputParser() != nullptr &&
```

src/test/llm/output_parsers/mistral_output_parser_test.cpp
```diff
@@ -111,14 +112,17 @@ TEST_F(MistralOutputParserTest, ParseToolCallOutputWithThreeToolCalls) {
     EXPECT_NE(secondToolCallId, thirdToolCallId);
 }
 
-TEST_F(MistralOutputParserTest, ParseToolCallOutputWithOneValidToolCallAndTwoInvalid) {
+// TODO: Re-enable after final behavior for mixed valid/invalid tool calls is
+// agreed. Current streaming path surfaces a parser exception when an invalid
+// call starts generating arguments before a valid "name" field.
```
przepeck: TODO to resolve

src/test/llm/output_parsers/gemma4_output_parser_test.cpp
```diff
@@ -189,7 +190,9 @@ TEST_F(Gemma4OutputParserTest, ParseToolCallOutputWithNoToolsInTheRequest) {
     for (auto& input : inputs) {
         auto generatedTensor = gemma4Tokenizer->encode(input).input_ids;
         std::vector<int64_t> generatedTokens(generatedTensor.data<int64_t>(), generatedTensor.data<int64_t>() + generatedTensor.get_size());
-        ParsedOutput parsedOutput = outputParserWithRegularToolParsing->parse(generatedTokens, false);
+        // With no tools available this path should behave like plain text output,
+        // so Gemma control tokens are stripped during decode.
```
przepeck: isn't it obvious from the context? This comment is probably unnecessary

src/test/llm/output_parsers/hermes3_output_parser_test.cpp
```diff
@@ -141,7 +138,13 @@ TEST_F(Hermes3OutputParserTest, ParseToolCallOutputWithThreeToolCalls) {
     }
 }
 
-TEST_F(Hermes3OutputParserTest, ParseToolCallOutputWithTwoValidToolCallsAndOneInvalid) {
+// TODO: Behavior gap introduced by the streaming migration.
+// On master this test passed via the dedicated OutputParser::parse() unary path, which
+// silently skipped tool calls with missing names.  The streaming path (parseChunk) throws
+// instead, matching the stricter "fully correct deltas" contract.
+// Decision needed: should parseChunk gracefully skip malformed tool calls (skip logic),
+// or should callers be required to provide valid input?  Until decided, this test is disabled.
```
przepeck: TODO  to resolve

src/test/llm/output_parsers/lfm25_output_parser_test.cpp
```diff
@@ -161,6 +162,23 @@ TEST_F(LFM25OutputParserTest, ParseToolCallOutputWithSingleToolCall) {
...
+    std::vector<int64_t> generatedTokens(generatedTensor.data<int64_t>(), generatedTensor.data<int64_t>() + generatedTensor.get_size());
+    ParsedOutput parsedOutput = ovms::test::parseWithStreamer(*lfm25Tokenizer, *outputParserWithRegularToolParsing, generatedTokens, true, false);
+    EXPECT_EQ(parsedOutput.content, "");
+    EXPECT_EQ(parsedOutput.reasoning, "");
+    ASSERT_EQ(parsedOutput.toolCalls.size(), 1);
+    EXPECT_EQ(parsedOutput.toolCalls[0].name, "example_tool");
+    EXPECT_EQ(parsedOutput.toolCalls[0].arguments, "{\"arg1\":\"value1\",\"arg2\":42}");
+    EXPECT_EQ(parsedOutput.toolCalls[0].id.empty(), false);
+}
+
```
przepeck, line 170: what is this test for? we don't want to skip sepecial tokens for this model

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
```diff
@@ -120,15 +121,15 @@ class Minicpm5OutputParserTest : public ::testing::Test {
...
     }
 
     void assertReasoningVec(const std::vector<std::tuple<std::string, ov::genai::GenerationFinishReason, std::optional<std::string>>>& chunkToDeltaVec) {
         for (const auto& [chunk, finishReason, expectedDelta] : chunkToDeltaVec) {
             std::vector<int64_t> tokens = {};
-            if (chunk == Minicpm5ReasoningParser::reasoningStartTag) {
+            if (chunk == "<think>") {
                 tokens = {Minicpm5ReasoningParser::reasoningStartTokenId};
-            } else if (chunk == Minicpm5ReasoningParser::reasoningEndTag) {
+            } else if (chunk == "</think>") {
```
przepeck, line 132: why do we change these tags to strings here?

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
```diff
@@ -266,13 +267,28 @@ TEST_F(Minicpm5OutputParserTest, ParseMixedStringAndIntegerParams) {
     EXPECT_EQ(parsedOutput.reasoning, "");
 }
 
-// This scenario will be handled only in unary, in streaming it's not possible to parse reasoning without the starting tag
+// In production, MiniCPM5's chat template appends <think> at the end of the
+// prompt; detectAndSetImplicitReasoningStart detects it and sets the parser
+// into implicit reasoning mode. The model then outputs reasoning text directly
+// without emitting <think>, and terminates with </think>.
+// We simulate that here via setImplicitReasoningStart(true).
 TEST_F(Minicpm5OutputParserTest, ParseReasoningWithoutStartingTag) {
```
przepeck, line 263: aren't we testing scenario that won't happen here?

src/llm/io_processing/onyx/onyx_reasoning_parser.cpp
przepeck, line 1: Reviewing tests for this model made me think about this model's unusual behavior. It sometimes doesn't emit `<think>` tag, When this was easy to handle in unary it was impossible to handle gracefully in streaming scenario. Do we want to think about solution for that problem. Maybe we need to introduce some kind of WA when this model starts the reasoning phase at the beginning of generation unless `enable_thinking` is `false`?

src/test/llm/output_parsers/phi4_output_parser_test.cpp
```diff
@@ -99,14 +100,17 @@ TEST_F(Phi4OutputParserTest, ParseToolCallOutputWithThreeToolCalls) {
     EXPECT_NE(secondToolCallId, thirdToolCallId);
 }
 
-TEST_F(Phi4OutputParserTest, ParseToolCallOutputWithOneValidToolCallAndTwoInvalid) {
+// TODO: Re-enable after final behavior for mixed valid/invalid tool calls is
+// agreed. Current streaming path surfaces a parser exception when an invalid
+// call starts generating arguments before a valid "name" field.
```
przepeck: TODO to resolve

src/llm/io_processing/gptoss/reasoning_parser.cpp
przepeck, line 1: GPT's reasoning parser still handles content, please add content parser for this model

src/llm/apis/openai_rapidjson_delta_serializer.hpp
```diff
@@ -0,0 +1,42 @@
...
+// Serializes Delta variants to the OpenAI streaming delta JSON schema using rapidjson.
+//
+// Output shapes:
+//   ContentDelta   → {"delta":{"content":"<text>"}}
+//   ReasoningDelta → {"delta":{"reasoning_content":"<text>"}}
+//   ToolCallDelta  → first delta (id/name present):
+//                      {"delta":{"tool_calls":[{"id":"<id>","type":"function","index":<n>,"function":{"name":"<name>"}}]}}
+//                    argument delta (id/name nullopt):
+//                      {"delta":{"tool_calls":[{"index":<n>,"function":{"arguments":"<args>"}}]}}
+//   FinishDelta    → {}
```
przepeck, line 32: no AudioDelta in this comment

src/llm/apis/openai_responses.cpp
```diff
@@ -1490,96 +1453,62 @@ std::string OpenAIResponsesHandler::serializeStreamingChunk(rapidjson::Document
...
-                if (!argDelta.empty() && static_cast<int>(responsesState.toolCalls.size()) > tcIndex) {
-                    responsesState.toolCalls[tcIndex].arguments += argDelta;
-                    events.emplace_back(serializeFunctionCallArgumentsDeltaEvent(responsesState.toolCalls[tcIndex].id, argDelta, tcOutputIdx));
-                }
-            }
-        }
-        // Empty delta object (no recognized member) — finish-only chunk, no events to emit here.
-    }
-    // Empty Document (no "delta" member) — finish-only chunk; lifecycle events already emitted above.
+    std::visit(overloaded{
```
przepeck, line 1514: this kind of block is hard to read. I wonder if it wouldn't be better to extract it to some other delta_serializer

src/llm/apis/openai_responses.cpp
```diff
@@ -1045,72 +1045,35 @@ std::string OpenAIResponsesHandler::serializeUnaryResponseImpl(const std::vector
...
-    for (const auto& finishReason : results.finish_reasons) {
-        if (finishReason == ov::genai::GenerationFinishReason::LENGTH) {
-            responsesFinishReason = ov::genai::GenerationFinishReason::LENGTH;
-            break;
+    parsedOutputs.reserve(allDeltas.size());
+    for (const auto& deltas : allDeltas) {
+        parsedOutputs.push_back(parsedOutputFromDeltas(deltas));
+    }
+    const ov::genai::GenerationFinishReason finishReason = [&]() {
+        // Promote LENGTH if any sequence was truncated, to avoid hiding truncation.
```
michalkulakowski: not sure if we need such comment

dtrawins (approved):

michalkulakowski (approved):

src/test/llm/output_parsers/default_content_parser_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+        GTEST_SKIP();
+    DefaultContentParser parser(*tokenizer, {"<|eot|>"});
+    EXPECT_EQ(parseContent(parser, "a<|eot|>b<|eot|>c"), "abc");
+}
+
+// ── Partial match → hold (return nullopt) ──────────────────────────────────
+
+TEST_F(DefaultContentParserTest, PartialTag_Hold) {
+    if (!tokenizer)
+        GTEST_SKIP();
```
przepeck: I wonder if we should skip these test when facebook model doesn't exist.
Another tests are still running using this model so for the full test without it it will still fail.
It may cover some issues during development phase and come up in CI which is suboptimal

src/llm/io_processing/base_output_parser.hpp
mzegla: I think there is no other way without pushing this refactor even deeper. Both OutputParser and OutputParsingConfig are per request entities that are tied to specific request from the beginning i.e. API handlers.
To avoid this per request encoding we would need to connect output parser/output parsing config to the deployment so it persists between requests or use some static instances. 

Encoding from strings is something I would like to keep, to have the flexibility of providing same config to the model that tokenizes same tags to different ids. In general, I think it would make sense to have at least parsing configs held on deployment level, in a servable for example. But the cost of such encoding is small enough to keep it as is for this PR in my opinion, so the refactor does not grow endlessly in a single PR.

src/llm/io_processing/utils.cpp
mzegla: It helps handling possessive apostrophes in words:
https://github.com/openvinotoolkit/model_server/pull/4403/changes#diff-a094c181b6416aea2eb7a9647bd7684aa283321bcf452afcb7d843f96a48f914R119-R133

src/test/llm/output_parsers/lfm25_output_parser_test.cpp
mzegla: User might want that. We need special tokens for tool calling and this test checks that.
So that even though we decode without special tokens, streamer+parser detect special token on the boundary and switch mode, so even if user explicitly asks for `skip_special_tokens: true` to decode content/unknown phases without special tokens, we will not miss it in tool phase detection.

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
mzegla: Because those statics no longer exist. This hardcoded replacement is the easiest change to make it work. 
Other way would be to create those parsers explicitly and read from them, but it might require changing the API to suffice tests needs (the idea was to keep those tags hidden in output parser, there were no plans for getters as no other component is expected to need them).

src/test/llm/output_parsers/minicpm5_output_parser_test.cpp
mzegla: Looking at the MiniCPM chat template we have in src/test/llm/chat_templates it ends with such part:
```
{%- if add_generation_prompt %}
    {{- '<|im_start|>assistant\n' }}
    {%- if enable_thinking is defined %}
        {%- if enable_thinking is false %}
            {{- '<think>\n\n</think>\n\n' }}
        {%- elif enable_thinking is true %}
            {{- '<think>\n' }}
        {%- endif %}
    {%- endif %}
{%- endif %}
```

As far as I understand it, with `enable_thinking` prompt will render with `<think>\n` at the end in the last assistant turn. In such case we will detect it as implicit reasoning and enter reasoning phase immediately as soon as parsing starts. This scenario is supposed to simulate that.

src/llm/io_processing/utils.cpp
przepeck: Doesn't this check handle this case `str[i] == '\'' && (i == 0 || str[i - 1] != '\\'`?

przepeck (approved):

PR #4416, opened 2026-07-28
refresh performance and troubleshooting guide

dtrawins:
### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/docs/performance_tuning.md
+++ b/docs/performance_tuning.md
@@ -26,7 +26,7 @@ There are several important considerations for tuning LLM serving via the OpenAI
       Lower values can improve prefix-caching efficiency by using smaller chunks, but also increase memory consumption.
       Allowed range: `8-256`.
 
-- Set `--enable_prefix_cache: true` (default) in the graph configuration to reuse KV cache for sequential requests with repeated prompt tokens (for example, chat history). This avoids duplicated prompt evaluation.
+- Set `--enable_prefix_cache true` (default) in the graph configuration to reuse KV cache for sequential requests with repeated prompt tokens (for example, chat history). This avoids duplicated prompt evaluation.
 
 - Use lower precision via model quantization and KV cache precision settings to improve throughput and reduce memory usage.
 
--- a/docs/performance_tuning.md
+++ b/docs/performance_tuning.md
@@ -14,8 +14,8 @@ There are several important considerations for tuning LLM serving via the OpenAI
       Static cache sizing is safer when you need hard memory limits for KV cache usage. It can improve prompt reuse under steady load, while requests that do not fit may be preempted.
 
 - The `--max_num_batched_tokens` parameter influences how prompts are divided and grouped in the prefill phase.
-      The default value `256` is efficient for short contexts and high concurrency memory management.
-      Increasing this value to `4096`, `8192`, or even the model context limit can improve first-token latency, but may increase KV cache memory consumption.
+      The default value `256` is efficient for short contexts and high concurrency.
+      Increasing this value to `4096`, `8192`, or even the model context limit can improve first-token latency.
 
 - On multi-socket systems, text generation is typically scoped to a single NUMA node.
       Disabling virtual NUMA nodes can improve locality in some deployments (for example, one NUMA node per CPU socket), but this is workload- and platform-dependent.
@@ -26,7 +26,7 @@ There are several important considerations for tuning LLM serving via the OpenAI
       Lower values can improve prefix-caching efficiency by using smaller chunks, but also increase memory consumption.
       Allowed range: `8-256`.
 
-- Set `enable_prefix_cache: true` (default) in the graph configuration to reuse KV cache for sequential requests with repeated prompt tokens (for example, chat history). This avoids duplicated prompt evaluation.
+- Set `--enable_prefix_cache: true` (default) in the graph configuration to reuse KV cache for sequential requests with repeated prompt tokens (for example, chat history). This avoids duplicated prompt evaluation.
 
 - Use lower precision via model quantization and KV cache precision settings to improve throughput and reduce memory usage.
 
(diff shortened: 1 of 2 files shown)


docs/performance_tuning.md
```diff
@@ -4,18 +4,43 @@
...
+- On multi-socket systems, text generation is typically scoped to a single NUMA node.
+      Disabling virtual NUMA nodes can improve locality in some deployments (for example, one NUMA node per CPU socket), but this is workload- and platform-dependent.
+      Always benchmark with and without vNUMA before standardizing this setting.
+
+- Models with linear attention also use `--cache_interval_multiplier`, which affects linear attention cache behavior.
+      The default value is adaptive and generally optimized for memory usage.
+      Lower values can improve prefix-caching efficiency by using smaller chunks, but also increase memory consumption.
+      Allowed range: `8-256`.
+
+- Set `enable_prefix_cache: true` (default) in the graph configuration to reuse KV cache for sequential requests with repeated prompt tokens (for example, chat history). This avoids duplicated prompt evaluation.
```
przepeck: maybe to be consistent with other params use `--enable_prefix_caching true`

docs/performance_tuning.md
```diff
@@ -4,18 +4,43 @@
...
-```
+There are several important considerations for tuning LLM serving via the OpenAI API endpoint:
+
+- Choose a cache size that matches the model, context length and expected concurrency. A practical starting point is dynamic resizing with value `0`, then observing cache usage in server logs under normal load.
+      Dynamic allocation can minimize RAM usage when demand is low, but with unrestricted concurrency and long contexts it can still grow enough to cause out-of-memory issues.
+      Static cache sizing is safer when you need hard memory limits for KV cache usage. It can improve prompt reuse under steady load, while requests that do not fit may be preempted.
+
+- The `--max_num_batched_tokens` parameter influences how prompts are divided and grouped in the prefill phase.
+      The default value `256` is efficient for short contexts and high concurrency memory management.
+      Increasing this value to `4096`, `8192`, or even the model context limit can improve first-token latency, but may increase KV cache memory consumption.
```
mzegla: Not sure I get the phrasing right. KV cache memory consumption as a space needed for KV per token is related to processed tokens amount. If we process 1000 tokens in one batch or in 10 batches, 100 tokens each, we need the same amount of memory for the generation. In higher concurrency lower max num batched tokens can indeed work as a bottleneck, so less concurrent requests are processed with big throughput and they get cleaned up before new ones are accepted. 

The overall, peak memory consumption is strictly tied to this parameter though. The bigger the batch, the more math is happening, so we need to use more memory to execute single step. That's less impactful than KV cache growth, but might be relevant in some situations.

docs/troubleshooting.md
```diff
@@ -2,26 +2,69 @@
...
+    If a model fails to load on one device, verify whether the model card or export configuration lists device-specific constraints.
+
+- Some newer models may not be supported on older platforms or older OpenVINO runtimes.
+    Xe+ platforms usually provide good coverage, but there can still be exceptions depending on the exact model architecture, precision, or plugin version.
+    If you see compatibility errors, confirm the model export and runtime version first.
+
+- Newly enabled LLM/VLM models may not always support the continuous batching pipeline, especially in preview or early-access scenarios.
+    In that case, switch to the non-continuous-batching pipeline with `--pipeline_type LM`, `--pipeline_type VLM`, or `--pipeline_type OMNI` as appropriate.
+    Continuous batching variants are the defaults: `LM_CB`, `VLM_CB`, and `OMNI_CB`.
+    If the model loads only after changing the pipeline type, the issue is usually a pipeline-model compatibility mismatch rather than a corrupted model.
```
mzegla, line 30: Maybe worth noting that NPU does not support continuous batching pipelines at all.

docs/performance_tuning.md
dtrawins: yes, that is correct interpretation, with a single client, memory usage is the same. With concurrency it will be higher than using small chunks

docs/performance_tuning.md
dtrawins:
```suggestion
- Set `--enable_prefix_cache: true` (default) in the graph configuration to reuse KV cache for sequential requests with repeated prompt tokens (for example, chat history). This avoids duplicated prompt evaluation.
```

docs/troubleshooting.md
```diff
@@ -2,26 +2,69 @@
...
+
```
dtrawins:
```suggestion
    NPU device can't use continuous batching pipeline.
```

docs/performance_tuning.md
```diff
@@ -4,18 +4,43 @@
...
+- On multi-socket systems, text generation is typically scoped to a single NUMA node.
+      Disabling virtual NUMA nodes can improve locality in some deployments (for example, one NUMA node per CPU socket), but this is workload- and platform-dependent.
+      Always benchmark with and without vNUMA before standardizing this setting.
+
+- Models with linear attention also use `--cache_interval_multiplier`, which affects linear attention cache behavior.
+      The default value is adaptive and generally optimized for memory usage.
+      Lower values can improve prefix-caching efficiency by using smaller chunks, but also increase memory consumption.
+      Allowed range: `8-256`.
+
+- Set `--enable_prefix_cache: true` (default) in the graph configuration to reuse KV cache for sequential requests with repeated prompt tokens (for example, chat history). This avoids duplicated prompt evaluation.
```
przepeck:
```suggestion
- Set `--enable_prefix_cache true` (default) in the graph configuration to reuse KV cache for sequential requests with repeated prompt tokens (for example, chat history). This avoids duplicated prompt evaluation.
```

przepeck (approved):

mzegla (approved):

PR #4421, opened 2026-07-31
make OVMSTextStreamer delay_n_tokens configurable via OVMS_LLM_DELAY_…

Zhiwei-Lii:
…N_TOKENS env var

OVMSTextStreamer previously hardcoded a 3-token lookahead buffer (DELAY_N_TOKENS) before flushing a decoded chunk to the client. This delay trades off time-to-first-token (TTFT) and inter-token latency for more reliable multi-token resolution (e.g. incomplete UTF-8 sequences, parser markers).

Make this tunable at runtime via the OVMS_LLM_DELAY_N_TOKENS environment variable, following the existing convention used by OVMS_AUDIO_MAX_FILE_SIZE_BYTES:
- kDefaultDelayNTokens (3) is now a public static constexpr on OVMSTextStreamer, used as the fallback default.
- getDelayNTokensFromEnv() reads and parses the env var via stou64(), falling back to the default (with a warning) when unset, empty, invalid, or zero.
- m_delay_n_tokens replaces the previous static DELAY_N_TOKENS constant and is resolved once at streamer construction time.

This applies uniformly across all pipeline types (LM, LM_CB, VLM, VLM_CB, OMNI) since they all construct OVMSTextStreamer the same way, with no graph.pbtxt / node_options changes required.

Documented the new env var in docs/llm/reference.md.

### 🛠 Summary

JIRA/Issue if applicable.
Describe the changes.

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/docs/llm/reference.md
+++ b/docs/llm/reference.md
@@ -110,6 +110,9 @@ The calculator supports the following `node_options` for tuning the pipeline con
 -    `optional bool enable_tool_guided_generation` - enable enforcing tool schema during generation. Requires setting response parser. [default = false];
 -    `optional SparseAttentionConfig sparse_attention_config` - Sparse attention configuration. Disabled if not specified.
 
+### Streaming settings
+The number of tokens the streaming response withholds before flushing a decoded chunk to the client can be tuned with the `OVMS_LLM_DELAY_N_TOKENS` environment variable (default: 3). This lookahead buffer lets the tokenizer resolve multi-token sequences (e.g. incomplete UTF-8 characters or parser markers) correctly before emitting a chunk. Increasing the value can improve output correctness for some models/parsers, but increases time to first token (TTFT) and inter-token latency. The value must be a positive integer; invalid, empty, or zero values fall back to the default.
+
 ### Caching settings
 The value of `cache_size` might have performance and stability implications. It is used for storing LLM model KV cache data. Adjust it based on your environment capabilities, model size and expected level of concurrency.
 You can track the actual usage of the cache in the server logs. You can observe in the logs output like below:
--- a/src/llm/ovms_text_streamer.cpp
+++ b/src/llm/ovms_text_streamer.cpp
@@ -16,10 +16,15 @@
 #include "ovms_text_streamer.hpp"
 
 #include <algorithm>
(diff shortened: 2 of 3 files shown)


Zhiwei-Lii:
hi @mzegla @przepeck @atobiszei , would you pls give a review on this PR, thanks

mzegla (commented):
I don't think we should allow this low level configuration. 
The delay of 3 tokens is not random and it is in line with the reference implementation of GenAI text streamer: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/text_streamer.cpp

To me it's not a trade off - it's a correctness requirement. For metrics collection, OVMS sends empty chunk signaling end of prefill phase as soon as model generates first token - giving real TTFT. The moment user receives first text chunk can slightly differ, but sometimes we have to collect a few tokens to be able to reliably decode.

As for inter-token - during generation, if we can decode single token, we decode and return. It's not that we always wait for 3 tokens and always return chunks created with 3 tokens, so there is no real impact on TPOT.

Zhiwei-Lii:
> I don't think we should allow this low level configuration. The delay of 3 tokens is not random and it is in line with the reference implementation of GenAI text streamer: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/text_streamer.cpp
> 
> To me it's not a trade off - it's a correctness requirement. For metrics collection, OVMS sends empty chunk signaling end of prefill phase as soon as model generates first token - giving real TTFT. The moment user receives first text chunk can slightly differ, but sometimes we have to collect a few tokens to be able to reliably decode.
> 
> As for inter-token - during generation, if we can decode single token, we decode and return. It's not that we always wait for 3 tokens and always return chunks created with 3 tokens, so there is no real impact on TPOT.

Thanks for your review :D
I understand your concerns. Here are my thoughts on this issue:
1. While OVMS sends an empty chunk marking the end of prefill, the client still waits 2 tokens to get the first non-null streaming chunk.
2. Tests on Qwen3-VL-8B / MinCPM V4 show setting m_delay_n_tokens from 2 to 1 reduces this 2-token latency and improves TTFT.

PR #4424, opened 2026-08-03
Remove typer constraint

stevegrubb:
Hugging Face Hub 1.21 requires Typer >=0.20, while the OVMS Kokoro requirements want Typer <0.12. These cannot coexist.

The prior commit message says:
"spaCy imports typer.main, which was removed in Typer 0.12."

Typer 0.12.0 did have a packaging bug that could leave module files missing while upgrading from an older Typer. It was fixed in Typer 0.12.1. It was not an intentional removal of typer.main. Typer issue #790 (https://github.com/fastapi/typer/issues/790)


mzegla (commented):
@michalkulakowski - won't this break Kokoro demo?

stevegrubb:
It shouldn't. This change removes an obsolete direct upper bound. Kokoro 0.9.4 pulls Misaki’s English extra, which pulls spaCy. spaCy supports typer>=0.3,<1.0, while the current Hugging Face Hub requires typer>=0.20,<0.26, so their compatible intersection is >=0.20,<0.26. typer.main is still present in those releases. On Fedora it resolves to 0.25.1.

Regarding the current limit, there was a Typer 0.12 incident where an upgrade packaging bug was fixed in 0.12.1 - not removal of typer.main. With the existing <0.12 cap, the common export requirements cannot resolve at all. This is why it needed to be dropped.

michalkulakowski (approved):

mzegla (approved):

PR #4465, opened 2026-08-20
Idle models

dtrawins:
### 🛠 Summary

This is a follow up to the PR https://github.com/openvinotoolkit/model_server/pull/4332
with extension of the feature according to https://github.com/openvinotoolkit/model_server/discussions/4456

### 🧪 Checklist

- [ ] Unit tests added.
``


src/modelmanager.cpp
```diff
@@ -1550,6 +1629,16 @@ Status ModelManager::getModelInstance(const std::string& modelName,
...
 
+    // On-demand group loading for idle model management
+    if (groupManager_ && groupManager_->isEnabled()) {
+        auto status = groupManager_->ensureGroupLoaded(modelName, const_cast<ModelManager&>(*this));
+        if (!status.ok()) {
+            SPDLOG_ERROR("Failed to load group for model '{}': {}", modelName, status.string());
+            return status;
+        }
+        groupManager_->recordActivity();
+    }
```
atobiszei, line 1640: A lot of code in OVMS depend on a assumption that model loading is done in single thread in model manager either during startup or in watcher loop. I think adding model loading in getModelInstance & createPipeline is just unsafe without adding a lot of tests concurrency tests which is unrealistic.

I think what we should do is:
1) Make watcher loop handle loading tasks one by one instead of just calling loadConfig(). 
2) Make createPipeline & getModelInstance create and push task for watcher loop as the next task to process. 

With that approach we can minimize tests required to cover this implementation.

src/modelmanager.cpp
```diff
@@ -1550,6 +1629,16 @@ Status ModelManager::getModelInstance(const std::string& modelName,
     std::unique_ptr<ModelInstanceUnloadGuard>& modelInstanceUnloadGuardPtr) const {
     SPDLOG_DEBUG("Requesting model: {}; version: {}.", modelName, modelVersionId);
 
+    // On-demand group loading for idle model management
+    if (groupManager_ && groupManager_->isEnabled()) {
+        auto status = groupManager_->ensureGroupLoaded(modelName, const_cast<ModelManager&>(*this));
```
atobiszei, line 1634: this method is const for a reason (const_cast)

src/kfs_frontend/kfs_grpc_inference_service.cpp
```diff
@@ -153,6 +164,15 @@ Status KFSInferenceServiceImpl::getModelReady(const KFSGetModelStatusRequest* re
...
         if (instance == nullptr) {
+            // In idle management mode, report configured models as ready even if no instance is loaded
+            auto* groupMgr = manager.getGroupManager();
+            if (groupMgr && groupMgr->isEnabled()) {
+                std::string group = groupMgr->getGroupForServable(name);
+                if (!group.empty()) {
+                    response->set_ready(true);
+                    return StatusCode::OK;
+                }
+            }
```
atobiszei, line 175: KFS/CAPI should not know about this feature. We shoudl just convert status. So possibly we need some lower level status translation, or additional ModelStatus code that would translate as READY.

PR #4492, opened 2026-08-31
Disable input count validation flag

przepeck:
### 🛠 Summary

[CVS-187884](https://jira.devtools.intel.com/browse/CVS-187884)
Adding flag `--disable_input_count_validation` to match TFS behavior with number of sent input entries grater then model input count

### 🧪 Checklist

- [ ] Unit tests added.
``


src/test/predict_validation_test.cpp
```diff
@@ -123,6 +123,48 @@ TEST_F(KFSPredictValidation, RequestTooManyInputs) {
...
+    void SetUp() override {
+        KFSPredictValidation::SetUp();
+        originalServerSettings = ovms::Config::instance().getServerSettings();
+        originalModelsSettings = ovms::Config::instance().getModelSettings();
+    }
+
+    void setDisableInputCountValidation(bool value) {
+        ovms::ServerSettingsImpl testServerSettings = originalServerSettings;
+        testServerSettings.disableInputCountValidation = value;
+        ovms::Config::instance().parse(&testServerSettings, &originalModelsSettings);
```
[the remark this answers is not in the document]
przepeck: if will fail everytime with "Use config_path or model_path with model_name" so it's not applicable in this semi-mock example

src/predict_request_validation_utils.hpp
```diff
@@ -290,7 +291,9 @@ Status RequestValidator<RequestType, InputTensorType, choice, IteratorType, Shap
         return StatusCode::NOT_IMPLEMENTED;
     }
     Status finalStatus = StatusCode::OK;
-    RETURN_IF_ERR(validateNumberOfTensors());
+    if (choice == ValidationChoice::INPUT && !ovms::Config::instance().disableInputCountValidation()) {
+        RETURN_IF_ERR(validateNumberOfTensors());
+    }
     RETURN_IF_ERR(validateRequestCoherency());
```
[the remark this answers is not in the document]
przepeck: I cannot agree with this comment, error "INVALID_MISSING_INPUT" is better for described scenario

docs/security_considerations.md
```diff
@@ -42,3 +42,7 @@ OpenVINO Model Server has a set of mechanisms preventing denial of service attac
 
 - MediaPipe does not validate all the settings during graph initialization. Some settings are checked during graph creation phase (upon request processing). Therefore it is a good practice to always test the configuration by sending example requests to the KServe endpoints before deployment.
 
+---
+
+By default, OVMS rejects inference requests whose input count/names don't match the model/pipeline signature. The `--disable_input_count_validation` flag turns off the *extra input* check, allowing requests with additional, unrecognized inputs to be processed (extra inputs are ignored; required inputs are still validated). **This flag is not recommended for production use** - disabling it may expose the model to malformed or malicious requests carrying unexpected inputs and weakens the server's input validation guarantees. Only enable it if you fully trust the clients sending requests to the server.
```
dtrawins: I think this is a bit far-fetched statement about security impact. It has performance consequences.

docs/parameters.md
```diff
@@ -58,6 +58,7 @@ Configuration options for the server are defined only via command-line options a
 | `allowed_local_media_path` | `string` | Path to the directory containing images to include in requests. If unset, local filesystem images in requests are not supported.|
 | `allowed_media_domains` | `string` | Comma separated list of media domains from which URLs can be used as input for LLMs. Set to \"all\" to disable this restrictions. If unset, URLs in requests are not supported."
 | `verbose_response` | `NA` | When enabled, responses include an extra `__verbose` object with additional debug information. Applies for text generation models |
+| `disable_input_count_validation` | `bool` (default: false) | When enabled, OVMS allows inference requests to include additional, unrecognized inputs beyond the model/pipeline signature (extra inputs are ignored). Required inputs must still be present, and shape/precision validation is still performed for recognized inputs. **Not recommended** - see [Security Considerations](security_considerations.md). |
```
dtrawins:
```suggestion
| `disable_input_count_validation` | `bool` (default: false) | Disables enforcement for the KServe requests to match all the model inputs. It ignores all inputs which are not used in the model. Not recommended for performance reasons but in some cases might simplify the client. |
```

dtrawins (approved):

src/predict_request_validation_utils.hpp
```diff
@@ -290,7 +291,9 @@ Status RequestValidator<RequestType, InputTensorType, choice, IteratorType, Shap
         return StatusCode::NOT_IMPLEMENTED;
     }
     Status finalStatus = StatusCode::OK;
-    RETURN_IF_ERR(validateNumberOfTensors());
```
atobiszei, line 293: I am not too keen on using singleton config here.

Ideally we would construct validator here with proper flags:
https://github.com/openvinotoolkit/model_server/blob/main/src/inference_executor.hpp#L62

And not spoil config dependency here. But there again we have big interface to validate() method (too much params), which we could probably wrap up with helper struct to pass required things down the validation pipeline. 

Probably to do as a follow up.

przepeck: good point, I will change that

atobiszei (approved):

PR #4506, opened 2026-09-03
Tool definition - response removing input processing | Qwen3-Next, Qwen3-Coder-Next and more

przepeck:
### 🛠 Summary

[CVS-187934](https://jira.devtools.intel.com/browse/CVS-187934)
Input processing for Qwen3, Qwen3 Next and Qwen3 Coder Next gemma4's original template and possibly more of them in the future.
Removing response field from tool definitions.

### 🧪 Checklist

- [ ] Unit tests added.
``


src/llm/io_processing/chat_template/caps.hpp
```diff
@@ -27,14 +27,17 @@ struct ChatTemplateCaps {
 
     std::string missnamedReasoningField = "";
 
+    bool removeResponseFromToolDefinition = false;
```
mzegla: Minor naming concern - how about renaming it to `supportsResponseFieldInToolDefinition`?
IMO it better suits struct intent (Caps->Capability) and `removeResponseFromToolDefinition` is already "call to action" and it believe the action part should be left for input processor to decide.

src/llm/io_processing/input_processors/chat_template_adapter.cpp
```diff
@@ -71,6 +71,26 @@ void injectReasoningIntoMissnamedSection(ov::genai::ChatHistory& chatHistory, co
...
+            continue;
+        }
+        auto function = toolDefinition["function"];
+        if (!function.is_object() || !function.contains("response")) {
+            continue;
+        }
+        function.erase("response");
+    }
+
+    chatHistory.set_tools(toolDefinitions);
```
mzegla, line 91: Do we have to do it? I wonder if we don't operate on references all the way through this function, could you make sure?

src/llm/io_processing/input_processors/chat_template_adapter.cpp
```diff
@@ -71,6 +71,26 @@ void injectReasoningIntoMissnamedSection(ov::genai::ChatHistory& chatHistory, co
...
+    for (size_t i = 0; i < toolDefinitions.size(); ++i) {
+        auto toolDefinition = toolDefinitions[i];
+        if (!toolDefinition.is_object() || !toolDefinition.contains("function")) {
+            continue;
+        }
+        auto function = toolDefinition["function"];
+        if (!function.is_object() || !function.contains("response")) {
+            continue;
+        }
+        function.erase("response");
```
mzegla, line 88: No issue with any pointers/iterators invalidation when removing during iteration?

src/llm/servable_initializer.cpp
```diff
@@ -177,6 +177,10 @@ static void probeServableChatTemplateCaps(std::shared_ptr<GenAiServablePropertie
     if (!properties->reasoningParserName.empty() && !probeChatTemplateReasoning(properties->tokenizer, properties->chatTemplateCaps)) {
         SPDLOG_LOGGER_WARN(llm_calculator_logger, "Chat template does not support reasoning_content field");
     }
+
+    if (!probeChatTemplateToolResponse(properties->tokenizer, properties->chatTemplateCaps)) {
+        SPDLOG_LOGGER_WARN(llm_calculator_logger, "Chat template does not support tool response field correctly");
+    }
```
mzegla, line 183: Looking at the lines and comments above, won't this logic be limited to Minja only?
Is that intended?

src/llm/io_processing/input_processors/chat_template_adapter.cpp
przepeck: I didn't observe any issue with that

src/llm/servable_initializer.cpp
przepeck: no, it won't, the comment above may be confusing, but it only applies for the condition below the comment

src/llm/io_processing/input_processors/chat_template_adapter.cpp
przepeck: I will check that

przepeck: ChatHistory exposes only one function to get tools: https://github.com/openvinotoolkit/openvino.genai/blob/7264b7b289c2b3241401e293b7beb3782cb85a8f/src/cpp/src/chat_history.cpp#L110
It returns `const JsonContainer&`,  so we are using copy, cannot operate on reference here.

michalkulakowski (approved):

mzegla (approved):

PR #4521, opened 2026-09-08
Embeddings --max_length parameter - error message when value is grater then model's max length

przepeck:
### 🛠 Summary

OVMS was silently failing when the `--max_length` parameter was grater then model's max length. Adding error message and OV_THROW to stop server.

### 🧪 Checklist

- [ ] Unit tests added.
``


dtrawins (approved):

src/embeddings/embeddings_servable.cpp
```diff
@@ -371,8 +371,13 @@ void reshapeModel(std::shared_ptr<Model>& model,
 
 std::shared_ptr<ov::Model> EmbeddingsServable::applyPrePostProcessing(ov::Core& core, std::shared_ptr<ov::Model> model, ov::AnyMap& properties) {
     if (this->configuredMaxLength.has_value()) {
-        SPDLOG_DEBUG("Overriding detected max model length {} with configured value {}", this->maxModelLength.value_or(0), this->configuredMaxLength.value());
-        this->maxModelLength = this->configuredMaxLength;
+        if (this->maxModelLength.has_value() && this->configuredMaxLength.value() > this->maxModelLength.value()) {
+            SPDLOG_ERROR("Configured max length {} is greater than detected max model length {}", this->configuredMaxLength.value(), this->maxModelLength.value());
+            OPENVINO_THROW("Configured max length is greater than detected max model length");
+        } else {
```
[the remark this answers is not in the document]
przepeck: I think it should be covered by functional tests not unit tests

src/embeddings/embeddings_servable.cpp
```diff
@@ -371,8 +371,13 @@ void reshapeModel(std::shared_ptr<Model>& model,
...
+            SPDLOG_DEBUG("Overriding detected max model length {} with configured value {}", this->maxModelLength.value_or(0), this->configuredMaxLength.value());
```
mzegla: Is it possible that we don't have `maxModelLength`?

If we don't have it we get the log saying:
`Overriding detected max model length 0 with configured value <value>`
For better user experience we could log explicitly that max model length was not detected instead of putting 0 there which might be confusing.

przepeck: done

mzegla (approved):

PR #4532, opened 2026-09-10
Array parsing fix for gemma4 output parser 

przepeck:
### 🛠 Summary

[CVS-194540](https://jira.devtools.intel.com/browse/CVS-194540)
Changing the way of handling gemma4 arrays, RapidJson::Parse was called too early so arrays were converted to the json strings: `"[\"first_string\", \"second\", ... ]"`, helper functions were added to `utils.cpp` . 

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/llm/io_processing/gemma4/gemma4_tool_parser.cpp
+++ b/src/llm/io_processing/gemma4/gemma4_tool_parser.cpp
@@ -51,7 +51,7 @@ std::string Gemma4ToolParser::parseArrayParameter(const std::string& argumentStr
 
     std::string parsedArray = "[";
     bool firstElement = true;
-    for (const std::string& element : splitTopLevel(body, maskStringValues(body), TOOL_ARGS_SEPARATOR_STR)) {
+    for (const std::string& element : splitRespectingSpecialChars(body, TOOL_ARGS_SEPARATOR_STR, maskDelimitedStringValues(body, TOOL_ARGS_STRING_INDICATOR))) {
         if (!firstElement) {
             parsedArray += ",";
         }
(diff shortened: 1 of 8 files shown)


src/llm/io_processing/utils.cpp
```diff
@@ -219,4 +219,38 @@ std::string replaceSingleWithDoubleQuotes(const std::string& input) {
...
+    return buffer.GetString();
+}
+
+bool isWrappedByDelimiter(const std::string& value, const std::string& delimiter) {
+    return value.size() >= 2 * delimiter.size() &&
+           value.compare(0, delimiter.size(), delimiter) == 0 &&
+           value.compare(value.size() - delimiter.size(), delimiter.size(), delimiter) == 0;
+}
+
+std::vector<std::string> splitTopLevel(const std::string& content, const std::string& maskedContent, const std::string& separator) {
```
mzegla: What is the "top-level" aspect here? Doesn't this function simply split string into vector of strings? 
It feels to gemma specific to me since I guess the naming comes from the usage of `findInStringRespectingSpecialChars` with `maskedContent`. 
Not sure if it belongs to general utils - feels a bit out of context here.

src/llm/io_processing/gemma4/gemma4_tool_parser.cpp
```diff
@@ -43,144 +43,98 @@ const int64_t Gemma4ToolParser::reasoningTokenId = 100;     // <|channel>
...
+    bool firstElement = true;
+    for (const std::string& element : splitTopLevel(body, maskStringValues(body), TOOL_ARGS_SEPARATOR_STR)) {
+        if (!firstElement) {
+            parsedArray += ",";
         }
-        parsedArguments += "\"" + originalStr + "\",";
-
-        pos = stringEndPos + TOOL_ARGS_STRING_INDICATOR.size() + 1;
+        parsedArray += normalizeArgStr(element);
+        firstElement = false;
```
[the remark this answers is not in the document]
przepeck: Adding tests for that example, it works with unchanged PR

src/llm/io_processing/gemma4/gemma4_tool_parser.cpp
```diff
@@ -43,144 +43,98 @@ const int64_t Gemma4ToolParser::reasoningTokenId = 100;     // <|channel>
...
+    size_t errorOffset = tempDoc.GetErrorOffset();
+    SPDLOG_LOGGER_TRACE(llm_calculator_logger, "Failed to parse argument string as JSON, falling back to string value. Argument string: {}, Error: {} Offset: {}", normalized, errorMessage, errorOffset);
 
-    return normalized;
+    std::string fallback = arg;
+    trim(fallback);
+    if (fallback.size() >= 2 && fallback.front() == '\"' && fallback.back() == '\"') {
+        fallback = fallback.substr(1, fallback.size() - 2);
+    }
+    return escapeAsJsonString(fallback);
```
mzegla: Isn't this flow contradicting a bit? I mean we first strip quotes in fallback and then `escapeAsJsonString` runs this:
`doc.Parse(("\"" + rawValue + "\"").c_str());`

src/llm/io_processing/utils.cpp
przepeck: As long as, I agree with you about the name of the function, I feel the place of this function is right if we move `maskStringValues` to `utils.cpp`. What do you think about this?

mzegla (approved):

michalkulakowski (approved):

PR #4551, opened 2026-09-11
Do not divide by zero when a download reports no Content-Length

### LINKED ISSUE #4550: Download progress bar divides by zero without Content-Length and prints ~2.1 billion spaces per tick
**Describe the bug**

When a model download is served over a response with no `Content-Length` (chunked transfer encoding), the progress bar divides by zero and then prints roughly 2.1 billion spaces per progress tick. The pull appears to hang; the transfer itself is fine.

`src/pull_module/curl_downloader.cpp:49-63`:

```cpp
static void print_progress(size_t count, size_t max, bool first_run, size_t elapsed_time) {
    float progress = (float)count / max;          // :50   max == 0 -> inf
    if (!first_run && progress < 0.01 && count > 0)
        return;

    const int bar_width = 50;
    int bar_length = progress * bar_width;        // :55   (int)inf is undefined
    ...
    for (i = bar_length; i < bar_width; ++i) {    // :62
        printf(" ");
    }
```

`max` is libcurl's `dltotal`, which is `0` when the server does not announce a size. The guard in `progress_callback` (`src/pull_module/curl_downloader.cpp:133-135`) only short-circuits while `dltotal == dlnow`:

```cpp
if ((dltotal == dlnow) && dltotal < 10000) {
    return 0;
}
```

Once any bytes have arrived, `dlnow > 0 == dltotal`, so the callback falls through to `print_progress(dlnow, 0, ...)`.

`progress` becomes `inf`. Converting an infinite float to `int` is undefined behaviour; on x86-64 it yields `INT_MIN`. The bar-fill loop at `:59` therefore does not run at all, and the padding loop at `:62` runs from `INT_MIN` to `bar_width`.

Compiling exactly those two lines and counting the padding iterations:

```
bar_length = -2147483648
padding loop would run 2147483698 times
```

That is ~2.1 billion `printf(" ")` calls on every tick, and `progress_callback` prints once a second.

**Related, same function:** `bar_length` is never clamped, so a server reporting more bytes than it announced (`dlnow > dltotal`, which happens with some content encodings) makes `progress > 1` and overruns the bar in the other direction.

**To Reproduce**

Point a pull at any HTTP endpoint that responds with `Transfer-Encoding: chunked` and no `Content-Length` — a local mirror, a proxy, an internal registry, or a plain `python3 -m http.server` in front of the artifacts:

```
ovms --pull --source_model <model> --model_repository_path /tmp/repo
```

The download proceeds but the console stops being usable and the process appears stuck.

**Expected behavior**

With an unknown total size, print a running byte count (or a spinner) rather than a percentage bar, and never divide by `max` when `max == 0`. The bar length should also be clamped to the bar width so no ratio can drive the loops out of range.

**Logs**

Console output only — the terminal fills with whitespace and the `\r`-based progress line never completes.

**Configuration**

1. OVMS version: `main` @ `fadb3314`
2. Any `--pull` against a source served without `Content-Length`
3. CPU
4. N/A
5. Any model

**Additional context**

Suggested fix: early-return from `print_progress` when `max == 0` and report bytes received instead, and move the bar arithmetic into a small helper that returns 0 for an unknown total and clamps to `[0, barWidth]`.

I have a patch for this and will open a PR shortly.

Daksha1611:
### 🛠 Summary

Fixes #4550.

`print_progress()` computed `(float)count / max` with `max` taken straight from libcurl's `dltotal`. A response using chunked transfer encoding reports `dltotal == 0`, and the guard in `progress_callback` only short-circuits while `dltotal == dlnow`, so once any bytes arrive the callback falls through to `print_progress(dlnow, 0, ...)`.

The ratio is then infinite, and converting that to `int` is undefined — on x86-64 it yields `INT_MIN`, so the bar-fill loop does not run and the padding loop below it runs from `INT_MIN` to `bar_width`. That is roughly 2.1 billion `printf(" ")` calls per tick, once a second, so the pull looks frozen and floods the terminal.

Reports the running byte count when the total is unknown, and moves the bar arithmetic into `computeProgressBarCells()`, which returns 0 for an unknown total and clamps to `[0, barWidth]`. The clamp also covers a server reporting more bytes than it announced, which previously overran the bar the other way.

`computeProgressBarCells()` is declared in the header only so the arithmetic can be unit tested — `print_progress()` is a file-local static that writes to stdout, and there is no existing test surface for the pull module's console output. If you would rather keep the header minimal I am happy to make it static again and drop the tests. The tests live in `pull_hf_model_test.cpp`, whose target already links `curl_downloader` transitively, so no BUILD change was needed.

I have no OVMS build container available, so this is not compiled against the full tree and CI will need to confirm the build. I did check the helper in isolation, including reproducing the original `INT_MIN` result from the two upstream lines. Draft for that reason.

### 🧪 Checklist

- [ ] Unit tests added.

### DIFF
--- a/src/pull_module/curl_downloader.cpp
+++ b/src/pull_module/curl_downloader.cpp
@@ -46,13 +46,45 @@ static void print_download_speed_info(size_t received_size, size_t elapsed_time)
     printf(" [%.2f %s/s] ", rate, sizeUnits[rate_unit_idx]);
 }
 
+int computeProgressBarCells(size_t count, size_t max, int barWidth) {
+    if (max == 0 || barWidth <= 0) {
+        return 0;
+    }
+    const double ratio = static_cast<double>(count) / static_cast<double>(max);
+    // Written as a positive test so a NaN ratio also lands here rather than falling through.
+    if (!(ratio > 0.0)) {
+        return 0;
+    }
+    if (ratio >= 1.0) {
+        return barWidth;
+    }
+    return static_cast<int>(ratio * barWidth);
+}
+
 static void print_progress(size_t count, size_t max, bool first_run, size_t elapsed_time) {
+    // A response with no Content-Length reports dltotal == 0, so there is no ratio to show;
+    // report the running byte count instead. Dividing by max here yielded an infinite ratio
+    // whose conversion to int is undefined - in practice INT_MIN, which drove the padding
+    // loop below through roughly 2.1 billion putchar calls on every progress tick.
+    if (max == 0) {
+        double received = (double)count;
+        size_t receivedUnitId = 0;
+        while (received > 1000 && sizeUnits[receivedUnitId + 1]) {
+            received /= 1000.0;
+            receivedUnitId++;
+        }
+        printf("\rProgress: %.2f %s downloaded, total size unknown", received, sizeUnits[receivedUnitId]);
+        print_download_speed_info(count, elapsed_time);
+        fflush(stdout);
+        return;
+    }
+
     float progress = (float)count / max;
     if (!first_run && progress < 0.01 && count > 0)
         return;
 
     const int bar_width = 50;
-    int bar_length = progress * bar_width;
+    const int bar_length = computeProgressBarCells(count, max, bar_width);
 
     printf("\rProgress: [");
     int i;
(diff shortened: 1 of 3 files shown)


rasapala:
Thank you for the contribution.

PR #4560, opened 2026-09-14
Cvs 194742 progress bar fix

rasapala:
### 🛠 Summary

JIRA 194742
ISSUE 4550
Contribution: https://github.com/openvinotoolkit/model_server/pull/4551

### 🧪 Checklist

- [ ] Unit tests added.
``

### DIFF
--- a/src/pull_module/curl_downloader.cpp
+++ b/src/pull_module/curl_downloader.cpp
@@ -46,13 +46,45 @@ static void print_download_speed_info(size_t received_size, size_t elapsed_time)
     printf(" [%.2f %s/s] ", rate, sizeUnits[rate_unit_idx]);
 }
 
+int computeProgressBarCells(size_t count, size_t max, int barWidth) {
+    if (max == 0 || barWidth <= 0) {
+        return 0;
+    }
+    const double ratio = static_cast<double>(count) / static_cast<double>(max);
+    // Written as a positive test so a NaN ratio also lands here rather than falling through.
+    if (!(ratio > 0.0)) {
+        return 0;
+    }
+    if (ratio >= 1.0) {
+        return barWidth;
+    }
+    return static_cast<int>(ratio * barWidth);
+}
+
 static void print_progress(size_t count, size_t max, bool first_run, size_t elapsed_time) {
+    // A response with no Content-Length reports dltotal == 0, so there is no ratio to show;
+    // report the running byte count instead. Dividing by max here yielded an infinite ratio
+    // whose conversion to int is undefined - in practice INT_MIN, which drove the padding
+    // loop below through roughly 2.1 billion putchar calls on every progress tick.
+    if (max == 0) {
+        double received = (double)count;
+        size_t receivedUnitId = 0;
+        while (received > 1000 && sizeUnits[receivedUnitId + 1]) {
+            received /= 1000.0;
+            receivedUnitId++;
+        }
+        printf("\rProgress: %.2f %s downloaded, total size unknown", received, sizeUnits[receivedUnitId]);
+        print_download_speed_info(count, elapsed_time);
+        fflush(stdout);
+        return;
+    }
+
     float progress = (float)count / max;
     if (!first_run && progress < 0.01 && count > 0)
         return;
 
     const int bar_width = 50;
-    int bar_length = progress * bar_width;
+    const int bar_length = computeProgressBarCells(count, max, bar_width);
 
     printf("\rProgress: [");
     int i;
--- a/src/pull_module/curl_downloader.hpp
+++ b/src/pull_module/curl_downloader.hpp
@@ -14,6 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //*****************************************************************************
+#include <cstddef>
 #include <string>
 
 namespace ovms {
@@ -22,5 +23,10 @@ class Status;
 Status downloadFileWithCurl(const std::string& url, const std::string& filePath);
 Status downloadFileWithCurl(const std::string& url, const std::string& filePath, const std::string& authTokenHF);
 Status fetchUrlToString(const std::string& url, const std::string& authToken, std::string& responseBody);
+
+// Number of filled cells in a barWidth-wide progress bar for count out of max bytes,
+// clamped to [0, barWidth]. max == 0 means the server sent no Content-Length, so there is
+// no ratio to render and the result is 0. Declared here so the arithmetic can be unit tested.
+int computeProgressBarCells(size_t count, size_t max, int barWidth);
 
 }  // namespace ovms
--- a/src/test/pull_hf_model_test.cpp
+++ b/src/test/pull_hf_model_test.cpp
@@ -18,6 +18,7 @@
 #include <cstdlib>
 #include <fstream>
 #include <iomanip>
+#include <limits>
 #include <memory>
 #include <openssl/sha.h>
 #include <mutex>
@@ -47,6 +48,7 @@
 #include "src/test/test_file_utils.hpp"
 #include "src/test/test_with_temp_dir.hpp"
 #include "src/filesystem/filesystem.hpp"
+#include "src/pull_module/curl_downloader.hpp"
 #include "src/pull_module/hf_pull_model_module.hpp"
 #include "src/pull_module/libgit2.hpp"
 #include "src/pull_module/optimum_export.hpp"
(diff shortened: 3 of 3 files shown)


src/pull_module/curl_downloader.cpp
```diff
@@ -46,13 +48,45 @@ static void print_download_speed_info(size_t received_size, size_t elapsed_time)
     printf(" [%.2f %s/s] ", rate, sizeUnits[rate_unit_idx]);
 }
 
+int computeProgressBarCells(size_t count, size_t max, int barWidth) {
+    if (max == 0 || barWidth <= 0) {
+        return 0;
+    }
+    const double ratio = static_cast<double>(count) / static_cast<double>(max);
+    // Written as a positive test so a NaN ratio also lands here rather than falling through.
```
mzegla: "Written as positive test"?

src/pull_module/curl_downloader.cpp
```diff
@@ -46,13 +48,45 @@ static void print_download_speed_info(size_t received_size, size_t elapsed_time)
...
+    if (!(ratio > 0.0)) {
```
mzegla: Why not `if (ratio <= 0.0)`

src/pull_module/curl_downloader.cpp
```diff
@@ -46,13 +48,45 @@ static void print_download_speed_info(size_t received_size, size_t elapsed_time)
...
+        return 0;
+    }
+    if (ratio >= 1.0) {
```
mzegla: Could be `else if` with above condition?

src/pull_module/curl_downloader.cpp
```diff
@@ -46,13 +48,45 @@ static void print_download_speed_info(size_t received_size, size_t elapsed_time)
...
+        return barWidth;
+    }
+    return static_cast<int>(ratio * barWidth);
+}
+
 static void print_progress(size_t count, size_t max, bool first_run, size_t elapsed_time) {
+    // A response with no Content-Length reports dltotal == 0, so there is no ratio to show;
+    // report the running byte count instead. Dividing by max here yielded an infinite ratio
+    // whose conversion to int is undefined - in practice INT_MIN, which drove the padding
+    // loop below through roughly 2.1 billion putchar calls on every progress tick.
```
mzegla: Looks like leftover comment from development process. It's hard for me to read as this comment seems to assume I have knowledge about some external context `"...Content-Length reports dltotal == 0..."`.

src/pull_module/curl_downloader.cpp
```diff
@@ -46,13 +48,45 @@ static void print_download_speed_info(size_t received_size, size_t elapsed_time)
...
+    if (max == 0) {
+        double received = (double)count;
+        size_t receivedUnitId = 0;
+        while (received > 1000 && sizeUnits[receivedUnitId + 1]) {
+            received /= 1000.0;
+            receivedUnitId++;
```
mzegla, line 72: No risk of getting out of bounds with this index? Maybe we could have some defensive check against expected size of `sizeUnits`?

src/pull_module/curl_downloader.hpp
```diff
@@ -23,4 +24,9 @@ Status downloadFileWithCurl(const std::string& url, const std::string& filePath)
 Status downloadFileWithCurl(const std::string& url, const std::string& filePath, const std::string& authTokenHF);
 Status fetchUrlToString(const std::string& url, const std::string& authToken, std::string& responseBody);
 
+// Number of filled cells in a barWidth-wide progress bar for count out of max bytes,
+// clamped to [0, barWidth]. max == 0 means the server sent no Content-Length, so there is
+// no ratio to render and the result is 0. Declared here so the arithmetic can be unit tested.
```
mzegla:
```suggestion
// no ratio to render and the result is 0.
```
Unnecessary overhead. If it's not natural, we should change it. If it is, no need for a comment.

src/test/pull_hf_model_test.cpp
```diff
@@ -378,6 +382,75 @@ ::testing::AssertionResult interruptPosixWorkerAndExpectGracefulExit(pid_t child
 
 }  // namespace
 
+// A response without Content-Length makes libcurl report dltotal == 0. The progress bar must
+// not divide by it: the ratio becomes infinite and converting that to int is undefined, which
+// in practice produced INT_MIN and a ~2.1 billion iteration padding loop.
```
mzegla: I don't think we need it here.

src/test/pull_hf_model_test.cpp
```diff
@@ -378,6 +382,75 @@ ::testing::AssertionResult interruptPosixWorkerAndExpectGracefulExit(pid_t child
...
+TEST(CurlDownloaderProgressTest, FilledCellsClampToBarWidth) {
+    EXPECT_EQ(ovms::computeProgressBarCells(200, 100, 50), 50);
+    EXPECT_EQ(ovms::computeProgressBarCells(100, 100, 0), 0);
+    EXPECT_EQ(ovms::computeProgressBarCells(100, 100, -1), 0);
+}
+
+// Regression test for the call-site, not just the extracted helper: a real chunked-transfer
+// response (no Content-Length) drives libcurl's progress callback with dltotal == 0 on every
+// tick. Before the fix this hung the download in a ~2.1 billion iteration padding loop; here
+// we bound the wait so a reintroduced regression fails instead of hanging the test suite.
```
mzegla: Repetition of the comment above function signature.

src/pull_module/curl_downloader.cpp
```diff
@@ -15,10 +15,12 @@
 //*****************************************************************************
 #include "curl_downloader.hpp"
 
+#include <cstdlib>
 #include <ctime>
 #include <filesystem>
 #include <iostream>
 #include <memory>
+#include <mutex>
```
mzegla, line 24: Why do we need to include mutex in this PR?

src/pull_module/curl_downloader.cpp
```diff
@@ -117,6 +151,25 @@ static size_t file_write_callback(void* buffer, size_t size, size_t nmemb, void*
...
+// calling curl_global_cleanup() after every download tears down global TLS/engine state
+// still needed elsewhere (other in-flight curl users), which segfaults on next use.
+static Status ensureCurlGlobalInit() {
+    static std::once_flag initFlag;
+    static CURLcode initResult = CURLE_OK;
+    std::call_once(initFlag, []() {
+        initResult = curl_global_init(CURL_GLOBAL_DEFAULT);
+        if (initResult == CURLE_OK) {
+            std::atexit([]() { curl_global_cleanup(); });
+        }
```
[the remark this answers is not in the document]
rasapala: Production CLI use-after-cleanup: low / no demonstrated path.
Unit-test and C API restart sensitivity: possible, but current persistent initialization actually protects these paths.
Multiple independent lifecycle owners: maintenance risk, not demonstrated production failure.
First initialization after threads start: safe in the current Linux image because libcurl 8.5.0 advertises threadsafe.
Windows path verified, and not possibility there.

src/pull_module/curl_downloader.cpp
rasapala: std::once_flag and std::call_once

mzegla (approved):

src/test/pull_hf_model_test.cpp
```diff
@@ -378,6 +381,78 @@ ::testing::AssertionResult interruptPosixWorkerAndExpectGracefulExit(pid_t child
...
+}
+
+TEST(CurlDownloaderProgressTest, FilledCellsTrackRatio) {
+    EXPECT_EQ(ovms::computeProgressBarCells(0, 100, 50), 0);
+    EXPECT_EQ(ovms::computeProgressBarCells(50, 100, 50), 25);
+    EXPECT_EQ(ovms::computeProgressBarCells(100, 100, 50), 50);
+}
+
+// Some servers report more bytes transferred than announced; the bar must stay within its width
+// so the padding loop below it always runs a sane number of times.
```
atobiszei: Unneccesary comment - name of test suggests what is expected.

src/pull_module/curl_downloader.cpp
```diff
@@ -117,6 +147,25 @@ static size_t file_write_callback(void* buffer, size_t size, size_t nmemb, void*
...
+    });
+    if (initResult != CURLE_OK) {
+        SPDLOG_ERROR("curl error: {}. Error code: {}", curl_easy_strerror(initResult), (int)initResult);
+        return StatusCode::INTERNAL_ERROR;
+    }
+    return StatusCode::OK;
+}
```
atobiszei, line 167: This means that every unit test that did start OVMS before did curl init and curl cleanup. So ovms behaves now more differently in ovms gtest scenarios than in prod. Why could we not just move this initialization to pull module? 

Why we can't follow pattern used in httpservermodule.cpp?
Init curl in module init, cleanup in shutdown?

Then we could dispose whole part of curl_global_init from functions used here/.

src/test/pull_hf_model_test.cpp
```diff
@@ -2539,7 +2616,8 @@ TEST_F(HfDownloaderPullHfModel, DownloadImageGenModelWithLoRA) {
     std::string modelName = "OpenVINO/stable-diffusion-v1-5-int8-ov";
     std::string downloadPath = ovms::FileSystem::joinPath({this->directoryPath, "repository"});
     std::string task = "image_generation";
-    std::string sourceLoras = "pokemon=juliensimon/sd-pokemon-lora@pytorch_lora_weights.safetensors";
+    // juliensimon/sd-pokemon-lora was removed upstream; replaced with a repo verified to still exist.
+    std::string sourceLoras = "pokemon=MohamedAhmedAE/stable-diffusion-v1-5_lora_finetuning@pytorch_lora_weights.safetensors";
```
atobiszei: Just replace the lora name without adding commend - "some_lora"?

src/pull_module/curl_downloader.cpp
rasapala: First question - yes.
We could follow the HttpServerModule init/shutdown pattern for the cases where HfPullModelModule::start()/shutdown() actually runs — but curl_downloader's functions aren't only invoked through that module's lifecycle. The curl_downloader.cpp-local std::call_once + atexit guard exists precisely to make the downloader self-sufficient regardless of how it's invoked. That said, it is true this changes gtest behavior (every test that exercises this file now does one process-wide curl_global_init/atexit-cleanup pair instead of zero) — but that's a correctness fix, not a new inconsistency: previously the guard was broken.

atobiszei: It is used only in pull_module. The other place we use it is:
https://github.com/openvinotoolkit/model_server/blob/main/src/llm/io_processing/image_utils.cpp#L68

but its entirely different curl call with their own guards.

Why did previous solution not work?
If there was issue with guards it could be reused solution from image_utils if you want to keep it self-contained. If we want to keep it called only once (which is not required by curl) then we could have it pull module.

rasapala: Moved to one place in serversettings.

src/capi_frontend/server_settings.hpp
```diff
@@ -218,6 +218,10 @@ struct HFSettingsImpl {
 };
 
 struct ServerSettingsImpl {
+    // Ensures process-wide one-time initialization (currently: libcurl) happens as soon as
+    // server settings exist, regardless of entry point (CLI, C-API, tests) - see server_settings.cpp.
+    ServerSettingsImpl();
+
```
atobiszei: This is just simple struct with settings - we shouldnt give it another responsibility. 

IMO  we should keep the curl initializaiton in used modules, and since there is already hack that we don't wont to fix just now add another intialization guard in CLI parser.

atobiszei (approved):

PR #4568, opened 2026-09-15
Fix Windows service idle CPU spin and stop-event initialization order

0x484558:
Fix high idle CPU usage when OVMS is running as a Windows Service. The service supervisor invokes a non-blocking poll causing its thread to run even when the server itself is idle; replacing it with finite event wait that periodically checks server state and readiness while waking immediately when the stop event is signaled, with an atomic fallback, ensures that a failure to signal the event cannot leave the worker indefinitely running in SERVICE_STOP_PENDING.

Additionally, create the service stop event before starting the worker thread, removing a startup race in which the worker could see the wrapper's initial invalid handle value, and propagate failures through the Windows service status instead of always reporting a successful stop, while preserving the originating Win32 error for diagnostics.

### 🧪 Checklist

- [ ] Unit tests added. - OVMS lifecycle right now is not very unit-testable.
- [ ] The documentation updated. - No documentation update is necessary.


0x484558:
The Windows service idle CPU usage issue is trivially verifiable to be real. As for lifecycle change, I prepared a reproduction to show that the issue in current model server code is legitimate:

[error_ownership_repro.txt](https://github.com/user-attachments/files/32298493/error_ownership_repro.txt)
(GitHub did not allow me to upload it with .cpp extension)

The mock server returns only after TearDown requests shutdown, making the overwritten-error example deterministic and free of concurrent writes.

0x484558:
Closing in favor of #4578

PR #4575, opened 2026-09-16
Add Granite 4.2 thinking parser

0x484558:
### 🛠 Summary

Implement and register Granite 4.2 thinking parsing with request-level thinking controls, and make the embedded Jinja tojson filter compatible with the official template,

### 🧪 Checklist

- [x] Unit tests added.


src/llm/io_processing/granite42/granite42_reasoning_parser.hpp
przepeck, line 1: In OVMS we are using name `reasoning_parser`.
Additionally, please check MiniCPM5 reasoning parser for reference for qwen3-like reasoning parsers.

0x484558: I've changed the name.

0x484558: It's not qwen3-like reasoning parser, Granite 4.2 parser stays entirely within the original DeepSeek-R1/Nemotron lineage. I cannot subclass qwen3 because IBM parser has incompatible semantics, Nor does it need to cope with unstable garbage that Qwen models generate, which historically prompted Qwen3 reasoning parser to be special.

PR #4578, opened 2026-09-17
Change cpu idle on service and add stop signals

rasapala:
### 🛠 Summary

JIRA CVS-195009
Windows service lifecycle fixes: remove idle spin, fix the startup stop-event race, preserve/report real stop errors, and allow stop during START_PENDING/STOP_PENDING.. Check the WaitForSingleObject return code and report failure if any.

Based on https://github.com/openvinotoolkit/model_server/pull/4568

### 🧪 Checklist

- [ ] Unit tests added.
``


przepeck (approved):

src/main_windows.cpp
```diff
@@ -511,24 +572,46 @@ DWORD WINAPI OvmsWindowsServiceManager::serviceWorkerThread(LPVOID lpParam) {
...
+        }
+        if (waitResult == WAIT_TIMEOUT) {
+            continue;
+        }
+
+        const DWORD waitError = GetLastError();
+        DEBUG_LOG("serviceWorkerThread: WaitForSingleObject returned error.");
+        serviceReportEvent("WaitForSingleObject", waitError);
+        serviceWorkerWin32Error.store(waitError);
+        ovmsService->error = static_cast<int>(waitError);
```
0x484558, line 613: Remove this, it serves absolutely no semantic purpose but the presence of this line deliberately introduces a danger of data race because another thread owns `ovmsService->error` there is absolutely no reason to retain a known data race danger.

src/main_windows.cpp
```diff
@@ -731,7 +814,7 @@ void OvmsWindowsServiceManager::setPythonPathRegistry() {
 }
 
 void OvmsWindowsServiceManager::setServiceStopStatusPending() {
-    serviceStatus.dwControlsAccepted = 0;
+    serviceStatus.dwControlsAccepted = SERVICE_ACCEPT_STOP;
```
0x484558, line 817: Stop is already in progress here, this is not necessary. This isn't harmful but I just find the misunderstanding of windows service lifecycle amusing.

src/main_windows.cpp
```diff
@@ -511,24 +572,46 @@ DWORD WINAPI OvmsWindowsServiceManager::serviceWorkerThread(LPVOID lpParam) {
...
             DEBUG_LOG("serviceWorkerThread: Server thread is not running.")
             break;
         }
 
         if (!ovmsService->started && ovmsService->checkModulesStarted()) {
+            if (serviceStopRequested.load()) {
+                break;
+            }
             // Tell the service controller we are started
             OvmsWindowsServiceManager::setServiceRunningStatus();
```
0x484558, line 596: despite the check being immediately before the publication, it is still a check-then-act sequence across threads.

rasapala: Implemented atomic state transition.

atobiszei:
What happens in this scenario:?
Worker thread                         Stop-handler thread
-------------                         ------------------
CAS: Starting -> Running

                                      Sees START_PENDING
                                      Publishes STOP_PENDING
                                      Stores StopRequested

Writes serviceStatus = RUNNING
Calls SetServiceStatus(RUNNING)

